PR #452, opened 2013-10-01
Add gifs to track availability of JavaScript

tombye:
Add code to the homepage to track availability of JavaScript among our users.

To be removed one week from merge.

Dependant on this config change to stop the gifs having the default expires headers:

https://github.gds/gds/puppet/pull/978

Do not merge until the above is part of master.


jabley:
Did we consider using 204 No Content rather than actually sending images?

bradwright:
@jabley we didn't.

PR #712, opened 2015-01-05
Remove unused jQuery localscroll plugin

edds:
There was an attempt to remove this when the feature was removed ff50102
however due to a merge conflict it got left behind 882a5fb. It isn't
used so we can get rid of it.


tombye:
:+1:

PR #716, opened 2015-01-07
Add a `/random` route that redirects to a random page on GOV.UK

issyl0:
- Based on https://github.com/issyl0/govuk-random-page - if Wikipedia
  can have a random page link, why can't we? It will allow people to
  find out about stuff that they may never normally find out about.
- I was going to add a link to the footer for ease of repeatedly
  finding random pages, but that might be too prominent.

:fire:

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -32,6 +32,19 @@ class RootController < ApplicationController
       section_url: "/")
   end
 
+  def random_page
+    # Redirect to a random GOV.UK page, for fun.
+    base_url = "https://www.gov.uk/api/artefacts.json"
+    total_pages = JSON.parse(RestClient.get(base_url).body)['pages']
+
+    random_page_number = Random.rand(1..total_pages)
+
+    random_page = RestClient.get("#{base_url}?page=#{random_page_number}")
+    result = JSON.parse(random_page)['results'].shuffle.first['web_url']
+
+    redirect_to result.gsub("https://www.gov.uk","#{Plek.current.find('www')}")
+  end
+
   def jobsearch
     @publication = prepare_publication_and_environment
   end
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -34,6 +34,11 @@ class RootController < ApplicationController
 
   def random_page
     # Redirect to a random GOV.UK page, for fun.
+
+    # Disable caching for this route to ensure that users actually
+    # get a random page every time.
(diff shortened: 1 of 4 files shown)


edds:
Will this work as expected in production? Will the CDN cache the redirect and there will only be a random page every 30 mins?

alext:
Without setting explicit cache headers, there will be a default cache TTL applied at the CDN.

But, that's all moot because there's nothing to create any routes to allow requests to reach this endpoint.

bradleywright:
I think sorting the entire list of GOV.UK pages in memory might be a bit intensive. It would be easier to add a `/random` endpoint to content API or similar and fetch a single page from that, where the random slice could be done in the database.

issyl0:
Oh, I hadn't thought about caching. What do you mean, @alext?

issyl0:
@bradleywright It only searches the one random page number's content in memory (500 items), not everything. :-\

KushalP:
@issyl0 I think what @alext means here is that there's no _exact_ route being registered with the router for `/random`: https://github.com/alphagov/frontend/blob/master/lib/tasks/router.rake#L14-22

issyl0:
@alext, @KushalP: Done.

issyl0:
:question:

fatbusinessman:
I like this. I would like to see it go live on GOV.UK. I think there are a couple of things we’d need to consider before we can launch it:

* This will always get the artefact list from `www.gov.uk/api` (i.e. the production environment), even when it’s in preview or staging. This could potentially mean we try to redirect to something that doesn’t exist in preview or staging. I don’t think that’s a big problem, but we may be able to work around it using the [content API adapter](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/content_api.rb).
* We’re planning to replace the content API at the proverbial Some Point, and its replacement may no longer make support grabbing a random page. In the worst case, we may have to remove this functionality at that point; I don’t think that should stop this going live now.

And here’s the big one:

* There aren’t any tests; even with something that’s just for fun, I don’t think we can put it into production without some test coverage. I’m happy to work with you on this, @issyl0.

How does that sound?

issyl0:
@fatbusinessman Thanks. Tests: yes, absolutely - I gave it a go at first, but wasn't very successful. If we could work together on this (and I don't know anything about the content API adapters...), that'd be good.

issyl0:
Having paired with @fatbusinessman, there's now a test for this. :palm_tree:

PR #718, opened 2015-01-08
Remove unused helpers

rboulton:
None of the search helper methods were being used or tested, and the
readable_time helper in travel advice helper wasn't being used either.
Remove them.


edds:
Only red. I like your style.

rboulton:
It's a good ratio.

PR #722, opened 2015-01-09
Add metadata to non-edition search results

evilstreak:
Previously there were two classes of search results: results from the mainstream index, typically things like guides and smart answers which have browse sections as metadata, and results from the government index which have date, organisation and format as metadata.

We're slowly moving towards a model where everything is added to the mainstream index, and the new existing metadata approach doesn't work for these new content items. These new results are being added with a specific `document_type`, instead of having everything lumped together as an `edition`.

This commit adds a third type of search result presenter which is used for any results that are in the mainstream index and have a type other than `edition`. It shows a date, a humanized version of the document format, and the organisations the result is tagged to.

The humanized version of the document format is set manually for a few types because they contain acronyms which aren't properly handled by Rails' `#humanize` method.

I'm interested in reckons (especially from @rboulton) on adding `last_update` to the list of requested parameters. I've used this because it's available for most of the new formats whereas `public_timestamp` isn't populated. It might be better to stick with `public_timestamp` and ensure the timestamps are properly populated by the publishing apps that register the content with Rummager.

## Screenshots

Here's a couple of examples of metadata which is added by this change.

screen shot 2015-01-09 at 11 40 15

screen shot 2015-01-09 at 11 39 15

### DIFF
--- a/app/presenters/non_edition_result.rb
+++ b/app/presenters/non_edition_result.rb
@@ -1,0 +1,58 @@
+class NonEditionResult < SearchResult
+  include ERB::Util
+
+  def to_hash
+    super.merge({
+      metadata: metadata,
+      metadata_any?: metadata.any?,
+    })
+  end
+
+  def format
+    result["document_type"]
+  end
+
+  def metadata
+    data = [
+      formatted_public_timestamp,
+      display_type,
+      organisations,
+    ]
+
+    data.select(&:present?)
+  end
+
+private
+  def formatted_public_timestamp
+    public_timestamp && public_timestamp.to_date.strftime("%e %B %Y")
+  end
+
+  def public_timestamp
+    result["public_timestamp"] || result["last_update"]
+  end
+
+  def display_type
+    overrides = {
+      "aaib_report" => "AAIB report",
+      "cma_case" => "CMA case",
+      "maib_report" => "MAIB report",
+      "raib_report" => "RAIB report",
+    }
+
+    overrides.fetch(format, format.humanize)
+  end
+
+  def organisations
+    orgs = Array(result["organisations"]).reject(&:blank?)
+
+    org_titles = orgs.map { |org|
+      if org["acronym"] && org["acronym"] != org["title"]
+        "<abbr title='#{h(org["title"])}'>#{h(org["acronym"])}</abbr>"
+      else
+        org["title"] || org["slug"]
+      end
+    }
+
+    org_titles.join(", ")
+  end
+end
--- a/app/models/search_parameters.rb
+++ b/app/models/search_parameters.rb
@@ -51,6 +51,7 @@ class SearchParameters
         display_type
         document_series
         format
+        last_update
         link
         organisations
         organisation_state
--- a/app/presenters/search_results_presenter.rb
+++ b/app/presenters/search_results_presenter.rb
@@ -51,12 +51,16 @@ class SearchResultsPresenter
   end
 
   def results
-    search_response["results"].map do |result|
-      if result["index"] == "government"
-        GovernmentResult.new(search_parameters, result)
-      else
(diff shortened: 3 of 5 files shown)


rboulton:
I'm not sure exactly what the semantics of last_update are - is it the latest update of any type?  Does it correspond to the "updated_at" field in content store? https://github.com/alphagov/content-store/blob/master/doc/content_item_fields.md#updated_at

I'm okay with this in theory, but haven't looked at the code in detail yet.

app/presenters/non_edition_result.rb
```diff
@@ -0,0 +1,60 @@
...
+private
+  def last_update
+    if result["last_update"]
+      result["last_update"].to_date.strftime("%e %B %Y")
+    end
+  end
+
+  def display_type
+    case format
+    when "aaib_report"
```
rboulton: I think this would be more readable as a hash lookup:

```ruby
  FORMAT_NAMES = {
    "aaib_report" => "AAIB report",
    "cma_case" => "CMA case",
  }

  FORMAT_NAMES.fetch(format, format.humanize)
```

(Or with a bit more logic if you want to avoid the call to humanize in the default case.)

app/presenters/non_edition_result.rb
```diff
@@ -0,0 +1,60 @@
...
+      "MAIB report"
+    when "raib_report"
+      "RAIB report"
+    else
+      format.humanize
+    end
+  end
+
+  def organisations
+    orgs = Array(result["organisations"]).reject(&:blank?)
```
rboulton: For my education: I assume the call to Array here is to make a "nil" value or missing "organisations" entry become [].  We use things like `[*result["organisations"]]` elsewhere in our codebase.  Is one form better than the other?  As far as I can see, they have equivalent effects.

rboulton:
Added a couple of comments, but the code generally looks fine.

I would much prefer avoiding exposing new fields from rummager if we can avoid it. Particularly if the semantics of the date in public_timestamp would be appropriate to show for these results, it would be better if the publishers just started populating that field.

Relevant to this: when we move over to index documents from the content store, my plan was that the content store's 'public_updated_at' field will be used to populate the public_timestamp field (or we may even rename the field in rummager at that point).

If that's going to delay you massively, I'd be okay with using `last_update` for now, but would want to know which formats it actually exists for (is it just documents in finders?  Or all things from specialist-publisher?)

app/presenters/non_edition_result.rb
edds: Assuming they have equivalent effects I far prefer `Array(thing)` over `[*thing]`. Trying to google or make sense of `Array(..)` is much easier than the second.

evilstreak:
Interestingly we have three timestamps defined in Rummager's [core schema](https://github.com/alphagov/rummager/blob/master/config/schema/default/core.json#L42-L56). Looking through the history of Rummager and Specialist Publisher, it seems that `last_update` is intended to have the same semantics as `public_timestamp`.

I'll get Specialist Publisher to send the same value for both fields for now, and then we can remove `last_update` from the schema at some point in the future.

evilstreak:
After discussion with @rboulton I've changed this to prefer `public_timestamp` and fall back to `last_update` so that we can merge the changes now without having to republish the 10k+ documents before they'll get metadata.

app/presenters/non_edition_result.rb
evilstreak: Updated

app/presenters/non_edition_result.rb
evilstreak: They have the same effect. [`Kernel#Array`](http://ruby-doc.org/core-1.9.3/Kernel.html#method-i-Array) is just a bit more explicit about the intention, and as Edd points out, a little more searchable.

PR #725, opened 2015-01-12
[URGENT]Add service unavailability callout

jackscotti:
Make a SORN and Vehicle Tax online services will be
unavailable from 11:15pm on 12 January to 8am on 13 January

This is a hack but the deploy slot is at 16.30 and anyway we will need to revert the changes tomorrow morning.

https://www.agileplannerapp.com/boards/105200/cards/9111

### DIFF
--- a/app/assets/stylesheets/views/_make-a-sorn.scss
+++ b/app/assets/stylesheets/views/_make-a-sorn.scss
@@ -18,17 +18,37 @@ body.full-width .make-a-sorn header.page-header div {
 
 
 
+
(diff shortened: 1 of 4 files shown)


edds:
You will probably also need to copy over `icon-important-2x.png` and `icon-important.png` from static for the asset urls to work.

jackscotti:
Do I? It works on preview/

edds:
The images are being provided by the frontend_toolkit. LGTM.

PR #728, opened 2015-01-13
Read downtime from the artefact details

vinayvinay:
www.agileplannerapp.com/boards/173808/cards/8806

the artefact has an associated downtime object in the JSON representation that indicates when a transaction will be unavailable. the artefact downtime details are now moved under the details hash in the artefact JSON.

### DIFF
--- a/app/presenters/publication_presenter.rb
+++ b/app/presenters/publication_presenter.rb
@@ -11,13 +11,13 @@ class PublicationPresenter
   end
 
   PASS_THROUGH_KEYS = [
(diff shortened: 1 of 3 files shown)


edds:
Does this need anything else merging first or is it safe to go out?

alext:
@edds This is fine to go out as it's a no-op until the data is present in content API.

dsingleton:
(I think the story link might be wrong, it's a stats redirect link)

fofr:
(Updated story link)

PR #729, opened 2015-01-13
Update to work with slimmer wrapper layout

edds:
Now that by default the main element takes up 2/3rds of the page and the
related links take up 1/3rd there is extra padding which isn't needed on
a collection of elements. Remove this extra padding so the pages don't
change.

This is the change needed to go with https://github.com/alphagov/static/pull/530


dsingleton:
Can't see anything worrying in the changes, safe to ship if it looks right :+1:

edds:
Got the thumbs up, merging so it can be deployed.

PR #733, opened 2015-01-22
Add a Content-Security-Policy reporting header

alexmuller:
**This pull request is open for feedback only, not to be merged (yet).**

Add `Content-Security-Policy-Report-Only` to the homepage in development only.

Whitelist:

- static and assets hostnames for loading anything
- Google Analytics for loading JavaScript
- Inline JavaScript and styles (we have a lot of existing inline JavaScript and inline styles loaded with jQuery)
- Reporting to the event-store app

There is still one report firing for me that I haven't been able to get rid of:

> Refused to load plugin data from ''

This is related to the accessible video player, which tests Flash on every pageload in JavaScript. I don't think there's a way to allow this behaviour.

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -1,0 +1,10 @@
+# This file is overwritten during deployment
+#
+# Boolean indicating whether or not to enable a Content-Security-Policy
+# header in this application.
+#
+if Rails.env.development? or Rails.env.test?
+  Frontend::Application.config.enable_csp = true
+else
+  Frontend::Application.config.enable_csp = false
+end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,6 +40,28 @@ protected
     end
   end
 
+  def set_content_security_policy
+    return unless Frontend::Application.config.enable_csp
+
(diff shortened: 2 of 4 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,10 @@
+# This file is overwritten during deployment
```
bradleywright: If it's overwritten during deployment do we need the logic fork?

bradleywright:
This looks good. When we move to CSP2 we probably want to do some stuff with Slimmer or components to allow apps to send their individual inline hashes to the header generation method.

config/initializers/content_security_policy.rb
alexmuller: I was wondering that, but I was nervous about setting it to true here and then having it be enabled in production if the file wasn't overwritten during deployment. Happy to take it out if you want.

alexmuller:
That sounds sensible. In alphagov/govuk_template#128 there's also discussion of the template providing a list of inline hashes that consumers can use to set a response header.

bradleywright:
So I think we should merge this and run it on preview for a while. Any objections @edds @dsingleton ?

edds:
We can always back it out. And by default it will be a no-op anyway in production.

dsingleton:
Retro-active :+1:

PR #735, opened 2015-01-27
Use component for beta-label

edds:
Use the component for all the beta labels. This means we can remove the
bespoke styling for the old beta label.


dsingleton:
LGTM. Is the Slimmer upgrade test changes worth documenting somewhere for when someone hits it in another app?

edds:
The changelog in slimmer does explain the changes which should give people enough information to assess the changes for each app.

dsingleton:
:+1: then

PR #736, opened 2015-01-28
Remove unused Cucumber code

jamiecobbett:
The last Cucumber features seem to have been removed here:
fee7e5298d5e9abaa38b1463801a04a25b229ccd


alext:
Test failure.  It's still trying to invoke cucumber - https://github.com/alphagov/frontend/blob/master/lib/tasks/test.rake#L2

It's also worth removing the cucumber rake tasks (`lib/tasks/cucumber.rake`)

jamiecobbett:
Oops, I ran `bundle exec rake`, thinking that would run the whole suite. I also missed some other Cucumber stuff, so I've removed that too.

alext:
Hmmm... `bundle exec rake` _should_ run the whole suite.  I might put in another PR tomorrow.

dsingleton:
I think this looks OK to merge - and support another PR for fixing the rake issue.

Any objections to merging this?

PR #740, opened 2015-01-29
Merge all the common template parts into a base

edds:
Lots of the templates use the same base template. Merge them into a
single base page which everything can then extend from. This would make
updating global page layouts much easier.

This will let us iterate aspects of all pages in one go. Like using a component
for all the titles.

### DIFF
--- a/app/views/root/_base_page.html.erb
+++ b/app/views/root/_base_page.html.erb
@@ -1,0 +1,25 @@
+<main id="content" role="main" class="<%= main_class if local_assigns.include?(:main_class) %>">
+  <header class="page-header">
+    <div>
+      <h1>
+        <% if local_assigns.include?(:context) %><span><%= context %></span><% end %>
+        <%= title %>
+      </h1>
+    </div>
+  </header>
+  <div class="article-container group">
+    <%= render :partial => 'beta_label' if publication.in_beta %>
+
+    <article role="article" class="group">
+      <div class="inner">
+        <%= yield %>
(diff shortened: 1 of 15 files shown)


app/views/root/_base_page.html.erb
```diff
@@ -0,0 +1,25 @@
...
+      <div class="inner">
+        <%= yield %>
+      </div>
+    </article>
+    <% json_link ||= publication_api_path(publication, :edition => edition)  %>
+    <%= render 'publication_metadata', :publication => publication, :api_links => { 'application/json' => json_link } %>
+  </div>
+</main>
+
+<% if !local_assigns.include?(:related_items) || related_items != false %>
```
dsingleton: This is quite a confusing conditional, it's hard to understand whats it's doing at a glance, could you make it a bit easier to understand?

It looks like it should be a guard against `:related_items` being set, but doesn't seem to be?

dsingleton:
Weird `if` aside this looks good, and for other reviewers it's best to check the [diff with whitespace ignored](https://github.com/alphagov/frontend/pull/740/files?w=1), much more readable :+1:

app/views/root/_base_page.html.erb
edds: Fair point. I have updated to use a `fetch` with a default rather than two negatives. Is that better?

dsingleton:
`show_related_items` reads a lot better, LGTM.

PR #741, opened 2015-01-29
Cache /random for 5 seconds instead of 0 seconds

issyl0:
- Following on from the 130 requests in 5 minutes it got at its peak
  tonight, Brad suggested we add a small cache to discourage bots. Five
  seconds is small enough not to affect users being shown a different
  page every time.

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -46,9 +46,9 @@ class RootController < ApplicationController
   def random_page
     # Redirect to a random GOV.UK page, for fun.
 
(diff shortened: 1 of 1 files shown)


ajlanghorn:
This built as Job 592 in CI - https://ci-new.alphagov.co.uk/job/govuk_frontend_branches/592 - and passed fine.

samjsharpe:
I :heart: this.

bradleywright:
This could do with a test - I'd expect that changing explicit caching behaviour would cause a test to fail. We can work on that later though.

PR #742, opened 2015-01-30
Use the Rummager API instead of the Content API for /random

issyl0:
- Rummager is faster than the Content API and is better designed for
  this purpose of loading non-cached pages in quick succession.

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -50,15 +50,18 @@ class RootController < ApplicationController
     # new content.
     expires_in 5.seconds
 
-    base_url = "https://www.gov.uk/api/artefacts.json"
-    total_pages = JSON.parse(RestClient.get(base_url).body)['pages']
+    base_url = Plek.new.find('search') + "/unified_search.json"
+    total_documents = JSON.parse(RestClient.get("#{base_url}?count=0"))['total']
 
-    random_page_number = Random.rand(1..total_pages)
+    random_page_number = Random.rand(0..total_documents-1)
 
-    random_page = RestClient.get("#{base_url}?page=#{random_page_number}")
-    result = JSON.parse(random_page)['results'].shuffle.first['web_url']
+    random_document = RestClient.get("#{base_url}?count=1&fields=link&start=#{random_page_number}")
+    result = JSON.parse(random_document)['results'][0]['link']
 
-    redirect_to result.gsub("https://www.gov.uk","#{Plek.new.website_root}")
+    # Some paths don't have leading slashes, so add them.
+    result = "/#{result}" unless result.starts_with?("/")
+
+    redirect_to Plek.new.website_root + result
   end
 
   def jobsearch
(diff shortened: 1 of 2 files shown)


issyl0:
Thanks to @rboulton for helping with this. :sunny:

app/controllers/root_controller.rb
```diff
@@ -50,15 +50,17 @@ def random_page
...
 
-    random_page = RestClient.get("#{base_url}?page=#{random_page_number}")
-    result = JSON.parse(random_page)['results'].shuffle.first['web_url']
+    random_page = RestClient.get("#{base_url}?count=1&fields=link&start=#{random_page_number}")
+    result = JSON.parse(random_page)['results'][0]['link']
 
-    redirect_to result.gsub("https://www.gov.uk","#{Plek.new.website_root}")
+    puts Plek.new.website_root + result
+
+    redirect_to Plek.new.website_root + result
```
bradleywright: Do we need to do this? Redirecting to `/thing` would work fine too.

app/controllers/root_controller.rb
```diff
@@ -50,15 +50,17 @@ def random_page
...
 
     random_page_number = Random.rand(1..total_pages)
 
-    random_page = RestClient.get("#{base_url}?page=#{random_page_number}")
-    result = JSON.parse(random_page)['results'].shuffle.first['web_url']
+    random_page = RestClient.get("#{base_url}?count=1&fields=link&start=#{random_page_number}")
+    result = JSON.parse(random_page)['results'][0]['link']
 
-    redirect_to result.gsub("https://www.gov.uk","#{Plek.new.website_root}")
+    puts Plek.new.website_root + result
```
bradleywright: Rogue `puts` here - should be removed before merging.

issyl0: Oops. Gone now.

app/controllers/root_controller.rb
```diff
@@ -50,15 +50,17 @@ def random_page
     # new content.
     expires_in 5.seconds
 
-    base_url = "https://www.gov.uk/api/artefacts.json"
-    total_pages = JSON.parse(RestClient.get(base_url).body)['pages']
+    base_url = "https://www.gov.uk/api/search.json"
+    total_pages = JSON.parse(RestClient.get("#{base_url}?count=0"))['total']
```
tommyp: You might want to change the variable names as this isn't the total pages, it's the total documents in Rummager.

app/controllers/root_controller.rb
```diff
@@ -50,15 +50,15 @@ def random_page
...
 
     random_page_number = Random.rand(1..total_pages)
```
rboulton: The "start" parameter in rummager is 0-based, so you want a range from (0..total_pages - 1) here.

app/controllers/root_controller.rb
issyl0: Sorted this out with @alext and @fatbusinessman.

app/controllers/root_controller.rb
issyl0: Done.

PR #743, opened 2015-01-30
Remove article tags and reliance on them

edds:
Use article tags in a more semantic manor. There are two parts to this.
In the first instance I have removed a whole collection of questionable
article tags. The category for questionable was do they contain a
heading. The second was add a `content-block` to any article tags which
I couldn't make an easy decision on. Adding this class will enable us to
keep the styling when we remove the article tag from the CSS selector.

This is reliant on this pull request being merged and deployed first: https://github.com/alphagov/static/pull/540

### DIFF
--- a/app/assets/stylesheets/views/_campaigns.scss
+++ b/app/assets/stylesheets/views/_campaigns.scss
@@ -124,7 +124,7 @@ main.campaign {
     }
   }
 
-  article.campaign {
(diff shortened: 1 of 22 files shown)


dsingleton:
Pending deployment of the static change, LGTM.

dsingleton:
Static deployed.

PR #746, opened 2015-02-03
Increase the displayed number of Imminence results from 5 to 10

issyl0:
- Following a request from Liz Lutgendorff over the firebreak, and a
  story in the Core team's backlog.


issyl0:
:sunny:

PR #747, opened 2015-02-05
Campaign page iteration

russellthorn:
• Remove breadcrumb - only ever has home link in it
• update banner to not use 3 background images and only use large banner image in markup
• move banner image to top as this is simply that a branding, banner image
• make banner inline and responsive, no image cropping, therefore no loss of content
• Center h1 and increase font size to make heading more prominent
• Increase width of content to max width of 75% and center. Makes better use of space and layout.
• For desktop right align org logo (when present) to flow within the content. This fits in with the 'F' pattern reading structure of content promoted by our content designers as discussed Elena Findley-de Regt

Before and after images for mobile, tablet and desktop -
mobile before
screen shot 2015-02-05 at 12 02 00
mobile after
screen shot 2015-02-05 at 12 01 09

tablet before
screen shot 2015-02-05 at 12 02 20
tablet after
screen shot 2015-02-05 at 12 00 48

desktop before
screen shot 2015-02-05 at 12 01 31
desktop after
screen shot 2015-02-05 at 11 58 54

### DIFF
--- a/app/views/root/campaign.html.erb
+++ b/app/views/root/campaign.html.erb
@@ -1,51 +1,24 @@
 <% content_for :title, page_title(@publication) %>
+<div class="header-context group"><!-- deliberately empty, which removes the breadcrumb from the head of the page --></div>
+
 <main id="content" role="main" class="group campaign">
   <section id="landing">
-    <h1><%= @publication.title %></h1>
-    <% if @publication.details['organisation'] %>
-    <div class="organisation">
-      <a href="<%= organisation_url(@publication) %>" class="organisation-logo organisation-logo-<%= organisation_crest(@publication) %> <%= organisation_brand_colour(@publication) %>"><%= formatted_organisation_name(@publication) %></a>
+    <div class="campaign-image">
+      <img src="<%= campaign_image_url(@publication, 'large') %>">
     </div>
-    <% end %>
+
+    <h1><%= @publication.title %></h1>
 
     <div class="campaign-beta-label">
       <%= render :partial => 'beta_label' if @publication.in_beta %>
     </div>
 
-
-    <div class="campaign-image">
-      <span class="image-scope">
-        <style scoped>
-          /* fall-back styles for browsers without media query support (IE6-8) */
-          #banner {
-            background-image: url('<%= campaign_image_url(@publication, 'large') %>');
-          }
-
-          /* small image for mobile */
-          @media (min-width: 0px) {
-            #banner {
-              background-image: url('<%= campaign_image_url(@publication, 'small') %>');
-            }
-          }
-
-          /* medium image for tablets */
-          @media (min-width: 500px) {
-            #banner {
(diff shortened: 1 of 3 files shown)


edds:
Can you include screenshots of how this will look at desktop and mobile please.

russellthorn:
I will once I've fixed the tests.

app/views/root/campaign.html.erb
```diff
@@ -1,52 +1,25 @@
...
-            #banner {
-              background-image: url('<%= campaign_image_url(@publication, 'large') %>');
-            }
-          }
-        </style>
-        <span id="banner" role="presentation"></span>
-      </span>
-    </div>
-    <div class="content-block campaign">
+    <article class="campaign">
```
edds: I don't believe this content should be wrapped in an article tag. @tombye is going to write some fuller description as to what is appropriate for an article tag but until then a basic rule is does the content have both a header and a text block. Have you used it just to pull in the standard text formatting? You should be able to get the formatting by using the `content-block` class on any element, which was introduced with [this pull request on static](https://github.com/alphagov/static/pull/540)

app/views/root/campaign.html.erb
```diff
@@ -1,52 +1,25 @@
 <% content_for :title, page_title(@publication) %>
+<div class="header-context group"><!-- deliberately empty --></div>
```
edds: Would it be possible to extend this `deliberately empty` comment to explain why it is deliberately empty please.

app/views/root/campaign.html.erb
```diff
@@ -1,52 +1,25 @@
...
+
 <main id="content" role="main" class="group campaign">
   <section id="landing">
-    <h1><%= @publication.title %></h1>
-    <% if @publication.details['organisation'] %>
-    <div class="organisation">
-      <a href="<%= organisation_url(@publication) %>" class="organisation-logo organisation-logo-<%= organisation_crest(@publication) %> <%= organisation_brand_colour(@publication) %>"><%= formatted_organisation_name(@publication) %></a>
+    <div class="campaign-image">
+      <img src="<%= campaign_image_url(@publication, 'large') %>" alt="Banner">
```
edds: I don't think this alternative text adds anything useful for assistive technology. Unless we are going to get publishers to add in dynamic alternative text I would be inclined to just not have an `alt` attribute.

edds:
It would be good to take this opportunity to remove some of the redundant styles from the `campaigns.scss` file. That file is only currently targeting the page you have modified and the `capaign/uk_welcomes` page. So things like the `campaign-links` selector can be deleted.

russellthorn:
Hi @edds I have fixed the failing tests, made changes and squashed the commits. Are you happy to merge this now?

edds:
There is a collection of IE specific styles in the stylesheets for these pages. Just checking you have looked at it in the relevant IE versions to ensure it still works?

russellthorn:
I have checked these changes in IE6-8 and all works fine.

PR #748, opened 2015-02-06
Remove the tax disc V11 form image hint

issyl0:
- This was out of date anyway compared to the new-style form, and the
  reference number on the form is visible enough not to need an image
  hint.

Before:

screen shot 2015-02-06 at 11 47 11

After:

screen shot 2015-02-06 at 11 45 48


fofr:
Can we remove the asset too, or is it being used elsewhere?

issyl0:
@fofr: I was about to do that, but it's being used in https://github.com/alphagov/frontend/blob/master/app/views/root/make-a-sorn.html.erb#L40. As that's the same, outdated image, I could remove that?

issyl0:
@fofr: Removed both.

PR #750, opened 2015-02-10
Change /random cache from private to public

dcarley:
#### Change /random cache from private to public

By default `expires_in` will generate a `Cache-Control` header with the
`private` instruction, which allows the client to cache the response but not
any intermediate proxies such as our Varnish nodes or CDN:

- http://apidock.com/rails/ActionController/ConditionalGet/expires_in
- https://github.com/alphagov/cdn-acceptance-tests/blob/e4ee66728c99cedb8b510c7736df35b4ee5b6bd4/cdn_nocache_test.go#L71-L81

This means that we would probably see the same amount of traffic at and
within origin, especially from bots. Change the instruction to `public`
which will allow the response to be cached by those layers and soak up any
brief increases in traffic.

I've moved the comment about why we're doing this, minus the specific value,
to an additional test which asserts this behaviour is happening. This should
reduce the risk of the code diverging in behaviour from the comments. The
move of where we call `expires_in` is purely aesthetic.

#### Refactor random_page test to use setup

By separating the stubbing from the request and assertions it's a bit easier
to see what's going on and allows us to provide a short sentence to describe
what the test is observing.

I've also modified some whitespace related to line wrapping.

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -46,10 +46,6 @@ class RootController < ApplicationController
   def random_page
     # Redirect to a random GOV.UK page, for fun.
 
-    # Cache the page for 5 seconds - this won't affect users seeing
-    # new content.
-    expires_in 5.seconds
-
     base_url = Plek.new.find('search') + "/unified_search.json"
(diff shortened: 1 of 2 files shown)


issyl0:
Thanks!

PR #751, opened 2015-02-12
Remove services and information heading from homepage

rboulton:
The "Services and information" heading above the links to mainstream browse
isn't particularly illuminating, and makes the mainstream
browse entry points appear further down the page than they need to.

We also use the phrase "Services and Information" on (some) organisation
pages to mean something quite different (a link to a set of subtopics
containing documents related to the organisation), so it's good to avoid
labelling this section on the homepage in the same way.

Before:

screen shot 2015-02-12 at 18 57 27

After:

screen shot 2015-02-12 at 18 57 48

Since this is a highly visible change, I would like this reviewed and
approved by the following before merging:

 - [x] A front end developer
 - [x] Product (Ben Andrews?)
 - [x] Design (Mark Hurrell?)
 - [x] User research (Charlotte Clancy?)

### DIFF
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -49,7 +49,7 @@
       <section class="services-block" aria-labelledby="services-and-information-label" id="services-and-information">
         <div class="inner-block floated-children">
           <div class="two-column-heading">
-            <h2 id="services-and-information-label">Services and information</h2>
+            <h2 id="services-and-information-label" class="visuallyhidden">Services and information</h2>
           </div>
           <div class="categories-lists">
             <ol class="categories-list">


mrandrews:
@rboulton Looks good. Go for it.

app/views/root/index.html.erb
```diff
@@ -48,9 +48,6 @@
 
       <section class="services-block" aria-labelledby="services-and-information-label" id="services-and-information">
         <div class="inner-block floated-children">
-          <div class="two-column-heading">
-            <h2 id="services-and-information-label">Services and information</h2>
```
bradleywright: Now that we've removed the `h2` we're skipping header levels, which we shouldn't do for accessibility.

So we have a few options:

1. Make the following `h3`s into `h2`s
2. Make the `h2` visually hidden so that it's still picked up by screen readers

Given the original reasoning behind the PR is mainly visual I'd suggest doing 2 and updating the text a bit so we avoid the overloading of the `Services and information` term.

fofr: There is a precedent for 2), in the section directly below there is a hidden "Deparments and policy" heading:
`<h2 id="departments-and-policy-label" class="visuallyhidden">Departments and&nbsp;policy</h2>`

bradleywright: Oh and another thing, if we delete it we need to remove the [`aria-labelledby` attribute on line 49](https://github.com/alphagov/frontend/pull/751/files#diff-ffd72ca7720288aa96290dd748c1dfe1R49).

rboulton: I think making the `h2` visually hidden makes a lot of sense, thanks.  I'll gather feedback on what a good improvement to the working might be.

rboulton: Had a quick offline chat with John Turnbull, Charlotte Clancy and Mark Hurrell - we couldn't come up with anything which is an improvement, and it's not a terrible description of the following links, so we're going to keep the wording as it is, and just visuallyhide it.

PR #752, opened 2015-02-13
Fix Last updated on Done page

russellthorn:
Last updated paragraph below the send feedback button on pages like https://www.gov.uk/done/tax-disc breaks because the form above is floated with a width of 95%. Removing this float and width fixes this layout issue.

@benilovj can you remember why the float and 95% width was added in the first instance and do you have any opinion on me removing them?

https://www.pivotaltracker.com/story/show/87918506

Before -
screen shot 2015-02-13 at 13 44 22

After -
screen shot 2015-02-13 at 15 19 12

### DIFF
--- a/app/assets/stylesheets/views/_completed-transaction.scss
+++ b/app/assets/stylesheets/views/_completed-transaction.scss
@@ -1,8 +1,6 @@
 .transaction-done {
   form {
     @include core-19;
-    float: left;
-    width: 95%;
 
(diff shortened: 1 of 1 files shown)


benilovj:
As far as i remember, there isn't a particular reason why those were there (it was copied from somewhere else) - I'm happy for them to be removed.

edds:
:+1:

PR #754, opened 2015-02-16
Fix empty tag capmaign pages

russellthorn:
Organisation HTML was being included in the page even for campaigns without an organisation, showing a blank organisation logo and link with coloured bar (see https://www.gov.uk/yourvotematters).

To fix this we added a `has_organisation?` method which checks if any of the organisation values are present.

It's still possible to get weird HTML included on the page, if the organisation details contain some values but not all of them. We've spoken to the content team and they're happy to ensure organisationless campaigns have no values set for organisation.

Before (orange circle and arrow added to highlight issue) -
screen-shot-2015-02-16-at-12 52 42

After -
screen shot 2015-02-16 at 12 53 03

### DIFF
--- a/app/helpers/campaign_helper.rb
+++ b/app/helpers/campaign_helper.rb
@@ -22,6 +22,12 @@ module CampaignHelper
     organisation_attributes(publication)["brand_colour"]
   end
 
+  def has_organisation?(publication)
+    organisation_attributes(publication).any? { |key, value|
+      value.present?
+    }
+  end
+
 private
   def organisation_attributes(publication)
     publication.details.fetch("organisation", {})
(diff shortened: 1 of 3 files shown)


edds:
The tests have failed. Is it worth adding a test to ensure that the new method of checking for orgs works?

bradleywright:
Tests have failed because you've refactored the `organisation_attr` method - might want to grep for instances of that.

Also worth adding a test, as @edds says.

russellthorn:
Test added and test failure fixed.

edds:
:+1: Good changes. Good commit messages. Thank you.

PR #756, opened 2015-02-19
2i follow-up to homepage vehicle tax promo changes (#751)

benilovj:
The changes have been 2i'ed by John Oates.


evilstreak:
For the record, this is a follow up to #755 not #751.

PR #759, opened 2015-02-24
Changing Education category description to include apprenticeships

rooreynolds:
'Apprenticeships' is frequently searched from the homepage (in the top 10 searches from the GOV.UK home page for the last month)

screen_shot_2015-02-24_at_09_37_34

[zendesk #942523]

### DIFF
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -85,7 +85,7 @@
               </li>
               <li>
                 <h3><a href="/browse/education">Education and learning</a></h3>
-                <p>Includes student loans and admissions</p>
+                <p>Includes student loans, admissions and apprenticeships</p>
               </li>
               <li>
                 <h3><a href="/browse/employing-people">Employing people</a></h3>


fatbusinessman:
Are there other places that this description will need updating? Panopticon certainly has a record of section tags; I don’t know whether [collections](https://github.com/alphagov/collections) gets all its information through the content API or whether it adds anything of its own.

rboulton:
I'm pretty sure that collections gets its information (for mainstream browse) from content API.  It gets some information for other things it serves from the content store and from rummager and from whitehall's API, but doesn't have hardcoded data of its own.

fatbusinessman:
Great: thanks, @rboulton. I think this just leaves Panopticon that potentially needs changing, then.

rooreynolds:
The description text provided for https://www.gov.uk/browse/education in panopticon is not used on the homepage or on the browse page. It doesn't seem to be used anywhere so far.

I'll leave the text in panopticon alone for now, but perhaps in the future the homepage could pull the relevant description from panopticon rather than being hardcoded(!)

fatbusinessman:
:+1:

rboulton:
@rooreynolds Yes - the top level descriptions for browse pages aren't taken from panopticon/content-api currently - the second level ones are, though.

This is closely related to stuff the "finding stuff" team is thinking about - while nothing is likely to change massively soon, we're looking into how the mainstream browse areas should be defined.  Also, there's work 90% done (but not being worked on at present - it's a hangover from search+browse work in december) to move the ownership and definition of mainstream browse pages from panopticon into collections-publisher (and then into content store).

PR #760, opened 2015-02-26
Upgrade Ruby to version 2.1.5 as 1.9.3 is deprecated

issyl0:


evilstreak:
Closing until we can fix the test failures.

issyl0:
Thanks! I meant to do this last night but forgot all about it.

benilovj:
@evilstreak @issyl0 @alext

I've gotten the build to green by:

- upgrading `sass` and `sass-rails` to get asset compilation working
- introducing @alext's `invalid_utt8_rejector` gem to catch malformed UTF8 paths, meaning we shouldn't need the failing UTF8 tests anymore

bradleywright:
Since Ruby 2.2.1 is on the servers now should we go to that?

alext:
> Since Ruby 2.2.1 is on the servers now should we go to that?

Actually, 2.2.2 would be a better target (rolled out today).

benilovj:
@bradleywright @alext @evilstreak Rails 3.2.X has some [unresolvable warnings with ruby 2.2.2](https://github.com/rails/rails/pull/18160), so even though the tests pass, I am reluctant to upgrade ruby all the way to 2.2.2 until we've migrated this app to Rails 4.

evilstreak:
:+1: Sounds sensible to me

PR #761, opened 2015-02-26
Send analytics events through the GOV.UK Analytics API

benilovj:
https://www.pivotaltracker.com/n/projects/1261204/stories/88175032

This change wraps the GA event in the [newly introduced](alphagov/static#549)
analytics API. This will aid in the migration from GA Classic to Universal Analytics, while
remaining functionally equivalent during the migration.

/cc @fofr


benilovj:
@fofr this should be ready to merge

PR #763, opened 2015-03-03
Replace the GOV.UK blogs promo with the tax-free savings promo

issyl0:
This needs _deploying_ on Wednesday 4th March.


benilovj:
I think we need a different image:

* the others promo are all JPGs
* the other promo images are 608 x 405 (so that they look good on an iPad)

bradleywright:
The JPG thing is mostly because we can compress JPGs down a lot further than PNGs.

app/views/root/index.html.erb
```diff
@@ -162,10 +162,10 @@
             </div>
             <div class="promo-image">
               <div class="promo-content">
-                <a href="https://www.blog.gov.uk"><%= image_tag 'homepage/govuk-blogs.jpg', alt: 'GOV.UK blogs' %></a>
-                <h3>GOV.UK blogs</h3>
+                <a href="/taxfreesavings"><%= image_tag 'homepage/tax-free-savings.png', alt: 'Tax-free savings' %></a>
+                <h3>Tax-free savings</h3>
                 <p>
-                  <a href="https://www.blog.gov.uk">Search the list</a> of GOV.UK blogs, and find one to match your interest.
+                  More people will be able to get tax-free interest on their savings from 6 April - <a href="/taxfreesavings">find out if you'll qualify</a>.
```
benilovj: pedantic nitpick - would be good to use [smart quotes](http://smartquotesforsmartpeople.com/) for `you'll`

bradleywright:
Actually looking at the image it might be better as a GIF (it's vector art, not a photo). PNG would also be suitable - worth checking the relative file size of both.

edds:
If we are loosing the last reference to `homepage/govuk-blogs.jpg` can we also remove the old asset as part of this please.

app/views/root/index.html.erb
issyl0: Yup - amended. And there was me getting pedantic about the hyphen in the text to insert instead of a proper dash...

issyl0:
@edds: Done.

benilovj:
closing this until we have a bigger image

issyl0:
Closing this until we get a better image...

PR #764, opened 2015-03-03
Replace the GOV.UK blogs promo with the tax-free savings promo

issyl0:
GitHub wouldn't let me reopen https://github.com/alphagov/frontend/pull/763 once I'd force-pushed to the branch with a new, properly sized JPG of the tax-free savings image.


issyl0:
@edds: How does this image look?

edds:
As @bradleywright said in the other pull request I think that you are best of turning this image into a GIF as the file size is slightly smaller without introducing any compression artefacts, then crushing the size down using [ImageOptim](https://imageoptim.com/).

I did this by downloading the original image from your pivotal ticket then opening it in Preview > File > Export. Then if you old down the option key (the one that says alt) when clicking the format dropdown it gives you a larger set of export options which include GIF. Then dragging the newly created file into ImageOptim will reduce the file size by 10%.

The image I created is here:
zero-rated-savings-608x405

issyl0:
Oh, nice. Thanks! I didn't realise that it required another piece of software to do, or about the Option-click thing to select GIF. I've amended the commit to include the GIF instead. Is that all?

edds:
:+1: LGTM.

PR #765, opened 2015-03-06
Remove the words "tax disc" from titles of sections and links

issyl0:
- Tax discs aren't used anymore!
- Remove the words on Make a SORN and Vehicle Tax pages.


issyl0:
:sunny:

PR #766, opened 2015-03-10
Replace the HGV vehicle tax link with "vehicle tax refunds"

issyl0:


fofr:
Change looks good to me. Is there some context about why we are making this change?

issyl0:
@fofr: See the last couple of comments on https://www.pivotaltracker.com/story/show/89789284 and also http://govuk.zendesk.com/tickets/976343.

PR #768, opened 2015-03-13
Bump Slimmer

alicebartlett:
This commit does two things:

1. Support for the proposition header was removed from slimmer v8.0.0 so remove it from the codebase and remove the test for it.
2. Bump gem version.

This work needs to be done in advance of another PR which will make use of further changes to slimmer in this PR:  https://github.com/alphagov/slimmer/pull/121

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -42,7 +42,6 @@ protected
       result_count: result_count,
       format:       "search",
       section:      "search",
-      proposition:  "citizen"
     )
   end
 
--- a/test/functional/search_controller_test.rb
+++ b/test/functional/search_controller_test.rb
@@ -361,7 +361,7 @@ class SearchControllerTest < ActionController::TestCase
     end
   end
 
-  test "should send analytics headers for citizen proposition" do
+  test "should send analytics headers" do
     result = {
       "title" => "title",
       "link" => "/slug",
@@ -372,7 +372,6 @@ class SearchControllerTest < ActionController::TestCase
     get :index, {q: "bob"}
     assert_equal "search",  @response.headers["X-Slimmer-Section"]
     assert_equal "search",  @response.headers["X-Slimmer-Format"]
-    assert_equal "citizen", @response.headers["X-Slimmer-Proposition"]
     assert_equal "1",       @response.headers["X-Slimmer-Result-Count"]
   end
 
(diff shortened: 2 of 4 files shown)


fofr:
This gem bump has a dependency on https://github.com/alphagov/static/pull/558 and shouldn't be merged until that PR has been deployed.

test/functional/search_controller_test.rb
```diff
@@ -361,21 +361,6 @@ def no_results
...
-      "title" => "title",
-      "link" => "/slug",
-      "highlight" => "",
-      "format" => "publication"
-    }
-    stub_results([result], "bob", [], [], total: 1)
-    get :index, {q: "bob"}
-    assert_equal "search",  @response.headers["X-Slimmer-Section"]
-    assert_equal "search",  @response.headers["X-Slimmer-Format"]
-    assert_equal "citizen", @response.headers["X-Slimmer-Proposition"]
```
fofr: Should it be just this line that gets removed?

From what I can see the test is poorly named and was actually testing all analytics headers, the others are still being used.

alicebartlett: Bah, you're right. Thanks for catching this, I'll fix.

app/controllers/search_controller.rb
```diff
@@ -41,8 +41,7 @@ def fill_in_slimmer_headers(result_count)
     set_slimmer_headers(
       result_count: result_count,
       format:       "search",
-      section:      "search",
-      proposition:  "citizen"
+      section:      "search"
```
fatbusinessman: A point of moderate pedantry: it looks like this code [dates back](https://github.com/alphagov/frontend/commit/68d79169a77669e19d92b5f68b66c683d05ffb71) to before we [added trailing commas to our style guide](https://github.com/alphagov/styleguides/commit/4da34b1848772ee06084ef3204601f34615f85d6), so is it worth changing the style while we’re here and keeping the trailing comma after the last line to give us nicer diffs?

alicebartlett: I agree! will change.

alicebartlett:
Hello, I've addressed all changes (thanks @fofr and @fatbusinessman) and http://github.com/alphagov/static#558 has been merged now so this is good to go I think.

fofr:
Ah, already merged. I was going to suggest bumping to `8.1.0`

PR #776, opened 2015-03-19
Update the "are being merged" text on the homepage

rooreynolds:
It's now more accurate to say that they have been merged. Hurrah!

screen_shot_2015-03-19_at_08_28_06


bradleywright:
:+1:

PR #778, opened 2015-03-25
[DO NOT MERGE UNTIL EOB THURSDAY 26] updated homepage promo text and photo with vehicle tax

binaryberry:
Do not merge until Thursday evening


issyl0:
:+1: on your first PR being a change to the homepage! :sunny:

evilstreak:
The image has some pretty visible JPEG artefacts. Is it possible to get a higher quality image to use?

evilstreak:
To follow up on an IRL conversation – we have a higher quality image available, but there are concerns about file size.

I definitely don't have the necessary knowledge to be making calls on trade offs like that, but it would be nice to get some guidance written up to make this less of a back-and-forth judgement call each time we change a promo. A good place for that would be in our internal guidance about promo slots: https://gov-uk.atlassian.net/wiki/display/GOVUK/GOV.UK+homepage

@binaryberry @issyl0 Is that something you two could pick up with the people who generally weigh in on this stuff?

benilovj:
:+1: (apart from the open question on image quality)

binaryberry:
Cancelling this pull request as we are expecting changes to the other promo slots - we will be pushing all the changes at the same time when we receive the content from the content team.

PR #779, opened 2015-03-25
Add history mode note to search results

dsingleton:
When a search result is flagged as `historic` we want to highlight
this to a user browsering search results, so they know that result
may not be up to date and may not reflect the views of the current
government.

`historic` is defined as `political` content published under a
previous government. This only applies to Whitehall Editions
currently.

- [ ] Check against an updated rummager index

### DIFF
--- a/app/assets/stylesheets/views/_search.scss
+++ b/app/assets/stylesheets/views/_search.scss
@@ -285,12 +285,15 @@ main.search {
           margin: 0;
           padding: 0;
 
+          &.historic,
           &.meta {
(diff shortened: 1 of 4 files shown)


rboulton:
I think you'll need to add the "history mode" field to the allowed return fields in rummager - see https://github.com/alphagov/rummager/blob/master/lib/search_parameter_parser.rb#L93 ish

dsingleton:
Yep, this is missing some plumbing. Closing for now.

PR #780, opened 2015-03-26
Add scoped search view

alicebartlett:
Pivotal: https://www.pivotaltracker.com/story/show/88435634

(before)
screen shot 2015-03-26 at 17 30 19

(After)
screen shot 2015-03-26 at 18 27 08

This PR enables searching within a single document format (in the first instance, a manual, but this could be widened to other multi-page formats)

This PR contains (in order of commit (more info on in the commit messages)):
- Technical implementation to respond to `filter_manual[]=manuals/manual-link` with only results from that manual
- Removal bits of interface (filters, search result meta-data about the format/organisation it appears in)
- Format way of displaying the scoped result count and a way to de-scope that count to show all results on GOV.UK
- Wrap the rummager api so that we can hide the multiple requests needed to rummager to build this page

**Dependencies**
This work should not be deployed until this rummager change has gone out: https://github.com/alphagov/rummager/pull/388

**Further work**
1. Need to de-dupe search result titles, so at the moment rummager returns the title of the manual for every search result from a manual, when we are only showing search results from that manual this is not needed
2. We need to add an additional request to rummager for un-scoped results. This will allow us to populate the result count "view n other results from GOV.UK". We'll also show some results from this rummager call as a visual cue to the user that there are other results outside of this scope as per @rivalee's design.
3. There is a bug whereby if you search for `filter_manual[]=manuals/a-manual-that-doesnt-exist` you see a slighly odd looking list of manuals. This could happen if the manual url changes and someone is using an old link, so we should definitely fix this.

**Pull request feedback**

Suggestions on how to fix point 3. in the further work list.
@rivalee I'd like you to check you're happy with this design
@rboulton I'd like you to check the rummager end of this work

### DIFF
--- a/app/assets/javascripts/search.js
+++ b/app/assets/javascripts/search.js
@@ -29,7 +29,6 @@ $(function() {
     $searchResults.on('click', 'a', function(e){
       var $link = $(e.target),
           sublink = '',
-          gaParams = ['_setCustomVar', 21, 'searchPosition', '', 3],
           position, href, startAt;
 
       if($link.closest('ul').hasClass('sections')){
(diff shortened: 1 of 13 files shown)


alicebartlett:
ugh tests failing. giving up on this until tomorrow.

rivalee:
@alicebartlett The design looks good.

PR #781, opened 2015-03-27
Add scoped search view

alicebartlett:
Pivotal: https://www.pivotaltracker.com/story/show/88435634

(before)
screen shot 2015-03-26 at 17 30 19

(After)
screen shot 2015-03-26 at 18 27 08

This PR enables searching within a single document format (in the first instance, a manual, but this could be widened to other multi-page formats)

This PR contains (in order of commit (more info on in the commit messages)):
- Technical implementation to respond to `filter_manual[]=manuals/manual-link` with only results from that manual
- Removal bits of interface (filters, search result meta-data about the format/organisation it appears in)
- Format way of displaying the scoped result count and a way to de-scope that count to show all results on GOV.UK
- Wrap the rummager api so that we can hide the multiple requests needed to rummager to build this page

**Dependencies**
This work should not be deployed until this rummager change has gone out: https://github.com/alphagov/rummager/pull/388

**Further work**
1. Need to de-dupe search result titles, so at the moment rummager returns the title of the manual for every search result from a manual, when we are only showing search results from that manual this is not needed
2. We need to add an additional request to rummager for un-scoped results. This will allow us to populate the result count "view n other results from GOV.UK". We'll also show some results from this rummager call as a visual cue to the user that there are other results outside of this scope as per @rivalee's design.
3. There is a bug whereby if you search for `filter_manual[]=manuals/a-manual-that-doesnt-exist` you see a slighly odd looking list of manuals. This could happen if the manual url changes and someone is using an old link, so we should definitely fix this.

**Pull request feedback**

Suggestions on how to fix point 3. in the further work list.
@rivalee I'd like you to check you're happy with this design
@rboulton I'd like you to check the expectations this piece has of rummager are sensible. There is also an ope question about the behaviour of the app when the manual we're filtering doesn't exist. What should be presented to the user in this case?

### DIFF
--- a/app/presenters/search_results_presenter.rb
+++ b/app/presenters/search_results_presenter.rb
@@ -33,15 +33,19 @@ class SearchResultsPresenter
   end
 
   def filter_fields
-    search_response["facets"].map do |field, value|
-      external = SearchParameters::external_field_name(field)
-      facet_params = search_parameters.filter(external)
-      facet = SearchFacetPresenter.new(value, facet_params)
-      {
-        field: external,
-        field_title: FACET_TITLES.fetch(field, field),
-        options: facet.to_hash,
-      }
+    if is_scoped?
+      []
+    else
+      search_response["facets"].map do |field, value|
+        external = SearchParameters::external_field_name(field)
+        facet_params = search_parameters.filter(external)
+        facet = SearchFacetPresenter.new(value, facet_params)
+        {
+          field: external,
+          field_title: FACET_TITLES.fetch(field, field),
+          options: facet.to_hash,
+        }
+      end
     end
   end
 
@@ -75,7 +79,9 @@ class SearchResultsPresenter
   end
 
   def build_result(result)
-    if result["document_type"] == "group"
+    if is_scoped?
+      ScopedResult.new(search_parameters, result)
+    elsif result["document_type"] == "group"
       GroupResult.new(search_parameters, result)
     elsif result["document_type"] && result["document_type"] != "edition"
       NonEditionResult.new(search_parameters, result)
@@ -119,6 +125,10 @@ class SearchResultsPresenter
     end
   end
 
+  def is_scoped?
+    is_scoped_to_manual?
+  end
+
 private
 
   attr_reader :search_parameters, :search_response
@@ -158,5 +168,9 @@ private
 
   def previous_page_number
     current_page_number - 1
+  end
+
+  def is_scoped_to_manual?
+    search_response["facets"]["manual"].present?
   end
 end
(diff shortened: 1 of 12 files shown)


bradleywright:
Looks great :+1: 

Just discussed with @alicebartlett that I think this is a good way to drive out newer APIs we might want - meet some needs with it and then work back into the stack from there.

app/presenters/search_results_presenter.rb
```diff
@@ -159,4 +177,8 @@ def next_page_number
   def previous_page_number
     current_page_number - 1
   end
+
+  def is_scoped_to_manual?
```
edds: I could be missing something but what ties this method to manuals as oppose to an generic scope? As far as I can tell I think that the contents of this method could easily be moved to the `is_scoped?` method on line 130 and this method removed as it's all generic now?

app/presenters/search_results_presenter.rb
```diff
@@ -119,6 +127,16 @@ def previous_page_label
...
 
+  def is_scoped?
+    search_response[:scope].present?
+  end
+
+  def scope_title
+     if is_scoped_to_manual?
+       search_response[:scope][:title]
+     end
+   end
```
edds: Tiny nitpick but the spacing on this method is 1 space out.

lib/search_api.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def search(search_params)
+    Searcher.new(rummager_api, search_params).call
+  end
+
+private
+  attr_reader :rummager_api
+
+  class Searcher
+      def initialize(api, params)
```
edds: This is indented two spaces too many.

app/presenters/search_results_presenter.rb
```diff
@@ -119,6 +127,16 @@ def previous_page_label
     end
   end
 
+  def is_scoped?
+    search_response[:scope].present?
+  end
+
+  def scope_title
+     if is_scoped_to_manual?
```
edds: This needs updating to just call `is_scoped?`

edds:
Couple of super minor comments but this change looks good to me.

lib/search_api.rb
```diff
@@ -0,0 +1,59 @@
...
+        {}
+      end
+    end
+
+    def rummager_params
+      params.rummager_parameters
+    end
+
+    def scope_object
+      @scope_object ||= api.unified_search(filter_link: scope_object_link).results.first
```
rboulton: It would be good to limit the results returned by rummager (and thus transfer less data and be quicker) here, by adding: `, count: 1, fields: %w{title}`, to just get 1 result, and to just get the title field for it.

rboulton:
One note on efficiency, but other than that this looks good to go to me. :+1:

alicebartlett:
OK thanks all, I've made amendments based on your comments.

PR #782, opened 2015-03-27
Add history mode note to search results

dsingleton:
When a search result is flagged as `historic` we want to highlight
this to a user browsing search results, so they know that result
may not be up to date and may not reflect the views of the current
government.

`historic` is defined as `political` content published under a
previous government. This only applies to Whitehall Editions
currently.

screenshot 2015-03-27 16 59 50

### DIFF
--- a/app/presenters/government_result.rb
+++ b/app/presenters/government_result.rb
@@ -8,7 +8,9 @@ class GovernmentResult < SearchResult
       metadata_any?: metadata.any?,
       sections: sections,
       sections_present?: sections.present?,
-      government: true
+      government: true,
+      historic?: historic?,
+      government_name: government_name,
     })
   end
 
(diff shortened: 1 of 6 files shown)


app/presenters/government_result.rb
```diff
@@ -8,7 +8,9 @@ def to_hash
       metadata_any?: metadata.any?,
       sections: sections,
       sections_present?: sections.present?,
-      government: true
+      government: true,
+      historic?: historic?,
+      government_name: government_name
```
edds: Should probably add a trailing comma on this line to conform to style for cleaner diffs in the future.

edds:
Tiny style point. Other than that LGTM.

rboulton:
LGTM too.

Non merge-blocking comment though: it would be good to check with an analyst (ideally Tara) if we have sufficient monitoring to determine whether users click on such documents more (or less) than others, or if there are any other related things we want to be tracking.

bmwelby:
good point @rboulton, that's definitely going to be important when we've also downweighted history mode content in search so that we can judge whether or not it's working. I'll add something to that story.

PR #785, opened 2015-04-15
Google Content Experiment A/A test

binaryberry:
The A/A test assigns users to cohort "variant_0" or "variant_1". We want to roll it out to a small subset of pages and see if the updated multivariate test framework interferes with analytics.

Before merging:

- [x] push this branch to preview and test
- [x] update the experiment ID with an experiment created on production


benilovj:
@binaryberry and I have tested this on preview with different browsers. This PR has also been replaced with the production experiment ID so it should be ready to go.

PR #786, opened 2015-04-15
Bump govuk_frontend_toolkit to 3.4.1

benilovj:
This is done to not fall too far behind the latest gem version.

The tests all pass, and `rake asset:precompile` runs successfully as well.


benilovj:
I've pushed this branch to preview, and ran a few manual tests (the homepage and a few different formats) with:

- latest Chrome
- latest FF
- FF 3.6
- mobile Safari (iOS8)
- IE6
- IE8

There weren't any noticeable regressions, so this should be ready to go.

PR #787, opened 2015-04-16
Remove therubyracer gem

benilovj:
`therubyracer` was needed for asset precompilation, but since
each ruby app server has `nodejs` installed, it's not needed.


benilovj:
Closing because removing the dependency would break the app outside the VM.

PR #790, opened 2015-04-17
Fix rendering bug

alicebartlett:
This PR fixes a performance issue that is currently blocking deploys to frontend.

Calling

```
{{#results}}
{{>search/_result}}
{{/results}}
```

was causing high CPU usage. This commit moves the iteration inside the
results partial so that the CPU intensive act of loading the partial
with mustache is only done once, not once per result. In order to make
this work with the nested sublists the `results` partial calls itself
when it needs to present a sublist. This is a bit awkward but avoids
having the identical code for a single result in two places.

Below is an analysis of the number of calls made to relevant functions
(mostly mustache related) that this commit reduces.

```
name                                   old  new
<Class::Regexp>#escape                2445  693
Mustache::Generator#ev                 454  130
Mustache::Generator#on_fetch           450  126
Mustache::Generator#str                623  171
Mustache::Parser#compile                13    4
Mustache::Parser#otag                 1559  439
Mustache::Parser#position              663  189
Mustache::Parser#regexp               1095  313
Mustache::Parser#scan_tags             675  190
Mustache::Parser#scan_text             675  190
Mustache::Parser#scan_until_exclusive  675  190
String#==                             2249  734
String#===                            2085  591
String#[]                             2277  725
String#split                          1399  451
StringScanner#pos                     3198  935
StringScanner#scan                    2664  757
StringScanner#scan_until               675  190
StringScanner#skip                    2110  602
```

This problem has arisen due to a fundamental problem with the way that
mustache handles partials (it doesn't cache them meaning each time a
partial is included it parses the template anew which is a CPU
intensive operation). @evilstreak and I will open a PR against the
mustache repo to introduce caching which will fix this problem. A
previous pull request to introduce template caching was rejected due to
memory leak concerns, so we anticipate a bit of back-and-forth with
the maintainers of mustache to get a patch accepted.


bradleywright:
:clap:

fofr:
Awesome sleuthing. :clap:

edds:
:+1:

jennyd:
Well done on finding the cause - I'd really like to hear more about how you investigated this :mag:

PR #793, opened 2015-04-27
Fixes to print link and pagination

benilovj:
On guide and programmes pages, this PR:

- moves the next/previous widget and the print link out of the `content-block` into a `footer`
- converts the next/previous widget to use components
- in the process, fixes the alignment issues with the next/previous buttons

## Before

image

## After

image

/cc @edds


evilstreak:
There are quite a few visual changes here which you haven't talked about, so I'm not sure if they're intentional or not.

- Text for the previous and next links has changed. "Previous" seems like a better option than "Navigate to previous part" – what are the reasons behind updating that?
- Moving the print link and navigation links into the footer has had the effect of reordering these extra bits so that the last updated date is above the print link and navigation links. Is that OK? Is the last updated date part of the content in a way the navigation isn't?
- Using the navigation component has had the effect of losing the horizontal line dividing the main content from the footer. In manuals we use a large amount of whitespace to separate the two. Do we need to either put the horizontal line back or add whitespace here?

benilovj:
Thanks for the feedback, @evilstreak.

> There are quite a few visual changes here which you haven't talked about

I've provided more information in 2785d18

> Text for the previous and next links has changed. "Previous" seems like a better option than "Navigate to previous part" – what are the reasons behind updating that?

Thanks for spotting this, this was a bug. I've since updated the code and the screenshot.

> Moving the print link and navigation links into the footer has had the effect of reordering these extra bits so that the last updated date is above the print link and navigation links. Is that OK? Is the last updated date part of the content in a way the navigation isn't?

This is intentional, I've documented it in the commit comment and has been OK'ed by @wryobservations. This will be revisited at a later stage.

> Using the navigation component has had the effect of losing the horizontal line dividing the main content from the footer. In manuals we use a large amount of whitespace to separate the two. Do we need to either put the horizontal line back or add whitespace here?

That's a good question. It'd be good for a designer (ccing @markhurrell or @rivalee) to make a call on this.

benilovj:
I've gotten some feedback from @futurefabric on the design, closing this PR until I've update the code accordingly.

PR #794, opened 2015-04-28
Display 20 results on search page

tijmenb:
Since the introduction of pagination there is no need to show an
excessive amount of results on the search page.

This commit changes the maximum number of results to 20.

Some tests had to be updated to reflect this, because they either
relied on the default value (they don't anymore) or relied on the fact
that when searching with the default number of results we don't show
the `count` in the URL.

From now on, changing the default number of results to anything but 50
wil *not* make the tests fail.

### DIFF
--- a/test/unit/presenters/search_results_presenter_test.rb
+++ b/test/unit/presenters/search_results_presenter_test.rb
@@ -70,7 +70,7 @@ class SearchResultsPresenterTest < ActiveSupport::TestCase
       presenter = SearchResultsPresenter.new(response, params)
 
       assert presenter.has_next_page?
-      assert_equal '/search?q=my-query&start=50', presenter.next_page_link
+      assert_equal '/search?count=50&q=my-query&start=50', presenter.next_page_link
       assert_equal '2 of 4', presenter.next_page_label
     end
 
@@ -97,7 +97,7 @@ class SearchResultsPresenterTest < ActiveSupport::TestCase
       presenter = SearchResultsPresenter.new(response, params)
 
       assert presenter.has_previous_page?
-      assert_equal '/search?q=my-query&start=50', presenter.previous_page_link
+      assert_equal '/search?count=50&q=my-query&start=50', presenter.previous_page_link
       assert_equal '2 of 4', presenter.previous_page_label
     end
 
@@ -127,7 +127,7 @@ class SearchResultsPresenterTest < ActiveSupport::TestCase
       assert_equal nil, presenter.previous_page_link
       assert_equal nil, presenter.previous_page_label
       assert presenter.has_next_page?
-      assert_equal '/search?start=50', presenter.next_page_link
+      assert_equal '/search?count=50&start=50', presenter.next_page_link
       assert_equal '2 of 4', presenter.next_page_label
     end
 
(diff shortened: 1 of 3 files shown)


KushalP:
> From now on, changing the default number of results to anything but 50
> wil _not_ make the tests fail.

:clap:

test/unit/presenters/search_results_presenter_test.rb
```diff
@@ -138,14 +138,7 @@ class SearchResultsPresenterTest < ActiveSupport::TestCase
         start: 100,
       })
 
-      assert_equal 50, params.count
-      presenter = SearchResultsPresenter.new(response, params)
```
rboulton: I'm not sure we want to drop these lines of testing. Are they duplicated somewhere else?

tijmenb: They are indeed duplicated. Only removing them here wasn't really enough though. Continued here: https://github.com/alphagov/frontend/pull/795

edds:
Would you be able to elaborate on the reason behind this change? When we built some of the [filter pages on GOV.UK](https://www.gov.uk/government/announcements) we initially had infinite scroll on the page. We added some tracking to that page to see how far people scrolled and found out that a large number of people scrolled to somewhere within the first 40 results but few went beyond that. So when we removed the infinite scroll and reverted to standard pagination we showed 40 results per page and were pretty confident that most users (more than 98% from memory) would never need to use the pagination.

I know that there is much more sophisticated analytics on the search pages though so we be interesting to see what the effect this change has on how users use search results and what we can learn from it. It would be good to also be consistent about the number of documents we show per page on all our document result style pages and iterate them with this page.

alicebartlett:
+1 to @edds comment, it would be really good to feed this change into our other design patterns wrt result count.

tijmenb:
Good point.

Here's a graph of the click positions in the past month:

screen shot 2015-04-29 at 15 03 26

From Tara:

> The main point is that 95% of clicks are on the top 20 results. There is a clear dropoff after 50, so I'd expect a similar dropoff at 20 instead, and we can monitor how much the use of pagination increases.

This sprint we're also improving on the search results itself, so we're also going to track what results appear for any given search. This should help us to improve the results in the future.

This might get changed in the future, depending on what the data says. For now we want to keep this change to this search interface because that's what we're working on.

CC @KushalP

KushalP:
In addition to what @tijmenb said, we're not going to propagate this change across all similar search interfaces.

We're intending to capture specific data for this interface which we can then use to tweak and improve it. Hopefully the same data points can be used on similar interfaces in the future (with hypotheses) and we can improve those too.

rboulton:
I agree with all the above, but In further addition ... there's good reason to think that the appropriate number of results to display for a "finder" might be different than for the appropriate number to display for the main search interface.  A reasonable hypothesis is that users on finder pages will often want to see all the results which match their search, whereas most users on the main search interface are searching for the answer to a particular question, so only want 1, or a few, documents.

This all depends on the quality of our ranking, of course, and we'll be monitoring closely with analytics.

There is a strong technical benefit to showing fewer results:
 - less data for users to download, and a faster response
 - we're adding tracking of which results have been shown, and this is easier with fewer results

But, I think the most compelling reason to reduce the results per page is that some users get overwhelmed by long lists of documents, and even if the top few are what they need, they feel they have to read all of them to be sure.  I've observed this happen in user research sessions, and it's painful to see.  I'd like to get to a state where we're happy to only show 10 results per page (or at least on the first page), but I don't think our ranking quality is high enough yet.

PR #797, opened 2015-05-08
Remove outline box from recommended links.

alext:
These should be styled in the same was as all the other search results
(except with the standard external link icon on the actual link).

Trello: https://trello.com/c/XJsHhfZV/85

Before:

image

After: 

image

### DIFF
--- a/app/assets/stylesheets/views/_search.scss
+++ b/app/assets/stylesheets/views/_search.scss
@@ -218,16 +218,6 @@ main.search {
           }
         }
 
-        li.external {
-          border: 1px solid $border-colour;
-          padding: 9px;
-          margin-bottom: 25px;
-
-          @include media(tablet) {
-            padding: 14px;
-          }
-        }
-
         li.descoped-results {
           border-left: 4px solid $govuk-blue;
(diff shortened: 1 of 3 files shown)


alicebartlett:
I haven't tested it, but I've had a look and I think you could also remove the `external` class from the list item in the template: https://github.com/alphagov/frontend/blob/6cb5b64e07ae7c70e462180893f129c52f76b542/app/views/search/_results.mustache#L13

fofr:
I can't see the Trello card, but was wondering what the user need/design reasoning was.

rboulton:
There's not a lot of detail to see in the Trello card (but I'll see about getting that Trello opened up, anyway).

There are two bits of reasoning here:

 - we've observed users (quite a long time ago, now) skipping over these items because they assume they're "adverts" or similar non-relevant stuff due to the different formatting. (I assume because they're used to adverts appearing in a similar way in web search engines.)

 - there's actually no user need we've identified for styling these differently anyway.

alext:
@alicebartlett I've updated to remove that class.

@fofr I've added more detail to the commit message at the same time.

PR #800, opened 2015-05-14
Remove .you-will-need CSS

bradleywright:
The HTML was removed in b2e1e4b but we didn't remove the CSS as well.

Refs #799


bradleywright:
This is wrong.

PR #804, opened 2015-05-26
Track search results shown to users

KushalP:
To improve the search results we show, we want to record:
- the list of results shown, in order
- any spelling suggestions made
- be able to represent groupings of results (eg, for the new "scoped" manuals searches)

This PR is split into **three major sections**:

### 1. Creating the testable code
Commits b3bafd0 and 061dc58 are where some tests are created around functions that do the main work associated with extracting the search results URLs.

### 2. Inlining the testable code
Commits 1f3eff9, 58707a1, and 58707a1 are where the code that was in a separate module is inlined into the `search.js` file and a `GOVUK.search` namespace is created.

### 3. Tracking the search results and spelling suggestions
Commits f197d7f and f1e9076 are where the tracking gets added.

Story: https://trello.com/c/TBJyZJDp


KushalP:
Closing this briefly. I want to check the naming of these tracking events with Tara.

edds:
Thank you for adding some tests to this :+1:. Would you however, be able to structure it more like how we describe in our [JavaScript style guide](https://github.com/alphagov/styleguides/blob/master/js.md#modules). The big change is your collection of named self executing functions would ideally be exposed methods which are then called in some kind of constructor or init method. Possibly something like:

```javascript
GOVUK.search = {
  init: function (){
    ...
  },
  extractSearchURLs: function ($searchResults) {
    ...  
  },
  extractSearchURL: function (index, element) {
    ...
  },
  trackSearchResultURLs: function () {
    ...
  },
  trackSpellingSuggestion: function () {
    ...
  },
  trackSearchClicks: function (){
    ...
  },
  trackExternalSearchClicks: function (){
    ...
  }
}
```

KushalP:
@edds I've extracted all of the pre-existing self-executing functions into a single `GOVUK.search` namespace in 84a2deb. They're not triggered using a single `GOVUK.search.init()` function. I think this is what you meant, if not please let me know.

rboulton:
This is looking pretty good, though I'll happily defer to @edds on the javascript style stuff.

I think the `scoped` parameter is the wrong way round: specifically, you're setting it to "true" for results which are "unscoped" (ie, from a search across all of GOV.UK).  I'd expect that to be "false".

edds:
Thanks a lot @KushalP, that is exactly what I meant :+1:

KushalP:
@rboulton f11cb58 updates the labelling to use `descoped` and I've removed the second case to reduce the amount of information we transfer from the user's browser.

PR #808, opened 2015-06-02
Remove /vehicle-tax static page

tadast:
This static page is being replaced by a dynamic page published via the publisher application. We have it ready as a draft, but can't publish it under the `vehicle-tax` url slug because it is used by this static page.

### Deployment strategy

1) We :shipit: 
2) The old static page is served by cache for 30mins
3) We publish the new version via publisher

This means I will need to be participating in the deployment.

Can anybody suggest a better deployment strategy or warn if there's something that can go wrong with this approach?

Also, I had to submit this PR from a forked repo, perhaps it would be beneficial to run this on the CI  before merging.


evilstreak:
Which part of the system is stopping the new version from being published before the old code is removed?

It should be possible to publish the artefact and the content for the new version while the existing page is still being served (this will probably need some devopsing), and then when a new version of this app is deployed removing the slug from the list of exceptions, it should start rendering the new version.

tadast:
@evilstreak has shown me around panopticon and we found another old document published under this URL slug. It is required to have something published under /vehicle-tax, because the title, metadata and the start link from that publication is used by the static view.

We will swap the new and the old publications first, then it should be ok to deploy with without any worries.

Thanks @evilstreak

tadast:
The new publication has gone live. You can verify it here https://www.gov.uk/api/vehicle-tax.json (the html version is still overridden by the static page). So once this PR gets merged and deployed, it should start showing:

screenshot 2015-06-04 11 02 19

Could someone please review the changes?

edds:
Happy to be getting rid of one of these hard coded pages. :+1:

PR #810, opened 2015-06-02
Cross domain tracking for HMRC renew tax credits

fofr:
HMRC will be promoting https://www.gov.uk/renewtaxcredits which sends people to their service at
https://www.tax.service.gov.uk/tax-credits-service/start. Attempt to follow users from the paid-for promotions through to the completion of the service.

https://www.pivotaltracker.com/story/show/95971198

### DIFF
--- a/app/helpers/cross_domain_analytics_helper.rb
+++ b/app/helpers/cross_domain_analytics_helper.rb
@@ -1,0 +1,9 @@
+module CrossDomainAnalyticsHelper
+  def cross_domain_tracker(slug)
+    {
+      'register-to-vote' => 'UA-23066786-5',
+      'accelerated-possession-eviction' => 'UA-37377084-12',
+      'renewtaxcredits' => 'UA-43414424-1', # HMRC, https://www.pivotaltracker.com/n/projects/1261204
+    }[slug]
+  end
+end
--- a/app/helpers/cross_domain_analytics_helper.rb
+++ b/app/helpers/cross_domain_analytics_helper.rb
@@ -3,7 +3,7 @@ module CrossDomainAnalyticsHelper
     {
       'register-to-vote' => 'UA-23066786-5',
       'accelerated-possession-eviction' => 'UA-37377084-12',
-      'renewtaxcredits' => 'UA-43414424-1', # HMRC, https://www.pivotaltracker.com/n/projects/1261204
+      'renewtaxcredits' => 'UA-43414424-1',
     }[slug]
   end
 end
--- a/app/views/root/transaction.html.erb
+++ b/app/views/root/transaction.html.erb
@@ -27,11 +27,15 @@
     <%- end -%>
 
   </section>
-  <%# Enable cross-domain analytics for IER (register to vote) only - not generalised yet %>
-  <% if @publication.slug == 'register-to-vote' %>
-    <%= render :partial => 'transaction_cross_domain_analytics', :locals => { :ga_account_code => 'UA-23066786-5' } %>
-  <%# Enable cross-domain analytics for Civil claims - still not worth generalising yet %>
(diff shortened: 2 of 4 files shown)


app/views/root/transaction.html.erb
```diff
@@ -27,11 +27,15 @@
...
-  <% elsif @publication.slug == 'accelerated-possession-eviction' %>
-    <%= render :partial => 'transaction_cross_domain_analytics', :locals => { :ga_account_code => 'UA-37377084-12' } %>
-  <% end %>
+  <% #Enable cross-domain analytics for specific start pages based on slug
+    {
+      'register-to-vote' => 'UA-23066786-5',
+      'accelerated-possession-eviction' => 'UA-37377084-12',
+      'renewtaxcredits' => 'UA-43414424-1', # HMRC, https://www.pivotaltracker.com/n/projects/1261204
+    }.each do |slug, account_code|
+      if @publication.slug == slug %>
```
edds: This is starting to turn into quite a lot of logic in this view. Is it worth moving some of this logic into either the presenter or a helper? So the view could just call something like `if @publication.ga_account_code` or `if account_code = ga_account_code_for_publication(@publication)`.

fofr: I agree, I've added a helper in 6bc6afb644107c99f59f4a0a9dc6ad2b69af660d

bradleywright:
Should we add a test that the helper does what we expect?

fofr:
@bradleywright I've added a unit test

app/helpers/cross_domain_analytics_helper.rb
```diff
@@ -0,0 +1,9 @@
+module CrossDomainAnalyticsHelper
+  def cross_domain_tracker(slug)
+    {
+      'register-to-vote' => 'UA-23066786-5',
+      'accelerated-possession-eviction' => 'UA-37377084-12',
+      'renewtaxcredits' => 'UA-43414424-1', # HMRC, https://www.pivotaltracker.com/n/projects/1261204
```
edds: Can we lose this link to the pivotal board. If there is information in there which should be attached to this code it would be better to put that in the commit message or the comment. This code will probably outlive the pivotal board.

bradleywright:
:+1:

app/helpers/cross_domain_analytics_helper.rb
fofr: All the content contained within the story is already in the commit.

fofr: While the link is available, it is useful. Link rot for the history and context of our backlog is the problem that ought to be fixed. Not having a permanent link back to context/discussion for any given story makes excavation more difficult, even if commits themselves are descriptive and useful.

edds: In which case can this be updated to a link to the relevant story rather than a link to a board then.

fofr: Fixed

evilstreak: Having the link in the pull request is great, and Github makes it easy to go from a commit to a pull request. For example, https://github.com/alphagov/frontend/commit/141d9458b048ff886fe2c106134fd287601fb4f2 shows that this commit is part of #810.

Having the link in the commit is fine, it means we're not reliant on the Github history, though that's probably at least as persistent as our Pivotal Tracker project will be. Getting hold of that information is pretty straightforward using `git blame`.

Having the link as a comment in the code seems pointless, and suggests that to properly understand what's done here the reader should follow through that link.

I'm strongly in favour of dropping the link from the code.

fofr: I disagree, but have removed the link in favour of getting this merged.

PR #811, opened 2015-06-03
Add counterpart driving licence homepage promo

marksheldonGDS:
Add counterpart driving licence homepage promo, to replace the Vehicle tax has changed promo.

Ticket: https://www.pivotaltracker.com/story/show/96076884


edds:
Thanks Mark, two tiny things: would you be able to pass that image through [ImageOptim](https://imageoptim.com/) as you should be able to get that file 25% smaller without any quality loss. Also can you delete the old image that this one is replacing. Thanks.

PR #814, opened 2015-06-08
Track the search results and suggestions

KushalP:
Track the full list of search URLs and any spelling suggestions. This is now in a single event request, because:

   1. We'll be processing the data outside of Google Analytics
   2. It's one less request from a user's browser

Story: https://trello.com/c/TBJyZJDp/67-record-which-documents-were-shown-for-a-search

### DIFF
--- a/app/assets/javascripts/search.js
+++ b/app/assets/javascripts/search.js
@@ -113,7 +113,7 @@
 
       if (GOVUK.analytics !== undefined &&
           GOVUK.analytics.trackEvent !== undefined) {
-        GOVUK.analytics.trackEvent('Search Results', 'Shown', {
+        GOVUK.analytics.trackEvent('searchResults', 'resultsShown', {
           label: searchResultData,
           nonInteraction: true
         });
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/search.js
```diff
@@ -87,6 +97,27 @@
...
+        searchResultData['urls'] = searchURLs;
+      }
+
+      if (searchSuggestion !== null) {
+        searchResultData['suggestion'] = searchSuggestion;
+      }
+
+      if (GOVUK.analytics !== undefined &&
+          GOVUK.analytics.trackEvent !== undefined) {
+        GOVUK.analytics.trackEvent('Search Results', 'Shown', {
```
rboulton: Wasn't the event name meant to be in camelCase?

rboulton: From Tara's comment in the ticket:

    Category: searchResults
    Action: resultsShown

So I think this is should be `trackEvent('searchResults', 'resultsShown', {...`

rboulton:
Other than the event naming, this looks good to me.

It would be good for pull requests to reference the Trello tickets (now that our Trello is open).

app/assets/javascripts/search.js
KushalP: Yup, fixed in de36ba3.

KushalP:
@rboulton story now includes a link to the Trello ticket.

rboulton:
Top.  Look forward to seeing these coming through in GA.

PR #816, opened 2015-06-10
Only run search JavaScript on the /search path

KushalP:
The JavaScript in this file was being triggered in multiple places
across the site. We only want it to occur on `/search` pages. Use a
conditional to enforce this behaviour.

### DIFF
--- a/app/assets/javascripts/search.js
+++ b/app/assets/javascripts/search.js
@@ -7,12 +7,14 @@
 
   var search = {
     init: function () {
-      var $searchResults = $('#results .results-list');
+      if (window.location.pathname === "/search") {
+        var $searchResults = $('#results .results-list');
 
-      search.enableLiveSearchCheckbox($searchResults);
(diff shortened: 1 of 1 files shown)


KushalP:
Worth saying that this probably needs some discussion. I'm wrapping all of the previously existing events here as well, which might not be desired.

app/assets/javascripts/search.js
```diff
@@ -7,12 +7,14 @@
 
   var search = {
     init: function () {
-      var $searchResults = $('#results .results-list');
+      if (window.location.pathname === "/search") {
```
alext: Instead of tying this to the path in the URL, is there some other way we could tie this to the search page.  I worry that this is brittle.

Could we not trigger this based on the presence of the `#results .results-list` element (or maybe something even more specific).

KushalP:
Actually, this isn't solving the problem. I'm going to close the PR.

PR #818, opened 2015-06-10
Stringify JSON we send to Google Analytics

KushalP:
Google Analytics expects labels to be Strings instead of JSON
objects. Stringiy the data we send them otherwise the values will be
converted to null values.

Docs: https://developers.google.com/analytics/devguides/collection/analyticsjs/events


boffbowsh:
:+1:

PR #820, opened 2015-06-16
Change homepage promo from Queen's Speech to 2015 Budget

issyl0:


evilstreak:
What compression did you run it through with? You can find a slider in the JPEGOptim section of the Preferences window for ImageOptim.

232kB is pretty large still – is it possible to increase the compression without introducing noticeable compression artefacts?

issyl0:
@evilstreak Is that better? Got it down to ~60kb after I realised where the slider was...

evilstreak:
:+1: Looks good

PR #823, opened 2015-06-17
Add aria-hidden to horizontal rules

edds:
The homepage is split by sections anyway and doesn't need extra
splitting. In a screen reader it announces 'group services and
information, horizontal rule, group department and policy, horizontal
rule...". The horizontal rules are extra noise which are just
distracting.

### DIFF
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -122,7 +122,7 @@
         </div>
       </section>
 
-      <hr id="departments-and-policy" />
+      <hr id="departments-and-policy" aria-hidden="true" />
 
       <section class="departments-and-policy" aria-labelledby="departments-and-policy-label">
         <div class="inner-block floated-children">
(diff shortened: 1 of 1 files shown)


alextea:
If that's the case is there a reason why the rules aren't applied as styling to the sections?

edds:
@alextea the problem is that there are no elements to apply the styling to. This is due to us using negative margins on wrapping elements so that we can float inner elements with margins on them for gutters. Though this might be different with the new grid mixins. So for the moment we need to use a 'dummy' element and the HR at least makes sense as something to add a keyline onto the page.

tombye:
:+1: (sorry for the lateness.)

As a rule (no pun intended) I'd say the value of `hr`'s is their ability to provide a break in text when that text isn't already divided up into other groupings. It's like in print I guess, where they're used when there is a break between types of copy not indicated by their presentation (the containing chapters or pages).

I mention this as a note for future use, that there may be a need for announcing them in some of our longer-form documents. As for here, totally agree with the intent.

PR #824, opened 2015-06-19
Track search results when filtered by organisation

KushalP:
When a user filters the search results by organisation, trigger a new GA
event to record the search results on the filtered results page.

Story: https://trello.com/c/NuwDBZEz

### DIFF
--- a/app/assets/javascripts/live-search.js
+++ b/app/assets/javascripts/live-search.js
@@ -42,6 +42,7 @@
     pageTrack: function(){
       GOVUK.analytics.setResultCountDimension(liveSearch.cache().result_count);
       GOVUK.analytics.trackPageview();
+      $(document).trigger("liveSearch.pageTrack");
     },
     checkboxChange: function(e){
       var pageUpdated;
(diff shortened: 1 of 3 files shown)


rboulton:
This looks good.  Is there some way we could test this?  Perhaps by adding something to `test/javascripts/unit/live-search-test.js`?

edds:
I think this should be done as a callback to the updating of the results rather than an event listener on the inputs. That way we can be sure there are new/updated results before we sent the analytics beacon.

KushalP:
> This looks good. Is there some way we could test this?

My initial attempt at this used a callback to run the GA event. Then in the test the callback was replaced for another function, but that felt convoluted. I'll see if I can write a better test for this functionality.

> I think this should be done as a callback to the updating of the results rather than an event listener on the inputs

I'll update it to do that :+1:

KushalP:
This now has tests and listens to an event from `GOVUK.liveSearch` instead of waiting for a checkbox to be checked/unchecked.

PR #825, opened 2015-06-22
Housekeeping in search views

tijmenb:
We're going to be doing some work on the search pages soon. This PR removes some old code and duplication to make that easier.


alext:
Is `params[:organisation]` actually used by anything any more?

tijmenb:
It has not, and the test testing for its existence has also been deleted since. Removed the tag & rebased.

PR #828, opened 2015-06-26
Restrict description displayed in search results

bishboria:
Some descriptions for content can be very long and the full description
is shown to the user. This has the effect of obscuring other search
results by moving them further down the page

https://trello.com/c/GKDjmzlW/209-truncate-long-descriptions-in-search-results

### DIFF
--- a/app/presenters/government_result.rb
+++ b/app/presenters/government_result.rb
@@ -70,7 +70,7 @@ class GovernmentResult < SearchResult
       description = result["description"]
     end
 
-    description = description.truncate(215, :separator => " ") if description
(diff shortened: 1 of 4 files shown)


rboulton:
Would it be easy to make this use an &hellip; entity instead of three separate dots?

bishboria:
Yeah, it's straight forward to change it to `…` if that's preffered.

tijmenb:
Maybe we could extract the 215 value in `GovernmentResult` and here into a constant so that we don't inadvertently change one but not the other.

tijmenb:
Probably best to `--amend` the original commit w/ the changes now, otherwise we'll end up with a lot of commits named "Address pull request comments". Looks good otherwise! :+1:

PR #830, opened 2015-06-29
Update mustache

edds:
This bumps shared_mustache which in turn bumps mustache to the version with the performance fix for partials inside loops.

It also reverts the change which restructured the templates to get around the performance issues.


alicebartlett:
Woohoo!

PR #831, opened 2015-06-30
Render a custom logo for Environment Agency

edds:
The Environment Agency have an exemption from the single identity brand
and as such have their own logo which they want to use on their campaign
page. This adds a way to have custom image logos for departments.

This would ideally be using a component to render all logo images so
that we wouldn't need to keep two copies of department logos around the
place. However, that has been deemed to be out of scope for this bit of
work. This is the smallest amount of work needed to get the logo working
right now and we can look at department logos as part of a bigger bit of
work in the future.

<table>
  <thead>
  <tr><th>Desktop</th><th>Mobile</th></tr>
</thead>
<tr>
<td></td><td></td>
</tr>
</table>


app/views/root/campaign.html.erb
```diff
@@ -15,9 +15,17 @@
 
     <div class="campaign content-block">
       <% if has_organisation?(@publication) %>
-        <div class="organisation">
-          <a href="<%= organisation_url(@publication) %>" class="organisation-logo organisation-logo-<%= organisation_crest(@publication) %> <%= organisation_brand_colour(@publication) %>"><%= formatted_organisation_name(@publication) %></a>
-        </div>
+        <% if organisation_has_custom_logo?(@publication) %>
+          <div class="organisation custom-logo">
+            <a href="<%= organisation_url(@publication) %>" class="custom-logo">
```
fofr: Do the wrapper and the link both need the `custom-logo` class?

fofr:
One minor comment otherwise this looks good.

app/views/root/campaign.html.erb
edds: Now with less `custom-logo`.

PR #832, opened 2015-06-30
Promote tax credits renewal instead of job search on the home page

tadast:
The deadline for renewing tax credits is 31 July 2015. This should raise users' awareness about it.

The new image has correct dimensions and @issyl0's https://github.com/issyl0/govuk_homepage_promo_checker has been really helpful to ensure that.


edds:
Could we push that file size down slightly further please. You should be able to make it quite a bit smaller without much visible quality loss. [ImageOptim](https://imageoptim.com) can be really useful for making it smaller.

tadast:
OK, 176k => 56.6k :)

PR #834, opened 2015-07-03
Wrap the related sidebar in an aside

edds:
The aside element lets you mark content on the page which is
complementary to the page. This lets screen readers and other assistive
technology announce to the user that the content isn't part of the main
body of the page.

In user research we have seen users accidentally enter the related
sidebar without realising and assuming that they are still in the main
body of the page. This should help those users.

### DIFF
--- a/app/views/root/_base_page.html.erb
+++ b/app/views/root/_base_page.html.erb
@@ -21,5 +21,7 @@
 </main>
 
 <% if local_assigns.fetch(:show_related_items, true) %>
-  <div id="related-items"></div>
+  <aside>
+    <div id="related-items"></div>
+  </aside>
 <% end %>


app/views/root/_base_page.html.erb
```diff
@@ -21,5 +21,7 @@
 </main>
 
 <% if local_assigns.fetch(:show_related_items, true) %>
-  <div id="related-items"></div>
+  <aside>
+    <div id="related-items"></div>
```
fofr: Any reason why the `<div>` can't be straight swapped for an `<aside>`?

edds: slimmer looks for `div#related-items` for the replace. I don't fancy updating slimmer.

alext: Because of slimmer - https://github.com/alphagov/slimmer/blob/master/lib/slimmer/processors/related_items_inserter.rb.  It specifically looks for a `div#related-items`, although the selector there could probably be relaxed to just `#related-items`.

Also, from looking at that slimmer processor, I don't think this will do what you intend.  It just uses the presence of this div as a flag to trigger the insertion of related items into the static template.

edds: So the template in static doesn't actually have a `#related-items` element. The `BodyInserter` processor in slimmer runs before the `RelatedItemsInserter`. The  `BodyInserter` moves the whole of the `wrapper` over to the template and the `#related-items` div is inside the wrapper so will be in the page_template by the time the `RelatedItemsInserter` runs. The `RelatedItemsInserter` `filter` method in the could be re-written as:

```ruby
def filter(content_document, page_template)
    if page_template.at_css('body.mainstream div#related-items')
      page_template.at_css('body.mainstream div#related-items').replace(related_item_block)
    end
  end
```

and it would have the same effect.

Thinking about it though I should probably just update the [`related_item_block`](https://github.com/alphagov/static/blob/master/app/views/root/related.raw.html.erb) template to use an aside and then this change will be rolled out everywhere we use related sidebars.

edds:
Opened a pull request on static to supersede this: https://github.com/alphagov/static/pull/612

PR #837, opened 2015-07-15
Log search with zero results and suggestion

jackscotti:
If a search returns no results, but does have a spelling suggestion, a searchResults event should be sent to GA including the spellingSuggestions (but no urls). Currently, no event is sent to GA in this case.

Ticket: https://trello.com/c/zJ5wo5rc/183-no-results-logged-for-a-search-with-a-suggestion-but-0-results

~~PS: I've added an empty string for the suggestion parameter for searchResultData, this because I did not want to overcomplicate the IF clause in trackSearchResultsAndSuggestions. I am assuming that an empty string sent to GA is not going to be logged. Am I right? If not, I can renounce to this modification and make that IF clause even uglier.~~ (see comments)

PPS: this is the first time I write JS for GOV.UK. Be nice. :metal:

### DIFF
--- a/app/assets/javascripts/search.js
+++ b/app/assets/javascripts/search.js
@@ -121,7 +121,7 @@
 
       if (GOVUK.analytics !== undefined &&
           GOVUK.analytics.trackEvent !== undefined &&
-          $searchResults.length) {
+          (searchResultData.urls.length || searchResultData.suggestion)) {
         GOVUK.analytics.trackEvent('searchResults', 'resultsShown', {
           label: JSON.stringify(searchResultData),
           nonInteraction: true
--- a/test/javascripts/unit/search-test.js
+++ b/test/javascripts/unit/search-test.js
@@ -13,18 +13,62 @@ describe('GOVUK.search', function () {
   describe('extractSearchURLs', function () {
     var $resultsList;
 
+    describe('no results found', function () {
+      beforeEach(function () {
+        GOVUK.analytics = GOVUK.analytics || { trackEvent : function(args) {} };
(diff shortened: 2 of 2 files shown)


rboulton:
The empty string will be sent to GA - we're just serialising the searchResultData value to JSON and sending it to GA.  So, it would be better to make the `if` clause more complex...

app/assets/javascripts/search.js
```diff
@@ -16,7 +16,7 @@
       search.trackSearchResultsAndSuggestionsOnPageTrack($searchResults);
     },
     buildSearchResultsData: function ($searchResults) {
-      var searchResultData = {'urls': []},
+      var searchResultData = {'urls': [], 'suggestion': ""},
```
edds: not sure you need this change. You will still be able to call  `if(searchResultData.suggestion)` without it.

app/assets/javascripts/search.js
```diff
@@ -121,7 +121,7 @@
 
       if (GOVUK.analytics !== undefined &&
           GOVUK.analytics.trackEvent !== undefined &&
-          $searchResults.length) {
+          (searchResultData['urls'].length || searchResultData['suggestion'])) {
```
edds: This would be more JavaScript'y as `(searchResultData.urls.length || searchResultData.suggestion)`. If the key is a string it you don't need to use the array syntax and can just use dot-notation.

jackscotti:
Grand, thanks @edds and @rboulton. I'll address what you've said.

edds:
:+1: Thank you for adding some good testing around this also.

PR #839, opened 2015-07-16
Dont treat empty strings as a valid analytics profile

jamiecobbett:
See commits for more detail.

cc @dsingleton


dsingleton:
:zap:

PR #841, opened 2015-07-17
Set GA page variable before sending events

tijmenb:
(Don't merge yet, here for feedback.)

[From Trello](https://trello.com/c/zOts34TQ/184-old-url-is-logged-when-search-filters-are-changed):

> It looks like the GA events sent on ajax page changes in the search results use the original URL of the page, not the updated URL.
> - If I go to https://www.gov.uk/search?q=tax and then add an organisation filter, a GA tracking event is sent for the filtered page, but it uses the unfiltered URL as the url of the event.
> - If I do a full page refresh (eg, of https://www.gov.uk/search?q=tax&filter_organisations%5B%5D=hm-revenue-customs), the filter parameter is recorded, but if I then remove the filter, an event is recorded with the filter still applied.

GA sends the original URL because by default it uses the `window.location` during page load. Usually you can get around that by explicitly setting the `page` variable when sending the event:

```js
ga('send', 'pageview', '/new-page');
```

But better is to re-set the variable, so that all subsequent events and page tracks will send the correct `page`.  

> Specifying a page value in a pageview hit will send that page value to Google Analytics for that hit only; it will not update the page value stored on the tracker object itself. This can be problematic if you send other hits (e.g. events or social hits) and don't explicitly include the current page value. In such cases, Google Analytics will associate those hits with whatever URL was present at the time of the initial page load.

https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications

This commit adds that call before the page tracking. 

But I'm not 100% convinced this is the right place for this. For one thing, it doesn't use the `GOVUK.analytics` abstraction (which would need changing to support this). Maybe @edds / @fofr have opinions?

Note that using `pathname` + `search` is [taken from the Analytics implementation](https://developers.google.com/analytics/devguides/collection/analyticsjs/pages).

### DIFF
--- a/app/assets/javascripts/live-search.js
+++ b/app/assets/javascripts/live-search.js
@@ -41,7 +41,7 @@
     },
     pageTrack: function(){
       GOVUK.analytics.setResultCountDimension(liveSearch.cache().result_count);
-      GOVUK.analytics.trackPageview();
+      GOVUK.analytics.trackPageview(window.location.pathname + window.location.search);
       $(document).trigger("liveSearch.pageTrack");
     },
     checkboxChange: function(e){
--- a/app/assets/javascripts/search.js
+++ b/app/assets/javascripts/search.js
@@ -124,7 +124,8 @@
           (searchResultData.urls.length || searchResultData.suggestion)) {
         GOVUK.analytics.trackEvent('searchResults', 'resultsShown', {
           label: JSON.stringify(searchResultData),
-          nonInteraction: true
+          nonInteraction: true,
+          page: window.location.pathname + window.location.search
         });
       }
     },


rboulton:
Good sleuthing and write up.  I think we probably want our GOVUK.analytics abstraction to handle this.  I think there are two main ways we could do this:

 - Make the GOVUK.analytics abstraction have a call on it which resets the value
 - Make the GOVUK.analytics abstraction always reset the value when sending any tracking event

I think the latter is probably what we want - my concern is whether there is any risk of a performance impact from setting this every time.

app/assets/javascripts/live-search.js
```diff
@@ -40,6 +40,7 @@
       }
     },
     pageTrack: function(){
+      ga("set", "page", window.location.pathname + window.location.search);
       GOVUK.analytics.setResultCountDimension(liveSearch.cache().result_count);
       GOVUK.analytics.trackPageview();
```
edds: You should just be able to update this to `GOVUK.analytics.trackPageview(window.location.pathname + window.location.search);` rather than calling `ga` directly.

tijmenb: That's what I wanted to do at first, but this only works for this specific pageview, but not for the event that records the positions of the search results, [which is called via the `liveSearch.pageTrack` event](https://github.com/alphagov/frontend/blob/112eb28c5550b42af000f34772d496ec09103564/app/assets/javascripts/search.js#L125).

`GOVUK.analytics.trackEvent` unfortunately doesn't support forwarding the `page` variable.

floehopper:
This sounds as if it's related to https://github.com/alphagov/smart-answers/pull/1787.

/cc @chrisroos

chrisroos:
I haven't looked at this pull request in detail but the title suggests that it might also be related to https://github.com/alphagov/govuk_frontend_toolkit/pull/203.

fofr:
Good sleuthing, an interesting problem.

It looks like https://github.com/alphagov/govuk_frontend_toolkit/pull/203 partly solves it, but adding a method for setting the page from within the toolkit (the approach taken in this PR) would also make sense.

We should also consider how to avoid other apps stumbling on this problem in the future – perhaps a method that sets and tracks a page for when URLs are dynamically updated.

tijmenb:
Thanks everyone. I've updated this PR to use the new `trackEvent` argument from https://github.com/alphagov/govuk_frontend_toolkit/pull/203. It won't have effect until https://github.com/alphagov/static/pull/623 is merged and deployed.

fofr:
:+1:

PR #844, opened 2015-07-22
Enable liveSearchCheckbox with 0 or more search results

jackscotti:
Fixing a bug that made the checkbox component unusable when a search on GOV.UK
returned 0 results.

This meant that if the User managed to narrow down their search to 0 results,
they wouldn't have been able to widen their search by using the checkbox
component.

The bug has been removed by updating the bounds check around the checkbox.

Example URL of a search with 0 results: https://www.gov.uk/search?q=carbonara

Ticket: 
https://trello.com/c/hQqnoc4R/245-when-a-search-has-a-filter-and-no-results-the-remove-filters-link-doesn-t-work

Paired on this with the magic @KushalP !

🐛+💣 ==💀

### DIFF
--- a/app/assets/javascripts/search.js
+++ b/app/assets/javascripts/search.js
@@ -31,7 +31,7 @@
       return searchResultData;
     },
     enableLiveSearchCheckbox: function ($searchResults) {
-      if ($searchResults.length > 0) {
+      if ($searchResults.length >= 0) {
         $('.js-openable-filter').each(function(){
           new GOVUK.CheckboxFilter({el:$(this)});
         });
--- a/app/assets/stylesheets/views/_checkbox-filter.scss
+++ b/app/assets/stylesheets/views/_checkbox-filter.scss
@@ -93,18 +93,24 @@
     position:relative;
 
     .controls {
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/search.js
```diff
@@ -31,7 +31,7 @@
       return searchResultData;
     },
     enableLiveSearchCheckbox: function ($searchResults) {
-      if ($searchResults.length > 0) {
+      if ($searchResults.length >= 0) {
```
tommyp: As this is now `>=` do we want the value to be changed to `1`?

jackscotti: We want >= 0!

KushalP: This discussion was just had in person. Expect the commit message and PR message to be updated to make this clearer. In the mean time, I'm going to close this PR.

jackscotti: Thank you Tommy for the feedback and thank you Kush for closing the PR. I've updated the commit message, please let me know if it is still not clear.

PR #845, opened 2015-07-23
BUGFIX: Enable liveSearchCheckbox with 0 results

jackscotti:
### The problem
Some users frequently search GOV.UK by editing the URL and keeping their
choice of organisation filters in place, e.g.

  https://www.gov.uk/search?q=Housing&filter_organisations%5B%5D=land-registry

However, when no results are returned for a query, the organisation filter
(land registry in the example) is still selected and it's no longer possible
to use the "Remove filters" link to easily clear them, e.g.

  https://www.gov.uk/search?q=Immigration+Rules&filter_organisations%5B%5D=land-registry

### What we found
The "Remove filter" is driven by the `GOVUK.CheckboxFilter` JavaScript.
However, when there are no results, the component is not being activated.

### The fix
The bug has been fixed by updating the bounds check within the
`GOVUK.search.enableLiveSearchCheckbox` function to fire when there are
 0 or more results.

Story: https://trello.com/c/hQqnoc4R/245-when-a-search-has-a-filter-and-no-results-the-remove-filters-link-doesn-t-work

### DIFF
--- a/app/assets/javascripts/search.js
+++ b/app/assets/javascripts/search.js
@@ -31,12 +31,10 @@
       return searchResultData;
     },
     enableLiveSearchCheckbox: function ($searchResults) {
-      if ($searchResults.length > 0) {
-        $('.js-openable-filter').each(function(){
-          new GOVUK.CheckboxFilter({el:$(this)});
-        });
-        GOVUK.liveSearch.init();
-      };
+      $('.js-openable-filter').each(function(){
+        new GOVUK.CheckboxFilter({el:$(this)});
+      });
+      GOVUK.liveSearch.init();
     },
     extractSearchSuggestion: function () {
       var $suggestion = $('.spelling-suggestion a');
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/search.js
```diff
@@ -31,7 +31,7 @@
       return searchResultData;
     },
     enableLiveSearchCheckbox: function ($searchResults) {
-      if ($searchResults.length > 0) {
+      if ($searchResults.length >= 0) {
```
alext: Is there ever a case where this could be negative?  If not, then this conditional seems to be redundant.

jackscotti: You're right Alex, going to remove this redundant clause.

bishboria:
Minor issue when testing locally: when there are no results for the search, the organisations drop down shouldn't be clickable (and change height).

Sort that and you're good.

PR #847, opened 2015-07-23
BUGFIX: Enable liveSearchCheckbox with 0 results

jackscotti:
### The problem:
Some users frequently search GOV.UK by editing the URL and keeping their
choice of organisation filters in place, e.g.

https://www.gov.uk/search?q=Housing&filter_organisations%5B%5D=land-registry

However, when no results are returned for a query, the organisation filter
(land registry in the example) is still selected and it's no longer possible
to use the "Remove filters" link to easily clear them, e.g.

  https://www.gov.uk/search?q=Immigration+Rules&filter_organisations%5B%5D=land-registry

### What we found:
The "Remove filter" is driven by the `GOVUK.CheckboxFilter` JavaScript.
However, when there are no results, the component is not being activated.

### The fix:
The bug has been fixed by removing the bounds check within the
`GOVUK.search.enableLiveSearchCheckbox`.

Story: https://trello.com/c/hQqnoc4R/245-when-a-search-has-a-filter-and-no-results-the-remove-filters-link-doesn-t-work

### DIFF
--- a/app/assets/javascripts/search.js
+++ b/app/assets/javascripts/search.js
@@ -31,9 +31,11 @@
       return searchResultData;
     },
     enableLiveSearchCheckbox: function ($searchResults) {
-      $('.js-openable-filter').each(function(){
-        new GOVUK.CheckboxFilter({el:$(this)});
-      });
+      if($('.js-openable-filter input[type=checkbox]').length) {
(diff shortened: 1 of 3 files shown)


bishboria:
:+1:

PR #848, opened 2015-07-28
Fix simple smartanswers layout

tadast:
Issue: simple smartanswers questions are not aligned with the
prominent in mobile view.

This is because section.content does not have any styles applied.
Change to `main.content` (that is used all around the application) that adds
correct padding. Remove `.current` css class from `.step` as it doubles the padding.

Before/After:
screenshot 2015-07-28 14 50 32

Reported by @alextea


alextea:
:+1:

PR #849, opened 2015-07-28
Use preferred radio button styles in simple smartanswers

tadast:
As per http://govuk-elements.herokuapp.com/#guide-forms use radio buttons
with a large hit area indicated by a grey square. These are easier to select
with a mouse or touch devices.

Before/After

screenshot 2015-07-28 14 56 41


alextea:
:+1:

PR #851, opened 2015-07-29
Clean up list of transactions which are forced to open in new windows

daibach:
This list was created in 2012 to force some legacy transactions to open in new windows when the “start now” button was clicked. This was because some of those transactions had either been designed to be opened in new windows or broke when opened with toolbars enabled.

This pull request cleans up the list of transactions by removing those which no longer exist in their original form. These transactions have either been rebuilt or closed down.

### DIFF
--- a/lib/data/new_window_transactions.json
+++ b/lib/data/new_window_transactions.json
@@ -1,15 +1,9 @@
 {
   "new-window": [
-    "car-tax-disc-vehicle-licence",
-    "register-login-student-finance",
     "find-your-local-council",
-    "change-date-theory-test",
     "apply-blue-badge",
     "claim-state-pension-online",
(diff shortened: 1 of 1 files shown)


daibach:
It's possible that more transactions could be removed from the list, but I've only removed ones that don't have a start page any more.

bradleywright:
@marksheldonGDS ?

bradleywright:
Actually as @daibach says this can be merged as-is with no impact. Question for @marksheldonGDS: do we need this 'open in new window' functionality at all?

marksheldonGDS:
I don't see any reason why we need 'open in new window' functionality at all. It breaks convention and we'd only do this in exceptional circumstances, such as opening a page that is context-sensitive.

PR #855, opened 2015-08-13
Register special routes.

elliotcm:
https://trello.com/c/blLdEZN5/292-make-apps-register-special-routes-on-deploy

Requires deploy of
- [x] https://github.com/alphagov/url-arbiter/pull/24
- [x] https://github.com/alphagov/content-store/pull/141
- [x] https://github.com/alphagov/publishing-api/pull/54

We've introduced a new special_route format[1] to allow us to register
unusual routes which don't have a real format associated with them. This
supersedes the previous approach of registering such routes via the
router. The advantage of registering with publishing api is that the
routes will also be reserved with url-arbiter which should eliminate the
risk of clashes.

This adds a rake task to register these routes via that
mechanism, and removes the old method of registering directly with the
router.

The intention is to run this rake task on deploy (as we do for the route
registration at present).

[1] alphagov/govuk-content-schemas#106


lib/tasks/publishing_api.rake
```diff
@@ -0,0 +1,50 @@
...
+    publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new
+
+    routes = [
+      {
+        content_id: "f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a",
+        base_path: "/",
+        title: "GOV.UK homepage",
+      },
+      {
+        content_id: "ffcd9054-ee77-4539-978d-171a60eb4b2a",
```
rboulton: Is this meant to have the same base_path as the previous one?  If so, this looks quite odd, and I assume there's some "magic" happening here.  A comment to explain what path is being redirected to / here would be helpful.

lib/tasks/router.rake
```diff
@@ -1,33 +0,0 @@
...
-  end
-
-  task :register_backend => :router_environment do
-    @router_api.add_backend('frontend', Plek.current.find('frontend', :force_http => true) + "/")
-  end
-
-  task :register_routes => :router_environment do
-    routes = [
-      %w(/ exact),
-      %w(/search exact),
```
rboulton: See my comment here https://github.com/alphagov/rummager/pull/480#issuecomment-129410826 - I think these search routes might be better being registered by `frontend`.

lib/tasks/publishing_api.rake
alext: This should have a base_path of `/homepage`.  Additionally I wondwe whether this would be better being created as a redirect item int he content-store.  Then we could remove the handling of this from the frontend application.

elliotcm: Yeah no magic, just a copy pasta mistake.

I'll add the redirect idea to the list of things to review when all these PRs are ready.  We've another 7 to do.

lib/tasks/router.rake
elliotcm: Will take this into account at the end when we collate all the comments.

elliotcm:
@rboulton @alext I've fixed the `/homepage` mistake and included the search paths.

PR #856, opened 2015-08-14
Add guard against double prefixes in org description

tijmenb:
Currently, Frontend is in charge of appending the prefix to open organisations. We want Whitehall to take care of its data. This will also allow us to highlight the organisation description correctly.

The work for that is in: https://github.com/alphagov/whitehall/pull/2314

To prevent double prefix when Whitehall is migrated, we add a temporary guard here. This can be removed entirely once Whitehall has been migrated.


tijmenb:
This has been deployed now.

PR #858, opened 2015-08-14
Show highlighting in search results

tijmenb:
This commit adds support for highlighting to the search results. This is still under a debug flag to more easily review the changes.

## Screenshot

screen shot 2015-08-14 at 14 42 23

cc @alextea 

--

### Dependencies:

- [x] https://github.com/alphagov/rummager/pull/492 is the PR on rummager to add support for this feature
- [x] https://github.com/alphagov/rummager/issues/501 needs to be merged
- [x] https://github.com/alphagov/rummager/pull/492 and https://github.com/alphagov/rummager/issues/501 need to be deployed
- [x] https://github.com/alphagov/whitehall/pull/2314 must be deployed to correctly populate the descriptions for organisations
- [x] https://github.com/alphagov/frontend/pull/856 needs to be deployed to not show double descriptions

### DIFF
--- a/app/assets/stylesheets/views/_search.scss
+++ b/app/assets/stylesheets/views/_search.scss
@@ -133,6 +133,24 @@ main.search {
         float: left;
         width: $two-thirds;
       }
+
+      mark {
+        background-color: rgba(255, 255, 140, 0.5);
+        color: inherit;
+      }
+
+      a:visited mark {
+        color: $link-visited-colour;
+      }
+
+      a:hover mark {
+        color: $link-colour;
+      }
+
+      a:active mark {
+        background-color: transparent;
+      }
+
       .result-count {
         padding: 0;
       }
--- a/app/models/search_parameters.rb
+++ b/app/models/search_parameters.rb
@@ -49,6 +49,10 @@ class SearchParameters
     params[:debug_score]
   end
 
+  def enable_highlighting?
+    params[:enable_highlighting]
+  end
+
   # Build a link to a search results page.
   def build_link(extra = {})
     search_path(combine_params(extra))
@@ -61,6 +65,8 @@ class SearchParameters
       q: search_term,
       fields: %w{
         description
+        title_with_highlighting
+        description_with_highlighting
         display_type
         document_series
         format
--- a/app/presenters/search_result.rb
+++ b/app/presenters/search_result.rb
@@ -39,11 +39,14 @@ class SearchResult
   def to_hash
     {
       debug_score: debug_score,
+      enable_highlighting: @search_parameters.enable_highlighting?,
       title: title,
       link: link,
       description: description,
       examples_present?: result["examples"].present?,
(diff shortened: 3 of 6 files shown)


robinwhittleton:
Should we be using `mark` instead of `em`? http://www.w3.org/TR/html51/semantics.html#the-mark-element

tijmenb:
Yes, I really like that! :+1:

I'll update rummager as well.

robinwhittleton:
Just found this quote in the spec, so yeah, perfect:
>Another example of the mark element is highlighting parts of a document that are matching some search string.

alextea:
Have we checked the contrast with this? @tombye do you have any accessibility reckons?

tijmenb:
Updated PR. Using `<mark>` made the CSS and markup a little easier, so that's nice.

However, the default styling of the element (`background-color: yellow; color: black`) seems to interfere with the CSS:

screen shot 2015-08-20 at 17 56 43

I'll try to find a way around that next week (off tomorrow).

rboulton:
Merging, since this is behind a feature flag.  We can sort out any display issues before removing the flag, but merging will help resolve any other issues.

app/assets/stylesheets/views/_search.scss
```diff
@@ -133,6 +133,24 @@ main.search {
         float: left;
         width: $two-thirds;
       }
+
+      mark {
+        background-color: rgba(255, 255, 140, 0.5);
```
alextea: I think we should use `$focus-colour` here as it has better contrast than the pale yellow.

screen shot 2015-09-01 at 11 54 32

fofr: Doesn't this colour already have an inferred meaning, ie something that's focused? If my search term matches the title of the search result – which doesn't seem too far fetched, then it'd look like the title was focused and I could hit space or enter to visit that result.

Just my reckons.

alextea: I wonder if it would be a problem in the real world. Perhaps we could see how many of the top X search terms match the page title exactly.

I don't think it's a problem when it's a partial match.

The other solution is not highlight the title when it is a complete match.

quis: The colour contrast between the link text and the yellow background isn't great:
image

We've been using `$yellow-25` on the Digital Marketplace, which doesn't show up on high-contrast screens (the telly by the purple sofas for example).
image

We don't highlight search result page titles at the moment, but the tradeoff between keeping link text legible and having highlights visible is the tradeoff when you do.

alextea: @quis The large text (>=18px) passes the AA checks, so that is OK.

We are highlighting the titles because we are using synonyms and want people to see that their term may have matched a synonym. We also want it to be visible to those using high contrast screens (the same reason we use the focus styles as well).

PR #869, opened 2015-08-26
Make cross domain tracking support any destination

jamiecobbett:
Cross-domain tracking uses a magic _ga parameter appended to the link with a value that identifies the user.

The argument we were passing - with the value 'service.gov.uk' - is used to do a substring match for the links on the page to rewrite.

By changing it to use the domain in the Transaction's link, we should be able to support any destination, not just ones on *.service.gov.uk, for example `online.hmrc.gov.uk` or `petition.parliament.uk`.

Tested successfully in development.

/cc @bradleywright


bradleywright:
:+1:

fofr:
Lovely! :+1:

PR #871, opened 2015-08-26
Add some debugging gems

dsingleton:
Adds `better_errors` for web-console debugging, and `pry` for console-console debugging.

We use both on these on lots of other apps, it's better to specify them in the `Gemfile` so they're available as consistently as possible.


mattbostock:
pry :heart:

PR #872, opened 2015-08-26
[Discuss] Set app env-vars in startup.sh

dsingleton:
(This is more proprosal than strong opinion, discuss)

I hit an issue when running `bowl www frontend --without=frontend` and
running frontend in a separate tab with `./startup`, then hitting it
via `www.dev.gov.uk/a-frontend-slug`.

The issue is that when running under the router assets won't work unless
`GOVUK_ASSET_ROOT` is set (see https://github.com/alphagov/frontend/commit/77bb2cb3a05a8a87b23008f6fcc8b7f5ff9b88c5)

--

This is a slightly niche case, but it would be nice if we could use
startup.sh reliably to run an app, and get the same behaviour as if
it had been run with bowl (without depdencies).

--

I know there has been some talk of moving from the `Procfile` in
`development` containing the run command for an app in dev, to having
that be in the startup.sh per-app and have the Procfile call that. This
is a vague step in that direction, assuming we planned to remove the
duplication.

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 bundle install
-bundle exec rails s -p 3005
+govuk_setenv frontend bundle exec rails s -p 3005


bradleywright:
The obvious issue with this is that `startup.sh` will no longer run outside the VM. A potential solution would be to check if `govuk_setenv` exists before trying to use it.

Not sure how others feel about that - my preference is to make sure people are using the VM at all times for dev/prod parity, but I'm also aware people have different preferences and we should be respectful of that.

dsingleton:
> The obvious issue with this is that startup.sh will no longer run outside the VM. A potential solution would be to check if govuk_setenv exists before trying to use it.

Yeah, I assumed that would be one of the issues. AFAIK @elliotcm is the main non-VM user (that i'm aware of at least). 

I'm still keen that it's easy to run simple FE apps outside the VM, which is straight-forward at the moment, and I've had great success helping designers/non-GOV.UK folks run simple apps without a VM.

So :+1: to making it conditional/ignore `govuk_setenv` if it doesn't exist.

elliotcm:
I consider working outside the VM to be at my own risk and has only been worth the time because most of our applications have been added incrementally in the time I've been here.  I wouldn't recommend it to anyone starting afresh.

That said, @dsingleton your point about frontend apps is a good one, reinforced by the existence of the dummy content store.

jamiecobbett:
This has been open for nearly 40 days, are we likely to reach a decision?

dsingleton:
I don't think so. It's sparked a bit of discussion, which will be beneficial when we discuss this in the future, hopefully.

Happy to close this discussion.

PR #878, opened 2015-09-03
Scroll to the selected tab on page load

robinwhittleton:
This behaviour means that a selected tab retains the default fragment behaviour, i.e. people can copy links to a particular section of a site. An example url would be http://frontend.dev.gov.uk/view-driving-licence#other-ways-to-apply  - loading the page will jump to the tab strip with the correct tab selected and its content showing below.

### DIFF
--- a/app/assets/javascripts/transactions.js
+++ b/app/assets/javascripts/transactions.js
@@ -18,7 +18,9 @@ $(document).ready(function () {
   var $container = $('section.more');
 
   if ($container.find('.js-tabs').length) {
-    $container.tabs();
+    $container.tabs({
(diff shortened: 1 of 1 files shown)


robinwhittleton:
I can’t see any problems with this becoming the default behaviour, but I’d like to hear dissenting opinions if there are any.

robinwhittleton:
For what it’s worth, this section of code was last touched in 2012: https://github.com/alphagov/frontend/commit/19b222026456665a27fe426cb63b14422533e2e4

bradwright:
:+1:

PR #879, opened 2015-09-03
Use bold for search highlights

rboulton:
This is less intrusive than the yellow background highlighting previously
used, which is desirable because the highlights shouldn't get in the way
of reading the actual text.  Also, using a font weight instead of a
background colour means that the gaps between highlighted words aren't
visible.

The bold doesn't show up in titles, because titles are already bold.
However, keeping the html markup in there seems desirable, because then we
can easily try out other visual markup options.  It also lets us build debug
tools to help us understand why things are matching the query.

# Before

screen shot 2015-09-03 at 17 58 15

# After

screen shot 2015-09-03 at 17 56 01

### DIFF
--- a/app/assets/stylesheets/views/_search.scss
+++ b/app/assets/stylesheets/views/_search.scss
@@ -135,7 +135,7 @@ main.search {
       }
 
       mark {
-        background-color: inherit;
+        background-color: transparent;
         color: inherit;
         font-weight: bold;
       }
--- a/app/assets/stylesheets/views/_search.scss
+++ b/app/assets/stylesheets/views/_search.scss
@@ -135,8 +135,9 @@ main.search {
       }
 
       mark {
-        background-color: $focus-colour;
+        background-color: inherit;
         color: inherit;
+        font-weight: bold;
       }
 
       a:visited mark {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/views/_search.scss
```diff
@@ -135,8 +135,9 @@ main.search {
       }
 
       mark {
-        background-color: $focus-colour;
+        background-color: inherit;
```
robinwhittleton: Do we want `transparent` instead of inherit?

alextea: @robinwhittleton not if the mark will be inside an `<a>` tag, and may have a `:focus` state with a background colour.

robinwhittleton: If the `mark` is inside the `a` the `a`’s background colour will show through the transparent `mark`. Unless I’m really misunderstanding things?
It doesn’t really matter either way, but transparent seems more accurate to what we want to do. E.g. if you had a gradient on the `a` on top of the background colour then the inherited background colour on `mark` would obscure the gradient on focus.

alextea:
This gif cycles through the various (focus, hover, visited) link states

link-states

PR #881, opened 2015-09-07
Always highlight search results

rboulton:
Remove the enable_highlighting parameter - now that we're happy with the
highlighting, we always want to highlight results:
https://trello.com/c/GPInH4fj/303-remove-search-highlighting-feature-flag

Looks like:
screen shot 2015-09-07 at 06 49 01

### DIFF
--- a/app/models/search_parameters.rb
+++ b/app/models/search_parameters.rb
@@ -49,10 +49,6 @@ class SearchParameters
     params[:debug_score]
   end
 
-  def enable_highlighting?
-    params[:enable_highlighting]
-  end
-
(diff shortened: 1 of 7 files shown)


tijmenb:
We can stop requesting title & description and remove the code related to generating the description too (like https://github.com/alphagov/frontend/blob/c3d25ec780c15223bec869456a5a284b17c78c3a/app/presenters/government_result.rb#L67). I think there's no reason to leave that in, right?

rboulton:
Yup, that can go.  PR welcome!

PR #882, opened 2015-09-08
Remove code to modify search descriptions and titles

tijmenb:
Frontend previously modified the titles and descriptions of search results. To enable us to highlight the search description and title, we've moved this logic to the publishing apps. For example, Whitehall
now correctly prepends "The home of HMRC on GOV.UK" to HMRC's search description (alphagov/whitehall#2314).

`description` is no longer needed, so we don't request that from rummager. `title` is still used to generate nice looking metadata. 

The truncation of the description is now the responsibility of rummager (alphagov/rummager#492).

Since turning on highlighting for all users on this app we no longer need this special case code. However, in making this PR we've discovered that for closed organisations, "Closed organisation:" is prepended to the title. This will be fixed in a PR on Whitehall.

Trello: https://trello.com/c/GPInH4fj

Note that I've had to reformat the hashes in the test file because Rubocop started complaining.

### DIFF
--- a/test/unit/presenters/search_result_test.rb
+++ b/test/unit/presenters/search_result_test.rb
@@ -2,30 +2,8 @@
 require "test_helper"
 
 class SearchResultTest < ActiveSupport::TestCase
-  should "display a description" do
-    result = SearchResult.new(SearchParameters.new({}), "description" => "I like pie")
-    assert_equal "I like pie", result.description
-  end
-
-  should "not display a generic description for other formats which are missing them" do
-    result = SearchResult.new(SearchParameters.new({}), "format" => "edition", "title" => "VAT")
-    assert_nil result.description
-  end
-
-  should "truncate descriptions to a maximum of 215 characters" do
-    result = SearchResult.new(SearchParameters.new({}),
-                              "description" => "Long description is long "*100)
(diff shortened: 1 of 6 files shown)


test/unit/presenters/search_result_test.rb
```diff
@@ -2,30 +2,8 @@
...
-
-  should "end the description with ellipsis if truncated" do
-    result = SearchResult.new(SearchParameters.new({}),
-                              "description" => "Long description is long "*100)
-    assert result.description.end_with?('…')
-  end
-
   should "report when no examples are present" do
-    result = SearchResult.new(SearchParameters.new({}), "description" => "I like pie").to_hash
+    result = SearchResult.new(SearchParameters.new({}), {}).to_hash
```
jackscotti: I don't understand why this change.

tijmenb: The second argument is the search_result. We're now just passing in an empty result, which is enough to make the tests pass. I suppose this hash used to contain a value for `description` because we we modified the description a lot and didn't want it to be nil.

jackscotti: :+1:  thanks

PR #873, opened 2015-08-27
Upgrade to Rails 4.2.4

boffbowsh:
Best reviewed commit by commit.

Highlights:

- Convert everything to ActiveSupport test cases
- Routing and rendering changes for print formats, now uses "variants"
- Added govuk-lint in diff mode

PR #884, opened 2015-09-15
Allow pages to be framed

tijmenb:
We use an iframe to show the search result in search-admin. This stopped working since the upgrade to Rails 4 (https://github.com/alphagov/frontend/pull/873), because Rails sends the `X-Frame-Options: SAMEORIGIN` header by default now (https://github.com/alphagov/whitehall/pull/2080).

Trello: https://trello.com/c/1hGnjzaO

### Before

screen shot 2015-09-15 at 16 08 45

### After

screen shot 2015-09-15 at 16 07 54

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -77,5 +77,10 @@ module Frontend
 
     # Disable Content Negotiation
     config.action_dispatch.ignore_accept_header = true
+
+    # Override Rails 4 default which restricts framing to SAMEORIGIN.
+    config.action_dispatch.default_headers = {
+      'X-Frame-Options' => 'ALLOWALL'
+    }
   end
 end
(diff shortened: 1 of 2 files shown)


bradwright:
We're pro-framing (as the Whitehall PR shows) - `Allow-From` would potentially leak some things we don't want leaked.

bradwright:
Given we don't mind framing everywhere (except Transaction start pages: https://github.com/alphagov/frontend/pull/511), this pattern might be useful:

https://github.com/alphagov/specialist-frontend/pull/73

tijmenb:
Sorry, my description wasn't clear without the code - this PR does indeed add `ALLOWALL`, like you say. 

The comment about `Allow-From` was more intended as "this would've been nice but I didn't actually do it" (because the support for that is patchy).

tijmenb:
Updated to allow all pages to be framed. Thanks @bradwright.

bradwright:
:+1:

PR #886, opened 2015-09-28
Revert "Temporary workaround to fix search within a manual"

issyl0:
This reverts commit ef9e4dc as we've fixed the problem properly now
in https://github.com/alphagov/hmrc-manuals-api/pull/109,
https://github.com/alphagov/manuals-frontend/pull/159 and
https://github.com/alphagov/specialist-publisher/pull/564.

### DIFF
--- a/lib/search_api.rb
+++ b/lib/search_api.rb
@@ -54,11 +54,7 @@ private
     end
 
     def scope_object_link
-      @scope_object_link ||= with_leading_slash(params.filter('manual').first)
-    end
-
-    def with_leading_slash(slug)
-      slug.start_with?("/") ? slug : "/#{slug}"
+      @scope_object_link ||= params.filter('manual').first
     end
 
     def unscoped_results


test/unit/search_api_test.rb
```diff
@@ -26,7 +26,7 @@ class SearchAPITest < ActiveSupport::TestCase
 
   context "given a search scoped to a manual" do
     setup do
-      @manual_link = '/manual/manual-name'
+      @manual_link = 'manual/manual-name'
```
jennyd: This line shouldn't be reverted - the value used here is the section's `manual` field which does now have the leading slash, and the test setup data should reflect that. The tests don't make any assertions about this value so they still pass with the slash left in.

issyl0: Ah. I just did a blanket revert on the whole original commit. Amended to not change this line...

PR #892, opened 2015-11-05
Add postcode finder link to local transactions

davidbasalla:
Adds the Royal Mail's Postcode Finder as link to the location form for local transactions. It also adds the Finder to 'licences' pages (https://www.gov.uk/house-to-house-collection-licence) and 'places' pages (https://www.gov.uk/number-plate-supplier), as they use the same partial.

One of the ACs mentions to set up exit link tracking. It appears this comes free from the `static` app, which includes the gem `govuk_frontend_toolkit` and thereby JS code for [tracking clicks on external links](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/analytics/external-link-tracker.js#L6-L9). We've manually tested successfully that the CSS selector includes the new link to the Postcode Finder. We also used the [Google Analytics Debugger Chrome extension](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en) to check that the right event is triggered.

### DIFF
--- a/app/views/root/_location_form.html.erb
+++ b/app/views/root/_location_form.html.erb
@@ -7,14 +7,21 @@
       <% end %>
 
       <% if @location_error %>
-      <div class="location_error error-notification"><%= t @location_error %></div>
+        <div class="location_error error-notification"><%= t @location_error %></div>
       <% elsif params[:postcode] and @publication.places.nil? %>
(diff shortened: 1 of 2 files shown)


h-lame:
Looks good. :+1:

PR #893, opened 2015-11-06
Autocorrect obvious typos in postcodes before sending to MapIt

issyl0:
- We were seeing lots of errors for postcodes entered with obvious
  typos, eg. hyphens instead of spaces, 0 instead of O, I instead of 1,
  etc, so this is an attempt at fixing most common ones. Strip most
  non-alphanumerics and trailing whitespace. Also use the uk_postcode
  gem to fix transposition of I/1, O/0.
- This was originally done in Imminence, thought it would be better in
  GDS API adapters (https://github.com/alphagov/gds-api-adapters/pull/384) for a consistent experience everywhere, but then
  reverted to do it again not centrally, but in each app, as there are
  (at the moment) only two (Imminence and Frontend) that deal with
  user-inputted postcodes.
- Sometimes in frontend, `params[:postcode]` can be nil, so I've made
  `PostcodeSanitizer` not try to sanitize a nil postcode, so that
  everything will continue to work in that case.

Reopening https://github.com/alphagov/imminence/pull/116 as well. Third time lucky! :sunny:

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -3,6 +3,7 @@ require "authority_lookup"
 require "local_transaction_location_identifier"
 require "licence_location_identifier"
 require "licence_details_from_artefact"
+require "postcode_sanitizer"
 
 class RootController < ApplicationController
   include ActionView::Helpers::TextHelper
@@ -76,10 +77,10 @@ class RootController < ApplicationController
 
   def publication
     @publication = prepare_publication_and_environment
-    @postcode = params[:postcode]
+    @postcode = PostcodeSanitizer.sanitize(params[:postcode])
 
     if ['licence', 'local_transaction'].include?(@publication.format)
-      @location = fetch_location @postcode
+      @location = fetch_location(@postcode)
       if @location
         snac = appropriate_snac_code_from_location(@publication, @location)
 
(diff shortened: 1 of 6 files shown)


app/controllers/root_controller.rb
```diff
@@ -164,7 +165,7 @@ def block_empty_format
 
   def prepare_publication_and_environment
     publication = publication_with_places(
-      params[:postcode], params[:slug], params[:edition]
+      PostcodeSanitizer.sanitize(params[:postcode]), params[:slug], params[:edition]
     )
```
h-lame: This is a micro-optimization but it may be worth moving the result of calling `PostcodeSanitizer.sanitize` somewhere where we can memoize the result.  In `publication` we end up doing it twice, once in `prepare_publication_and_environment` and then once immediately below when we assign the result to `@postcode`.

issyl0: I was just about to change this to be better anyway. :+1:

issyl0: Done. Now we only call `PostcodeSanitizer` once as the `@postcode` assignment is above the `prepare_publication_and_environment` call, and that method now uses `@postcode` (and it's still OK if it's nil).

h-lame:
Redo to address [this comment](https://github.com/alphagov/frontend/pull/893#discussion_r44273224) looks good, but I just realised that the `jobsearch` and `legacy_completed_transaction` actions also use `prepare_publication_and_environment` and won't have prepared `@postcode` so we might break something.

In the spirit of making the smallest change and not affecting everything it may be worth just rolling back and having the double sanitization for now - we may be able to address it more cleanly when we merge in #894 which refactors this controller somewhat.

issyl0:
@h-lame Thanks - that makes sense. I've put it back to how my changes were originally.

PR #894, opened 2015-11-06
Log postcode errors

tommyp:
In order to improve the experience for users on local transactions we need to log the error states from the Postcode form. In order to know if the difference between a bad Postcode and a good Postcode with no Mapit match I've bumped `gds-api-adapters to `25.0.0` to include the change in https://github.com/alphagov/gds-api-adapters/pull/381. This means that when entering a bad postcode, the API adapter will return a exception with a 400 code. We can then use this to log the error state.

This PR does that by unfortunately duplicating the code in the `publication` action and added setting an instance var with the error of the variable state. This is also sent to Google Analytics in the template. These error state strings came from our Analyst. The idea is we will be able to scale back duplication as we continue to improve the local transaction format.

This change also picked up a load of lint offences, so the PR also fixes a bunch of those. [Ticket](https://trello.com/c/Db6csIA7/135-log-the-errors-users-see-when-entering-postcodes-on-local-transaction-pages).

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -128,7 +128,7 @@ class RootController < ApplicationController
           # Matching local authority and redirect to publication page
           # with the local authority name. This is the 100% success state.
           # The redirect below redirects back to this action with the `part`
-          redirect_to publication_path(:slug => params[:slug], :part => slug_for_snac_code(snac)) and return
+          redirect_to publication_path(slug: params[:slug], part: slug_for_snac_code(snac)) && return
         else
           # No matching local authority.
           # This points the user towards "Find your LA" which is an
@@ -145,7 +145,7 @@ class RootController < ApplicationController
           snac = AuthorityLookup.find_snac(params[:part])
 
           if request.format.json?
-            redirect_to "/api/#{params[:slug]}.json?snac=#{snac}" and return
+            redirect_to "/api/#{params[:slug]}.json?snac=#{snac}" && return
           end
 
           # Fetch the artefact again, for the snac we have
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -79,8 +79,9 @@ class RootController < ApplicationController
     @publication = prepare_publication_and_environment
     @postcode = PostcodeSanitizer.sanitize(params[:postcode])
 
-    if ['licence', 'local_transaction'].include?(@publication.format)
-      @location = fetch_location(@postcode)
+    if @publication.format == 'licence'
+      @location = fetch_location @postcode
+
       if @location
         snac = appropriate_snac_code_from_location(@publication, @location)
 
@@ -107,6 +108,45 @@ class RootController < ApplicationController
           assert_found(updated_artefact)
           @publication = PublicationPresenter.new(updated_artefact)
         end
+      end
+
+      @interaction_details = prepare_interaction_details(@publication, authority_slug, snac)
+    elsif @publication.format == 'local_transaction'
+      @location = fetch_location @postcode
+
+      if @location
+        # Valid postcode and matching location
+        snac = appropriate_snac_code_from_location(@publication, @location)
+
+        if snac
+          # Matching local authority and redirect to publication page
+          # with the local authority name. This is the 100% success state.
+          # The redirect below redirects back to this action with the `part`
+          redirect_to publication_path(:slug => params[:slug], :part => slug_for_snac_code(snac)) and return
+        else
+          # No matching local authority.
+          # This points the user towards "Find your LA" which is an
+          # England only service
+          @postcode_error = "No matching local authority, sent user to 'Find your LA'."
+          @location_error = "formats.local_transaction.no_local_authority_html"
+        end
+      elsif params[:authority] && params[:authority][:slug].present?
+        return redirect_to publication_path(slug: params[:slug], part: CGI.escape(params[:authority][:slug]))
+      elsif params[:part]
+        # Link to local authority but no direct link to service
+        authority_slug = params[:part]
+
+        snac = AuthorityLookup.find_snac(params[:part])
+
+        if request.format.json?
+          redirect_to "/api/#{params[:slug]}.json?snac=#{snac}" and return
+        end
+
+        # Fetch the artefact again, for the snac we have
+        # This returns additional data based on format and location
+        updated_artefact = fetch_artefact(params[:slug], params[:edition], snac, @location) if snac
+        assert_found(updated_artefact)
+        @publication = PublicationPresenter.new(updated_artefact)
       end
 
       @interaction_details = prepare_interaction_details(@publication, authority_slug, snac)
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -86,12 +86,12 @@ class RootController < ApplicationController
         snac = appropriate_snac_code_from_location(@publication, @location)
 
         if snac
-          redirect_to publication_path(:slug => params[:slug], :part => slug_for_snac_code(snac)) and return
+          return redirect_to publication_path(slug: params[:slug], part: slug_for_snac_code(snac))
         else
           @location_error = "formats.local_transaction.no_local_authority_html"
         end
       elsif params[:authority] && params[:authority][:slug].present?
-        redirect_to publication_path(:slug => params[:slug], :part => CGI.escape(params[:authority][:slug])) and return
+        return redirect_to publication_path(slug: params[:slug], part: CGI.escape(params[:authority][:slug]))
       elsif params[:part]
         authority_slug = params[:part]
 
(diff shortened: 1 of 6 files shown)


app/controllers/root_controller.rb
```diff
@@ -109,12 +110,58 @@ def publication
       end
 
       @interaction_details = prepare_interaction_details(@publication, authority_slug, snac)
+    elsif @publication.format == 'local_transaction'
+      @location = fetch_location @postcode
+
+      if @location.nil?
+        @postcode_error = "fullPostcodeNoMapitMatch"
```
h-lame: I'm not sure that `"fullPostcodeNoMapitMatch"` is the right error here: `fetch_location` returns `nil` when no postcode is provided and doesn't talk to mapit at all.

app/controllers/root_controller.rb
```diff
@@ -109,12 +110,58 @@ def publication
...
+      elsif @location.try(:code) == 400
+        @postcode_error = "invalidPostcodeFormat"
+      end
```
h-lame: Is a `400` the only possible code here?  What about other errors - don't we log them?

app/controllers/root_controller.rb
tommyp: When `fetch_location` returns `nil`, it's the equivalent of a 404 from Mapit. That case happens when there is no match but it is a Postcode.

app/controllers/root_controller.rb
tommyp: Yup. Mapit only returns `nil`, `400` or `200`.

app/controllers/root_controller.rb
h-lame: Oh. I see that it might also return `nil` when mapit can't find a match.  Maybe we could refactor `fetch_location` to return something more useful than a `nil`, a `Location`, or an `GdsApi::BaseError` subclass.

tommyp: This conversation happened over [here](https://github.com/alphagov/gds-api-adapters/pull/378).

h-lame: There's a difference between the `nil` you get from `fetch_location` when the user hasn't entered a postcode at all and the `nil` you get from `fetch_location` when they have entered a postcode but mapit can't find it because it's not a valid postcode.  

The `else` clause further down suggests that there are interactions where we don't have to provide a postcode to proceed (because we've provided a `part` param) and so we shouldn't log a "couldn't find it in mapit" postcode error in this case.

app/controllers/root_controller.rb
h-lame: Ah, ok.

app/controllers/root_controller.rb
tommyp: I get you now. Yeah, that would be the case then. What do you suggest returning from `fetch_location` when Mapit returns `nil`?

tommyp: I could only log a postcode error if `@location.nil? && !params.has_key?(:part)`?

h-lame: I might be inclined to return an object that wraps all this behaviour.

```
class FetchedLocation
  attr_reader :location, :error
  def initialize(location: l, error: e)
    @location = l
    @error = e
  end

  def no_postcode_provided?
    location.nil? && error.nil?
  end

  def location_found?
    location.present?
  end

File: app/views/root/local_transaction.html.erb [hunk 3]
```diff
@@ -58,3 +58,17 @@
     </div>
   </section>
 <% end %>
+
+<% if @postcode_error %>
+  <script type="text/javascript">
+    //<![CDATA[
+      GOVUK.analytics.trackEvent(
+        "userAlerts:LocalTransaction",
+        "postcodeErrorShown:<%= @postcode_error %>",
```
fofr: The format of the category and action being passed to GA is different to many of the existing events. It suggests a type of namespacing that doesn't currently exist, is this the intent? Will there be similar events added later?

The preference for category is a human readable string, eg "External Link Tracked", or "Print Intent". Perhaps something like "Postcode Error".

I'm not sure the `LocalTransaction` bit is needed, this is identified using a Custom Dimension via:
`<meta name="govuk:format" content="local_transaction">`

tommyp: These strings came from the Analyst on our team. I'll send him a link to this comment and see what he says.

app/controllers/root_controller.rb
```diff
@@ -109,12 +110,58 @@ def publication
...
+          # The redirect below redirects back to this action with the `part`
+          return redirect_to publication_path(slug: params[:slug], part: slug_for_snac_code(snac))
+        else
+          # No matching local authority.
+          # This points the user towards "Find your LA" which is an
+          # England only service
+          @postcode_error = "noLaMatchLinkToFindLa"
+          @location_error = "formats.local_transaction.no_local_authority_html"
+        end
+      elsif params[:part]
```
h-lame: We're missing the:

```ruby
      elsif params[:authority] && params[:authority][:slug].present?
        return redirect_to publication_path(slug: params[:slug], part: CGI.escape(params[:authority][:slug]))
```

part of this conditional that the `licence` format path will run through if no location is found.  Does that only apply to licence types?

app/views/root/local_transaction.html.erb
```diff
@@ -58,3 +58,17 @@
     </div>
   </section>
 <% end %>
+
+<% if @postcode_error %>
+  <script type="text/javascript">
+    //<![CDATA[
```
fofr: We don't need the `//<![CDATA` bits. See http://stackoverflow.com/questions/66837/when-is-a-cdata-section-necessary-within-a-script-tag for some details.

Is there a way we could avoid putting an inline script in the page?

tommyp: I'll remove the `CDATA`. As for not having an inline script, suggestions welcome.

app/controllers/root_controller.rb
tommyp: Ah yeah. Good catch. I don't think so. I think this is the state the user gets to when there's a local authority but no link to the service.

app/views/root/local_transaction.html.erb
timlb: Both points are entirely correct according to past examples and current guidance. But there's a project under way to revise our approach to GA events and this is an early example.

The "local_transaction" detail is indeed duplicated in the "format" dimension. Duplication here was 1) to surface it in standard GA reports for users who do not use custom reporting 2) to allow splitting potential "userAlerts" at a high level in reports and so leave more scope for other data at  lower levels.

Moving to camelCase is intended to compress the string slightly because we're adding details here and want to save space in reports.

I've reviewed the comment with Paul Cronk and Peter Jordan. We took the comment on board but would prefer to stick with the new format please.

issyl0:
This will eventually need a rebase due to https://github.com/alphagov/frontend/pull/893, which could be pretty bad. Sorry!

tommyp:
Thanks @issyl0. It was actually pretty painless. Just one line needed changing.

test/functional/local_transactions_controller_test.rb
```diff
@@ -59,6 +59,30 @@ class LocalTransactionsControllerTest < ActionController::TestCase
...
+      setup do
+        mapit_does_not_have_a_postcode("WC1E 9ZZ")
+
+        post :publication, slug: "send-a-bear-to-your-local-council", postcode: "WC1E 9ZZ"
+      end
+
+      should "log the invalid postcode error" do
+        assert_equal assigns(:postcode_error), "fullPostcodeNoMapitMatch"
+      end
+    end
```
h-lame: There's 3 possible postcode errors, but only 2 specs.  We might also want to explicitly test that we don't set postcode errors for non local transaction formats (this might be overkill, I'd be open to being talked down from that).

app/views/root/local_transaction.html.erb
```diff
@@ -58,3 +58,15 @@
...
+<% if @postcode_error %>
+  <script type="text/javascript">
+      GOVUK.analytics.trackEvent(
+        "userAlerts:LocalTransaction",
+        "postcodeErrorShown:<%= @postcode_error %>",
+        {
+          label: "<%= t @location_error || 'Please enter a valid full UK postcode.' %>"
+        }
+      )
+  </script>
```
h-lame: There was a comment from @fofr on a previous version of this diff about avoiding dropping a script tag directly on the page.  Could we (do we have a way to) communicate with analytics from the server?  Even if we did would it be bad because this would a) mess up request/response time or b) not include useful client-side details?

fofr: My preference would be a way of sending analytics data through `data-` attributes. There is a pattern for this in the `govuk_admin_template` that I'm currently looking at porting to the frontend toolkit.

fofr: It shouldn't block this PR though.

davidbasalla:
I've added a test for correct assigning of postcode error for missing local authorities [here](https://github.com/alphagov/frontend/blob/log-postcode-errors/test/functional/local_transactions_controller_test.rb#L86-L96). After discussing with @fofr we decided to address his comment regarding the inline script tags at a later point once all the work for https://github.com/alphagov/govuk_frontend_toolkit/pull/227 is in place.

h-lame:
Looks good.  Merging.

PR #895, opened 2015-11-17
Add LocationError class to encapsulate error logic

davidbasalla:
Previously the JavaScript would break in the script tag inside the `location_transaction` view, as when `@location_error` was `nil`, running `I18n.translate` on `nil` creates a span with missing translation and does not use the second argument of the `||` expression.

To fix this, we refactor by adding a new LocationError class to group the two error related variables set from the root controller and move the logic out of the view and into the class. We also add a locale definition for the 'invalid postcode' error case.


h-lame:
Looks good, seems like a nice approach too.

PR #896, opened 2015-11-19
Log postcode error for missing interaction

davidbasalla:
Adding postcode error code `laMatchNoLink` for where a postcode has a location from MapIt and a Local Authority but no Location Interaction/Link.

Also implements tracking in Kibana by using `LogStasher` gem to log postcode errors in JSON format. Had to upgrade `LogStasher` to release version `0.6.2` to get notification subscriptions to work. Latest version of the gem is `0.8.6` but various things break or change (see commit message for details).

### DIFF
--- a/app/views/root/local_transaction.html.erb
+++ b/app/views/root/local_transaction.html.erb
@@ -21,10 +21,10 @@
        <span class="destination"> on <%= @interaction_details['local_authority']['name'] %></span>
     </p>
 
-  <% elsif @interaction_details['local_authority'] %>
+  <% elsif @location_error && @location_error.no_location_interaction? %>
 
     <div class="application-notice help-notice">
-      <p>We don't have a direct link to this service from <%= @interaction_details['local_authority']['name'] %><% if @interaction_details['local_authority']['contact_url'].present? %>, but you could try the <%= link_to @interaction_details['local_authority']['name'] + " website", @interaction_details['local_authority']['contact_url'], :rel => :external %><% end %>.</p>
+      <p><%= t(@location_error.message, @location_error.message_args) %></p>
     </div>
 
   <% else %>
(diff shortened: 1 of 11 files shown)


app/views/root/local_transaction.html.erb
```diff
@@ -21,10 +21,10 @@
        <span class="destination"> on <%= @interaction_details['local_authority']['name'] %></span>
     </p>
 
-  <% elsif @interaction_details['local_authority'] %>
+  <% elsif @location_error && @location_error.postcode_error == "laMatchNoLink" %>
```
h-lame: Minor nitpick: consider wrapping the 2nd part of this in a method on `LocationError`.  "laMatchNoLink" isn't super clear to the next developer, but `@location_error.no_location_interaction?` would be.

config/locales/en.yml
```diff
@@ -31,6 +31,8 @@ en:
     local_transaction:
       invalid_postcode: "Please enter a valid full UK postcode."
       no_local_authority_html: "Sorry, we can't find the local council for your postcode. Try using <a href='/find-your-local-council'>the local council directory</a>."
+      no_local_interaction: "We don't have a direct link to this service from %{local_authority_name}"
+      no_local_interaction_with_link: "We don't have a direct link to this service from %{local_authority_name}, but you could try the <a href=%{contact_url} rel='external'>%{local_authority_name} website</a>."
```
h-lame: If we named these with `_html` suffixes then we could drop the `.html_safe` bit when we draw them in the view.  See: http://guides.rubyonrails.org/i18n.html#using-safe-html-translations

davidbasalla: oh neat, will do!

app/views/root/local_transaction.html.erb
davidbasalla: sounds good, will add :+1:

PR #897, opened 2015-11-20
Add Autumn statement promo

jackscotti:
What it will look like:

screen shot 2015-11-20 at 15 06 49

Zendesk:
https://govuk.zendesk.com/agent/tickets/1194049


fofr:
The image is 200kb, which feels a bit big. Putting it through https://compressor.io/ (or your lossless image compression tool of choice) reduces it to 50kb.

jackscotti:
I had already pushed that through `imageoptim`. I tried using yours but with the loseless algorithm I do not get any dimension reduction:

screen shot 2015-11-20 at 15 31 40

jackscotti:
Add a chat in real life with Paul and we decided to go for the lossy algorithm. The image now is ~50KB.

alextea:
@jackscotti ImageOptim only does lossless compression, so won't have much effect on jpegs. You need to go into the preferences and enable lossy compression (80% is usually good enough and gives you a 42K image in this instance).

fofr:
:+1: This is good to merge on Monday.

PR #899, opened 2015-11-27
Revert "Nokogiri 1.6.6.5 pre release"

boffbowsh:
Reverts alphagov/frontend#898

Nokogiri will be cutting an [official release soon](https://github.com/sparklemotion/nokogiri/pull/1378#issuecomment-159447729) after more testing. This hacked together release is causing some developers issues with building Nokogiri, and the security vulnerabilities are no longer thought to be too severe.

Reverting all PRs that use this temporary release.

/cc @bradwright


bradwright:
I still think we should be going to `v1.6.6.4`, as without that we're exposed to the more severe CVEs.

PR #905, opened 2016-01-13
Update poltergeist to 1.8.1

steventux:
[Older versions of poltergeist do not always return the correct exit code](https://github.com/teampoltergeist/poltergeist/issues/497) and can lead to test suites failing but CI passing, this behaviour has been seen in [travel advice publisher branch builds](https://ci.dev.publishing.service.gov.uk/job/govuk_travel_advice_publisher_branches/107/console).

See https://github.com/alphagov/travel-advice-publisher/pull/98 for detail.


test/integration/simple_smart_answers_test.rb
```diff
@@ -333,7 +333,7 @@ class SimpleSmartAnswersTest < ActionDispatch::IntegrationTest
         end
 
         # Ensure the edition param isn't added to the form URL as it needs to be in a hidden field.
-        assert_equal "/the-bridge-of-death/y", page.find('form')["action"]
+        assert_match /\/the-bridge-of-death\/y$/, page.find('form')["action"]
```
benilovj: would this be a bit more readable if it were `%r{/the-bridge-of-death/y$}`?

benlovell: Can we use `%r(...)` here, so we can avoid escaping those forward-slashes?

benlovell:
Thanks ✨

dsingleton:
:+1: :100:

PR #906, opened 2016-01-13
Travel advice index migration

tuzz:
This work makes it so that frontend speaks to the content store in order to render the travel advice index page rather than the content api.

Notes:

- We will need to run the republishing rake task in travel-advice-publisher (pre-deploy) (**DONE**)
- The tests now depend on the govuk-content-schemas to supply fixture data
- We should set up a trigger to build frontend downstream of the content schemas project (post-merge)
- We should manually check that the entry ids have not changed in the atom feed
- There's a bit of hardcoding for breadcrumbs, but this will be replaced soon
- The show pages for individual countries are handled by the multipage-frontend application


elliotcm:
@tuzz Probably worth deploying this branch to integration to check the atom entry IDs.

elliotcm:
All in all looking pretty good.  As I was reading through the commits I found a number that felt like they should be merged (and didn't always leave a comment) so probs worth giving them a once over yourself.

steventux:
I've had a look over this as well having paired briefly with @tuzz on some of the slimmer work, looks good once comments are addressed. :+1:

tuzz:
I have addressed pull request comments. Please could you take another look @elliotcm.

I've diff'ed the atom feeds on integration running this branch vs. release branch. I will repeat this test once more on my local machine.

elliotcm:
@tuzz You've got two commits called "Render the atom feed for the index page".  Other than that, good to merge when you're ready.

tuzz:
Good spot @elliotcm, fixed.

Please don't merge yet. We're deploying travel-advice-publisher soon and I want to do another round of testing once that's out.

elliotcm:
:+1:

tuzz:
We have now completed the travel-advice-publisher deploy, so this can be merged if you're happy. :-)

PR #908, opened 2016-01-22
Sort the groups and countries on the index page…

tuzz:
Previously, we were relying on the list in the
content item, which probably isn’t a good idea.
This makes it explicit.

### DIFF
--- a/app/helpers/travel_advice_helper.rb
+++ b/app/helpers/travel_advice_helper.rb
@@ -3,9 +3,11 @@ require 'htmlentities'
 module TravelAdviceHelper
 
   def group_by_initial_letter(countries)
-    countries.group_by do |country|
+    groups = countries.group_by do |country|
       country.name[0] if country and country.name
     end
+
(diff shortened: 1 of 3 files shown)


app/helpers/travel_advice_helper.rb
```diff
@@ -3,9 +3,11 @@
 module TravelAdviceHelper
 
   def group_by_initial_letter(countries)
-    countries.group_by do |country|
+    groups = countries.group_by do |country|
       country.name[0] if country and country.name
     end
+
+    groups.sort_by { |name, _| name }
```
benlovell: Would `groups.sort_by(&:name)` suffice?

tuzz: No,

`sort_by(&:name)` is equivalent to
`#sort_by { |element| element.name }` which is not the same as
`#sort_by { |name, _| name }`

benlovell: :+1:

PR #907, opened 2016-01-21
Update to Local transactions based on prototype

sihugh:
[Trello card](https://trello.com/c/fT1OKmgb/228-make-changes-to-the-shared-content-of-local-transactions-based-on-the-postcode-prototype)

1. Changes to wording/structure of the shared postcode lookup form based on [this prototype](http://local-gov-links.herokuapp.com/pay-council-tax_2)
 - Emphasises "Enter a postcode" text
 - Removes placeholder and places example above instead

2. Changes to "Start now" (or "happy path") page based on [this prototype](http://local-gov-links.herokuapp.com/start-now-simple_2)
 - Emphasises which service provider we've matched to
 - Change "Start now" wording to "Go to their website"

3. Changes to errors / non-happy paths
 In general, we've changed the wording and appearance of the error information to match the following prototypes:
  - [Invalid postcode format](http://local-gov-links.herokuapp.com/invalid-postcode-format)
  - [Correct looking postcode but it doesn't exist in Mapit](http://local-gov-links.herokuapp.com/full-postcode)
  - [We find the postcode but Mapit has no mapping to a local authority](http://local-gov-links.herokuapp.com/no-la-match)

The prototype links are indicative, as is [this google doc](https://docs.google.com/document/d/1u1VG_ytCI8SsxGnIWX0VeIP8TdooPih4Y1M4K_lgGrM/edit)

Screenshots are below

PR #914, opened 2016-01-29
Rework local transaction result pages for "LA found, but no transaction url present" result

h-lame:
In #907 we changed most of the result pages for local transactions based on [the prototype](http://local-gov-links.herokuapp.com/), this PR changes the [remaining page](http://local-gov-links.herokuapp.com/la-match-no-link).

We did this page separately because it relies on changes to be made elsewhere to support it:

1. https://github.com/alphagov/govuk_content_models/pull/356 - to give LocalAuthority models a homepage_url attribute
2. https://github.com/alphagov/publisher/pull/443 - to have publisher import that attribute from the local directgov feed.
3. https://github.com/alphagov/govuk_content_api/pull/237 - to make govuk_content_api put the homepage_url into the JSON representation

**While reviewing this PR check for defensive coding - we should be able to merge and deploy this PR without necessarily merging and releasing/deploying those other PRs**

Trello: https://trello.com/c/td7NzjPG/261-use-local-authority-home-page-url-when-we-don-t-have-a-matching-transaction

### Before

before-lanolink

#### With a "the"

after-lanolink-da-the

#### Without a "the"

after-lanolink-da

## Edgecase

As part of doing this PR we also realised that there was one final extreme edgecase where we find the LA and it has no service link, but it also has no homepage or contact url that we can direct the user towards.  The existing branch was sub-optimal because it would be on a `/<service-slug>/<authority-slug>` URL but would tell the user we didn't know what council they were from.  We changed the messaging to match the url and point them at the local council finder (it's not a great continuation point, but better than nothing).

### Before

before-lanolinknoauthurl

#### With a "the"

after-lanolinknourl-da-the

#### Without a "the"

after-lanolinknourl-da

## Definite Articles

We also remove the extra "the" when talking about the councils.  Some councils already start with "The" (e.g. The Highland Council), for others it sounds strange to use a "the" (e.g. Sunderland City Council) whereas for all where it sounds ok to talk about "The <council>" (e.g. Royal Borough of Greenwich) it sounds ok to not use "the" anyway.  We leave it in though when talking about "The <council> website".

This work a part of https://trello.com/c/wROG0UOv/284-remove-the-from-local-transaction-start-page

#### Without a "the"

before-success-la

#### With a "the"

before-success-the

#### Without a "the"

after-success-nothe

#### With a "the"

after-success-the

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -156,8 +156,11 @@ class RootController < ApplicationController
       end
 
       @interaction_details = prepare_interaction_details(@publication, authority_slug, snac)
-      if local_authority_match_but_no_interaction?(@interaction_details)
-        @location_error = error_for_missing_interaction(@interaction_details)
+      if local_authority_match?(@interaction_details)
+        @local_authority = LocalAuthorityPresenter.new(@interaction_details['local_authority'])
+        if no_interaction?(@interaction_details)
+          @location_error = error_for_missing_interaction(@local_authority)
+        end
       end
 
     elsif @publication.empty_part_list?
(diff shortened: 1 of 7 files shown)


app/controllers/root_controller.rb
```diff
@@ -156,8 +156,11 @@ def publication
...
 
       @interaction_details = prepare_interaction_details(@publication, authority_slug, snac)
-      if local_authority_match_but_no_interaction?(@interaction_details)
-        @location_error = error_for_missing_interaction(@interaction_details)
+      if local_authority_match?(@interaction_details)
+        @local_authority = LocalAuthorityPresenter.new(@interaction_details['local_authority'])
+        if local_authority_match_but_no_interaction?(@interaction_details)
+          @location_error = error_for_missing_interaction(@local_authority)
+        end
       end
```
erikse: This code is new to me, so maybe I missed something, but is it necessary to check if local_authority_matches again, given it's inside the block that already did? 
```ruby
if local_authority_match?(@interaction_details)
  @local_authority = LocalAuthorityPresenter.new(@interaction_details['local_authority'])
  if no_interaction?(@interaction_details)
    @location_error = error_for_missing_interaction(@local_authority)
  end
end
```

h-lame: Good point!

h-lame: Rebased away, thanks for pointing it out.

sihugh:
Looks good to me

PR #915, opened 2016-02-10
Handle No Location Exception from Imminence

brenetic:
# Why
For certain locations, such as the Channel Islands, the Find My Nearest postcodes searches using the Mapit API in Imminence were returning results having found the postcode but would not have found any locations. This resulted in a error below:

# Before:
before

Frontend now shows a more descriptive message for the user:

# After:
after

The work in Imminence related to this is [here](https://github.com/alphagov/imminence/pull/124).

[Trello card](https://trello.com/c/TC6mZatW/265-imminence-errors-on-jersey-postcodes-for-find-my-nearests).

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -262,7 +262,8 @@ protected
       Frontend.imminence_api.places_for_postcode(artefact.details.place_type, postcode, 10)
     end
   rescue GdsApi::HTTPErrorResponse => e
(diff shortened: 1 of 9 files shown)


h-lame:
Looks good.  Minor comment about the GA tracking code, but apart from that :clap: :clap:

brenetic:
@h-lame Added the GA partial for tracking and rebased.

PR #916, opened 2016-02-18
Add report child abuse views

davidbasalla:
For https://trello.com/c/l4RcbEfd/300-front-end-report-child-abuse-changes

This PR prepares the frontend app to display search results for "Report Child Abuse" with a customised view and will only take full effect once the 'report-child-abuse' page has been turned into a `place` format (known as 'find-my-nearest') in [this story](https://trello.com/c/ycz55OuN/292-back-end-report-child-abuse-changes). We also implement GA tracking for clicking the submit button on the postcode search location form, as well as for if any results get shown.

Screenshots below with dummy data to show the visual changes:

# Before:
screen shot 2016-02-18 at 15 51 18

# After:
screen shot 2016-02-18 at 15 48 29

### DIFF
--- a/app/assets/javascripts/modules/track-submit.js
+++ b/app/assets/javascripts/modules/track-submit.js
@@ -1,0 +1,20 @@
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function(Modules) {
+  "use strict";
+
+  Modules.TrackSubmit = function () {
+    this.start = function (element) {
+      element.on('submit', 'form', trackSubmit);
+
+      var category = element.data('track-category'),
+          action = element.data('track-action');
+
+      function trackSubmit() {
+        if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
+          GOVUK.analytics.trackEvent(category, action);
+        }
+      }
+    };
+  };
+})(window.GOVUK.Modules);
--- a/app/assets/stylesheets/views/_report-child-abuse.scss
+++ b/app/assets/stylesheets/views/_report-child-abuse.scss
@@ -1,0 +1,8 @@
+.phone-numbers-report-child-abuse {
+  padding-top: $gutter-half;
+  padding-bottom: $gutter-half;
+
+  p.tel-report-child-abuse {
+    @include core-24;
+  }
+}
--- a/app/views/root/_ga_postcode_error_tracking.html.erb
+++ b/app/views/root/_ga_postcode_error_tracking.html.erb
@@ -1,9 +1,11 @@
-<script type="text/javascript">
-  GOVUK.analytics.trackEvent(
-    "userAlerts:<%= alert_type %>",
-    "postcodeErrorShown:<%= @location_error.postcode_error %>",
-    {
-      label: "<%= t(@location_error.message, @location_error.message_args) %>"
-    }
-  );
-</script>
+<% if @location_error && @location_error.postcode_error %>
+  <script>
+    GOVUK.analytics.trackEvent(
+      "userAlerts:<%= alert_type %>",
+      "postcodeErrorShown:<%= @location_error.postcode_error %>",
+      {
+        label: "<%= t(@location_error.message, @location_error.message_args) %>"
+      }
+    );
+  </script>
+<% end %>
--- a/app/views/root/_location_form.html.erb
+++ b/app/views/root/_location_form.html.erb
@@ -1,4 +1,7 @@
-<div class="find-nearest">
+<div class="find-nearest"
+      data-module="track-submit"
+      data-track-category="placeSearch"
+      data-track-action="postcodeSearchStarted">
 
   <% if @location_error %>
     <div class="error-summary">
--- a/app/views/root/_option_report_child_abuse.html.erb
+++ b/app/views/root/_option_report_child_abuse.html.erb
@@ -1,0 +1,43 @@
+<% places ||= [] %>
+<% places.each do |place| %>
+  <li>
+    <div class="place group">
+      <div class="information">
+        <div class="address vcard">
+          <div class="adr org fn">
+
+            <p class="adr">
+              <% if place["name"].present? %>
+                <span class="fn">You can call the children's social care team at the council in <%= place["name"] %></span>
+              <% end %>
+            </p>
+
+            <div class="phone-numbers-report-child-abuse">
+              <% if place["phone"].present? %>
+                <p class="tel-report-child-abuse">
+                  <% phone_digits = phone_digits(place["phone"]) %>
+                  <a class="tel" href="tel://<%= phone_digits %>"><%= phone_digits %></a>
+                  <%= phone_text(place["phone"]) %>
+                </p>
+                <% end %>
+
+              <% if place["general_notes"].present? %>
+                <p class="tel-report-child-abuse">
+                  <% out_of_hours_digits = phone_digits(place["general_notes"]) %>
+                  <a class="tel" href="tel://<%= out_of_hours_digits %>"><%= out_of_hours_digits %></a>
+                  <%= phone_text(place["general_notes"]) %>
+                </p>
+              <% end %>
+            </div>
+
+            <% if place["url"].present? %>
+              <p class="url">
+                <a href="<%= place["url"] %>">Go to their website</a>
+              </p>
+            <% end %>
+          </div>
+        </div>
+      </div>
+    </div>
+  </li>
+<% end %>
--- a/app/views/root/place.html.erb
+++ b/app/views/root/place.html.erb
@@ -37,6 +37,4 @@
   <% end %>
 <% end %>
 
-<% if @location_error && @location_error.postcode_error %>
-  <%= render partial: 'ga_postcode_error_tracking', locals: { alert_type: 'place' } %>
-<% end %>
+<%= render partial: 'ga_postcode_error_tracking', locals: { alert_type: 'place' } %>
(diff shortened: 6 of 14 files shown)


app/views/root/_option_report_child_abuse.html.erb
```diff
@@ -0,0 +1,43 @@
...
+<% places.each do |place| %>
+  <li>
+    <div class="place group">
+      <div class="information">
+        <div class="address vcard">
+          <div class="adr org fn">
+
+            <p class="adr">
+              <% if place["name"].present? %>
+                <span class="fn">You can call the children's social care team at <%= place["name"] %> Council</span>
```
h-lame: Would it be better to leave "Council" off here and make sure it's included in the names of all the places in the data-set?  This allows for the names to be the full council name and include their own preferred way of naming themselves.  Also means we don't have to have the name be "Blahville District" or "Areaton Borough" if that part of the name is important.

app/helpers/report_child_abuse_helper.rb
```diff
@@ -0,0 +1,11 @@
+module ReportChildAbuseHelper
```
h-lame: Could this be `PhoneNumberHelper` ?  I know it's only used in the report child abuse partial for now, but maybe it can be used elsewhere?  I guess there's a line to walk around how custom this regexp parsing is for the exact data we know we're getting from this dataset.

app/assets/javascripts/modules/track-submit.js
```diff
@@ -0,0 +1,21 @@
+window.GOVUK.Modules = window.GOVUK.Modules || {};
```
brenetic: @fofr ^^ Unsure about this?

fofr: That makes sense. I was wondering whether this should be done in just the tests. It adds some protection against JS errors if this didn't run, for whatever reason: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/modules.js#L4

So I think it makes sense.

app/assets/javascripts/modules/track-submit.js
```diff
@@ -0,0 +1,21 @@
...
+
+(function(Modules) {
+  "use strict";
+
+  Modules.TrackSubmit = function () {
+    this.start = function (element) {
+      element.on('submit', 'form', trackSubmit);
+
+      var options = {nonInteraction: 1}, // automatic events shouldn't affect bounce rate
```
fofr: When someone interacts with a form, is that really a `nonInteraction`?

app/assets/javascripts/modules/track-submit.js
```diff
@@ -0,0 +1,21 @@
...
+          category = element.data('track-category'),
+          action = element.data('track-action');
```
fofr: Should we be capturing a label too?

app/assets/stylesheets/views/_report-child-abuse.scss
```diff
@@ -0,0 +1,8 @@
+.phone-numbers-report-child-abuse {
+  padding-top: 15px;
+  padding-bottom: 25px;
+
+  p.tel-report-child-abuse {
+    font-size: 25px;
```
fofr: Can this use one of the govuk_frontend_toolkit core mixins?
Perhaps this one: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_typography.scss#L70

More details:
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/mixins.md#typography

app/assets/stylesheets/views/_report-child-abuse.scss
```diff
@@ -0,0 +1,8 @@
+.phone-numbers-report-child-abuse {
+  padding-top: 15px;
+  padding-bottom: 25px;
```
fofr: Rather than magic numbers, could we use the standard `$gutter` values? Or a 5px variant of?

app/assets/javascripts/modules/track-submit.js
brenetic: In this instance the label was not requested to be used in the story.

app/views/root/_ga_postcode_error_tracking.html.erb
```diff
@@ -1,9 +1,11 @@
...
-  GOVUK.analytics.trackEvent(
-    "userAlerts:<%= alert_type %>",
-    "postcodeErrorShown:<%= @location_error.postcode_error %>",
-    {
-      label: "<%= t(@location_error.message, @location_error.message_args) %>"
-    }
-  );
-</script>
+<% if @location_error && @location_error.postcode_error %>
+  <script type="text/javascript">
```
fofr: `type="text/javascript"` is redundant here.

Can this use the `auto-track-event` module, and the values placed as attributes elsewhere?

Inline script tags are one of the reasons we can't adopt CSP https://en.wikipedia.org/wiki/Content_Security_Policy

app/views/root/_location_form.html.erb
```diff
@@ -1,4 +1,7 @@
-<div class="find-nearest">
+<div class="find-nearest"
+      data-module="track-submit"
+      data-track-category="placeSearch"
+      data-track-action="postcodeSearchStarted">
```
fofr: Minor style point: HTML attribute values should be lowercase and hyphenated

test/javascripts/unit/modules/track-submit.spec.js
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it('tracks submit events', function() {
+    spyOn(GOVUK.analytics, 'trackEvent');
+
+    element = $('\
+      <div \
+        data-track-category="category"\
+        data-track-action="action">\
+        <form method="post" id="formshit">\
```
fofr: :neutral_face:

app/assets/javascripts/modules/track-submit.js
fofr: It means a module that looks and behaves like the `auto-track-event` one has an inconsistency. As this is limited to the frontend app I guess it's fine, but if it was moved into static or the toolkit this could catch someone out.

app/views/root/_ga_postcode_error_tracking.html.erb
brenetic: That is something that we should definitely change this to, but for this story have opted to keep it as it is since there are a number of different postcode errors used in both local authorities and find my nearest. We were intending on writing another story to make this change.

test/javascripts/unit/modules/track-submit.spec.js
brenetic: :flushed:

app/assets/stylesheets/views/_report-child-abuse.scss
brenetic: :+1:

app/assets/stylesheets/views/_report-child-abuse.scss
brenetic: :100:

app/views/root/_ga_postcode_error_tracking.html.erb
brenetic: Good point. We will keep things consistent and send something in the label.

app/views/root/_location_form.html.erb
brenetic: Category and action are what the analytics guys are searching for specifically. Making them hyphenated here would make the inconsistent with other events in GA.

app/assets/stylesheets/views/_report-child-abuse.scss
```diff
@@ -0,0 +1,8 @@
+.phone-numbers-report-child-abuse {
+  padding-top: $gutter-half;
+  padding-bottom: $gutter-half;
+
+  p.tel-report-child-abuse {
+    @include core-24
```
fofr: Missing a closing `;` here

davidbasalla: :+1:

fofr:
Changes LGTM :+1:

app/views/root/place.html.erb
```diff
@@ -10,17 +10,21 @@
...
 
   <% if @postcode.present? and !@publication.places.nil? %>
     <% if @publication.places.any? %>
-      <section class="places">
-        <h2>Results near <strong><%= @postcode %></strong>:</h2>
+      <section class="places"
+               data-module="auto-track-event"
+               data-track-category="placeSearch"
+               data-track-action="postcodeResultShown"
+               data-track-label="<%= places_names(@publication.places) %>">
```
h-lame: As discussed face-to-face: I'm mildly wary of this.  For a standard "find-my-nearest" we would be sending the names of 10 things which could be fairly large, and names are the sort of thing that might change over time.  We also have the issue that we join the names on `,`, but `,` can be present within the name, so performing an unjoin operation would be hard.  We could use the `CSV` library to get around this - but maybe other punctuation could be considered (wrapping in `[]` and joining on `,` might work).

davidbasalla: After talking to Tim (our analyst) we settled on setting the label to be the name of the first result. This addresses most of the concerns around this issue. See commit notes for [rebased commit](https://github.com/alphagov/frontend/commit/380eb25c60c5a2dae368df3128aa9a47cc0d97b3): 

"... 

We also add the name of the first places result as a label for GA tracking. This allows for specific tracking of shown results in the DfE campaign to promote the "Report child abuse" page. For non-report-child-abuse cases this is not necessarily useful, but it is deemed as the best compromise to allow tracking for this specific campaign and may be extended in future (by sending all results) for other formats if useful."

PR #917, opened 2016-02-22
Make Travel Advice index format match leaf pages

bradwright:
The leaf pages of Travel Advice are served with a format of `travel_advice`. This makes the index page format match that name (and also all other underscore separated formats).


bradwright:
Actually this is wrong - changing this value here will affect CSS all over the site. Closing to re-work.

PR #918, opened 2016-02-23
Use GSS code in place of (non-existent) SNAC codes for new NI councils (except Belfast)

sihugh:
Recent Northern Irish authorities have no SNAC code which breaks compatibility
with several of our applications.  We will use the GSS code as a SNAC as a temporary fix until GOV.UK moves wholesale to GSS codes.

Note that this PR does not change Belfast because its slug (which must match with Licensing) do not change and we don't want to stop sending traffic to Belfast before we switch over to use the new mapit database that will include these GSS codes (see: https://github.com/alphagov/mapit/pull/13)

This is for https://trello.com/c/kjurG3i8/302-support-new-northern-ireland-local-authorities-for-licensing

### DIFF
--- a/lib/data/authorities.json
+++ b/lib/data/authorities.json
@@ -2163,5 +2163,10 @@
     "name": "Belfast City Council",
     "ons": "95Z",
     "gss": null
+  },
+  "north-down-and-ards": {
+    "name": "North Down and Ards",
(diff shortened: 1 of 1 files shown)


sihugh:
This looks fine :-)

brenetic:
:+1:

PR #920, opened 2016-02-25
Fix IE11 text overlapping bug on done pages

issyl0:
- Users were reporting a bug on IE11 where the first done page
  feedback question was overlapping the related links on the left.
  This change fixes it with some new CSS for the `legend` HTML, and
  applies it everywhere. It makes it no worse in Chrome etc., but a
  lot better in IE11.

Before:

before

After:

after

Trello: https://trello.com/c/sQQiAeyy/309-done-page-ie11-rendering-issue-5

### DIFF
--- a/app/assets/stylesheets/views/_completed-transaction.scss
+++ b/app/assets/stylesheets/views/_completed-transaction.scss
@@ -12,6 +12,10 @@
   form {
     @include core-19;
 
+    legend {
+      width: 100%;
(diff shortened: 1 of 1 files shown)


issyl0:
Thanks to @dsingleton for narrowing this down with me and recommending adding `width` instead of `float`.

dsingleton:
:100:

PR #921, opened 2016-02-25
Fix full width for campaign pages

davidbasalla:
This fixes the bug where 'campaign' pages were not displayed at full width as intended. The reason was that `RootController` did not discern between custom slugs and custom formats correctly, as it only ever checked if a publications 'slug' was in the keys of the CUSTOM_FORMATS hash. In the case of 'campaign' pages, it's the actual 'format' (not the 'slug') that's different.

Besides updating `RootController` to account for the bug, we add integration tests to check that the appropriate pages like 'campaign' format pages are displayed at full width. We also add functional tests to check that `RootController`'s calls the `render` method with the correct parameters for custom cases.

# Before:
screen shot 2016-02-25 at 18 18 42

# After:
screen shot 2016-02-25 at 18 19 17

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -46,6 +46,9 @@ class RootController < ApplicationController
     }
   }
 
+  def self.custom_formats; CUSTOM_FORMATS; end
+  def self.custom_slugs; CUSTOM_SLUGS; end
+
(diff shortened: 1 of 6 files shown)


h-lame:
Looks like the right fix to me.

PR #922, opened 2016-02-26
Use test helper for places for postcode

brenetic:
- Bumped gds-api-adapters gem to 28.3.1. This contains an additional `imminence_has_places_for_postcode` test helper which was added [here](https://github.com/alphagov/gds-api-adapters/commit/34be7baf9bc798149edec3c0687649885dd9830d)
- Places test has been changed to use the above. The reasoning for this was that there was already a test helper for areas but not for places.


h-lame:
Nice.

PR #927, opened 2016-03-09
Remove `test-for-content-store` code

tijmenb:
This was added in 2014 as a proof-of-concept (https://github.com/alphagov/frontend/pull/613). It's only activated in development and integration (via initializer-overwrite on deploy), so safe to remove.


tijmenb:
https://github.gds/gds/alphagov-deployment/pull/1158 removes the feature flag from the initialisers.

PR #929, opened 2016-03-09
Extract random page into own controller

tijmenb:
As part of some upcoming work I'm going through this app to try to understand all the things it does. I'm taking the opportunity to do some minor refactoring / extraction to make that job easier.

This PR extracts the gov.uk/random functionality from the `RootController`, which has grown way too large. No change in functionality intended.

Previously: https://github.com/alphagov/frontend/pull/927


issyl0:
:star: Thanks! And I'm glad the RootController is being split up!

benlovell:
Nice :heart_decoration:

PR #930, opened 2016-03-09
Remove ActionMailer

tijmenb:
This application does not send emails. Presumably the action_mailer framework with SES config was added to support error notification (we used to use a email-based thing, but that has all been replaced by
Errbit/Airbrake).

This will speed up the load time for the app slightly.

Previously: https://github.com/alphagov/frontend/pull/929


benlovell:
Another good one to remove is `therubyracer` :+1:

PR #938, opened 2016-04-14
Tweaks to homepage promo banner design

dsingleton:
- More prominent heading
- Fix rendering of external link icon at smaller type sizes (which are automatic from `core-x` mixins).

bigger banner

CC @boffbowsh @fofr @joelanman


fofr:
:+1:

PR #939, opened 2016-04-14
Add register-to-vote callout to homepage

dsingleton:
Add to the browse categories as a short term highlight.

As there are alreadys some register-to-vote links on the page
add a query param to distinguish which was used when looking
at traffic analytics.

rtv-category


fofr:
:+1:

boffbowsh:
👍

PR #940, opened 2016-04-15
Add Register to vote link to popular links block

boffbowsh:
Remove bottom link temporarily
Use same source tracking as other vote links

/cc @dsingleton @fofr


fofr:
Can we add tracking to all those links while we are here?

boffbowsh:
Done.

dsingleton:
:+1: 

Register to vote has been one of the most popular pages on GOV.UK over
the last week or two, and will likely continue to be, at least until registration closes.

Longer term we should look at updating these links, tho they do roughly represent popular content, they've not been properly updated since late 2014.

PR #942, opened 2016-04-18
Add register to vote done page promotion

h-lame:
For: https://trello.com/c/dXgKpUBZ/340-register-to-vote-promotion-on-done-pages-3

In publisher we allow choosing the organ donation promo or the register to vote promo.  This PR changes frontend to understand this choice and show the appropriate promo.  It has to handle two ways of an artefact storing that choice, as seen in https://github.com/alphagov/govuk_content_models/pull/374 and https://github.com/alphagov/publisher/pull/468.

## Screenshot

get information about your fleet vehicles - gov uk

### DIFF
--- a/app/presenters/publication_presenter.rb
+++ b/app/presenters/publication_presenter.rb
@@ -19,7 +19,7 @@ class PublicationPresenter
   PASS_THROUGH_DETAILS_KEYS = [
     :body, :short_description, :introduction, :need_to_know, :video_url,
     :summary, :overview, :name, :video_summary, :continuation_link, :licence_overview,
-    :link, :will_continue_on, :more_information, :downtime, :presentation_toggles,
+    :link, :will_continue_on, :more_information, :downtime,
     :alternate_methods, :place_type, :min_value, :max_value, :organiser, :max_employees,
     :eligibility, :evaluation, :additional_information, :contact_details, :language, :country,
     :alert_status, :change_description, :caption_file, :nodes, :large_image, :medium_image, :small_image,
@@ -129,9 +129,46 @@ class PublicationPresenter
     end
   end
 
+  def promotion_choice
+    choice = promotion_choice_details['choice']
+    if choice.empty?
+      if has_legacy_organ_promotion_details?
+        "organ_donor"
+      else
+        "none"
+      end
+    else
+      choice
+    end
+  end
+
+  def promotion_url
+    url = promotion_choice_details['url']
+    url.empty? ? legacy_organ_promotion_details['organ_donor_registration_url'] : url
+  end
+
   def to_json
     {
       places: places
     }.to_json
+  end
+
+
+private
+
+  def has_legacy_organ_promotion_details?
+    legacy_organ_promotion_details['promote_organ_donor_registration']
+  end
+
+  def legacy_organ_promotion_details
+    presentation_toggles.fetch('organ_donor_registration', {'promote_organ_donor_registration' => false})
+  end
+
+  def promotion_choice_details
+    presentation_toggles.fetch('promotion_choice', {'choice' => '', 'url' => ''})
+  end
+
+  def presentation_toggles
+    details.fetch('presentation_toggles', {})
   end
 end
--- a/app/views/root/completed_transaction.html.erb
+++ b/app/views/root/completed_transaction.html.erb
@@ -18,11 +18,21 @@
       </p>
     </div>
   <% elsif @publication.promotion_choice == 'register_to_vote' %>
-    <div id="register-to-vote-promotion">
+    <div id="register-to-vote-promotion"
+         data-module="auto-track-event"
+         data-track-category="linkTrack"
+         data-track-action="linkDisplayed"
+         data-track-label="Register">
       <p>You must <a href="/register-to-vote">register to vote</a> by 7 June if you want to take part in the EU referendum. You can register online and it only takes 5 minutes.</p>
       <p>
         <%= link_to 'Register', @publication.promotion_url,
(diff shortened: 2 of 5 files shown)


app/presenters/publication_presenter.rb
```diff
@@ -129,9 +129,46 @@ def video_embed_url
...
+      end
+    else
+      choice
+    end
+  end
+
+  def promotion_url
+    url = promotion_choice_details['url']
+    url.empty? ? legacy_organ_promotion_details['organ_donor_registration_url'] : url
+  end
```
h-lame: This is a faff and we've duplicated some logic already in publisher, would it be better to do this elsewhere?

app/views/root/completed_transaction.html.erb
```diff
@@ -8,18 +8,27 @@
...
       <p>
-        <%= link_to 'Join', organ_donor_registration_attributes['organ_donor_registration_url'],
+        <%= link_to 'Join', @publication.promotion_url,
               title: "Register to become an organ donor", rel: "external", class: "button", role: "button" %>
       </p>
     </div>
-    <h2 class="satisfaction-survey-heading">Satisfaction survey</h2>
+  <% elsif @publication.promotion_choice == 'register_to_vote' %>
+    <div id="register-to-vote-promotion">
+      <p>Lorem ipsum.</p>
```
issyl0: We'll need to change this to be non-placeholder text.

h-lame: Argh, I had, but lost it in a rebase.  Thanks for spotting.

h-lame:
Marked as [WIP] because we still have to decide about GA tracking on the links.

app/presenters/publication_presenter.rb
davidbasalla: Agreed, but where would that be? Would it help to introduce some sort of `Promotion` class that encapsulates this logic, and attributes such as display_text, url, etc?

h-lame: Well, yeah, that's the question I think.  I'm not sure where would be good.  Maybe we should do more in Publisher so that upstream doesn't need to care, or maybe we should do it in content-api?  Or maybe we just live with it?

h-lame:
With the addition of 376e95187f9a8d4aa7634a49634a3c7fa336758e and a4533c2f8e45bbe45b2bd6665df872463b14ec65 this PR is now feature complete and no longer WIP.  Your reviews may indicate otherwise ;)

fofr:
Is there a screenshot?

app/presenters/publication_presenter.rb
davidbasalla: I can’t think of any improvements to this that don’t raise more questions. 

I was wondering whether the `frontend` app knows too much about promotions in general, and that all ‘promotion’ related data should come from the artefact (i.e. title, text, url, link_text). That would imply the promotion data would be set in `Publisher` already, but causes the issue of how to update promotions across multiple sites, once they have been set (although in theory we already have that problem as the promotion URL is set in ‘Publisher'). Also the promotion ‘data’ should probably be identical for each done page, so each promotion should come with a default URL/text.

https://github.com/alphagov/govuk_content_models/pull/374 mentions “...so that eventually we can remove the old, organ donor-specific code in favour of "any promo you want as long as it's in the allowed list of `promotion_choice`s”, so maybe it’s best to hold out for that and do a refactor then.

Would like to have another chat face-to-face on Monday, but happy to merge otherwise.

h-lame: After discussion what we think is a good solution is:

1. deploy these complicated versions of frontend and publisher that support both old and new style of promo storage
2. deploy a data-migration that re-writes promo storage for all documents into new style only
3. deploy new simpler versions of frontend and publisher that only support new style of promo storage

Note that 2 and 3 are, as yet, unwritten.

h-lame:
@fofr screenshot added

fofr:
Will we also remove the vote spot on the right? Is it _a bit much_?

app/views/root/completed_transaction.html.erb
```diff
@@ -8,18 +8,37 @@
...
-    <h2 class="satisfaction-survey-heading">Satisfaction survey</h2>
+  <% elsif @publication.promotion_choice == 'register_to_vote' %>
+    <div id="register-to-vote-promotion"
+         data-module="auto-track-event"
+         data-track-category="linkTrack"
+         data-track-action="linkDisplayed"
+         data-track-label="Register">
+      <p>You must <a href="/register-to-vote">register to vote</a> by 7 June if you want to take part in the EU referendum. You can register online and it only takes 5 minutes.</p>
+      <p>
+        <%= link_to 'Register', @publication.promotion_url,
```
fofr: Why "Register" and not "Register to vote" as the button text?

There are no headings above the button to suggest that "Register" relates to voting, and not the transaction you just completed.

issyl0:
@fofr The register to vote spot in the sidebar was @dsingleton's thing and only meant to be temporary - what does he think?

app/views/root/completed_transaction.html.erb
h-lame: These are good questions - I'll see if I can find out.  The current form does follow the example of the organ donor though (no header, brief button text), maybe it's for consistency?

dsingleton:
The register to vote spot in the sidebar is intended to be short-lived, but I can't give an exact date of when it will come down.

In the mean time i'd be fine with this being merged as long as editors know not to pick register-to-vote while the sidebar promo is active.

brenetic:
👍

PR #943, opened 2016-04-19
Remove source tracking on non RTV links

dsingleton:
According to our analytics teams this isn't super helpful and can
actually cause some problems once the links with source tracking
are copied/used elsewhere.

The source tracking is still needed for RTV links as we have 3 on the
same page, at least while we establish which perform better.
Tracking for the other popular links was added at the same time, but
isn't needed right now, so i'm removing it to avoid potential issues.


boffbowsh:
Fair enough

PR #945, opened 2016-04-22
Improve EU referendum homepage link copy for SRs.

tvararu:
Screen reader users have the option to navigate by jumping between anchor tags and reading out the content. "More information" does not provide a lot of context about what the link does.

### After

screen shot 2016-04-22 at 15 22 26


fofr:
The "about the EU referendum" could be wrapped in a `<span class="visuallyhidden">`

tvararu:
Great suggestion @fofr, I've updated it.

I'm going to open an equivalent PR on `static` for the global bar.

tvararu:
Done: https://github.com/alphagov/static/pull/775

fofr:
:claps:

edwardhorsford:
Nice work guys!

PR #948, opened 2016-04-28
Update GSS codes for Northumberland, East Herts and Gateshead councils

issyl0:
- Our upgrade of MapIt meant that these GSS codes changed.

(Trello: https://trello.com/c/iVFolVCb/376-gss-code-fixes-3.)


jennyd:
Could you add some more of the context around this change to the commit message, so that we can see in future what happened?

issyl0:
@jennyd Yep, I will do once I've finished working it out myself. Let's talk in person.

PR #949, opened 2016-05-03
Update tour page

alexmuller:
Lots of this hasn't been modified in ages. I've updated the bits which sound the most old.


fofr:
LGTM :+1:

PR #951, opened 2016-05-10
Make register-to-vote callout copy clearer

dsingleton:
We've seen some issues with users mistakenly thinking they need to register
for the referendum specifically. Be clearer.

### DIFF
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -62,7 +62,7 @@
             <ol class="categories-list">
               <li>
                 <h3><a href="/register-to-vote?source=homepage-category">Register to vote</a></h3>
-                <p>Register to vote in elections and referendums. It takes less than 5 minutes.</p>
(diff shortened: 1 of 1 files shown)


app/views/root/index.html.erb
```diff
@@ -62,7 +62,7 @@
             <ol class="categories-list">
               <li>
                 <h3><a href="/register-to-vote?source=homepage-category">Register to vote</a></h3>
-                <p>Register to vote in elections and referendums. It takes less than 5 minutes.</p>
+                <p>You need to be registered to vote in the EU referendum. It takes less than 5 minutes.</p>
```
fofr: I don't think this makes it clearer. I'm not sure what clarity we're seeking though.

fofr: Doesn't this imply that you _do_ need to register to vote for the EU referendum specifically?

dsingleton: Updated, my bad on grabbing the wrong copy.

PR #955, opened 2016-05-16
Remove scoping by manual

tijmenb:
This feature was added in https://github.com/alphagov/frontend/pull/781.

This doesn’t actually work, because the filter box is not displayed correctly:

For example:

https://www.gov.uk/search?q=a&filter_manual[]=guidance/content-design
https://www.gov.uk/search?q=x&filter_manual[]=service-manual

It will show this:

screen shot 2016-05-16 at 18 07 04

After talking with @rboulton, I’m confident nothing is using this functionality. It can therefore be removed to avoid confusion.

### DIFF
--- a/app/models/search_parameters.rb
+++ b/app/models/search_parameters.rb
@@ -6,7 +6,7 @@ class SearchParameters
   DEFAULT_RESULTS_PER_PAGE = 20
   MAX_RESULTS_PER_PAGE = 100
   ALWAYS_FACET_FIELDS = %w{organisations}
-  ALLOWED_FACET_FIELDS = %w{organisations topics manual}
+  ALLOWED_FACET_FIELDS = %w{organisations topics}
 
   # specialist_sectors will be renamed to topics at some point.  To avoid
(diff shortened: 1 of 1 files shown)


jennyd:
Search within a manual does work at the moment, and this change breaks it. Manuals have a search box in the header which is scoped to the manual eg [Compliance Operational Guidance](https://www.gov.uk/hmrc-internal-manuals/compliance-operational-guidance):

screen shot 2016-05-17 at 10 24 00

Searching via that box gives you a page like this, which presents results from that manual and the top 3 results from all of GOV.UK:

screen shot 2016-05-17 at 10 26 13

In dev on this branch that page is formatted in the same way, but the results which should be scoped to the manual are from all of GOV.UK:

screen shot 2016-05-17 at 10 27 44

tijmenb:
Argh. I assumed that the scoping with the multi-select box would be separate from the "mixed scoping" for manuals. Thanks for catching this.

fofr:
Could the removal of that facet raise some test errors to indicate its use?

PR #956, opened 2016-05-17
Change <h1> to <h2> for transaction additional info block

tvararu:
Transactions have blocks of text that follow the main content. The main content has a `<h1>` level header already, and HTML5 limits the total suggested number of those to just 1.

The styling for `<h1>`s inside the transactional blocks are all overriden to look almost identical to `<h2>`s anyway, so it seems like they should have been `<h2>`s to begin with.

Will affect pages such as http://frontend.dev.gov.uk/register-to-vote.

### Before

screen shot 2016-05-17 at 14 16 50

### After (very subtle margin differences)

screen shot 2016-05-17 at 14 16 56


nickcolley:
LGTM 👍

PR #957, opened 2016-05-20
Update the organ donation registration text with info about Wales

issyl0:
- A while ago (#901 and #925), we added some text about organ donation
  in Wales being opt-out, then removed it because the wording was
  confusing. This is different wording, as requested in
  https://govuk.zendesk.com/agent/tickets/1333304.


fofr:
Does the link need a `rel="external"`?

issyl0:
@fofr: Thanks. Done.

PR #953, opened 2016-05-13
[Do not merge] Remove banner promo slot from homepage

fofr:
__Deploy after May 26 4pm__
https://trello.com/c/npd5H4IW/405-sitewide-banner-remove-supporting-banner-code

Reverts PRs #936 #952 #938

We don’t intend to use this again soon.

PR #959, opened 2016-05-24
[do-not-merge] Remove referendum/register to vote promos from homepage

dsingleton:
**Removed**
- Register to vote from popular links, brings back the link that it replaced
- Register to vote "browse category"

**Replaced**
- EU ref promo slot, with Job search
- Register to vote blue box, with bank holidays

**Screenshot**
new

Note, the light-blue banner is removed separately in https://github.com/alphagov/frontend/pull/953

CC @nickcolley, @fofr, @boffbowsh


boffbowsh:
👍

app/views/root/index.html.erb
```diff
@@ -173,10 +169,10 @@
           <div class="departments-promo-sections">
             <div class="promo-image">
               <div class="promo-content">
-                <a href="/government/topical-events/eu-referendum"><%= image_tag 'homepage/eu-referendum.jpg', alt: 'EU referendum' %></a>
-                <h3>EU referendum</h3>
+                <a href="/jobsearch"><%= image_tag 'homepage/find-a-job.png', alt: 'Find a job' %></a>
+                <h3>Find a job</h3>
                 <p>
-                  <a href="/government/topical-events/eu-referendum">Vote on Thursday 23 June</a> and help to decide if the UK remains part of the European Union.
+                  Use <a href="/jobsearch">Universal Jobmatch</a> to find full or part-time work in Great Britain and abroad.
```
nickcolley: In terms of a11y do we want to duplicate links?

Think it'd be better to reduce the amount of duplicate links, could consider putting `aria-hidden`?

cc @alphagov/accessibility

dsingleton: Yeah, I considered this - and putting an empty `alt` as the solution. But it's the same pattern as the other promo slots on the page, and I feel it doesn't block this change.

I'd be happy with a separate PR to improve the a11y for all the promo slots.

PR #960, opened 2016-05-31
[Do not merge] Remove authorities file

davidbasalla:
This removes the `authorities.json` file and its dependencies from Frontend, having moved the file to MapIt (https://github.com/alphagov/mapit/pull/20). We take the time to refactor some code to make it easier to understand its intention.

Note: Relies on https://github.com/alphagov/gds-api-adapters/pull/515, will need to wait until that PR is merged and the new version is used in this PR before the tests will pass.

Note: LocalLinksManager will also have to be updated before this gets merged as it currently imports the `authorities.json` file to attach slugs to LocalAuthorities

WIP - Needs a refactor to use MapIt routes for `code/govuk_slug/`


lib/authority_lookup.rb
```diff
@@ -1,39 +1,32 @@
...
-
-  private
+private
 
-  def self.find_code(type, slug)
-    self.load_authorities unless self.authorities
-    return self.authorities[slug][type] if self.authorities[slug]
-    nil
+  def authority_by_slug
+    # fairly expensive call at the moment. Could this be a supported mapit route?
```
jennyd: Once we've got LA slugs in our MapIt, we should be able to use [an existing MapIt endpoint](https://github.com/mysociety/mapit/pull/242) to find the authority by slug. That endpoint is at `/code/<code_type>/<code>`, so we'd make a request to MapIt at `/code/govuk_slug/lambeth.json` which would redirect to the area with that slug, and that response would include the snac. For example http://mapit.mysociety.org/code/gss/E06000057.json redirects to http://mapit.mysociety.org/area/2248.json.

davidbasalla: Oh nice, that should clean things up quite a bit

davidbasalla:
Closed in favour of https://github.com/alphagov/frontend/pull/970

PR #968, opened 2016-06-12
Fix focus outlines on homepage images

robinwhittleton:
The links surrounding the images on the homepage are `display: inline`. Different browsers treat this in different ways with our default focus styles. Most look really ugly; Chrome doesn’t show anything, which is why I assume it’s gone unnoticed. Making it `display: block` fixes this. Before / after screenshots:

Browser | Before | After
------- | ------ | -----
Chrome | chrome_before | chrome_after
Firefox | firefox_before | firefox_after
IE8 | ie8_before | ie8_after
IE9 | ie9_before | ie9_after
IE10 | ie10_before | ie10_after
IE11 | ie11_before | ie11_after
Edge | edge_before | edge_after

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -124,6 +124,9 @@ body.homepage {
     }
     .promo-content {
       padding: 0 15px;
+      > a {
+        display: block;
+      }
     }
     img {
       width: 100%;


tvararu:
Not a huge fan of the child selector as it's usually code smell, but looks fine in this use-case. 👍

robinwhittleton:
Either needed to use it or add class to the link. Would happily do that if people thought it was worthwhile…

tombye:
Think the child seelctor is justified here 👍

robinwhittleton:
Great, thanks for the review. Will leave it to a product owner to do the merge.

PR #969, opened 2016-06-17
[Do not merge] Add Northern Ireland authorities (LGD) to the tier lookup

sihugh:
For some reason LGD (Northern Ireland) authorities were added to the list of
area types but not the tier lookup.

As part of the fix, we'll try and reduce duplication of lists of types.  It's also hard
 to check which ones are in the list if they're randomly ordered.

The ordering of the list of `authority_types` is important as licences need to
be able to select the most specific match in the case when two local
authorities are returned for a postcode.  This means that districts must come 
before counties.  It doesn't really matter where unitary authorities come in the list.

[This test describes the case](https://github.com/alphagov/frontend/blob/3b12a066f0787f9a926912becf04a8b440792f12/test/unit/licence_location_identifier_test.rb#L50L61)

### DIFF
--- a/lib/location_identifier.rb
+++ b/lib/location_identifier.rb
@@ -1,14 +1,21 @@
 class LocationIdentifier
   private
+    DISTRICT = ['DIS']
+    COUNTY = ['CTY']
+    UNITARY = %w(COI LBO LGD MTD UTA)
+
+    #It is important that DIS (district) is before CTY (county) in the list of
+    # authority types as licences want the most specific provider first. All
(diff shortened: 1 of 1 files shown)


jennyd:
I think there are still [more things we'd need to fix](https://gov-uk.atlassian.net/wiki/display/TCT/Understand+the+extent+to+which+we+should+support+Northern+Ireland+postcode+lookups) apart from this to be able to support Northern Ireland fully for local transactions. What's the effect for users of making this change in isolation?

sihugh:
Seems fair, I'll withdraw this

PR #970, opened 2016-06-20
Use MapIt for authority lookups

davidbasalla:
We recently added govuk_slugs for local authorities to MapIt. This means we can get these slugs directly from MapIt rather than rely on the local 'authorities.json' file. We remove the file in this PR and update the  `AuthorityLookup` class to make MapIt requests. As a consequence we have to stub these MapIt calls in our tests. We remove some unused methods from `AuthorityLookup`. Finally we update the LocationIdentifier subclasses to be more readable and remove the legacy term `geostack` from the code.

As part of this PR we add some code only to delete it in later commits. We've chosen to keep this in the git commit history (rather than squash commits) so that the refactoring is easier to follow.

For https://trello.com/c/YZEWeHwX/404-switch-frontend-to-get-la-slugs-from-mapit-instead-of-authorities-json-3

Note that the tests will not pass before https://github.com/alphagov/gds-api-adapters/pull/524 is merged and we bump 'gds-api-adapters', which exposes the MapIt endpoint for `/code/<code_type>/<code>`.

Paired with @issyl0


lib/licence_location_identifier.rb
```diff
@@ -5,27 +5,44 @@ class LicenceLocationIdentifier < LocationIdentifier
...
-      return authorities[type] unless authorities[type].nil?
+  def providing_tier
+    return unless artefact
+
+    artefact["details"].try(:[], "licence").try(:[], "local_service").try(:[], "providing_tier")
+  end
+
+  def matching_authority_by_tier
+    @_authority_by_tier ||= areas.detect do |area|
+      providing_tier.include?(LocationIdentifier.identify_tier(area["type"]))
```
h-lame: Given:
```ruby
providing_tier: ['county', 'unitary']
areas: [{type: 'unitary', name: 'foo'}, {type: 'county', name: 'bar'}]
```
In the old implementation we'd get "bar" for `matching_authority_by_tier`, but in the new implementation we'd get "foo".  When the service is provided by the "all" tier, which maps to `["district", "unitary", "county"]` for providing_tier (see: https://github.com/alphagov/publisher/blob/master/lib/local_service_importer.rb#L27-L36) then this would mean we might get the "county" instead of the "district" if it appears first in the list from mapit.  This might be bad.

lib/licence_location_identifier.rb
```diff
@@ -5,27 +5,44 @@ class LicenceLocationIdentifier < LocationIdentifier
...
+    @artefact = artefact
+  end
 
-      return appropriate_authority unless appropriate_authority.nil?
+  def find_slug
+    if providing_tier && matching_authority_by_tier
+      matching_authority_by_tier.try(:[], "codes").try(:[], "govuk_slug")
+    elsif matching_authority_by_area_type
+      matching_authority_by_area_type.try(:[], "codes").try(:[], "govuk_slug")
     end
```
h-lame: Another difference between the old and new implementations is that in the `if providing_tier` conditional in the old version our result could be `nil` if we don't have an area that maps to the providing tier *or* if we do have one but it has no slug.  If the result of that conditional is `nil` we fall out of the conditional and try the other way of getting a slug.  In the new implementation we'd only fall through if we couldn't find an area in `matching_authority_by_tier`, *not* if we do find one but it has no slug.

I'm less sure this is a problem based on the data we actually use, but still, edgecases are what caused bugs.

lib/licence_location_identifier.rb
```diff
@@ -5,27 +5,57 @@ class LicenceLocationIdentifier < LocationIdentifier
...
 
+    artefact["details"].try(:[], "licence").try(:[], "local_service").try(:[], "providing_tier")
+  end
+
+  def matching_authority_by_area_type_slug
+    matching_authority_by_area_type.try(:[], "codes").try(:[], "govuk_slug")
+  end
+
+  def matching_authority_by_area_type
+    LocationIdentifier.authority_types.each do |type|
```
davidbasalla: Is there a more elegant way to do this instead of using nested`each` loops?

lib/licence_location_identifier.rb
davidbasalla: @h-lame We've added tests around the case you described and changed the code accordingly. Do you mind having another look when you get the chance?

lib/licence_location_identifier.rb
davidbasalla: We've added a test to guard against this case and updated the code.

lib/licence_location_identifier.rb
h-lame: The inner `each` could probably be written with `detect`.

```ruby
LocationIdentifier.authority_types.each do |type|
  found = areas.detect { |area| area["type"] == type && area_has_slug?(area) }
  return found if found
end
```

But I'm not hugely convinced it's much clearer.  I reckon these nested `each` loops are probably ok.

h-lame:
Good stuff!  Happy to merge once we've done the gds-api-adapters stuff.

test/unit/licence_location_identifier_test.rb
```diff
@@ -2,93 +2,109 @@
...
+      },
+    }
+
+    @shrewsbury_civil_parish = {
+      "name" => "Shrewsbury_civil_parish",
+      "type" => "CPC",
+    }
+  end
+
+  # SOMETHING WRONG HERE, ONLY ONE ARTEFACT BEING CREATED
```
jennyd: What's this comment about?

test/unit/licence_location_identifier_test.rb
```diff
@@ -2,93 +2,109 @@
...
-    should "select the closest authority from geostack if district not provided" do
-      geostack = {
-        "council" => [
-          {"id" => 2, "name" => "South Ribble Borough Council", "type" => "DIS", "govuk_slug" => "south-ribble-borough-council"},
-          {"id" => 3, "name" => "Leyland South West", "type" => "CED" },
-          {"id" => 4, "name" => "South Ribble", "type" => "WMC" }
-        ]
-      }
-      slug = LicenceLocationIdentifier.find_slug(geostack, @artefact)
+    should "select the closest authority from areas if district not provided" do
```
jennyd: It took me a while to figure out what this test and the next few are for. I _think_ the intent is to test the fallback behaviour when there are no areas for any of the providing tiers. In this case the artefact specifies `"providing_tier" => %w(county unitary)` but neither of those are in `areas`, so instead `LicenceLocationIdentifier.find_slug` goes through [`LocationIdentifier.authority_types`](https://github.com/alphagov/frontend/blob/c8a03ac693e4036ffc04f254aaf82e01da0e4ecb/lib/location_identifier.rb#L4) looking in `areas` for an area of each type in turn, and returns the slug of the first one it finds. This fallback is also what happens when the artefact doesn't contain any providing tiers.

The "closest" phrasing of these tests was pretty confusing to me - it sounds like it's doing a geographical query rather than relying on the ordering of area types in an array. I know you didn't choose that word, but is there a better way of describing these contexts so that this behaviour is better documented?

jennyd:
Your refactoring of the `LocationIdentifier` subclasses makes the intent much clearer 👍 

There's now quite a bit of similarity between the two subclasses - I wonder if it's worth moving the shared methods into `LocationIdentifier` to avoid the duplication?

test/unit/licence_location_identifier_test.rb
jennyd: Ah, it's removed in the next commmit :)

davidbasalla:
@jennyd We pulled the common methods from the LocationIdentifier subclasses into the parent class in https://github.com/alphagov/frontend/pull/970/commits/30e417b7457681d8fc2b88660abfed23f2a0aec1. However we're undecided whether this sacrifices readability for less code duplication. When reading one of the subclasses, one has to jump between parent and child class a couple of times now in order to understand what it's doing. What do you think?

h-lame:
Looks good.  Merging.

PR #973, opened 2016-06-23
Remove make-a-sorn custom slug

brenetic:
`make-a-sorn` was previously a custom slug with a hard corded html
template. This removes all the custom code associated with that which
makes it fall back to a regular transaction. There is already a start
now page for `make-a-sorn` in Publisher.

# Before

before

# After

after

[Trello card](https://trello.com/c/IkNiy7qi/410-convert-make-a-sorn-start-page-into-a-standard-start-page)

cc @keithiopia


h-lame:
It's all 🔴 ! Big 👍 for delete-only PRs :)

h-lame:
On closer inspection it also looks to do the correct thing.  Happy to merge once we get the ok from content that the new version of the make-a-sorn document has been published.

brenetic:
This is waiting on the ok from either @keithiopia or John Ploughman in the content team before merging.

brenetic:
New make-a-sorn page will be published 5th July. This PR is now ready for merging

PR #976, opened 2016-07-21
Update department count

surminus:
This should match what is shown in the following two places:
https://www.gov.uk/government/organisations
https://www.gov.uk/government/organisations#agencies-and-government-bodies


rjbate:
👍

deanwilson:
Merge away.

PR #977, opened 2016-07-22
[DISCUSS] Update root controller to make request to LocalLinksManager for local transactions

davidbasalla:
Please see the comments in the code. We assume a wrapper for LocalLinksManager in gds-api-adapters. For demo purposes we use a stubbed response from LocalLinksManager.

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -133,23 +133,9 @@ class RootController < ApplicationController
         return redirect_to publication_path(slug: params[:slug], part: CGI.escape(params[:authority][:slug]))
       elsif params[:part]
         authority_slug = params[:part]
-
-        unless non_location_specific_licence_present?(@publication)
-          snac = AuthorityLookup.find_snac_from_slug(params[:part])
-
-          if request.format.json?
-            return redirect_to "/api/#{params[:slug]}.json?snac=#{snac}"
-          end
-
-          # Fetch the artefact again, for the snac we have
-          # This returns additional data based on format and location
-          updated_artefact = fetch_artefact(params[:slug], params[:edition], snac) if snac
-          assert_found(updated_artefact)
-          @publication = PublicationPresenter.new(updated_artefact)
-        end
       end
 
-      @interaction_details = prepare_interaction_details(@publication, authority_slug, snac)
+      @interaction_details = prepare_interaction_details(@publication, authority_slug)
       if local_authority_match?(@interaction_details)
         @local_authority = LocalAuthorityPresenter.new(@interaction_details['local_authority'])
         if no_interaction?(@interaction_details)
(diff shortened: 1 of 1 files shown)


app/controllers/root_controller.rb
```diff
@@ -273,10 +259,42 @@ def treat_as_standard_html_request?
     !request.format.json? and !request.format.print? and !request.format.video?
   end
 
-  def local_transaction_details(artefact, authority_slug, snac)
-    raise RecordNotFound if snac.blank? && authority_slug.present?
+  def local_transaction_details(artefact, authority_slug)
```
jennyd: Instead of putting this in the controller, we could use a separate object to construct the local links manager request and add the details into the artefact, as [finder-frontend does](https://github.com/alphagov/finder-frontend/blob/master/app/lib/finder_api.rb).

app/controllers/root_controller.rb
```diff
@@ -273,10 +259,47 @@ def treat_as_standard_html_request?
...
+        "name"=>"London Borough of Camden",
+        # "snac"=>"00AG",
+        # "tier"=>"unitary",
+        "homepage_url"=>"http://www.camden.gov.uk/"
+      },
+      "local_interaction"=>{
+        # "lgsl_code"=>432,
+        # "lgil_code"=>0,
+        "url"=>"https://www.westminster.gov.uk/stray-and-lost-dogs"
+      }
```
davidbasalla: So we're returning 2 different sets of data of here, one for the local authority and one for the local interaction. Wondering whether this should then be 2 different requests. One that hits a LocalLinksManager `/api/local_authorities` endpoint and one that hits `api/local_interaction_details`. So we would make 2 requests and still merge the data in here. But maybe that would be a cleaner separation of the 2 types of data. As a bonus, this could give us a API endpoint in the LocalLinksManager API for later 'Find my local authority'?

issyl0: Given that we need `authority_slug` to do the `local_interaction_details` call in the first place (and then in this implementation we get back the LA details too, splitting them up to get the LA details first makes sense to me - then we're not muddling the names of things either.

h-lame: Bit wary of adding another API request - we're already making one to mapit so that would bump this up to 3 requests for the council and link information.  Maybe we just need to think about the structure of the data and the name of the endpoint so it makes more sense as I agree it feels a bit disjointed at the moment.

h-lame:
Approach seems sound :+1:

app/controllers/root_controller.rb
davidbasalla: We've taken out one MapIt request [here](https://github.com/alphagov/frontend/pull/977/files/15ed19e09e28695d5c592d1331f45a5165385137#diff-132a19fd5e7a1edc0d9127204dfa48eaL138), which converted 'slug' to 'snac', which we don't need anymore as we don't use snacs for LocalInteraction lookups (and decided that providing support for the `/api/` route wasn't necessary). So the only MapIt request remaining would be the one from the initial postcode lookup I think. But yeah, let's have another think about the approach for sure!

app/controllers/root_controller.rb
davidbasalla: We can definitely try this when we get the LocalLinksManager API working. I'm a bit unclear how much the code would have to change. Currently we just produce a hash of local interaction data which we ultimately assign to `@interaction_details` [here](https://github.com/alphagov/frontend/pull/977/files#diff-132a19fd5e7a1edc0d9127204dfa48eaR138) and then access in various views. So unsure of the impact of merging this kind of data back into the artefact and accessing the relevant data from it instead of `@interaction_details`.

PR #979, opened 2016-07-25
Proper label for homepage search form

robinwhittleton:
@cjforms has evidence that ‘placeholder’ style labels are detrimental to accessibility. While we don’t necessarily have the luxury of space in the search form in the header, there’s no reason not to have a proper label on the homepage. This uses the [label styles from govuk_elements](http://govuk-elements.herokuapp.com/form-elements/#form-labels) (19px font, 2px padding). It also fixes the no-JS problem where the input was smaller than the expectations set by the visual display.

The [JavaScript that controlled the old label behaviour](https://github.com/alphagov/static/blob/c26b0992dc7f847c469d99b24d7d1b7fc660b532/app/assets/javascripts/core.js#L4-L17) also controls the header search form so needs to remain in place.

New look:
screen shot 2016-07-25 at 13 16 08

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -221,32 +221,16 @@ body.homepage {
       .header-search-content {
         @extend %contain-floats;
         position: relative;
-        background: #fff;
+        padding-right: 40px;
       }
       label {
-        float: left;
-        height: 40px;
-        line-height: 40px;
-        text-indent: 15px;
-        overflow: hidden;
         display: block;
-        margin-right: 10px;
-        color: $secondary-text-colour;
-
-        .js-enabled & {
-          float: none;
-          position: absolute;
-          left: 0;
-          top: 1px;
-          z-index: 1;
-          width: 100%;
-          margin-right: 0;
-        }
+        padding-bottom: 2px;
+        @include core-19;
       }
       input {
         @include box-sizing(border-box);
         float: left;
-        width: 50%;
         min-width: 0;
         display: block;
         margin: 0;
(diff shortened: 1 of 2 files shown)


robinwhittleton:
Summoning @markhurrell to comment.

markhurrell:
there's no evidence that ‘placeholder’ style labels are detrimental to accessibility for single search inputs like this

cjforms:
The accessibility people are adamant that placeholder text should not be used a label, which is what's happening (visually at least, I haven't checked the markup) in this case. For example, see this from the Paciello Group: https://www.paciellogroup.com/blog/2011/02/html5-accessibility-chops-the-placeholder-attribute/

(Many other similar articles are available, only a couple of them written by me). 

GOV.UK is widely used as an exemplar by many other governments and organisations. They won't consider the subtleties like 'we're using a non-recommended thing here on our home page because this particular usage is a special case that probably isn't causing any problems. They'll just point to our home page and say 'Look, GDS uses text inside the box as a label for the box right here, so it must be fine'. 

That's why I'm worried about it.

robinwhittleton:
For what it's worth we're using a proper `label` styled to look like a `placeholder` attribute.

cjforms:
Thanks. That would mean that it works OK for people using screenreaders which is good. 

Visually the problems with 'stuff inside boxes' exist for anyone who doesn't use a screenreader.

robinwhittleton:
Going to close this based on @markhurrell’s evidence that this isn’t a problem.

robinwhittleton:
We decided that the existing style is OK for search boxes. Will revisit this when we redesign the homepage to make sure it’s still working for everyone.

PR #982, opened 2016-07-27
Fetch links for local transactions from local links manager api

davidbasalla:
We don't want to use content api for this because local links manager is the new canonical source of link data.

Requires https://github.com/alphagov/gds-api-adapters/pull/542 so the tests will fail on jenkins until we merge that and bump it in the gemfile here.

Note: PR should only be merged/deployed on Wednesday 03/08/2016 after we mention it at tech lead standup and to give us a chance to do any other sort of communication.

### DIFF
--- a/config/initializers/local_links_manager.rb
+++ b/config/initializers/local_links_manager.rb
@@ -1,0 +1,3 @@
+require 'gds_api/local_links_manager'
+
+Frontend.local_links_manager_api = GdsApi::LocalLinksManager.new(Plek.current.find('local-links-manager'))
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,7 @@ end
 if ENV['API_DEV']
   gem 'gds-api-adapters', :path => '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '32.2.0'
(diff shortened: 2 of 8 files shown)


davidbasalla:
@sihugh Having looked through the code, they are guards (https://github.com/alphagov/frontend/pull/982/files?diff=unified#diff-132a19fd5e7a1edc0d9127204dfa48eaR137) in place that will ensure that the authority_slug and LGSL will be present and valid at the point of making the request to LocalLinksManager, so we think there's no need to add any extra tests around the case of missing/nil parameters in the request call to LocalLinksManager.

sihugh:
Sounds fine to me :-)

config/initializers/local_links_manager.rb
```diff
@@ -0,0 +1,3 @@
+require 'gds_api/local_links_manager'
+
+Frontend.local_links_manager_api = GdsApi::LocalLinksManager.new(Plek.current.find('local_links_manager'))
```
sihugh: This needs to be `local-links-manager`

sihugh:
Merge when ready! :+1:

PR #984, opened 2016-08-01
Add search parameter to display the finder as open

rubenarakelyan:
The finder on the search results page is normally shown closed, unless at least one filter has been selected, in which case it is shown open. This adds a search parameter that forces the finder to be shown open even if no filters are selected. If the parameter is not present, then the logic on whether or not to display the finder reverts to the previous behaviour.

This feature has been added to allow searches originating from Whitehall (and possibly other applications in future) to always display the finder.

Trello: https://trello.com/c/t9R5QDCf/42-when-a-search-happens-from-whitehall-content-open-the-organisation-filter-by-default

screen shot 2016-08-01 at 09 37 43

### DIFF
--- a/app/models/search_parameters.rb
+++ b/app/models/search_parameters.rb
@@ -25,6 +25,10 @@ class SearchParameters
     params[:q]
   end
 
+  def show_filter_organisations?
+    params[:show_filter_organisations] == "true"
+  end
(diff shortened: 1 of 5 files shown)


MatMoore:
LGTM

PR #986, opened 2016-08-05
Build find local council

davidbasalla:
We add a separate page for 'Find-my-local-council'. We reuse the templates used by 'Find my nearest' and Local Transactions. We've decided to hardcode both related links and breadcrumbs for the moment, and will decide at a later point whether to store this sort of information on a content item.


app/controllers/find_local_council_controller.rb
```diff
@@ -50,6 +53,11 @@ def result
 
 private
 
+  def set_artefact_headers
+    artefact = fetch_artefact(BASE_PATH)
+    set_slimmer_artefact_headers(artefact)
+  end
```
h-lame: I wonder if it's more consistent for now to just hardcode them like we do with the related links.  Or if now that we're using the artefact we should also use it to populate the related links sidebar too? 

My vote would be for not using the artefact at all and hardcoding everything as a stopgap until we work out how to make this entire thing use managed content.

davidbasalla:
@h-lame I have updated the last commit to use hardcoded breadcrumbs

app/views/find_local_council/one_council.html.erb
```diff
@@ -0,0 +1,22 @@
+<%= render layout: 'base_page', locals: {
+  title: 'Find Your Local Council'
```
robinwhittleton: We don’t generally capitalise page titles, probably want sentence case (e.g. `Find your local council`).

app/views/find_local_council/one_council.html.erb
```diff
@@ -0,0 +1,22 @@
...
+        We don't have a link for their website.
+      <% else %>
+        Website: <%= link_to extract_host(@authority["homepage_url"]), @authority["homepage_url"] %>
+      <% end %>
+    </p>
+  </div>
+  <hr class="separator">
+
+  <div class="search-again">
+    <a title="Search again using a different postcode." href="/find-your-local-council">Back</a>
```
nickcolley: I think this may pose problems with screenreaders potentially? 

https://silktide.com/i-thought-title-text-improved-accessibility-i-was-wrong

davidbasalla:
Added Do not merge tag, as we want to add some analytics before this goes live. See comments in https://trello.com/c/yoz4mg1B/436-3-of-3-build-find-my-local-council-3

config/locales/en.yml
```diff
@@ -32,9 +32,8 @@ en:
       invalid_postcode: "This isn't a valid postcode."
       invalid_postcode_sub: "Try checking it and entering it again."
       valid_postcode_no_match: "We couldn't find this postcode."
-      valid_postcode_no_match_sub_html: "Try entering it again or <a href='http://local.direct.gov.uk/LDGRedirect/Start.do?mode=1'>look up the council</a>."
+      valid_postcode_no_match_sub_html: "Try checking it and entering it again."
```
h-lame: We should get content team to review the words here to make sure it's an ok message.

davidbasalla: Good point, have emailed Mark and Paula about this

brenetic:
@davidbasalla Updated with new route for find-local-council

lib/tasks/rummager.rake
```diff
@@ -1,5 +1,26 @@
...
+        title: "Find your local council",
+        description: "Find your local authority in England, Wales, Scotland and Northern Ireland",
+        link: "/find-local-council",
+        indexable_content: "Find your local authority in England, Wales, Scotland and Northern Ireland",
+        public_timestamp: Time.zone.now.iso8601,
+      },
+    ]
+
+    rummager_payloads.each do |rummager_payload|
+      # custom-application is ultimately what the /help route ends up as when it
```
davidbasalla: How do 'custom-application' and the '/help' route relate to this bit of code? Maybe the comment could add some more context?

h-lame: Oh, yes, this comment isn't really relevant anymore.  Will remove.

PR #990, opened 2016-08-10
Update ruby version to 2.3.0

mgrassotti:
Trello https://trello.com/c/tCwG3wqV/449-upgrade-frontend-ruby-dependency-small

Update ruby version to 2.3.0.

A new warning is appearing now when running tests:
```
[warning] The response contained in an RestClient::Exception is now a RestClient::Response instead of a Net::HTTPResponse, please update your code
```
This was already referred here:
https://github.com/alphagov/specialist-publisher-rebuild/pull/706/commits/9c395728b9f628113ba2c0319a2a2a2f1a8ad81e

It should be safe to ignore it, although it is not great to have such a lot of warnings in tests results.


boffbowsh:
Did you try updating `gds-api-adapters` to `rest-client` 2.0.0?

boffbowsh:
@andrewgarner is looking into upgrading gds-api-adapters at the moment, speak to him

binaryberry:
As discussed in standup, I've amended this pull request to keep the deprecation warnings from gds-api-adapters. This is not ideal as they are very noisy when running tests, but the only way to silence them is to change the configuration of gds-api-adapters as recommended [here](https://github.com/alphagov/gds-api-adapters/commit/4ad659ff0becbc059e19bac07d9bd091dac7f788), which means when there is a 404 or 410 an error gets raised instead of returning nil. Some of our code relies on a nil being returned, and we are unsure we are consistently testing for this nil being returned, therefore changing the configuration might break things silently. That's why keeping the deprecation warnings for the time being is preferable.
I am preparing a story to fix this by the 1st of October at the very latest as that's the date gds-api-adapters will adopt the new behaviour by default.

binaryberry:
[Here](https://trello.com/c/JuCEOTnC/671-handle-404s-and-410s-in-frontend) is the proposed story to fix the deprecation warnings / handle the new behaviour of gds-api-adapters.
[Here](https://trello.com/c/TLsvlW9C/670-refactor-frontend-s-root-controller) is the proposed story to refactor the root controller.
@h-lame fyi ^ and thanks for the help yesterday.

h-lame:
@binaryberry refactoring root_controller is definitely something we've wanted to do for a while, but we've also been swallowing that pain on the hope we'll just 🔥 it entirely when we do migration of the formats it handles.

binaryberry:
@h-lame Mmmh, good point. I will include this thought in the card, might not be worth doing.

gpeng:
Should we just update to 2.3.1 now it is available on the boxes?

boffbowsh:
What Graham said, otherwise 👍

PR #991, opened 2016-08-10
[DO NOT MERGE] Add ga tracking to find your local council

davidbasalla:
We add Google Analytics tracking to 'Find your local council'. This aims to provide similar error tracking as already present for 'Local Transactions'.

We add tracking for:
- `postcodeSearchStarted`
- `postcodeResultsShown` (with number of results of shown)
- exit link tracking for unitary/district/county homepage URLs
- postcodeError tracking, specifcally:
  - `postcodeErrorShown:invalidPostcodeFormat`
  - `postcodeErrorShown:noLaMatch`
  - `postcodeErrorShown:fullPostcodeNoMapitMatch`

For https://trello.com/c/qOtW0fJB/444-analytics-for-find-my-local-council-2 

Paired with @emmabeynon 

This branched off of https://github.com/alphagov/frontend/pull/986, so will need to be rebased after that branch gets merged.


app/models/mapit_postcode_response.rb
```diff
@@ -19,6 +19,10 @@ def invalid_postcode?
     postcode && location.nil? && error.present?
   end
 
+  def blank_postcode?
+    postcode.nil? || postcode.empty?
```
h-lame: `postcode.blank?` (a rails addition) will achieve the same thing.

app/views/application/_location_form.html.erb
```diff
@@ -4,7 +4,12 @@
       data-track-action="postcodeSearchStarted">
 
   <% if @location_error %>
-    <div class="error-summary">
+    <div class="error-summary"
+         data-module="auto-track-event"
+         data-track-category="userAlerts:<%= alert_type %>"
+         data-track-action="postcodeErrorShown:<%= @location_error.postcode_error %>"
+         data-track-label="<%= t(@location_error.message, @location_error.message_args) %>">
+
```
h-lame: :+1: for this, moving to using the auto tracking is good and we get to burn an extra partial.  yay!

h-lame:
Good stuff.  GH seems to have the commits in a weird order, but I'm sure it'll be fine when we merge it into the other branch.

One thought is that we could merge the commit where we introduce `alert_type` and the subsequent commit where we change it to `publication_format` as the change only occurs a commit apart - we could just pretend we got it right all along.

davidbasalla:
thanks @h-lame, have addressed your comments and rebased!

davidbasalla:
Merging this into 'build-find-local-council' branch which already has an open PR

PR #997, opened 2016-08-24
Add GA tracking for no results shown for place format

davidbasalla:
We add Google Analytics tracking in the case that no results are returned from Imminence for a valid postcode. We add integration tests to ensure that the correct GA parameters are being set.

We also update the track category for Find local council with the correct value. It was previously set to 'placeSearch' which in the context of this app refers to the 'find-my-nearest' format.

For https://trello.com/c/ZWX0Apjj/469-add-ga-tracking-in-frontend-for-when-find-my-nearest-shows-no-results-2

Paired with @surminus


h-lame:
🚜 all the things! :+1:

PR #1000, opened 2016-08-26
Add accessibility changes to find local council

davidbasalla:
A small number of changes to meet accessibility standards as per an informal review. They are all related to the location form. While the review was specifically for the new 'Find local council' page, the location form partial is also used by the licence, place and local transaction formats so those formats will benefit from the changes as well.

See commits for individual updates.

For https://trello.com/c/ldjGC28P/476-fix-the-things-from-the-accessibility-review

### DIFF
--- a/app/assets/javascripts/frontend.js
+++ b/app/assets/javascripts/frontend.js
@@ -9,3 +9,7 @@
 //= require templates
 //= require search
 //= require_tree ./modules
+
+$(document).ready(function() {
+  $('.error-summary').focus();
+});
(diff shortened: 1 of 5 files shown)


nickcolley:
Went through this with @davidbasalla in person, works well in VoiceOver.. 👍 

We thought maybe just changing the name to .error-summary would make it more generic.

Ideally this functionality would live at the govuk_component level so we don't duplicate this work, but I think this is better suited to be resolved by the frontend rationalisation work coming up soon.

davidbasalla:
thanks @nickcolley, have renamed to `.error-summary`

brenetic:
💭

PR #1004, opened 2016-09-05
Use content-store & component to render the breadcrumbs

tijmenb:
Currently, we pass in the "artefact" from the content-api into slimmer. Slimmer will use an elaborate process of middleware to add a breadcrumb onto the pages.

Instead, this commit:

- Fetches the content item from the content-store
- Passes the content item into a "navigation helper" library (`govuk_navigation_helpers` gem).
- Uses the generated payload to render a GOV.UK Component (http://govuk-component-guide.herokuapp.com/components/breadcrumbs)

Part of: https://trello.com/c/mOeDK914

Before merge:

- [x] Deploy https://github.com/alphagov/travel-advice-publisher/pull/161
- [x] Republish travel advice index page 
- [x] Breadcrumbs render with the correct spacing


tijmenb:
cc @nickcolley @carolinegreen

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,17 @@ class ApplicationController < ActionController::Base
   slimmer_template 'wrapper'
 
 protected
+
+  # Set up the navigation helpers. This will make breadcrumbs appear for pages
+  # that have a content item. In the near future this will also provide data to
+  # render the sidebar.
+  def setup_navigation_helpers
+    content_item = content_store.content_item(request.path)
```
jennyd: There won't be a content item for every request path, for example local transactions have many separate [results pages](https://www.gov.uk/pay-council-tax/lambeth) but there's only one content item. [Guides](https://www.gov.uk/adoption-pay-leave/eligibility) and any other mainstream multipart formats also don't have content items for each page. You might be able to use only the first path segment instead, but I don't know whether that would work for all formats.

test/integration/find_local_council_test.rb
```diff
@@ -8,6 +8,10 @@ class FindLocalCouncilTest < ActionDispatch::IntegrationTest
 
   context "when visiting the start page" do
     setup do
+      example_generator = GovukSchemas::RandomExample.for_schema("placeholder", schema_type: "frontend")
```
jennyd: Find your local council is a [special route](https://github.com/alphagov/frontend/blob/master/lib/tasks/publishing_api.rake#L72-L77), not a placeholder.

app/controllers/application_controller.rb
jennyd: That is yet to be decided, we're starting mainstream migration discovery today :)

test/integration/find_local_council_test.rb
tijmenb: That's interesting observation. Special routes don't have a frontend schema:

> Note that this schema has a restricted set of properties at the metadata level, and no details or links hashes at all. Therefore, it is not constructed by combining sub-schemas. In addition, there is no frontend version, because redirects are never seen by the frontend: everything is handled by the router.

(https://github.com/alphagov/govuk-content-schemas/blob/master/formats/special_route/README.md#L22)

I'm wondering if we should change the format, since the page is tagged with a mainstream browse page and parent.

tijmenb:
https://www.gov.uk/api/content/foreign-travel-advice now has a parent!

app/controllers/application_controller.rb
gpeng: We came across this too with guides that link to multi-page formats. We had to remove links to 'chapters' or 'sub pages' as they don't have a `content_id`. I'm not aware of any immediate plan to change that.

tijmenb: The way we handle this behaviour now is in most cases to use the same URL this app uses for content-api. For example for publications it uses the `:slug` from the URL.

test/integration/find_local_council_test.rb
tijmenb: I've kept this as a placeholder, because `special_route` formats cannot have a frontend schema. It's on my list to change that. After that I'll update this test. For now this is fine to merge because there wouldn't be an actual difference.

tijmenb:
@MatMoore & I paired on this, so we can't review. Anybody else?

PR #1006, opened 2016-09-08
Find my nearest: show 'no results' alert above the postcode form

issyl0:
- Currently when a user searches for a 'find-my-nearest' place and gets
  no results, they see the "Sorry, no results were found near you."
  message underneath the location form.
- This fixes that by making an error message in the same style as the
  local transactions ones: red box above the postcode input.
- The existing validPostcodeNoLocation message wasn't that informative,
  so make it tell the user more about what the error actually was.

Trello: https://trello.com/c/YNKOcWqr/485-show-no-results-alert-above-location-search-form-for-find-my-nearest-2

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -260,7 +260,10 @@ protected
 
   def fetch_places(artefact, postcode)
     if postcode.present? and artefact.format == 'place'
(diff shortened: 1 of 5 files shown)


h-lame:
:+1: good stuff.

PR #1007, opened 2016-09-09
Improve CSS for the search box on the GOV.UK homepage

36degrees:
This is based on work done to create a very similar search box within the yet-to-be-published service manual homepage. We needed a different solution for users without Javascript as the label was larger, and tackled a few other quirks at the same time.

## Javascript Disabled Behaviour
- Users without javascript will now see the label above the search box, rather than the label and input together next to each other within the white box.

### Before

js disabled - before

### After 
js disabled - after

## Focus State
- Fixes the focus state of the input and search box such that they do not obstruct each other.
- Alters the offset of the focus on the input to 'match' the focus of the submit button.
- Removes usage of `calc` so that the input is now full width in browsers that don't support calc (e.g. IE8, Safari 5) rather than 86% of the box.

### Before
focus state - before

### After
focus state - after

## Misc
- Increases the font size of the label from 16 to 19px (responsive)
- Simplifies the search box in IE6 / IE7 to match the non javascript behaviour. This allows us to get rid of the usage of `calc`. There might be a way to make it look more like it used to, but I don't believe it's worth spending time on. 

## Browser Before/After Comparisons
The most affected browsers are IE6 and IE7, which are now less pretty but remain functional, and IE8 which now looks and behaves in the same way as modern browsers (the input is the full width of what you think it should be). All other browsers should only exhibit the changes detailed above.

https://www.dropbox.com/sh/a0o8ypn6jz041vo/AABI8VEti1x2kgYmN9tJaJQ-a?dl=0

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -285,11 +285,47 @@ body.homepage {
           }
 
           &:focus {
-            z-index: 10;
+            z-index: 15;
             position: relative;
           }
 
           @include border-radius(0);
+        }
+
+        // If JS is enabled, and browser is not IE6 or IE7, position label
+        // as a placeholder
+        .js-enabled & {
+          position: relative;
+          background: $white;
+
+          // Absolutely position the label behind the input
+          label {
+            @include core-19($line-height: 40px, $line-height-640: 40px);
+            position: absolute;
+            left: 0;
+            top: 1px;
+            z-index: 5;
+
+            text-indent: 15px;
+
+            color: $secondary-text-colour;
+          }
+
+          // Input is transparent and sits above the label
+          .search-input {
+            position: relative;
+            z-index: 10;
+
+            background: transparent;
+
+            // Javascript adds a confusingly named `focus` class to the input
+            // when it contains text. We use this to set the background to
+            // white, hiding the label.
+            &:focus,
+            &.focus {
+              background: $white;
+            }
+          }
         }
       }
     }
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -260,6 +260,7 @@ body.homepage {
             margin: 0;
 
             border: 0;
+            outline-offset: 0;
           }
         }
 
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -216,108 +216,76 @@ body.homepage {
       }
     }
     form#header-search {
-      padding-bottom: 30px;
+      @extend %contain-floats;
+      margin-bottom: 30px;
 
-      .header-search-content {
-        @extend %contain-floats;
-        position: relative;
-        background: #fff;
-      }
-      label {
+      .search-input-wrapper {
         float: left;
-        height: 40px;
-        line-height: 40px;
-        text-indent: 15px;
-        overflow: hidden;
-        display: block;
-        margin-right: 10px;
-        color: $secondary-text-colour;
-
-        .js-enabled & {
-          float: none;
-          position: absolute;
-          left: 0;
-          top: 1px;
-          z-index: 1;
+        width: 100%;
+        
+        .search-input {
           width: 100%;
-          margin-right: 0;
+          margin: 0.5em 0;
+          padding: 6px;
+
+          @include core-19($line-height: (28 / 19), $line-height-640: (28 / 16));
         }
       }
-      input {
-        @include box-sizing(border-box);
+
+      .submit {
         float: left;
-        width: 50%;
-        min-width: 0;
-        display: block;
-        margin: 0;
-        border: 0;
-        height: 40px;
-        padding: 6px 0;
-        @include ie-lte(7){
-          height: 28px;
-        }
-        @include appearance(none);
-      }
-
-      input#search-main {
-        @include core-19($line-height: (28/19), $line-height-640: (28/16));
-        position: relative;
         padding: 6px;
-        z-index: 3;
-        background: transparent;
-
-        &.focus,
-        &:focus {
-          background: #fff;
-        }
-
-        @include ie-lte(7){
-          padding-left: 0;
-        }
-        .js-enabled & {
-          width: 86%;
-          @include calc(width, "100% - 37px");
-        }
-      }
-
-      input.submit {
-        position: absolute;
-        z-index: 4;
-        right: 0;
-        top: 0;
-        width: 40px;
-        height: 40px;
-
-        border: 1px solid $black;
-        border-width: 0 0 0 1px;
-        @include ie-lte(7){
-          border-width: 1px;
-        }
-        border-left-color: #222;
-        @include border-radius(0);
-
-        overflow: hidden;
-        text-indent: -5000px;
-        cursor: pointer;
-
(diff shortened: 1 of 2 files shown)


robinwhittleton:
Haven’t reviewed the code yet but I’m very happy with the general approach. The fallback in IE6/7 (the search button dropping below the field) is absolutely fine given that we don’t technically support them any more.

thehenster:
It'll be nice to fix that iOS version. That's probably seen quite a lot I'd guess!

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,134 @@ body.homepage {
...
       label {
-        float: left;
-        height: 40px;
-        line-height: 40px;
-        text-indent: 15px;
-        overflow: hidden;
-        display: block;
-        margin-right: 10px;
-        color: $secondary-text-colour;
+        color: $white;
```
selfthinker: Minor: I don't think you need this as all text inside the header is already white. The label would just inherit that.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,134 @@ body.homepage {
...
+      @if ($is-ie == false) or ($ie-version >= 8) {
+        .search-input-wrapper {
+          @include box-sizing(border-box);
 
-        &.focus,
-        &:focus {
-          background: #fff;
-        }
+          padding-right: 40px;
+          margin-right: -40px;
```
selfthinker: Minor: To make it clearer where the `40px` are coming from, you could extract that into a (local) sass variable (e.g. `$search-icon-size`). That is mostly to improve readability and therefore maintainability of the code.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,134 @@ body.homepage {
...
         }
-        border-left-color: #222;
-        @include border-radius(0);
 
-        overflow: hidden;
-        text-indent: -5000px;
-        cursor: pointer;
+        .submit {
+          @include box-sizing(border-box);
+          float: left;
```
selfthinker: Minor: It is already set to `float: left` above, so would inherit it and doesn't need the line. I'm not sure it's necessary to float at all? (Haven't tested in all the browsers, so some might need it?)

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,134 @@ body.homepage {
...
+          height: 40px;
+          width: 40px;
+        
+          border: none;
```
selfthinker: Minor: The border was already set to `0` above, so this line is not needed.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,134 @@ body.homepage {
...
+        // If JS is enabled, and browser is not IE6 or IE7, position label
+        // as a placeholder
+        .js-enabled & {
+          position: relative;
+          background: $white;
+
+          // Absolutely position the label behind the input
+          label {
+            @include core-19($line-height: 40px, $line-height-640: 40px);
+            display: block;
```
selfthinker: Minor: All absolutely positioned elements are already block elements, so that line is not needed.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,134 @@ body.homepage {
...
+            position: absolute;
+            width: 100%;
```
selfthinker: Minor: I don't think this needs to be 100%?

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,134 @@ body.homepage {
...
+            color: $secondary-text-colour;
+          }
+
+          // Input is transparent and sits above the label
+          .search-input {
+            position: relative;
+            z-index: 10;
+
+            // Internet Explorer does not play well with "empty" elements.
+            // url(#) somehow makes them work. ¯\_(ツ)_/¯
```
selfthinker: It would be great to reference a URL explaining that further. I tried to google but Google only searches for "url", not "url(#)", although I put everything inside quotation marks.
I also don't understand the sentence. What are "empty" elements? And what exactly does the workaround fix?

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,134 @@ body.homepage {
...
+            background: transparent url(#);
+
+            // Javascript gives input a confusingly named `focus` class when it
```
selfthinker: I don't understand this. I suspect the `.focus` class is for older browsers? Why does `.focus` not need the same CSS as `:focus` then?

selfthinker:
Sorry for bombarding you with all those minor issues. None of them are show stoppers and there are only for your consideration. I'm happy enough with the code and think it can be shipped as is.

The only thing I would improve is the size of the commit. As it is quite big I think it would be better if it could be split it to several smaller commits, especially as you already acknowledge the smaller changes as separate list items in the commit message.
I appreciate that splitting changes is not always possible. But if you can do it within 15 minutes or so I think it will be worth it. If you think it would take longer or it would be a real pain, it might not be worth it and we can ship it as it is.

app/assets/stylesheets/views/_homepage.scss
36degrees: It's needed for IE6 and IE7 – without them the search box ends up in a very weird place indeed.

screen shot 2016-09-15 at 10 13 09

36degrees:
Thanks @selfthinker! Lots of good feedback there, thanks. I also realised whilst implementing your changes that there was also some other unnecessary code to juggle z-indexes that wasn't required, so removed htat.

I've rewritten the commit history to try and reflect the process I went through. Obviously it's a bit artificial as it's done retrospectively, but you're right that lumping it into one commit wasn't the right approach.

Please can you take another look and let me know if you're happy with the changes I've made?

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,127 @@ body.homepage {
...
+
+            color: $secondary-text-colour;
+          }
+
+          // Input is transparent and sits above the label
+          .search-input {
+            position: relative;
+            z-index: 10;
+
+            // Internet Explorer does not play well with "empty" elements. The
```
selfthinker: Minor: Having the links is much better now and helps to understand the issue. But I think the way it is explained it not great (already on the SO post which you half-copied here). Instead of '"empty" elements' I think it would be clearer to say "form fields without an empty content/value". It is not clear from the original post if it really applies to **all** elements, I would suspect it's only true for form fields. It would also be good to know which IEs are affected and only apply to those. But will probably be difficult to find out. And the code doesn't cause any harm even if it gets applied to versions which have no issues here.
I also like the solution with the data url better than the `#` before.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -216,108 +216,127 @@ body.homepage {
...
-        @include border-radius(0);
 
-        overflow: hidden;
-        text-indent: -5000px;
-        cursor: pointer;
+        .submit {
+          @include box-sizing(border-box);
+          height: $search-button-size;
+          width: $search-button-size;
+
```
selfthinker: There are some rogue spaces there.

PR #1008, opened 2016-09-12
The validPostcodeNoLocation error message shouldn't mention PO Boxes

issyl0:
- Mark thought this was confusing when we were going through the result
  of #1006, so simplify it. This makes it even more consistent with
  local transactions.


sihugh:
Is it worth checking with Paula that we're happy with it?  LGTM

sihugh:
Merging anyway.  We can amend later if necessary

PR #1009, opened 2016-09-15
Add (UKVI) AB testing setup for September 2016 session

ikennaokpala:
This commit sets up AB test for the following ukvi page sections:

* /eea-registration-certificate/apply
* /eea-registration-certificate/permanent-residence

Cohorts across the aforementioned ukvi pages section fall under the following.
* original (with a 90% weighting)
* applyTextAndLink (with a 10% weighting)

The isSection checks that the current path matches the the given path that has
been defined for A/B testing.

It basically does an (string based) equality check.

The AddtoApplySection and AddtoPermanentResidenceSection function inserts
the given html before the respectively identified html tags.

### Before
screen shot 2016-09-15 at 13 34 21

### After

screen shot 2016-09-15 at 13 33 39
 

### Before

screen shot 2016-09-15 at 13 38 09

### After
screen shot 2016-09-15 at 13 37 33

*NB: Factcheck was carried out on integration after deploying just this branch.*

### DIFF
--- a/app/assets/javascripts/modules/ukvi-ab-test-september-2016.js
+++ b/app/assets/javascripts/modules/ukvi-ab-test-september-2016.js
@@ -1,0 +1,57 @@
+//= require govuk/multivariate-test
+
+var UkviABTest = UkviABTest || {};
+
+UkviABTest.isSection = function(pathname) {
+  return window.location.pathname === pathname;
+}
+
+UkviABTest.AddtoApplySection = function() {
+  var html = "<p>You can <a " +
+   "href='https://visas-immigration.service.gov.uk/product/eea-qp'>apply online "+
+   "as a qualified person</a> but not if you’re a student or self-sufficient "+
+    "person and you're either:"+
+  "<ul>"+
+    "<li>reliant on a family member for financial support</li>"+
+    "<li>financially responsible for any other family members</li>"+
+  "</ul></p>";
+
+  $(html).insertBefore('h2:contains("Apply as a ‘family member’")');
+}
+
+UkviABTest.AddtoPermanentResidenceSection = function() {
+  var html = "<p>If you're from the EEA, you can also <a "+
+   "href='https://visas-immigration.service.gov.uk/product/eea-pr'>apply "+
+    "online</a> but not if you’re a student or self-sufficient person and "+
+    "you're either: "+
+  "<ul>"+
+    "<li>reliant on a family member for financial support</li>"+
+    "<li>financially responsible for any other family members</li>"+
+  "</ul></p>";
+
+  $(html).insertBefore('h3:contains("Supporting documents")');
+}
+
+$(function(){
+  if(UkviABTest.isSection("/eea-registration-certificate/apply")) {
+    new GOVUK.MultivariateTest({
+      name: 'ukvi_apply-201609',
+      customDimensionIndex: 13,
+      cohorts: {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/ukvi-ab-test-september-2016.js
```diff
@@ -0,0 +1,53 @@
+//= require govuk/multivariate-test
+
+var UkviABTest = UkviABTest || {};
+
+UkviABTest.isSection = function(pathname) {
+  return window.location.pathname == pathname;
+}
+
+UkviABTest.AddtoApplySection = function() {
+  var html = `<p>You can <a href='https://visas-immigration.service.gov.uk/product/eea-qp'>apply online as a qualified person</a> but not if you’re a student or self-sufficient person and you're either:
```
nickcolley: Template Literals are an ES6 feature, so we'll have to do it like

```js
var html = '' +
'<h1>Hello, World</h1>' +
'<p>Test</p>' +
''
```

or

```js
var html = [
  '<h1>Hello, World</h1>',
  '<p>test</p>'
].join('')
```
(Hopefully we'll be able to think about transpiling ES6 in the future though since it's a much nicer syntax for things like this!)

http://caniuse.com/#feat=template

app/assets/javascripts/modules/ukvi-ab-test-september-2016.js
```diff
@@ -0,0 +1,53 @@
+//= require govuk/multivariate-test
+
+var UkviABTest = UkviABTest || {};
+
+UkviABTest.isSection = function(pathname) {
+  return window.location.pathname == pathname;
```
nickcolley: If we're comparing strings we should probably use '===' instead of '=='

app/assets/javascripts/modules/ukvi-ab-test-september-2016.js
ikennaokpala: @nickcolley thanks for this.. I have made changes

app/assets/javascripts/modules/ukvi-ab-test-september-2016.js
ikennaokpala: @nickcolley thanks for this.. I have made changes

steventux:
@nickcolley I've checked your review comments have been addressed. This looks good to me.

app/assets/javascripts/modules/ukvi-ab-test-september-2016.js
```diff
@@ -0,0 +1,57 @@
...
+
+UkviABTest.isSection = function(pathname) {
+  return window.location.pathname === pathname;
+}
+
+UkviABTest.AddtoApplySection = function() {
+  var html = "<p>You can <a " +
+   "href='https://visas-immigration.service.gov.uk/product/eea-qp'>apply online "+
+   "as a qualified person</a> but not if you’re a student or self-sufficient "+
+    "person and you're either:"+
```
gemmaleigh: The paragraph here needs to be closed </p> before the opening unordered list <ul>, as list elements cannot be children of p elements.

steventux:
In the past we have tended to insert templated html from `<script>` blocks rather than writing them out in js. This approach was arbitrarily dependent on how much html needs inserting and in this case it's pretty marginal. So I'm happy to merge.

PR #1010, opened 2016-09-15
Fix paragraph wrapping unordered list

ikennaokpala:
This PR is related to #1009 

This commit addresses the issue raised by @gemmaleigh, where the p tag wrapped an unordered list.


steventux:
cc @gemmaleigh this fixes [the problem you noticed here](https://github.com/alphagov/frontend/pull/1009#r78967487)

PR #1012, opened 2016-09-20
Use new GDS API Adapters behaviour

tijmenb:
This opts-in to the new recommended behaviour of gds-api-adapters, which always raises a `GdsApi::HTTPNotFound` exception when it encounters a 404 or 410 - instead of returning nil.

### DIFF
--- a/app/controllers/find_local_council_controller.rb
+++ b/app/controllers/find_local_council_controller.rb
@@ -27,8 +27,11 @@ class FindLocalCouncilController < ApplicationController
 
   def result
     authority_slug = params[:authority_slug]
-    authority_results = Frontend.local_links_manager_api.local_authority(authority_slug)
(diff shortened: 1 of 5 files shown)


binaryberry:
LTGTM but I'd prefer another pair of eyes. Thanks for doing this!

jennyd:
We think there may be more to do on this, particularly around calls to [Imminence](https://github.com/alphagov/frontend/blob/4c39f57cb313ab83d8ba197051070ab0d3c7fd96/app/controllers/root_controller.rb#L267-L278) and [content API](https://github.com/alphagov/frontend/blob/master/app/controllers/root_controller.rb#L16). This may break some of our error tracking around postcodes and locations - we need to look at those areas more closely. Let's chat @tijmenb!

PR #1018, opened 2016-09-28
Correct some linting issues

tijmenb:
This application is currently configured to only test changed files on lint. When doing large scale refactoring (like #1004) this causes a lot of noise, since you're forced to change a lot of unrelated code. It makes the diff hard to read and is fairly frustrating, since linting runs before tests when running rake.

Originally I intended to fix all the linting warnings in this app. However, it's so creaky at some places that Rubocop's autocorrect will actually break the tests.

Instead, this adds a .rubocop.yml file (generated with --auto-gen-config). It allows us to fix some issues selectively, while not introducing any more.


tijmenb:
Updated PR with a new approach, since running `--autocorrect` breaks the tests. The commits now fix a subset of linting issues that mostly about whitespace. This should solve most of the irritation working with the code base. I invite everyone to pitch in and fix the remaining linting issues.

PR #1021, opened 2016-10-03
Revert "Remove travel advice country page code"

tijmenb:
Reverts alphagov/frontend#1020.

Apparently FCO is still using these pages via `private-frontend`.

cc @danielroseman @jennyd @boffbowsh


jennyd:
Here's where travel advice publisher [links to these pages on private-frontend](https://github.com/alphagov/travel-advice-publisher/blob/b8e4a8e0a6f2f24f98f2f898140991b2b82139ba/app/helpers/application_helper.rb#L7) - it's all the "preview" links on the right:

screen shot 2016-10-03 at 16 43 04

danielroseman:
FWIW, we are talking with FCO about whether these "historical previews" are actually useful; I would dearly like to remove them, so we can move to the standard preview via draft-origin.

PR #1026, opened 2016-10-12
Remove set_slimmer_dummy_artefact usage

tijmenb:
`set_slimmer_dummy_artefact` was used to configure slimmer to send appropriate analytics tags. This has long ago been replaced by the slimmer headers (and will be replaced again by the analytics component from govuk-components).

Removing this has no effect on the analytics tags.


carvil:
👍

PR #1029, opened 2016-10-19
Replace Content API with Content Store for meta tags

rubenarakelyan:
The Content Store will be used to render meta tags instead of the Content API. This commit changes meta tags to be rendered with data from the Content Store.

Trello: https://trello.com/c/FBmZSc1J/177-render-meta-tags-from-content-store-in-frontend

### DIFF
--- a/app/controllers/find_local_council_controller.rb
+++ b/app/controllers/find_local_council_controller.rb
@@ -1,7 +1,7 @@
 require "postcode_sanitizer"
 
 class FindLocalCouncilController < ApplicationController
-  before_filter :setup_navigation_helpers
+  before_filter -> { setup_content_item_and_navigation_helpers BASE_PATH }
   before_filter :set_artefact_headers
   before_filter :set_expiry
 
@@ -126,10 +126,5 @@ private
       end
     end
     MapitPostcodeResponse.new(postcode, location, error)
-  end
-
-  def setup_navigation_helpers
-    content_item = content_store.content_item(BASE_PATH)
-    @navigation_helpers = GovukNavigationHelpers::NavigationHelper.new(content_item)
   end
 end
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -10,7 +10,7 @@ class RootController < ApplicationController
   include ActionView::Helpers::TextHelper
 
   before_filter :validate_slug_param, only: [:publication]
-  before_filter :setup_navigation_helpers
+  before_filter :get_content_item
   before_filter :block_empty_format, only: [:jobsearch, :publication]
 
   rescue_from RecordNotFound, with: :cacheable_404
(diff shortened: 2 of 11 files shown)


app/controllers/find_local_council_controller.rb
```diff
@@ -129,7 +129,14 @@ def fetch_location(postcode)
   end
 
   def setup_navigation_helpers
-    content_item = content_store.content_item(BASE_PATH)
-    @navigation_helpers = GovukNavigationHelpers::NavigationHelper.new(content_item)
+    @content_item = content_store.content_item(BASE_PATH).to_hash
+    # Remove the organisations from the content item - this will prevent the
```
tijmenb: Perhaps we can move this to a helper or the template?

tijmenb: Would also be good to explain why we don't want the organisations meta tag in this application.

test/test_helper.rb
```diff
@@ -31,6 +31,8 @@ class ActiveSupport::TestCase
   include ContentStoreHelpers
 
   setup do
+    stub_request(:get, %r{#{Plek.new.find("static")}/templates/locales/(en|cy)})
```
tijmenb: Do we need this?

app/controllers/find_local_council_controller.rb
rubenarakelyan: I've moved this to the parent ApplicationController

test/test_helper.rb
rubenarakelyan: 1. I've added some more comments about why we remove this key
2. I've removed this since it's not needed

app/controllers/find_local_council_controller.rb
```diff
@@ -1,7 +1,7 @@
 require "postcode_sanitizer"
 
 class FindLocalCouncilController < ApplicationController
-  before_filter :setup_navigation_helpers
+  before_filter -> { setup_navigation_helpers BASE_PATH }
```
jackscotti: I don't understand this change.

rubenarakelyan: The `setup_navigation_helpers` method has been centralised in the `ApplicationController`, therefore it needs to be provided with a base path to work with. This change passes in the base path from this controller to the method in the `ApplicationController` using the `before_filter`.

app/controllers/root_controller.rb
```diff
@@ -354,8 +354,7 @@ def custom_format_locals
   end
 
   def setup_navigation_helpers
-    content_item = content_store.content_item("/" + params[:slug])
-    @navigation_helpers = GovukNavigationHelpers::NavigationHelper.new(content_item)
+    super("/" + params[:slug])
```
tijmenb: Can you rename the `setup_navigation_helpers` here and call `setup_navigation_helpers` from that? That way we don't have to use `super`.

tijmenb:
`setup_navigation_helpers` might be confusing now, because it's also setting the instance variable `@content_item`. Could we come up with a better name?

app/controllers/root_controller.rb
rubenarakelyan: Done

rubenarakelyan:
`setup_navigation_helpers` is now called `setup_content_item_and_navigation_helpers`

PR #1033, opened 2016-10-27
Remove special case for business support pages

tijmenb:
Business support pages (like https://www.gov.uk/apprenticeshipsni) explicitly don't have a sidebar. This is redundant, because the sidebar won't show anyway when there's nothing to show (like related mainstream pages and external links).

I haven't found any good reason in the commit history why this page is a special case, so I think it would make sense to remove it. This means that if we'd ever want something in the sidebar here we could make it so.

🚫❄️

## Before (or after)

screen shot 2016-10-27 at 11 43 41

## After (or before)

screen shot 2016-10-27 at 11 43 48

### DIFF
--- a/app/views/root/_base_page.html.erb
+++ b/app/views/root/_base_page.html.erb
@@ -20,6 +20,4 @@
   </div>
 </main>
 
-<% if local_assigns.fetch(:show_related_items, true) %>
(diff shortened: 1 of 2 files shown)


davidbasalla:
❤️

PR #1036, opened 2016-11-01
Pre-Migration Frontend Refactor - Help format

whoojemaflip:
The goal of the refactoring is to speed up and de-risk the process of migrating formats to publishing-api. It would be challenging to unroll all the logic in the `RootController#publication` catch-all handler so instead we re-route requests on a per-format basis. By enabling us to write idiomatic controllers that wrap the rendering cycle of each format, we should drive out the specific rendering requirements for each format making it easier to migrate them to new content-store schemas when the time comes.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,9 +13,6 @@ Frontend::Application.routes.draw do
   # http://stackoverflow.com/a/3443678
   get "*path.gif", to: proc { |env| [404, {}, ["Not Found"]] }
 
-  get "/help", to: "help#index"
-  get "/tour", to: "help#tour"
-
   get "/find-local-council" => "find_local_council#index"
   post "/find-local-council" => "find_local_council#find"
   get "/find-local-council/:authority_slug" => "find_local_council#result"
@@ -46,8 +43,13 @@ Frontend::Application.routes.draw do
 
   get ":slug/y(/*responses)" => "simple_smart_answers#flow", :as => :smart_answer_flow
 
+  # Help pages
+  get "/help", to: "help#index"
+  get "/tour", to: "help#tour"
+  get "*slug", slug: %r{help/.+}, to: "help#show", constraints: FormatRoutingConstraint.new('help_page')
+
   with_options(to: "root#publication") do |pub|
-    pub.get "*slug", slug: %r{(done|help)/.+}
+    pub.get "*slug", slug: %r{done/.+}
     pub.get ":slug/print", variant: :print
     pub.get ":slug/:part/:interaction", as: :licence_authority_action
 
--- a/lib/format_routing_constraint.rb
+++ b/lib/format_routing_constraint.rb
@@ -1,0 +1,23 @@
+class FormatRoutingConstraint
+  def initialize(format, artefact_retriever: ArtefactRetrieverFactory.artefact_retriever)
+    @format = format
+    @artefact_retriever = artefact_retriever
+  end
+
+  def matches?(request)
+    slug = request.params.fetch(:slug)
+    edition = request.params.fetch(:edition, nil)
+    artefact = cacheable_artefact(slug, edition, request)
+    artefact.format == @format if artefact && artefact.respond_to?(:format)
+  end
+
+  private
+
+  def cacheable_artefact(slug, edition, request)
+    request.env[:__artefact_cache] ||= {}
+    request.env[:__artefact_cache][slug] ||=
+      @artefact_retriever.fetch_artefact(slug, edition)
+  rescue
+    nil
+  end
+end
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -2,6 +2,7 @@ require "slimmer/headers"
 
 class HelpController < ApplicationController
   before_filter :set_expiry
+  before_filter :redirect_if_api_request
 
   def index
     setup_content_item_and_navigation_helpers("/help")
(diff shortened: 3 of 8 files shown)


binaryberry:
Thanks for having a go at slaying the dragon - a few rubocop things make CI fail, if you fix them we could see if anything else brakes?

app/controllers/front_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+      # legacy
+      RootController.new
+    end.tap do |controller|
+      controller.request = request
+      controller.response = response
+    end
+  end
+
+  # https://gdstechnology.blog.gov.uk/2017/06/01/how-we-implemented-the-front-controller-pattern-in-rails-and-why-you-shouldnt-do-that
+  def action_from_method(method)
```
davidbasalla: I think we only need to support GET and POST for frontend, not necessarily important at this point though...

config/routes.rb
```diff
@@ -35,18 +35,18 @@
   # Jobssearch form override (English and Welsh variants)
   constraints(slug: /(jobsearch|chwilio-am-swydd)/) do
     get "/:slug.json"      => redirect("/api/%{slug}.json")
-    get "/:slug(.:format)" => "root#jobsearch"
+    get "/:slug(.:format)" => "jobsearch#show"
   end
 
   # Transaction finished pages
```
davidbasalla: Oh, are these some sort of special case of a done page? I wonder if we could clean this up somehow (probably as a different piece of work...)

h-lame:
I quite like the idea of this, but I admit to being pretty surprised that it actually works (does it?).  Rails controllers are weird, and I'm not entirely sure that we can instantiate one, call the action method on it, and expect it to work (even if we are setting the request and response).  It feels like something that might break as and when rails versions change.

Still, I do really like the obvious separation of concerns here - there's a thing that deals with each type of content request and that feels good.  Would it be possible to achieve this by using some other object instead of controllers?  We'd have to define an interface to let these non-controllers tell the `FrontController` what to do interms of the HTTP response and view rendering, but hopefully it's a minimal set of the total controller API.  How much controller-y-ness do we really need to show a page; decide to redirect or render, decide which template to render, maybe set some headers?

I could be way wrong about how much controller-y-ness we'd need in this API (if it's a lot then we probably should just use controllers), and I could also be wrong about instantiating a controller within another controller being bad/impossible/painful (if it works then maybe it's ok).  Keen to hear what other folk think.

whoojemaflip:
I quite like the idea that - although this is a weird pattern - the weirdness doesn't spread beyond this class. I was thinking we might want to wrap this functionality up in a Middleware object, so it's executed before the Rails router. That would let us do format-based routing like a normal app, although it would hide some important implementation away in `/lib`. Might be worth prototyping if this front-controller thing doesn't have legs. Thoughts?

The problem I found with using a not-controller object to handle the per-format requests is that I wound up reimplementing a subset of the ActionController DSL. It's what led me to this approach.

That said, there are plenty of ways of tackling this. I'm not particularly attached to this one, and as it's still sorta pseudo-code it doesn't actually work yet. Lets see if I can get it to run for a simple example this morning.  I agree that it might be an even worse idea than it looks.

config/routes.rb
whoojemaflip: they'll need doing at some point. Might back out of this for now though.

h-lame:
Oh! Making this into middleware, sounds promising as at that point our controllers would be invoked normally by rails, right?  I wonder if it'd be possible to have middleware re-write the request path internally so that `/*slug` becomes `/:format/*slug` and our routes.rb becomes more standard (sort of like apache re-write rules)?  Wonder if that might be even weirder though, and yes, hides some core routing functionality in `/lib`.

whoojemaflip:
The middleware approach is also a bit gnarly. Possible, but gnarly.

Owing to all the edge-cases captured in the [routes config](https://github.com/alphagov/frontend/blob/master/config/routes.rb) it would be tricky to determine the `slug` without repeating at least some of that logic.

There's also the potential for additional API calls to be caused by router/middleware logic disparities, potentially leading to 4x calls to backend systems to retrieve a content item that would currently only take 1. This is fixable, but would require a substantial rewrite of the app, rather than the step-by-step approach favoured here.

So not impossible, but definitely potential for gnar. I'm happy to go deeper on it if people think it's a really good idea.

app/controllers/front_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    # set the params hash to the expected values
+    # not sure if this is useful
+    request.params[:controller] = format_controller(edition.format).class.to_s.downcase.gsub('controller', '')
+    request.params[:action] = action_from_request_method(request.method).to_s
+
+    # delegate this request to the appropriate controller
+    status, headers, body = format_controller(edition.format).dispatch(
+      action_from_request_method(request.method),
+      request
+    )
```
pixeltrix: The `MyController#dispatch` method is not documented API - the method has changed signature from Rails 4.2 to Rails 5.0. I would recommend not using this private API.

lib/format_routing_constraint.rb
```diff
@@ -0,0 +1,23 @@
+class FormatRoutingConstraint
+  def initialize(format, artefact_retriever = PreconfiguredArtefactRetriever)
```
thehenster: I don't know have loads of context so this is more of a 'how I read it' comment than a suggestion.

Passing in `artefact_retriever = PreconfiguredArtefactRetriever` confused me for a bit because it made me think `PreconfiguredArtefactRetriever` was a type of `ArtefactRetriever`, or at least had the same interface. The signatures of `#initialize` and `#fetch_artefact` are not quite compatible.

I dunno what might be better. Renaming? Making them a bit more polymorphic? Or maybe this is just a transitional thing?

whoojemaflip: You're absolutely right, my thinking was a bit muddy here as I was imagining something quite different when I originally built the thing.

I've replaced the `PreconfiguredArtefactRetriever` class with an `ArtefactRetrieverFactory` static class which simply configures and returns an `ArtefactRetriever`. 

Thanks for the nudge!

davidbasalla:
As we plan on going with this (second) approach, I'm wondering whether we should clean up the commits, especially the first one? That commit introduced both the FrontController (which we steered away from afterwards) and extracted some template files, so it might be good to split those two things out into separate commits. It also added a test helper which we removed in the second commit without ever using it. So maybe just massaging the commits a little bit so they become less confusing for future us?

whoojemaflip:
~~ much rebasing occurs ~~

davidbasalla:
Looks good! Merging

PR #1041, opened 2016-11-10
Convert check vehicle tax

brenetic:
`check-vehicle-tax` was a custom slug. It is now editable in Publisher as a standard local transaction start page.

[Trello card](https://trello.com/c/qzrzVSAP/545-convert-check-vehicle-tax-to-normal-start-page-2)

# Before

before

# After

after


sihugh:
🎉

PR #1043, opened 2016-11-15
Move answer pages to own controller

davidbasalla:
This PR moves answer pages to their own controller, following the same pattern as Help pages (https://github.com/alphagov/frontend/pull/1036).

The first three commits do a bit of clean up for things that we came across while implementing this. Have a look at the commit messages for more details.

NOTE: There is a root_controller_test.rb file that uses an artefact of format 'answer' to test how the RootController works. We thought about updating this to use another format, but eventually (after we finish with  https://trello.com/c/phpe3PAO/546-refactor-frontend-5) all formats will have their own controllers anyway, so we can just remove this whole test file along with the RootController. So in the meantime we're happy to leave this as is.


test/integration/answer_test.rb
```diff
@@ -0,0 +1,74 @@
+# encoding: utf-8
```
whoojemaflip: I know it's copy-pasta but for Ruby > 2.0 there's no need to explicitly set the encoding to UTF-8 like this.

app/controllers/answer_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  before_filter :redirect_if_api_request
+
+  def show
+    setup_content_item_and_navigation_helpers("/" + params[:slug])
+    @publication = PublicationPresenter.new(artefact)
+    @edition = params[:edition]
+    set_language_from_publication(@publication)
+  end
+
+  private
```
brenetic: ◀️  (sorry)

PR #1045, opened 2016-11-16
Deal with some style comments hanging over from PR 1043

issyl0:
- The PR was merged before we got the chance to deal with https://github.com/alphagov/frontend/pull/1043#discussion_r88240341 and https://github.com/alphagov/frontend/pull/1043#discussion_r88241206. They're
tiny, but so they don't hang around forever, I'm fixing them now.
- The `answer_controller.rb` change was a copy/paste, so fix it at source as well in `help_controller.rb`.


issyl0:
Welcome back @fofr! ⭐️

PR #1046, opened 2016-11-16
Move video pages to own controller

davidbasalla:
This PR moves video pages to their own controller, following the same pattern as Help pages (https://github.com/alphagov/frontend/pull/1036).

The first two commits do a bit of clean up for things that we came across while implementing this. Have a look at the commit messages for more details.


issyl0:
This will also make it easier to know where the video code is when we come to delete the format.

PR #1047, opened 2016-11-16
Remove unused content security policy code

tijmenb:
This was introduced in https://github.com/alphagov/frontend/pull/733. It hasn't made it to production yet and we have no immediate plans to do so. I really hope we pick this up again, but in the meantime I think it's good to remove the code to avoid clutter and unnecessary maintenance.

Allows us to remove one more thing from the deprecated alphagov-deployment (https://github.gds/gds/alphagov-deployment/pull/1289).

https://trello.com/c/WkJ6qbE6/224-12-factorise-frontend


boffbowsh:
@davbo might have opinions on this. Was there a reason it never got turned on?

tijmenb:
Time. It was created by @alexmuller, so I'll let him merge.

davbo:
I am interested! It would be great if we could deploy CSP more widely on gov.uk. 

That said it will take effort and I suspect this particular code isn't how that ends up happening so removal for now doesn't set us back much, especially if it clears up some of the said "deprecated alphagov-deployment" repo. What do you think @alexmuller ?

Does the service on `#{Frontend.govuk_website_root}/e` exist in production?

davbo:
@boffbowsh the reasons it was never switched on are in here https://gdstechnology.blog.gov.uk/2015/02/12/experimenting-with-content-security-policy-on-gov-uk/ in particular

> This part of the CSP 2 specification works well, but because it's not widely available in browsers yet we're unable to use CSP to prevent inline JavaScript from running.

The policy here allows `'unsafe-inline'` so I can understand not pushing for it in production without the CSP2 support (which is now available in pretty much all browsers).

fofr:
We have made steps toward removing some of those inline scripts. eg Analytics is no longer an inline script.

PR #1050, opened 2016-11-17
Move smart answer start pages to own controller

davidbasalla:
This PR moves simple smart answer start pages to their own controller, following the same pattern as Help pages (https://github.com/alphagov/frontend/pull/1036).

The first commit does a bit of clean up for things that we came across while implementing this. Have a look at the commit message for more details.

### DIFF
--- a/app/controllers/simple_smart_answers_controller.rb
+++ b/app/controllers/simple_smart_answers_controller.rb
@@ -2,15 +2,16 @@ require 'simple_smart_answers/flow'
 
 class SimpleSmartAnswersController < ApplicationController
   before_filter :validate_slug_param
-  before_filter :set_expiry
-  before_filter :setup_edition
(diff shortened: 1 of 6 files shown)


app/controllers/simple_smart_answers_controller.rb
```diff
@@ -40,6 +47,17 @@ def change_completed_question_path(question_number)
     smart_answer_path_for_responses(@flow_state.completed_questions[0...question_number], previous_response: @flow_state.completed_questions[question_number].slug)
   end
 
+  def artefact
```
sihugh: What's the difference between this and `fetch_artefact`?  We use both here, but they seem to do similar things

davidbasalla:
@sihugh Addressed your comment, turns out the `fetch_artefact` call was redundant so we removed the line

PR #1052, opened 2016-11-21
Move guides to own controller

davidbasalla:
This PR moves guide pages to their own controller, following the same pattern as Help pages (https://github.com/alphagov/frontend/pull/1036).

This format is different from the previous ones in that it provides printing functionality.


issyl0:
Seems OK, though the linter is unhappy about travel advice page tests.

davidbasalla:
Yeah that's to do with an unrelated update to govuk_content_schemas. Will need to wait for https://github.com/alphagov/frontend/pull/1053 to get merged, then this should be happy

PR #1053, opened 2016-11-21
Sort countries by utf-8

NeilvB:
The pre-exisitng sort method did not take utf-8 characters such as
a-tilde into account. This commit uses a method [1] to sort by utf-8.

1. http://stackoverflow.com/a/17172055

Note that this PR follows updates to the associated `gov_uk_content_schemas` test fixture (https://github.com/alphagov/govuk-content-schemas/pull/439)


davidbasalla:
Looks good!

PR #1060, opened 2016-11-23
Move programme format to controller

emmabeynon:
This PR moves guide pages to their own controller, following the same pattern as Help pages (#1036).

This format is similar to guides, including that it provides printing functionality.

Paired with @davidbasalla


app/controllers/programme_controller.rb
```diff
@@ -0,0 +1,61 @@
+require "slimmer/headers"
+
+class ProgrammeController < ApplicationController
+  before_filter -> { set_expiry unless viewing_draft_content? }
+  before_filter :redirect_if_api_request
```
whoojemaflip: would it make sense for the redirect to come before the expiry header?

config/routes.rb
```diff
@@ -56,11 +56,14 @@
...
 
+  # Programme pages
+  get ":slug/print", to: "programme#show", variant: :print, constraints: FormatRoutingConstraint.new('programme')
+  get ":slug(/:part)", to: "programme#show", constraints: FormatRoutingConstraint.new('programme')
+
   # Transaction pages
   get ":slug", to: "transaction#show", constraints: FormatRoutingConstraint.new('transaction')
 
   with_options(to: "root#publication") do |pub|
-    pub.get ":slug/print", variant: :print
```
whoojemaflip: nm, just seen the next commit :)

PR #1061, opened 2016-11-23
Move completed transaction to controller

emmabeynon:
This PR moves completed transaction pages to their own controller, following the same pattern as Help pages (#1036).

We also move legacy completed transactions to the completed transaction controller as they are similar.

Paired with @davidbasalla

### DIFF
--- a/app/controllers/completed_transaction_controller.rb
+++ b/app/controllers/completed_transaction_controller.rb
@@ -1,0 +1,29 @@
+require "slimmer/headers"
+
+class CompletedTransactionController < ApplicationController
+  before_filter -> { set_expiry unless viewing_draft_content? }
+  before_filter :redirect_if_api_request
+
+  def show
+    setup_content_item_and_navigation_helpers("/" + params[:slug])
+    @publication = PublicationPresenter.new(artefact)
+    @edition = params[:edition]
+  end
+
+private
+
+  def artefact
+    @_artefact ||= ArtefactRetrieverFactory.artefact_retriever.fetch_artefact(
+      params[:slug],
+      params[:edition]
+    )
+  end
+
+  def redirect_if_api_request
+    redirect_to "/api/#{params[:slug]}.json" if request.format.json?
+  end
+
+  def viewing_draft_content?
+    params.include?('edition')
+  end
+end
--- a/app/controllers/completed_transaction_controller.rb
+++ b/app/controllers/completed_transaction_controller.rb
@@ -1,13 +1,24 @@
 require "slimmer/headers"
 
 class CompletedTransactionController < ApplicationController
-  before_filter -> { set_expiry unless viewing_draft_content? }
   before_filter :redirect_if_api_request
+  before_filter -> { set_expiry unless viewing_draft_content? }
+
+  # These 2 content items have the format of 'CompletedTransaction' but do not
+  # follow the convention of their slug being prefixed with '/done'`. They also
+  # use a different template.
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -49,8 +44,13 @@
...
   get ":slug", to: "video#show", constraints: FormatRoutingConstraint.new('video')
 
+  # Legacy done pages
+  constraints(slug: /(transaction-finished|driving-transaction-finished)/) do
+    get "/:slug(.:format)" => "completed_transaction#show"
+  end
+
   # Done pages
-  get "*slug", slug: %r{done/.+}, to: "root#publication"
+  get "*slug", slug: %r{done/.+}, to: "completed_transaction#show", constraints: FormatRoutingConstraint.new('completed_transaction')
```
davidbasalla: Spoke to @whoojemaflip about removing the '/done' slug constraint so that we can remove the Legacy done pages constraints from the routes.rb file. Will attempt this after https://github.com/alphagov/frontend/pull/1060 is merged

davidbasalla:
Marking as 'Do not merge' as https://github.com/alphagov/frontend/pull/1062 should be merged first

davidbasalla:
This is ready for review again, I've rebased on top of latest changes to master.

app/controllers/completed_transaction_controller.rb
```diff
@@ -0,0 +1,37 @@
+require "slimmer/headers"
+
+class CompletedTransactionController < ApplicationController
+  before_filter :redirect_if_api_request
+  before_filter -> { set_expiry unless viewing_draft_content? }
+
+  LEGACY_SLUGS = ["transaction-finished", "driving-transaction-finished"].freeze
```
whoojemaflip: it's not at all obvious why this is here. Could there be a comment to explain? It will be tricky to investigate the git history of this line, once this is merged.

config/routes.rb
```diff
@@ -30,12 +30,11 @@
...
   # Done pages
-  get "*slug", slug: %r{done/.+}, to: "root#publication"
-
-  # Transaction finished pages
-  constraints(slug: /(transaction-finished|driving-transaction-finished)/) do
-    get "/:slug.json"      => redirect("/api/%{slug}.json")
-    get "/:slug(.:format)" => "root#legacy_completed_transaction"
+  constraints FormatRoutingConstraint.new('completed_transaction') do
+    with_options(to: "completed_transaction#show") do
+      get "/:slug(.:format)", slug: /(transaction-finished|driving-transaction-finished)/ # Legacy
```
whoojemaflip: Could this not have it's own controller method? Seems that it's entirely different to the common case.

app/controllers/completed_transaction_controller.rb
davidbasalla: Added a comment

config/routes.rb
davidbasalla: So I've updated the routes file in the next commit, what do you think. It removes the duplication of those 2 specific slugs and just leaves it in the Controller. I'd say that there's not a lot of difference between the legacy and the normal case, other than the missing '/done' slug prefix and rendering a different partial. So the controller methods would be identical apart from the render call. The partials are actually also very similar, and the legacy ones should be removed once we do some house keeping. I've added [a story for that](https://trello.com/c/MsS4wrrf/479-convert-legacy-done-transaction-pages), what do you think?

whoojemaflip: 🌮  🎉

whoojemaflip:
Squash and merge!

PR #1062, opened 2016-11-24
Fix error handling for unsupported routes

davidbasalla:
This PR aims to fix the following issues:
- more comprehensive test coverage for 'parts' for Guides and Programmes
- avoid ['refactored' formats](https://trello.com/c/phpe3PAO/546-refactor-frontend-overview-card-5) accidentally being handled by RootController 
- 500 errors for POST requests that actually 'GET' content ([as seen in Errbit](https://errbit.publishing.service.gov.uk/apps/53611a4b0da1151271001055/problems/583454d5657863164c0d0000))
- update RootController tests to no longer use refactored formats as test data
- reinstate functionality for certain formats to reroute to the base slug when a 'part' is requested

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -24,9 +24,13 @@ Frontend::Application.routes.draw do
     country.get "/foreign-travel-advice/:country_slug(/:part)", as: :travel_advice_country
   end
 
-  # Campaign pages.
-  get "/ukwelcomes", to: "campaign#uk_welcomes" # This is a special case.
-  get ":slug", to: "campaign#show", constraints: FormatRoutingConstraint.new('campaign')
+  # Help pages
+  get "/help", to: "help#index"
+  get "/tour", to: "help#tour"
+  get "*slug", slug: %r{help/.+}, to: "help#show", constraints: FormatRoutingConstraint.new('help_page')
+
+  # Done pages
+  get "*slug", slug: %r{done/.+}, to: "root#publication"
 
   # Transaction finished pages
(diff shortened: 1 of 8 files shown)


config/routes.rb
```diff
@@ -36,32 +40,47 @@
...
 
-  # Help pages
-  get "/help", to: "help#index"
-  get "/tour", to: "help#tour"
-  get "*slug", slug: %r{help/.+}, to: "help#show", constraints: FormatRoutingConstraint.new('help_page')
+  # Campaign pages.
+  get "/ukwelcomes", to: "campaign#uk_welcomes" # This is a hardcoded page in Frontend and does not have an associated artefact
+  constraints FormatRoutingConstraint.new('campaign') do
+    get ":slug", to: "campaign#show"
+    get ":slug/:part", to: redirect('/%{slug}') # Support for campaigns that were once a format with parts
```
sihugh: Weirdly [Transactions and Campaigns register exact routes](https://github.com/alphagov/publisher/blob/c1506fc0abeab2b5d88eb06582ab897e552b1db2/lib/registerable_edition.rb#L30) (along with Help pages) so a request for `:slug/:part` won't get as far as Frontend for pages with these formats.  I'm not sure why these ones in particular.

davidbasalla: Ah good find! So that means we don't have to worry about doing the 'part' redirect for Transactions, Campaigns and Help Pages...

davidbasalla: Thanks @sihugh, have removed the routing for those formats and updated the commit message

PR #1063, opened 2016-11-25
Bump Slimmer and update code

Davidslv:
Finding Things has been working on migrating related links, part of that
work was to replace the breadcrumbs on a number of different frontend
applications like:

- calendars
- calculators
- business-support-finder
- frontend
- licence-finder
- smart-answers

That work has already been done, this commit aims to clean up things we
have missed during the migration.

This includes

- upgrading Slimmer to 10.0.0
- deletion of govuk-component-helpers.scss
- rename [Slimmer::SharedTemplates to Slimmer::GovukComponents](https://github.com/alphagov/slimmer/pull/173)

Trello: https://trello.com/c/5Utjs9qw/300-related-links-cleanup

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,7 @@ gem 'sprockets-rails', "2.3.3"
 if ENV['SLIMMER_DEV']
   gem 'slimmer', path: '../slimmer'
 else
-  gem 'slimmer', '~> 9.0.0'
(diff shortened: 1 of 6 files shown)


tijmenb:
For context, the CSS can be deleted because that now lives in static: https://github.com/alphagov/static/pull/855

PR #1035, opened 2016-10-31
Add a new assisted digital satisfaction survey

h-lame:
This survey is specifically about help with fees and we only want to show
it on certain done pages.  For now this is a bit experimental, so we don't
want to add it to the documents themselves as a toggle selectable in
publisher.  We've left it as a hardcoded list in
lib/assisted_digital_satisfaction_survey.rb.

Note that the backend for this is handled by the feedback app (see https://github.com/alphagov/feedback/pull/179 and https://github.com/alphagov/feedback/pull/189 and https://github.com/alphagov/feedback/pull/190 - all of these should be merged and deployed before this is).

Trello: https://trello.com/c/kZCX6dGB/53-assisted-digital-support-on-done-page-survey

PR #1066, opened 2016-11-29
Remove support for legacy completed transactions

davidbasalla:
We will have updated the content for the 2 legacy completed transactions to be normal completed transactions, so can remove some from code from Frontend. We still do not want to show the user satisfaction survey for them, as many transactions might be linking to them. 

For https://trello.com/c/MsS4wrrf/575-refactor-frontend-convert-legacy-done-transaction-pages

Marked as Do Not Merge as this will require some content changes first. Will self-merge before we deploy this. Deploy scheduled for 11am on 1/12/2016

# Before
screen shot 2016-11-30 at 15 45 36

# After
screen shot 2016-11-30 at 15 45 09


app/views/completed_transaction/show.html.erb
```diff
@@ -38,47 +38,50 @@
     </div>
   <% end %>
 
-  <h2 class="satisfaction-survey-heading">Satisfaction survey</h2>
+  <% if show_survey? %>
+    <h2 class="satisfaction-survey-heading">Satisfaction survey</h2>
```
whoojemaflip: might it be clearer if this survey markup live in a partial?

h-lame: *cough* https://github.com/alphagov/frontend/pull/1035 ;)  If you can leave this as-is it might make my eventual rebase merge nightmare easier.

davidbasalla: Ah nice. Okay I won't extract this into a partial then.

PR #1067, opened 2016-11-30
Improve the handling of content API errors

whoojemaflip:
The format routing constraint didn't handle errors coming back from the
content API. This introduced potential issues around content API timeouts
which wouldn't be cached and therefore could trigger a cascading error in
Frontend.

Here we introduce a new class which is responsible for memoising the
response from the content API, and catches any errors. Future requests
will return early if an error was encountered, saving us a long wait.

The error-caching behaviour makes this new class usable in our controllers
too, where we can rethrow or interrogate the exception and handle it
accordingly.

[Trello](https://trello.com/c/2LhEQXWF/568-refactor-frontend-cleanup)


app/controllers/error_controller.rb
```diff
@@ -0,0 +1,8 @@
+class ErrorController < ApplicationController
```
davidbasalla: We have controller tests for the other controllers, so should do the same for this one?

davidbasalla:
Should we add some sort of integration tests? There are some for [checking timeouts in page_rendering_test.rb](https://github.com/alphagov/frontend/blob/master/test/integration/page_rendering_test.rb#L4-L46). We could have similar ones for each refactored format (sort of like Rspec shared examples), since the page_rendering_test.rb will disappear once we're done refactoring... ?

lib/content_api_error_routing_constraint.rb
```diff
@@ -0,0 +1,10 @@
+class ContentApiErrorRoutingConstraint
```
davidbasalla: We could add a unit test for this class

lib/artefact_retriever_with_request_cache.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    begin
+      cache_artefact slug, edition
+    rescue => err
+      cache_error err
+      false
+    end
+  end
+
+  def cache_artefact(slug, edition)
```
davidbasalla: `cache_artefact` and `cache_error` could be private methods since they're only called from within this class and because we don't have unit tests for these methods?

whoojemaflip:
@davidbasalla :eyes:?

app/controllers/error_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+  def handler
+    # Our error is cached against the request object.
+    content_api_cached_error = ArtefactRetrieverFactory.caching_artefact_retriever
+    content_api_cached_error.set_request(request)
+
+    # This should never happen - we are only routed to this
+    # controller in the event of an API error occurring.
+    error 500 unless content_api_cached_error.error?
```
davidbasalla: What's the benefit of having this line if this will never happen?

davidbasalla:
With [this commit](https://github.com/alphagov/frontend/pull/1067/commits/0278247f83768cc3d02dd2d9cd33e0b6aaca6827) we can now remove `require 'application_controller'` from https://github.com/alphagov/frontend/blob/master/test/unit/artefact_retriever_test.rb#L2

PR #1068, opened 2016-11-30
Move business support to own controller

davidbasalla:
This PR moves business support pages to their own controller, following the same pattern as Help pages (#1036).

### DIFF
--- a/app/controllers/business_support_controller.rb
+++ b/app/controllers/business_support_controller.rb
@@ -1,0 +1,34 @@
+require "slimmer/headers"
+
+class BusinessSupportController < ApplicationController
+  before_filter :redirect_if_api_request
+  before_filter -> { set_expiry unless viewing_draft_content? }
+
+  def show
+    setup_content_item_and_navigation_helpers("/" + params[:slug])
+    @publication = PublicationPresenter.new(artefact)
+    @edition = params[:edition]
+    set_language_from_publication(@publication)
+  end
+
+private
+
+  def artefact
+    @_artefact ||= ArtefactRetrieverFactory.artefact_retriever.fetch_artefact(
+      params[:slug],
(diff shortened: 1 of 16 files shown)


test/integration/business_support_test.rb
```diff
@@ -0,0 +1,81 @@
...
+      within '#content' do
+        within 'header' do
+          assert page.has_content?("Business support example")
+        end
+      end
+
+      assert_current_url "/business-support-example?edition=5"
+    end
+
+    should "render an answer in Welsh correctly" do
```
emmabeynon: Should this be 'business support' instead of 'answer'?

app/controllers/business_support_controller.rb
```diff
@@ -0,0 +1,34 @@
+require "slimmer/headers"
+
+class BusinessSupportController < ApplicationController
+  before_filter -> { set_expiry unless viewing_draft_content? }
+  before_filter :redirect_if_api_request
```
emmabeynon: I think in previous pull requests we decided to redirect before setting the expiry.

test/functional/business_support_controller_test.rb
```diff
@@ -0,0 +1,40 @@
+require "test_helper"
+
+class BusinessSupportControllerTest < ActionController::TestCase
+  context "GET show" do
+    setup do
+      @artefact = artefact_for_slug('business-support-example')
+      @artefact["format"] = "answer"
```
emmabeynon: Should this be 'business-support' instead of 'answer'?

davidbasalla:
Thanks for the review @emmabeynon, I've fixed and rebased. Added another commit to reorder the before_filters for all of our new controllers so that the api redirect always happens before we set the expiry headers.

PR #1069, opened 2016-12-01
Avoid rendering the related items sidebar when empty

Davidslv:
When there are no related items we should not render the sidebar.
At the moment it is rendering an horizontal blue bar, which shouldn't because there's no related items.

Trello: https://trello.com/c/5TCdhKVN/320-fix-the-hovering-blue-sidebar-on-pages-without-related-links

## Before
screen shot 2016-12-01 at 14 57 01

## After

screen shot 2016-12-01 at 14 57 12

### DIFF
--- a/app/views/find_local_council/_base_page.html.erb
+++ b/app/views/find_local_council/_base_page.html.erb
@@ -21,8 +21,8 @@
   </div>
 </main>
 
-<% if @navigation_helpers %>
-<div class="related-container">
-  <%= render partial: 'govuk_component/related_items', locals: @navigation_helpers.related_items %>
-</div>
+<% if @navigation_helpers.related_items[:sections].any? %>
+  <div class="related-container">
+    <%= render partial: 'govuk_component/related_items', locals: @navigation_helpers.related_items %>
+  </div>
 <% end %>


tijmenb:
This problem applies to all applications that use the component. Have you considered changing the component itself instead?

fofr:
@tijmenb Good question. Which part should be responsible? Should the component always expect to receive good data?

tijmenb:
I guess you could say that the data passed to the component is "good" already (`sections: []`). But the [readme doesn't specify](http://govuk-component-guide.herokuapp.com/components/related_items) so it's anyones guess actually :). 

But it seems practical to make the component able to handle this case, it saves a lot of `empty?` checks across GOV.UK.

nickcolley:
> Nick Colley
I’m not sure but my gut feeling is that the component should warn if it does not have the right data for a minimal render.
Then in your application you can make the decision if you want to render it or not, since you may need to accommodate for it not being there.
Whereas the component itself couldn’t make an assumption on where it’s used

> last time this came up, the outcome was apps are responsible for ensuing good data, components should loudly complain


>David Singleton [3:58 PM]  
last time this came up, the outcome was apps are responsible for ensuing good data, components should loudly complain
https://github.com/alphagov/static/blob/master/app/views/govuk_component/related_items.raw.html.erb#L2 =>
```raise unless sections```

screen shot 2016-12-01 at 16 57 31

Davidslv:
While I was doing this task I came to the same problem, should I fix it on the "frontend" application or on the component side?
Once I stepped into https://github.com/alphagov/static/blob/master/app/views/govuk_component/related_items.raw.html.erb#L2 I felt that someone already tried to avoid chaos with `sections ||= []` which I'm not a big fan of, this clearly misses a contract (https://en.wikipedia.org/wiki/Design_by_contract) between the "frontend" application and the component (the partial).

Given the situation I felt that it was more appropriate to fix it on this repo as I think there's no interest in rendering the div `.related-container` without any content.

Although I'm happy to change this to the component side but I think there's a need to clarify the component needs so all the rendering apps meet the requirements.
Also whatever decision is (or was) made it should be documented on the appropriate repos so everyone is on the same page.

Davidslv:
This PR is now being superseded by https://github.com/alphagov/static/pull/859

PR #1070, opened 2016-12-05
Move places to separate controller

davidbasalla:
This PR moves place/find-my-nearest pages to their own controller, following the same pattern as Help pages (#1036).

This is the first of the interactive formats so the controller is a bit more complex. Previously it supported GET and POST (for the LocationForm) requests, we've removed the POST as we're not updating any data.

### DIFF
--- a/app/controllers/place_controller.rb
+++ b/app/controllers/place_controller.rb
@@ -1,0 +1,78 @@
+require "slimmer/headers"
+
+class PlaceController < ApplicationController
+  before_filter :redirect_if_api_request
+  before_filter -> { set_expiry unless viewing_draft_content? }
+
+  CUSTOM_SLUGS = {
+    "report-child-abuse-to-local-council" => {
+      locals: {
+        option_partial: "option_report_child_abuse",
+        preposition: "for",
+      }
+    },
+  }.freeze
+
+  def self.custom_slugs; CUSTOM_SLUGS; end
+
+  def show
+    setup_content_item_and_navigation_helpers("/" + params[:slug])
+    @edition = params[:edition]
+
+    if !postcode_provided?
+      @publication = PublicationPresenter.new(artefact)
+    else
+      @publication = PublicationPresenter.new(artefact, places)
+      @postcode = params[:postcode]
+    end
+
+    if is_custom_slug?
+      render :show, locals: custom_slug_locals
+    else
+      render :show
+    end
+  end
+
+private
+
+  def artefact
+    @_artefact ||= ArtefactRetrieverFactory.artefact_retriever.fetch_artefact(
+      params[:slug],
+      params[:edition]
+    )
+  end
+
+  def postcode
+    PostcodeSanitizer.sanitize(params[:postcode])
+  end
+
+  def postcode_provided?
+    params[:postcode].present?
+  end
+
+  def redirect_if_api_request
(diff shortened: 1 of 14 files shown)


app/controllers/place_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    },
+  }.freeze
+
+  def self.custom_slugs; CUSTOM_SLUGS; end
+
+  def show
+    setup_content_item_and_navigation_helpers("/" + params[:slug])
+    @edition = params[:edition]
+
+    if !postcode_provided?
```
whoojemaflip: could we flip the logic on this `if..else` to remove the need for negation?

davidbasalla: Sure! My initial thinking was that it would be clearer to sort this 'chronologically', so the first request would likely be without the postcode, but happy to change it!

app/controllers/place_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    redirect_to "/api/#{params[:slug]}.json" if request.format.json?
+  end
+
+  def viewing_draft_content?
+    params.include?('edition')
+  end
+
+  def places
+    places = Frontend.imminence_api.places_for_postcode(artefact.details.place_type, postcode, Frontend::IMMINENCE_QUERY_LIMIT)
+    @location_error = LocationError.new(NO_LOCATION_ERROR) if places.blank?
```
whoojemaflip: if Imminence finds no places for a postcode, will it return an empty body or will it throw `NO_LOCATION_ERROR`?

app/controllers/place_controller.rb
```diff
@@ -0,0 +1,93 @@
+require "slimmer/headers"
+
+class PlaceController < ApplicationController
+  before_filter :redirect_if_api_request
+  before_filter -> { set_expiry unless viewing_draft_content? }
+
+  INVALID_POSTCODE_ERROR = "invalidPostcodeError".freeze
+  NO_LOCATION_ERROR = "validPostcodeNoLocation".freeze
+
+  CUSTOM_SLUGS = {
```
whoojemaflip: Might this be a lot simpler if it was handled in `routes.rb` and given it's own controller method?

app/controllers/place_controller.rb
davidbasalla: If Imminence finds no places in a given dataset it will return an empty body. Imminence can also throw the 'validPostcodeNoLocation' error for when MapIt can't find a location for a postcode AFAIK.

app/controllers/place_controller.rb
davidbasalla: Good point. I'll give it a go

davidbasalla: @whoojemaflip Have a look at https://github.com/alphagov/frontend/pull/1070/commits/09aa005b29376fccedbe3620993a3a323215cdf6. Initially we moved everything into a separate controller method but found there was quite a bit of duplication in the controller and controller tests, as the only way the 'report-child-abuse' differs is in the templates that it renders. What do you think?

davidbasalla:
@whoojemaflip Addressed your comments, see last commit

whoojemaflip:
Needs rebasing, but 👍 

:sheep: :it:

davidbasalla:
@whoojemaflip Have rebased, do you mind approving and merging if you're happy with it?

PR #1071, opened 2016-12-06
Move Local Transactions into their own controller

whoojemaflip:
- Local transactions now use their own controller.
- Removed all methods associated with local transactions from the root controller. 

[Trello card](https://trello.com/c/2QmNuV5e/578-refactor-frontend-localtransaction)


app/controllers/local_transaction_controller.rb
```diff
@@ -0,0 +1,132 @@
...
+private
+
+  def postcode_search
+    if postcode.nil?
+      @location_error = LocationError.new(INVALID_POSTCODE)
+      return
+    end
+
+    if local_authority_slug
+      return redirect_to local_transaction_results_path(local_authority_slug: local_authority_slug)
```
whoojemaflip: I don't think that the return value of `redirect_to` is part of it's public API.  Might it be a little more comfortable to phrase this differently so it always returns truthy, regardless of how `redirect_to` responds?

app/controllers/local_transaction_controller.rb
```diff
@@ -0,0 +1,132 @@
...
+    end
+  end
+
+  def local_authority_slug
```
whoojemaflip: This method is still pretty gruesome.

app/controllers/local_transaction_controller.rb
```diff
@@ -0,0 +1,129 @@
...
+    if request.post?
+      @postcode = postcode
+      la_slug = local_authority_slug
+
+      return redirect_to local_transaction_results_path(local_authority_slug: la_slug) if la_slug
+    end
+
+    respond_to do |format|
+      format.html { render }
+      format.json { render json: @publication.to_json }
```
davidbasalla: We're redirecting to '/api/slug.json' for the other formats, I'm pretty sure we could do the same here so that we're consistent. Also I think this only returns an empty `places` hash so isn't really useful.

app/controllers/local_transaction_controller.rb
```diff
@@ -0,0 +1,129 @@
+class LocalTransactionController < ApplicationController
+  include ActionView::Helpers::TextHelper
+
+  before_filter :validate_slug_param, only: :search
```
davidbasalla: Validating the slug is now a bit weird, because we already use the slug in the routes.rb file with the advanced constraints. So doing it here is happening too late. We've written up a story to push this into middle ware of some sort -> https://trello.com/c/leLnBdG4/580-refactor-frontend-cleanup-slug-validation. So unless this breaks any tests, I'd say remove the line for now.

app/controllers/local_transaction_controller.rb
```diff
@@ -0,0 +1,129 @@
...
+
+    find_la_slug
+  end
+
+  def find_la_slug
+    mapit_response = location_from_mapit
+
+    if mapit_response.location_found?
+      la_slug = LocalTransactionLocationIdentifier.find_slug(mapit_response.location.areas, artefact)
+      return la_slug if la_slug
```
davidbasalla: It feels a bit unusual to return out of the method here, was there a specific reason for it not to return at the end of the method?

davidbasalla:
For the other formats, we added integration tests for preview and translation functionality. Might be worth doing it for this format too for consistency's sake

app/controllers/local_transaction_controller.rb
brenetic: I'm eager to return 🐎

whoojemaflip:
<3

davidbasalla:
Could we remove 'app/views/root/local_transaction.html.erb'?

PR #1073, opened 2016-12-13
Remove the `application_controller#validate_slug_param` method

whoojemaflip:
There's no need for this anymore. The functionality was extracted into
a middleware component years ago. While there is an argument for some
way of improving the quality of URLs that get passed to upstream systems
this partial and confusing implementation isn't it.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -94,15 +94,6 @@ protected
     )
   end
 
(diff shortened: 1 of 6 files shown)


davidbasalla:
Looks fine, but let's make sure we investigate whether we need to do more work to guard against these sorts of inputs

PR #1074, opened 2016-12-14
Update GdsApiAdapters to 38.0.0

whoojemaflip:


test/unit/format_routing_constraint_test.rb
```diff
@@ -21,7 +21,7 @@ def request
   end
 
   def artefact(format)
-    stub(format: format)
+    stub(:[] => format)
```
davidbasalla: What does this do? Why did it need changing?

davidbasalla:
Looks good! Could you add a few details to the commit message about the changes (the change regarding OpenStructs and why you chose to not use the part helper any more), would be good for reviewing and if we ever look back at this commit.

test/unit/format_routing_constraint_test.rb
whoojemaflip: The `GdsApiAdapters::Response` object that is being doubled in this test has changed. It no longer implements a `method_missing` approach to interrogating the underlying JSON data, but instead gives us the [square brackets method as a way of querying an internal Hash of parsed data.](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/response.rb#L24)

PR #1075, opened 2016-12-15
Move licence format to separate controller

davidbasalla:
This PR moves licence pages to their own controller, following the same pattern as LocalTransactions Controller (https://github.com/alphagov/frontend/pull/1071).

We also remove RootController now that it's not used anymore! Party!

Paired with @emmabeynon 

TODO - test in integration


app/controllers/licence_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+
+  def search
+    @publication = publication
+    @interaction_details = licence_details
+
+    @postcode = postcode if postcode_search_submitted?
+    @location_error = location_error if postcode_search_submitted?
+
+    if postcode_search_submitted? && local_authority_slug
+      return redirect_to authority_path(slug: params[:slug], authority_slug: local_authority_slug)
```
whoojemaflip: it doesn't look like these `return` statements are necessary here.

app/controllers/licence_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    'permission-to-place-tables-and-chairs-on-the-pavement',
+    'pavement-or-street-display-licence',
+    'petroleum-storage-licence',
+    'weighbridge-operator-certificate',
+    'performing-animals-registration',
+    'approval-of-premises-for-civil-marriage-or-civil-partnership',
+    'licence-projection-over-highway-england-wales',
+  ].freeze
+
+  def search
```
whoojemaflip: Would it be possible to further highlight which code-paths are handling the `post` method? I assume they are almost entirely seperate

config/routes.rb
```diff
@@ -96,15 +96,17 @@
     get ":slug/:part", to: redirect('/%{slug}') # Support for places that were once a format with parts
   end
 
+  # Licence pages
+  constraints FormatRoutingConstraint.new('licence') do
+    get ":slug", to: "licence#search", as: "licence"
+    post ":slug", to: "licence#search" # Support for postcode submission which we treat as confidential data
+    get ":slug/:authority_slug(/:interaction)", to: "licence#authority", as: "authority"
```
whoojemaflip: `licence_authority` maybe? `authority_path` might be a little opaque.

app/controllers/licence_controller.rb
```diff
@@ -37,15 +37,19 @@ def search
...
     @interaction_details = licence_details
 
-    @postcode = postcode if postcode_search_submitted?
-    @location_error = location_error if postcode_search_submitted?
-
-    if postcode_search_submitted? && local_authority_slug
-      redirect_to licence_authority_path(slug: params[:slug], authority_slug: local_authority_slug)
-    elsif authority_choice_submitted?
-      redirect_to licence_authority_path(slug: params[:slug], authority_slug: CGI.escape(params[:authority][:slug]))
+    if request.post?
```
davidbasalla: @whoojemaflip Do you think this makes it more clear in terms of what's happening when a POST happens?

whoojemaflip: Much better I think. It might look gnarly, but that's because it is! I feel like this is much more understandable on an immediate level.

brenetic:
💯

PR #1082, opened 2016-12-16
Add an integration test for the /search route

whoojemaflip:
The SearchAPI class was exploding a bit, so here it is covered by an integration test with WebMocked data. This should ensure a smoother upgrade path for the GdsApiAdapters.

The unit test was mainly testing a complex system of mock objects, so that's gone.


davidbasalla:
The fixture files are pretty massive and hard to read. Would it be possible to reduce those to more toy-like examples so it's clearer what to watch out for in the tests?

test/integration/search_test.rb
```diff
@@ -7,11 +7,36 @@ class SearchTest < ActionDispatch::IntegrationTest
...
+      assert page.has_content? "Offshore sandwich device"
+      assert page.has_content? "flood risk in Sandwich"
+    end
+  end
+
+  context "given a search scoped to a manual" do
+    setup do
+      stub_search_to_return_results_scoped_to_manual
+      stub_scope_object
+      stub_unscoped_rummager_request
```
davidbasalla: I wasn't quite sure why we need to stub an unscoped rummager request in a test for a scoped search, but it appears to be needed. Good to have this spelled out more 👍

whoojemaflip:
Will anyone ever want to read the fixture json?

davidbasalla:
Nice! Merging

PR #1084, opened 2016-12-19
Extract similarities from mainstream formats

davidbasalla:
Now that we've refactored out the mainstream formats into separate controllers, we can remove duplication that piled up during the refactor.

For https://trello.com/c/7Tf6oIK6/590-refactor-frontend-tidy-up-duplicate-controller-support-methods


app/controllers/concerns/api_redirectable.rb
```diff
@@ -0,0 +1,15 @@
+module ApiRedirectable
+  extend ActiveSupport::Concern
+
+  included do
+    before_filter :redirect_if_api_request
+  end
+
+  def redirect_if_api_request
+    redirect_to "/api/#{slug_param}.json" if request.format.json?
```
whoojemaflip: for the avoidance of uncertainty, is it also worth guarding this with a test for `request.get?`?

davidbasalla:
Is there a way to reduce the duplication in the integration tests? We are testing the same thing there as well over and over again... 🤔

app/controllers/campaign_controller.rb
```diff
@@ -1,33 +1,17 @@
 class CampaignController < ApplicationController
-  before_filter :redirect_if_api_request
+  include ApiRedirectable
+  include Previewable
+
   before_filter -> { set_expiry unless viewing_draft_content? }
+  before_filter -> { setup_content_item_and_navigation_helpers("/" + params[:slug]) }, only: :show
```
whoojemaflip: Could this be a mixin too?

app/controllers/campaign_controller.rb
```diff
@@ -1,33 +1,17 @@
 class CampaignController < ApplicationController
-  before_filter :redirect_if_api_request
+  include ApiRedirectable
+  include Previewable
+
   before_filter -> { set_expiry unless viewing_draft_content? }
```
whoojemaflip: Could this be a mixin too?

app/controllers/licence_controller.rb
```diff
@@ -153,24 +141,7 @@ def local_authority_slug
     end
   end
 
-  def appropriate_slug_from_location(publication, location)
-    tier_override = :county_unitary if LICENCE_SLUGS_WITH_COUNTY_TIER_OVERRIDE.include?(publication.slug)
-    LicenceLocationIdentifier.find_slug(location.areas, publication.artefact, tier_override)
-  end
-
   def postcode
```
whoojemaflip: Could this be a mixin too?

app/controllers/place_controller.rb
```diff
@@ -23,19 +22,12 @@ def show
 
   def publication
     if postcode_provided?
-      PublicationPresenter.new(artefact, places)
+      PublicationWithPlacesPresenter.new(artefact, places)
```
whoojemaflip: 🎉

app/presenters/publication_with_parts_presenter.rb
```diff
@@ -0,0 +1,66 @@
+require 'gds_api/part_methods'
```
whoojemaflip: This will conflict with an incoming PR, fwiw

app/controllers/campaign_controller.rb
davidbasalla: Definitely could be! I guess I was wondering if it was worth it for a one-liner like this...

davidbasalla:
@whoojemaflip Pushed 3 more commits to address your comments, what do you think?

whoojemaflip:
🌾

PR #1085, opened 2016-12-19
Invalid URL fragment handling

whoojemaflip:
Due to the unusual way in which this app parses URL fragments, it is
possible for it to construct an invalid URL for a backend resource.

Invalid URLs will be caught by the version of GdsApiAdapters referenced
here, and now the application controller can handle this case like any
of the other failure cases and return a cacheable 404.


davidbasalla:
Looks good to me! ☀️

PR #1087, opened 2016-12-20
Tidy up the Gemfile

whoojemaflip:
It was well out of hand.  Nothing has changed here except for the order of the gems, which should now be sorted alphabetically.


davidbasalla:
👍

PR #1088, opened 2016-12-20
Delete TrackClick module

carvil:
This module has been moved to `static` in
https://github.com/alphagov/static/pull/864

That version of `static` is now live.


davidbasalla:
Seems good, checked in Integration

PR #1089, opened 2016-12-20
Upgrade a handful of dependencies

whoojemaflip:
Also taking this opportunity to remove the ENV switches in the gemfile, which are old and inconsistent between projects.


davidbasalla:
Seems fine, tested on Integration on a couple of formats

PR #1090, opened 2016-12-20
Fix bug with scoped search results

tijmenb:
In #1082 the `SearchApi` class was refactored. The internal `Searcher` class was inlined into `SearchApi`. However, we didn't realise that `SearchApi` isn't a normal class, but an object that is only instantiated in an initialiser.

This caused the instance variables to be persisted between requests, which in turn caused very odd behaviour, since the scoped search results would be remembered (but differently on different
machines/rails processes).

This PR makes a number of changes that fixes the issue will hopefully prevent it from happening again.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -12,7 +12,7 @@ class SearchController < ApplicationController
     if search_params.no_search? && params[:format] != "json"
       render action: 'no_search_term' and return
     end
-    search_response = search_client.search(search_params)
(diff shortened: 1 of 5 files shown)


davidbasalla:
How easy would it be to write a test for the odd behaviour that you were seeing?

tijmenb:
I'm not sure there would be a good test to write for that. The only thing that I can think of would just assert that different searches return different things, which seems odd. I don't think there's the danger of doing the same thing again here.

PR #1097, opened 2017-01-13
Add ‘Publishing’ to the name of the product

robinwhittleton:
The readme refers to ‘GOV.UK Frontend’ which collides with the other GOV.UK Frontend product (currently in alpha at https://github.com/alphagov/govuk_frontend_alpha). To reduce confusion among people familiar with GOV.UK applications I’ve amended the heading. This is the only place in the repo where alphagov/frontend is referred to as ‘GOV.UK Frontend’, and I don’t think the name change hampers understanding of what the repo contains.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# GOV.UK Frontend
+# GOV.UK Publishing Frontend
 
 Application to serve mainstream formats, the homepage, and search vertical for
 the GOV.UK single domain.


nickcolley:
Another option is leave this as is unless we rename this repo/codebase instances, this keeps historical context and does not distract from the fact that 'GOV.UK Publishing Frontend' can refer to the wider set of frontend applications.

I'd propose we add a sentence at the top of the README 'If you're not working on GOV.UK Publishing, you're probably looking for [GOV.UK Frontend](https://github.com/alphagov/govuk_frontend_alpha)' type affair.

We definitely need a long term plan to sort these naming conflicts though.

robinwhittleton:
I guess we may want to change the intro para to something like this too?
>This application renders the homepage of GOV.UK along with search and other common templates. If you’re looking for GOV.UK Frontend (the package to build GOV.UK-styled services) then you can find it at https://www.github.com/alphagov/govuk_frontend

fofr:
@robinwhittleton Any updates on this?

tijmenb:
I've updated the title in https://github.com/alphagov/frontend/pull/1128, so I think this can be closed. Feel free to reopen if not.

PR #1098, opened 2017-01-16
Move to new ci

mgrassotti:
Trello: https://trello.com/c/KSnWfoTy/574-move-frontend-to-new-ci-infrastructure-1

The old CI infrastructure is in the process of being wound down so we need to move our apps to the new one.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,35 +1,62 @@
 #!/usr/bin/env groovy
 
 REPOSITORY = 'frontend'
+DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
 
-node {
+node('mongodb-2.4') {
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -1,60 +1,99 @@
...
-    recipients: 'govuk-ci-notifications@digital.cabinet-office.gov.uk',
-    sendToIndividuals: true])
+          notifyEveryUnstableBuild: true,
+          recipients: 'govuk-ci-notifications@digital.cabinet-office.gov.uk',
+          sendToIndividuals: true])
     throw e
   }
-
-}
+}
```
gpeng: Needs a newline

mgrassotti: 👍 I fear I need to do that on every single other repo...

PR #1099, opened 2017-01-16
Add example page for A/B testing

suzannehamilton:
Adds a new help page which acts as an example of an A/B test and provides a way to manually and automatically check that A/B testing is working. The main concern is that A/B testing cookies are being correctly converted to custom headers by the CDN.

This simple example gives us a way to check A/B testing before we create a real (and more complicated) A/B test of the navigation.

alphagov/smokey#222 adds regression tests for this page.

There are a couple of FIXMEs around the wording on the page which I need to update before this can be merged.

https://trello.com/c/23lOrO6h/308-create-a-test-page-for-a-b-testing

### DIFF
--- a/app/views/help/ab_testing.html.erb
+++ b/app/views/help/ab_testing.html.erb
@@ -1,0 +1,23 @@
+<% content_for :title, "A/B testing - GOV.UK" %>
+<% content_for :extra_headers do %>
+  <meta name="robots" content="noindex">
+<% end %>
+
+<main id="content" role="main" class="group">
+  <header class="page-header group">
+    <div>
+      <h1>This is a test page</h1>
+    </div>
+  </header>
+
+  <div>
+    <p>We sometimes test different versions of pages with different users to see which work better.
+      This page shows us if the process is working.</p>
+    <p>There are 2 versions of this page. You are viewing the
+      <span class="ab-example-group">
+        <strong><%= @ab_variant %></strong>
+      </span>
+      version.</p>
+    <p>Visit the <a href="/">GOV.UK homepage</a> to find government services and information.</p>
+  </div>
+</main>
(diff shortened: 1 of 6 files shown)


app/views/help/ab_testing.html.erb
```diff
@@ -0,0 +1,25 @@
+<% content_for :title, "A/B testing - GOV.UK" %>
+<% content_for :extra_headers do %>
+  <meta name="description"
+  <% # FIXME: Does this page need specific wording? %>
```
tijmenb: Can we add a `noindex` tag so that Google won't pick this up? Perhaps the description can be left off then.

suzannehamilton:
@tijmenb - can I just generate a new UUID to use as a content ID, or does it need to be generated by a different govuk system first?

app/views/help/ab_testing.html.erb
```diff
@@ -0,0 +1,25 @@
...
+<main id="content" role="main" class="group">
+  <header class="page-header group">
+    <div>
+      <h1>A/B testing on GOV.UK</h1>
+    </div>
+  </header>
+
+  <div>
+    <% # FIXME: Discuss this with a content designer %>
+    <p>Some text explaining what A/B testing is.</p>
```
tijmenb: ?
> This page is being used internally by GOV.UK to make sure A/B testing works.

suzannehamilton: Updated.

suzannehamilton:
I've added a `noindex` header and publishing API route.

lib/tasks/publishing_api.rake
```diff
@@ -39,6 +39,12 @@ namespace :publishing_api do
           description: "Find out about GOV.UK, including the use of cookies, accessibility of the site, the privacy policy and terms and conditions of use.",
         },
         {
+          content_id: "a4d4e755-3d75-4b19-b120-a638a6d79ba8",
```
suzannehamilton: Generated using `SecureRandom::uuid` as discussed.

suzannehamilton:
I've renamed the header to `GOVUK-ABTest-Example` to match alphagov/govuk-cdn-config#15.

suzannehamilton:
I've updated the text on the page based on a discussion with a Content Designer.

So unless anyone has any more review comments, this is ready to go once the Varnish changes have been deployed.

PR #1105, opened 2017-01-24
Revert "Move to new ci"

tijmenb:
Reverts alphagov/frontend#1098.

`ci/reporter` is required on production, where the gem isn't available. This is preventing deploys:

https://deploy.staging.publishing.service.gov.uk/job/Deploy_App/2804/console


tijmenb:
Argh, reverting here won't work because it would mess up CI, obviously.

https://ci.integration.publishing.service.gov.uk/job/frontend/job/revert-1098-move-to-new-ci/1/console

PR #1108, opened 2017-01-26
Prevent duplicate caching of non-AB-tested pages

suzannehamilton:
Move the `Vary: GOVUK-ABTest-Example` header from every response to just the A/B testing page, since this is the only page with variants that should be cached separately.

This will save Varnish and the CDN from having to cache _every_ frontend page twice.

As far as I know, this is isn't currently causing a problem, but it's such a small change that it seemed worth just doing. It might be more helpful in the future when there are more A/B tests going on, since each `Vary` header on a page multiplies the cache size for that page.

### DIFF
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -20,6 +20,8 @@ class HelpController < ApplicationController
 
   def ab_testing
     @ab_variant = request.headers["HTTP_GOVUK_ABTEST_EXAMPLE"] == "B" ? "B" : "A"
+
+    response.headers['Vary'] = 'GOVUK-ABTest-Example'
   end
 
 private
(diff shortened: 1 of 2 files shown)


app/controllers/help_controller.rb
```diff
@@ -20,6 +20,8 @@ def show
 
   def ab_testing
     @ab_variant = request.headers["HTTP_GOVUK_ABTEST_EXAMPLE"] == "B" ? "B" : "A"
+
+    response.headers['Vary'] = 'GOVUK-ABTest-Example'
```
suzannehamilton: It seems a bit dangerous to _set_ the `Vary` header rather than add it to existing headers, but it currently doesn't overwrite anything, and it's only affecting the A/B test page. The `Vary:Accept-Encoding` header is still being added correctly by Rails.

Rails 5 has an `add` method which lets you set multi-valued headers more safely.

PR #1109, opened 2017-01-27
Show taxon breadcrumbs

alecgibson:
Show taxon breadcrumbs when present, and when a user is in the correct A/B testing bucket. This is hidden behind a feature flag.

Trello: https://trello.com/c/DrBm0dU3/334-changes-to-education-related-content-pages-as-part-of-the-new-navigation-in-frontend


app/controllers/application_controller.rb
```diff
@@ -86,4 +89,13 @@ def set_publication
...
   end
+
+  def breadcrumbs
+    return {} if @navigation_helpers.nil?
+    if should_present_new_navigation_view?
+      @navigation_helpers.taxon_breadcrumbs
+    else
+      @navigation_helpers.breadcrumbs
+    end
+  end
```
alecgibson: @carvil What do you think about moving this logic into the Navigation Helpers itself?

carvil: That sounds like a good idea. However, that gem would have to know about the A/B testing API. Maybe @tijmenb has an idea on this?

alecgibson: Mm. I think in the longer term possibly we should move these sorts of tests into an A/B Testing Gem, that eg Navigation Helpers can rely on. Probably for a different PR though.

tijmenb:
Marking DNM until we have the client library.

app/presenters/education_navigation_ab_testable.rb
```diff
@@ -0,0 +1,17 @@
+module EducationNavigationABTestable
```
mobaig: An alternative implementation would be to make this a class that you instantiate with a content item and a request object. You'd then create an instance of it in ApplicationController#breadcrumbs:

```
ab_testing_policy = EducationNavigationABTestingPolicy.new(@content_item, request)
...
return @navigation_helpers.taxon_breadcrumbs if ab_testing_policy.should_present_new_navigation_view?
return @navigation_helpers.breadcrumbs
...
```

Given none of this logic is doing any actual presenting, a better home for it might be `app/models`.

Benefits: 
* The resulting class is simpler to unit test.
* Its dependencies are obvious within the first few lines of its implementation.
* You can mark the latter three methods private, hinting they aren't used outside the class.
* Renaming and relocating the file makes its purpose a little clearer/easier to remember without needing to read the implementation.

Better still would be to extract out the entire notion of a NavigationPresenter from the controller layer and letting it handle breadcrumbs and related links, but that's probably a bit much for one PR.

alecgibson: Yes I've thought about this approach. I decided against it mainly because of verbosity. Any controllers using this class would have to:
* Instantiate the class
* Attach `helper_method` to make methods available on view
* Attach `after_filter` to mutate the `response` object

I think my ideal API (as discussed with @tijmenb) would be to have a method similar to things like `helper_method` etc., and declare at the top of the controller: `ab_test :educationnavigation`.

However, this is all pretty temporary code, so I don't think we should worry about it too much. Let me know if you think differently.

alecgibson: @mobaig as discussed, I've moved the module into the `models` folder, but will probably leave the rest as-is, given that the code will be removed again after the A/B test is over.

mobaig: @alecgibson cool - although if you're keeping the module implementation a better place still would be `app/controllers/concerns`. It's quite similar to other modules in there already.

Otherwise fine with this, given we're nuking it all after the AB test's finished.

alecgibson: Done

carvil:
This has been successfully tested in integration.

PR #1111, opened 2017-01-30
Use govuk-components for help page related links

MatMoore:
In static we removed the CSS for the old sidebar, but the help page index was stilling using it:

https://www.gov.uk/help

It should have looked like this:

http://webarchive.nationalarchives.gov.uk/20151109153250/https://www.gov.uk/help

Using the govuk-component fixes this.

Before:
screen shot 2017-01-30 at 13 55 16

After:
screen shot 2017-01-30 at 13 52 06

Trello: https://trello.com/c/QFUjiqHs/449-fix-related-links-for-help-pages
Paired with @klssmith @Davidslv 

We've left out the "return to top" link as it wasn't there beforehand, and it doesn't seem to serve any purpose.

### DIFF
--- a/app/views/help/index.html.erb
+++ b/app/views/help/index.html.erb
@@ -56,25 +56,23 @@
 
 </main>
 
-<!-- related -->
-<div class="related-positioning">
-  <div class="related-container">
-    <div class="related" id="related">
-      <div class="inner group">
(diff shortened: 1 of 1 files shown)


tijmenb:
👍

PR #1113, opened 2017-02-01
Fix a bug in the routing of Help Page JSON requests

whoojemaflip:
The segment constraint `slug: %r{help/.+}` was matching on the entire
path including the .json format identifier. This caused an issue where
JSON requests were being interpreted as if for HTML content and would
404 as the slug would not be available in the content-api.


davidbasalla:
We'll look into retiring the JSON API soon, but it's good to have this fixed until we retire it properly

PR #1117, opened 2017-02-06
Getting Help Pages to rendered

whoojemaflip:
**In 2 parts:** 
 - First, we allow the app to call the content-store to find out if the requested document has been migrated. 
 - Then, we've modified the HelpPageController so that it can render documents from the content-store.

A notable addition is the ContentFormatInspector class which is responsible for finding the
content format for a requested slug (and handling API errors that might come up). This is
used by the FormatRoutingConstraint in the router to help determine which
controller to route the request to. It memoizes the result of any necessary API calls against the `request` object, so that we don't multiply the number of backend call required.

Since we're now in the middle of migrating mainstream content to
content-store there is a fair amount of complexity being introduced. Since
content might need to be pulled from either the content-api or the content-
store, the Frontend app needs to be able to decide which to call. It's hoped
that it will be relatively easy to reduce this complexity once all of the
mainstream content has been migrated.

https://trello.com/c/1zlDYRH8/625-migrate-help-page-format-to-rendered-1-3-3


lib/services.rb
```diff
@@ -1,7 +1,12 @@
 require 'gds_api/rummager'
+require 'gds_api/content_store'
 
 module Services
   def self.rummager
     @rummager ||= GdsApi::Rummager.new(Plek.new.find('search'))
   end
+
+  def self.content_store
+    @content_store ||= GdsApi::ContentStore.new("http://content-store.dev.gov.uk")
```
boffbowsh: Should use Plek here

whoojemaflip: Good spot! Was working round a now-fixed issue in a dev environment.

app/presenters/content_item_presenter.rb
```diff
@@ -0,0 +1,48 @@
...
+    date = @content_item["updated_at"]
+    DateTime.parse(date).in_time_zone if date
+  end
+
+  def format
+    @content_item['schema_name']
+  end
+
+  def short_description
+    nil
```
davidbasalla: I never got around to checking what `short_description` referred to with artefacts...

app/presenters/content_item_presenter.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  PASS_THROUGH_DETAILS_KEYS.each do |key|
+    define_method key do
+      details[key.to_s] if details
+    end
+  end
+
+  def in_beta
+    @content_item['phase'] == 'beta'
```
davidbasalla: We could add something to the integration test for testing beta content? After a quick glance I don't think we have anything that tests this for help pages (although we do for the other formats)

test/unit/api_error_routing_constraint_test.rb
```diff
@@ -0,0 +1,24 @@
+require 'test_helper'
+require 'api_error_routing_constraint'
+
+class ApiErrorRoutingConstraintTest < ActiveSupport::TestCase
+  should "return true if there's a cached error" do
+    request = stub(env: { __api_error: StandardError.new })
+    assert subject.matches?(request)
+  end
+
+  should "return false if there's no error in content API" do
```
davidbasalla: content API/ content store?

app/presenters/content_item_presenter.rb
davidbasalla: Ah, looks like we're just using description now instead of short_description, so seems fine to leave as is -> https://github.com/alphagov/frontend/blob/master/app/views/shared/_publication_metadata.html.erb#L15-L21

davidbasalla:
Tested it in integration and it renders gloriously, so all good apart from some minor comments!

app/presenters/content_item_presenter.rb
whoojemaflip: For help pages this is only used if there is no description, [to populate the meta-description tag](https://github.com/alphagov/frontend/blob/master/app/views/shared/_publication_metadata.html.erb#L15)

app/presenters/content_item_presenter.rb
whoojemaflip: Added a unit test for the `ContentItemPresenter` which tests this function.

whoojemaflip:
@boffbowsh anything further to add?

boffbowsh:
If someone who's had a proper review of this can approve, I'm happy with the changes I requested. @davidbasalla

davidbasalla:
Great, merging! 👍

PR #1120, opened 2017-02-09
Fall back to English when content is in an unsupported locale

suzannehamilton:
The content schema locales include some locales which are not supported by Rails by default, such as `tk` and `zh`. This triggers an exception in ApplicationController when it attempts to set the application locale from the content locale.

I think this is an edge case of #1117, which fixed the way the application locale was extracted from the content.

This was causing flaky tests when a random content item was generated. For example, help_controller_test would fail about 20% of the time when a content item with an unsupported language was generated. Example seed: 47089.

This fix checks whether the content's locale is in the list locales supported by the Rails application, and falls back to English if not.

A better long-term fix might be to align the locales in govuk-content-schemas with the available locales in GOVUK Rails applications. I'm just proposing this as a short-term fix to prevent errors on any real pages in these languages, and also fix the flaky test.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -83,7 +83,11 @@ protected
   end
 
   def set_language_from_publication
-    I18n.locale = @publication.locale || I18n.default_locale
+    I18n.locale = if @publication.locale && I18n.available_locales.map(&:to_s).include?(@publication.locale)
+                    @publication.locale
+                  else
+                    I18n.default_locale
+                  end
   end
 
   def content_item
--- a/test/functional/transaction_controller_test.rb
+++ b/test/functional/transaction_controller_test.rb
@@ -105,5 +105,23 @@ class TransactionControllerTest < ActionController::TestCase
         get :show, slug: "chwilio-am-swydd"
       end
     end
+
+    context "given a version in an unsupported language exists" do
+      setup do
+        @details = {
+          'id' => 'https://www.gov.uk/api/document-in-turkmen.json',
+          'web_url' => 'https://www.preview.alphagov.co.uk/document-in-turkmen',
+          'format' => 'transaction',
+          'details' => { "need_to_know" => "", "language" => "tk" },
+          'title' => 'Some title'
+        }
+        content_api_and_content_store_have_page("document-in-turkmen", @details)
+      end
+
+      should "set the locale to the English default" do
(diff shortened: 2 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -83,7 +83,11 @@ def set_publication
   end
 
   def set_language_from_publication
-    I18n.locale = @publication.locale || I18n.default_locale
+    I18n.locale = if @publication.locale && I18n.available_locales.map(&:to_s).include?(@publication.locale)
```
whoojemaflip: Might I suggest setting the Rails `I18n.available_locales` to the list of locales supported by the schema instead of adding this complexity to the controller?

The list is
```
[
  "ar", "az", "be", "bg", "bn", "cs", "cy", "de", "dr", "el", "en", "es", 
  "es-419", "et", "fa", "fr", "he", "hi", "hu", "hy", "id", "it", "ja", "ka", 
  "ko", "lt", "lv", "ms", "pl", "ps", "pt", "ro", "ru", "si", "sk", "so", "sq", 
  "sr", "sw", "ta", "th", "tk", "tr", "uk", "ur", "uz", "vi", "zh", "zh-hk", 
  "zh-tw"
]
```

suzannehamilton:
In case it's useful for anyone: the list of content locales is in [definitions.json](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/definitions.json) in govuk-content-schemas.

Of those, the locales not supported by Rails by default are: dr, hy, ka, ps, si, so, tk, zh, zh-hk, zh-tw.

tijmenb:
That's really odd, because I got the list of locales in definitions.json from checking Rails! 

https://github.com/alphagov/govuk-content-schemas/pull/364

Perhaps it would be better to update that list?

suzannehamilton:
@tijmenb Looks like government-frontend configures its own list of locales and translations, whereas most apps use the Rails defaults: https://github.com/alphagov/government-frontend/tree/master/config/locales

tijmenb:
Ah, that's of course where `available_locales` comes from.. Can we force/override `available_locales` somehow? Would be great to have one canonical list of languages that GOV.UK supports.

h-lame:
Maybe we all already know this but the reason for the list being different to rails defaults is that it's the list of languages available in whitehall, mostly for embassies and such to provide content in the languages widespread in their locality.

PR #1121, opened 2017-02-09
Frontend can now render migrated Answers from the content-store

whoojemaflip:
We removed the ability to render draft editions in `private-frontend` for Answers and HelpPage content. Now we have a working draft-stack, this seems redundant.


davidbasalla:
Looks great! :+1: We should only deploy this once we have frontend working on the draft stack. We'll also have to look into the issues with govuk_content_schemas...

PR #1124, opened 2017-02-10
Render migrated local transaction

davidbasalla:
Switch frontend to render local transactions from the content store (instead of content api).

Should only be merged once https://github.com/alphagov/publisher/pull/560 has been merged and all LocalTransaction content has been republished


davidbasalla:
We'll have a go at setting the `PLEK_SERVICE_#{service}_URI`s instead of that last commit...

PR #1130, opened 2017-02-16
Render Guide format content

whoojemaflip:
Bumped the schemas gem in order to give us the Example for testing.

Deleted `data-protection.json` fixture as it was not being used anymore.

Trello: https://trello.com/c/r2gA1lxN/640-migrate-guide-format-to-rendered


app/controllers/guide_controller.rb
```diff
@@ -1,11 +1,11 @@
 class GuideController < ApplicationController
   include ApiRedirectable
-  include Previewable
   include Cacheable
```
davidbasalla: Quick thought re including the `Cacheable` module. The main thing it does is `set_expiry` unless there's an edition param present. That's not really relevant for the migrated formats anymore. I guess we could leave it in the controllers for now (since it doesn't do any harm) and just make sure we remove it once we've migrated all formats and we're not sending user from Publisher to private-frontend anymore.

app/controllers/guide_controller.rb
```diff
@@ -1,11 +1,11 @@
...
   include Navigable
   include EducationNavigationABTestable
 
-  before_filter :set_publication
+  before_filter :set_content_item
+  before_filter -> { setup_content_item_and_navigation_helpers("/" + params[:slug]) }
```
davidbasalla: I think this is already being done by the `Navigable` module

app/controllers/guide_controller.rb
whoojemaflip: to follow on from that: you think we should try and replicate that functionality in draft-frontend? i.e. not set a future expiry + maybe set no-cache

lib/content_format_inspector.rb
```diff
@@ -3,7 +3,7 @@ class ContentFormatInspector
 
   attr_reader :error
 
-  MIGRATED_SCHEMAS = %w(answer help_page local_transaction).freeze
+  MIGRATED_SCHEMAS = %w(answer help_page guide local_transaction).freeze
```
davidbasalla: 🔺 pedantry alert 🔺  

'guide' should be before 'help_page' if we're sorting alphabetically 😁

whoojemaflip: `a..b..c..d..e..f..g..h` omg you're right - don't know how we ballsed that up

app/presenters/content_item_with_parts_presenter.rb
```diff
@@ -0,0 +1,83 @@
+class ContentItemWithPartsPresenter < ContentItemPresenter
```
davidbasalla: For the other formats we opted to name them after the specific format (ie AnswerPresenter, LocalTransactionPresenter etc). We thought that would be simpler to separate the formats in future (should anyone want to pull out any format into another app). So we could rename this accordingly. Also the other parted format Programmes will go away, so we won't benefit from reusing this anyway.

test/functional/guide_controller_test.rb
```diff
@@ -10,139 +10,88 @@ def prevent_implicit_rendering
...
   context "GET show" do
-    setup do
-      @artefact = artefact_for_slug('data-protection')
-      @artefact["format"] = "guide"
-    end
-
     context "for live content" do
       setup do
-        content_api_and_content_store_have_page('data-protection', @artefact)
+        @content_item = content_store_has_example_item('/foo', schema: 'guide')
```
davidbasalla: This is cool!

app/controllers/guide_controller.rb
whoojemaflip: I'll save that change for a future PR

davidbasalla:
Looks great!

PR #1133, opened 2017-02-22
Set token in default url options for fact check content

brenetic:
For content that is in fact check a JWT token is generated and the authentication proxy will pass that through with the request to the frontend.

This token needs to be carried through on the get and post requests for those formats that are a little more interactive.

The authentication proxy no longer strips the token itself as a result of [this PR](https://github.com/alphagov/authenticating-proxy/pull/21)

https://trello.com/c/RoHFOwid/646-enable-postcode-search-for-local-transactions-in-factcheck

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -121,4 +121,19 @@ protected
     end
   end
   helper_method :present_new_navigation?
+
+private
+
+  def default_url_options
+    {}.merge(token)
+      .merge(cache)
+  end
+
+  def token
+    params[:token] ? { token: params[:token] } : {}
+  end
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -121,4 +121,18 @@ def present_new_navigation?
     end
   end
   helper_method :present_new_navigation?
+
+private
+
+  def default_url_options
+    token.merge(cache)
```
whoojemaflip: this is a little confusing at first glance. How would you feel about
```
{}
  .merge(token)
  .merge(cache)
```

or similar?

PR #1134, opened 2017-02-22
Render simple smart answer format

davidbasalla:
Update controller to read content from content store, add specialised presenter for the schema. We remove any tests to do with draft, as we don't use editions anymore to determine if a draft item is being viewed. Also there is no cacheing happening on the draft stack so we can remove the code that conditionally set the expiry headers. We also remove the fixtures for 'bridge of death' and prefer to move it into the test itself.

For https://trello.com/c/2ucbq949/643-migrate-simple-smart-answer-format-to-rendered

PS: Hot tip from @issyl0, use 'w=1' as query param to ignore white space changes, it'll make the diff more pleasant to look at 🎉  -> https://github.com/alphagov/frontend/pull/1134/files?w=1


davidbasalla:
We checked with @boffbowsh about the snyk failure and it's fine to ignore for this PR!

sihugh:
Yep

PR #1136, opened 2017-02-23
Add taxonomy sidebar to pages displaying the new navigation

mobaig:
Use the taxonomy sidebar component in static and rework the tests to
assert for the presence of the sidebar and breadcrumbs where
appropriate.

https://trello.com/c/0BD3POzu/403-show-parent-taxons-in-the-side-navigation-on-all-education-content-pages


Gemfile.lock
```diff
@@ -348,4 +348,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.13.6
+   1.14.5
```
carvil: I'm not sure if this matters, but perhaps it's best to not include it in the changes.

mobaig: Doing so encourages future committers to keep their local bundler up to date, which seems reasonable. Assuming we don't just merge changes to a lower version in further PRs, that is. WDYT?

carvil: Sounds good. I wasn't sure if we had to have specific versions of bundler in CI etc, but it looks like it only gives a warning.

LGTM

PR #1139, opened 2017-02-27
Render completed transactions from content store

brenetic:
This format has now been migrated.

https://trello.com/c/unTNbCHi/658-migrate-completed-transaction-format-to-rendered

Mobbed with @emmabeynon @binaryberry

### DIFF
--- a/app/views/completed_transaction/show.html.erb
+++ b/app/views/completed_transaction/show.html.erb
@@ -8,34 +8,36 @@
   publication: @publication,
   edition: @edition,
 } do %>
-  <% if @publication.promotion_choice == 'organ_donor' %>
-    <div id="organ-donor-registration-promotion">
-      <p>Please join the NHS Organ Donor Register.</p>
-      <p>If you needed an organ transplant would you have one? If so please help others.</p>
-      <p>If you live in Wales and want to be an organ donor, you don’t need to do anything. Find out about your choices for <a href="https://www.organdonation.nhs.uk/supporting-my-decision/welsh-legislation-what-it-means-for-me/" rel="external">organ donation in Wales</a>.</p>
-      <p>
-        <%= link_to 'Join', @publication.promotion_url,
-              title: "Register to become an organ donor", rel: "external", class: "button", role: "button" %>
-      </p>
-    </div>
-  <% elsif @publication.promotion_choice == 'register_to_vote' %>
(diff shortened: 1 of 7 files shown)


test/integration/completed_transaction_test.rb
```diff
@@ -2,11 +2,32 @@
 require 'integration_test_helper'
 
 class CompletedTransactionTest < ActionDispatch::IntegrationTest
+  setup do
+    @payload = {
```
davidbasalla: We could remove some of the details here to make it easier to read, as I'm guessing we don't test for each one (eg analytics_identifier). Not sure what we've done previously for the other formats

davidbasalla:
We could clean up the PublicationPresenter a little bit now that this is done, as it has a number of promotion related methods. Looks great though! 👍

app/views/completed_transaction/show.html.erb
```diff
@@ -8,34 +8,36 @@
...
-              data: {
-                module: 'track-click',
-                track_category: 'linkTrack',
-                track_action: 'linkClicked',
-                track_label: @publication.promotion_url
-              } %>
-      </p>
-    </div>
+  <% if @publication.promotion %>
+    <% if @publication.promotion['category'] == 'organ_donor' %>
```
davidbasalla: What was your reasoning for not using the presenter methods here? I don't have any strong opinions in it, just curious

test/integration/completed_transaction_test.rb
emmabeynon: Thanks - we've removed the unnecessary stuff, and kept what's required ⚡️

app/views/completed_transaction/show.html.erb
emmabeynon: There aren't any methods available in the CompletedTransactionPresenter.  @binaryberry and I did discuss extracting this out in to methods, but felt there wasn't enough duplication going on to warrant it.

test/integration/completed_transaction_test.rb
davidbasalla: Great, could do a similar clean up in 'test/functional/completed_transaction_controller_test.rb' to be consistent?

emmabeynon: Should now be taken care of 👍

PR #1140, opened 2017-02-28
add live flag to be consistent with other repos

mcgoooo:
this enables the repo to be run off of the live api servers

this makes it consistent with
[government frontend](https://github.com/alphagov/government-frontend/blob/master/startup.sh)
[contacts frontend](https://github.com/alphagov/contacts-frontend/blob/master/startup.sh)
[specialist frontend](https://github.com/alphagov/specialist-frontend/blob/master/startup.sh)
and a few others i can't remember off the top of my head

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -1,4 +1,13 @@
 #!/bin/bash
-
 bundle install
-bundle exec rails s -p 3005
+
+if [[ $1 == "--live" ]] ; then
+  GOVUK_APP_DOMAIN=www.gov.uk \
+  GOVUK_WEBSITE_ROOT=https://www.gov.uk \
+  PLEK_SERVICE_CONTENT_STORE_URI=https://www.gov.uk/api \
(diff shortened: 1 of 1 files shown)


startup.sh
```diff
@@ -1,4 +1,14 @@
 #!/bin/bash
 
 bundle install
-bundle exec rails s -p 3005
+
+if [[ $1 == "--live" ]] ; then
+  GOVUK_APP_DOMAIN=www.gov.uk \
+  GOVUK_WEBSITE_ROOT=https://www.gov.uk \
+  PLEK_SERVICE_CONTENT_STORE_URI=https://www.gov.uk/api \
```
tijmenb: You perhaps also want the Rummager URL here, so the search results can be displayed.

mcgoooo: still haven't figured out what that should actually be?
i presume
search.publishing.service.gov.uk
or
rummager.publishing.service.gov.uk

but nether of them seem to exist

PR #1141, opened 2017-02-28
Get licence data from Licensify

davidbasalla:
Switch Frontend to talk to Licensify directly (as opposed to via the content-api since we'll be turning that off soon). Also allows us to remove a hack for county licences since `is_offered_by_county` now is returned from Licencify.

Also... REFACTOR EVERYTHING (LicenceLocationIdentifier, LicenceController, etc). The plan was to remove the `@interaction_details` hash which encapsulated the licence details from Licensify, but was difficult to work with due to its nested natures. Instead we chose to create a LicenceDetailsPresenter which handles the response from Licensify. 

We had to go through a number of steps to get to the final refactoring, so while reviewing the individual commits might be useful we seeing how we got there, the individual commits might not pass the tests 100%. We tested it in Integration and everything appears to work as before. Note that we have to make changes in Licensify to correct the data for licences that are offered by Counties.

For https://trello.com/c/H8HhjucN/659-update-frontend-to-get-licence-data-from-licensify

Paired with @issyl0 

---------------------------

For reference:

Published LicenceEditions on gov.uk: 466

With continuation link (means they link off to another site and don't talk to Licensify): 290

LicenceEditions that get info from Licensify: 176
Licences that can't be found in Licensify: 66
LocalAuthoritySpecific licences: 88
LocalAuthoritySpecific & offered by Counties (not Districts): 9
NonLocationSpecific & SingleAuthority: 22
NonLocationSpecific & MultipleAuthorities (even if the code supports it): 0


app/controllers/licence_controller.rb
```diff
@@ -96,17 +96,15 @@ def single_authority_interaction_details_present?
   end
 
   def location_specific_licence?
-    artefact['details']['licence']['location_specific']
+    licence_details[:licence]['location_specific']
```
whoojemaflip: I appreciate the playful interleaving of `:symbol` and `'string'` hash keys.

app/controllers/licence_controller.rb
```diff
@@ -136,8 +134,8 @@ def local_authority_slug
     return nil unless mapit_response.location_found?
 
     @_la_slug ||= begin
-      tier_override = :county_unitary if LICENCE_SLUGS_WITH_COUNTY_TIER_OVERRIDE.include?(params['slug'])
-      LicenceLocationIdentifier.find_slug(mapit_response.location.areas, artefact, tier_override)
+      # tier_override = :county_unitary if LICENCE_SLUGS_WITH_COUNTY_TIER_OVERRIDE.include?(params['slug'])
```
whoojemaflip: there's something awful about really long `SCREAMING_SNAKE_CASE` constant names.

davidbasalla: Luckily `LICENCE_SLUGS_WITH_COUNTY_TIER_OVERRIDE` goes away in the next commit I think

lib/licence_location_identifier.rb
```diff
@@ -1,30 +1,36 @@
-require 'location_identifier'
+class LicenceLocationIdentifier
+  AUTHORITY_TYPES = %w(DIS LBO UTA CTY LGD MTD COI).freeze
+  ALLOWED_REQUESTED_AREA_TYPES = %(CTY).freeze
+
+  def self.find_slug(areas, requested_area_type = "")
```
davidbasalla: Wondering whether it would be better to have `county_requested` here as the second parameter (instead of `requested_area_type`). Otherwise any class calling this one would have to know about MapIt style area descriptions (well only "CTY")...

lib/licence_details_from_licensify.rb
```diff
@@ -1,11 +1,11 @@
-class LicenceDetailsFromArtefact
+class LicenceDetailsFromLicensify
   attr_accessor :licence, :snac_code, :licence_authority_slug, :interaction
 
   def initialize(artefact, licence_authority_slug, snac_code, interaction = nil)
-    self.licence = artefact['details']['licence']
-    self.snac_code = snac_code
-    self.licence_authority_slug = licence_authority_slug
-    self.interaction = interaction
+    @artefact = artefact
```
whoojemaflip: Does this object need an `Artefact`?  Might it be simpler if instantiated with the `licence_identifier` instead?

lib/licence_location_identifier.rb
whoojemaflip: Looks like it's called with `@licence_details.offered_by_county?` as the second param. Is that a `boolean`?

app/controllers/licence_controller.rb
```diff
@@ -14,101 +14,62 @@ class LicenceController < ApplicationController
...
-    'permission-to-place-tables-and-chairs-on-the-pavement',
-    'pavement-or-street-display-licence',
-    'petroleum-storage-licence',
-    'weighbridge-operator-certificate',
-    'performing-animals-registration',
-    'approval-of-premises-for-civil-marriage-or-civil-partnership',
-    'licence-projection-over-highway-england-wales',
-  ].freeze
-
   def search
```
whoojemaflip: Not 100% related to your PR, but is this really a good name for the controller method?

lib/licence_location_identifier.rb
davidbasalla: 👍  Yup it should be, currently isnt :/

app/controllers/licence_controller.rb
davidbasalla: Quite probably not, as it's only a 'search' for when a licence is local authority specific! That's only true for ~20% of our published LicenceEditions

whoojemaflip: `search` and `authority` => `show` and `detail`?

app/controllers/licence_controller.rb
```diff
@@ -14,101 +14,64 @@ class LicenceController < ApplicationController
...
-  def artefact_with_snac
-    return nil if snac_from_slug.blank?
+  def set_publication
+    @publication = LicencePresenter.new(artefact)
+    @licence_details = LicenceDetailsPresenter.new(licence_details_from_api, params["authority_slug"], params[:interaction])
+    set_language_from_publication
+  end
+
+  def licence_details_from_api(snac = nil)
+    return {} if @publication.continuation_link.present?
```
whoojemaflip: It's a bit of a code smell that this check is repeated so many times in different places. It suggests that maybe we've missed something important. What does this represent?

app/controllers/licence_controller.rb
```diff
@@ -14,101 +14,64 @@ class LicenceController < ApplicationController
...
 
-    @_artefact_with_snac ||= ArtefactRetrieverFactory.artefact_retriever.fetch_artefact(
-      params[:slug],
-      params[:edition],
-      snac_from_slug
-    )
+    begin
+      Services.licensify.details_for_licence(artefact["details"]["licence_identifier"], snac)
+    rescue GdsApi::HTTPErrorResponse, GdsApi::TimedOutException
```
whoojemaflip: Is it still worth continuing with the response if we get an error result from Licensify?

app/controllers/licence_controller.rb
davidbasalla: Possibly... generally it represents the fact that a LicenceEdition will either have a link stored on it which links to another website. It does not need to talk to Licensify in that case, so the idea is to save API calls. To me it's part of the weirdness of the format that some info about the licence is stored in the LicenceEdition where as other info is stored in Licensify

app/controllers/licence_controller.rb
davidbasalla: Do you mean continuing by returning an empty hash? Licences pages show a message about the Licence not being available online if Licensify does not return a response

app/controllers/licence_controller.rb
davidbasalla: we've gone with `start` which sort of has to encapsulate the possible outcomes on this page:

app/controllers/licence_controller.rb
davidbasalla: We thought about adding a wrapper predicate method but it always just sounded very similar to 'continuation link' so gave up.

app/controllers/licence_controller.rb
whoojemaflip: Hmm, yeah so it does. The template has a big conditional in it, which suggests that maybe it could be two templates. Is there any way of straightening out this little tangle?

app/controllers/licence_controller.rb
```diff
@@ -14,101 +14,64 @@ class LicenceController < ApplicationController
...
-        @postcode = postcode
-        @location_error = location_error
-
-        if local_authority_slug
-          redirect_to licence_authority_path(slug: params[:slug], authority_slug: local_authority_slug)
-        end
-      end
-    elsif @publication.continuation_link.present?
+  def start
+    if @publication.continuation_link.present?
```
whoojemaflip: This method is super hard to follow. in part because it's actually two controller actions mashed together. 

Might it be easier to read if there's a single top-level conditional that differentiates between the `GET` and `POST` handlers? 

I know that this leads to some potential duplication, but IMO it's preferable to have duplicated code than complicated logic.

app/controllers/licence_controller.rb
issyl0: Done.

app/controllers/licence_controller.rb
issyl0: Done.

issyl0: It doesn't seem, from talking to @whoojemaflip, that this is easier to read or understand at all. The "multiple authorities" case is a GET, not a POST, despite the `authority_submitted?` method being named `submitted` (and so one assuming that it is a POST), so it has to be in the GET part. This differentiation doesn't help we don't think, it just makes it more confusing and there's duplication, so I'm going to delete this commit. Then we should be good to get on and migrate licences.

issyl0:
Tested on integration and it all looks good. :crossed_fingers:

PR #1144, opened 2017-03-01
Migrate Place

cbaines:
Migrate the Place format, such that it is rendered through the Publishing API.

Do not merge yet, as the merging/release of this needs to be coordinated with Publisher.

https://trello.com/c/t3DXuS2t/660-migrate-place-format-to-rendered

### DIFF
--- a/app/controllers/place_controller.rb
+++ b/app/controllers/place_controller.rb
@@ -14,6 +14,7 @@ class PlaceController < ApplicationController
   REPORT_CHILD_ABUSE_SLUG = "report-child-abuse-to-local-council".freeze
 
   def show
+    set_content_item(PlacePresenter)
     if request.post?
       @location_error = location_error
       if @location_error
--- a/app/controllers/place_controller.rb
+++ b/app/controllers/place_controller.rb
@@ -4,8 +4,6 @@ class PlaceController < ApplicationController
   include Cacheable
   include Navigable
 
-  before_filter :set_publication
-
   helper_method :postcode_provided?, :postcode
 
   INVALID_POSTCODE = "invalidPostcodeError".freeze
@@ -15,12 +13,13 @@ class PlaceController < ApplicationController
 
   def show
     set_content_item(PlacePresenter)
+
     if request.post?
       @location_error = location_error
       if @location_error
         @postcode = postcode
       elsif imminence_response.places_found?
-        @publication = PublicationWithPlacesPresenter.new(artefact, imminence_response.places)
+        @publication = PlacePresenter.new(content_item, imminence_response.places)
       end
     end
 
@@ -60,7 +59,7 @@ private
   def places_from_imminence
     if postcode.present?
       begin
-        places = Frontend.imminence_api.places_for_postcode(artefact["details"]["place_type"], postcode, Frontend::IMMINENCE_QUERY_LIMIT)
+        places = Frontend.imminence_api.places_for_postcode(@publication.place_type, postcode, Frontend::IMMINENCE_QUERY_LIMIT)
       rescue GdsApi::HTTPErrorResponse => e
         error = e
       end
--- a/app/presenters/place_presenter.rb
+++ b/app/presenters/place_presenter.rb
@@ -1,0 +1,14 @@
+class PlacePresenter < ContentItemPresenter
+  PASS_THROUGH_DETAILS_KEYS = %i(
+    introduction
+    more_information
+    need_to_know
+    place_type
+  ).freeze
+
+  PASS_THROUGH_DETAILS_KEYS.each do |key|
+    define_method key do
+      details[key.to_s] if details
+    end
+  end
+end
--- a/app/presenters/place_presenter.rb
+++ b/app/presenters/place_presenter.rb
@@ -11,4 +11,24 @@ class PlacePresenter < ContentItemPresenter
       details[key.to_s] if details
     end
   end
+
+  attr_reader :places
+
+  def initialize(content_item, places = [])
+    @content_item = content_item
+    @places = format_places(places)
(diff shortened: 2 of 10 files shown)


whoojemaflip:
The `PlaceController` is still making a call to `content-api` [here](https://github.com/alphagov/frontend/blob/f0d6a9c16ae3147307b6692733ebe82dd42c67a6/app/controllers/place_controller.rb#L7)

whoojemaflip:
The `PublicationWithPlacesPresenter` class is still accepting an `artefact` [here](https://github.com/alphagov/frontend/blob/f0d6a9c16ae3147307b6692733ebe82dd42c67a6/app/controllers/place_controller.rb#L23)

whoojemaflip:
Tests?

app/presenters/place_presenter.rb
```diff
@@ -11,4 +11,19 @@ class PlacePresenter < ContentItemPresenter
       details[key.to_s] if details
     end
   end
+
+  def initialize(content_item, places = nil)
+    @content_item = content_item
+    @places = places if places
```
whoojemaflip: there's no need for the `if` guard here - it makes very little difference to the program logic if an instance variable is initialised to `nil`

app/presenters/place_presenter.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def places
+    if @places
+      @places.map do |place|
+        place['text']    = place['url'].truncate(36) if place['url']
```
whoojemaflip: any idea why this truncates to 36 chars?

cbaines:
The tests for frontend are failing at the moment, I've proposed https://github.com/alphagov/frontend/pull/1168 to fix this.

app/presenters/place_presenter.rb
cbaines: I believe its a style thing, e.g. take a look at the result in the following screenshot.

truncated-url

app/controllers/place_controller.rb
```diff
@@ -14,12 +12,15 @@ class PlaceController < ApplicationController
...
+    set_content_item(PlacePresenter)
+
     if request.post?
       @location_error = location_error
       if @location_error
         @postcode = postcode
       elsif imminence_response.places_found?
-        @publication = PublicationWithPlacesPresenter.new(artefact, imminence_response.places)
+        @publication = PlacePresenter.new(content_item, imminence_response.places)
+        set_language_from_publication
```
davidbasalla: I think `set_language_from_publication` should be happening in all cases. Its effect is fairly minimal I believe in that it only translates the 'Last updated:' field at the bottom of the start page

cbaines: The code is confusing, as it actually is. `set_content_item` called at the top of this method calls `set_language_for_publication`. I believe its called again here as `set_language_from_publication` uses the value of `@publication` that has just been changed.

It could be moved below the end of the if, which might read better, but it would be called twice sometimes.

davidbasalla: Oh right! But do we need this second call to set the language at all then? Presumably the content item it's referring to is still the same so the language wouldn't have changed?

cbaines: That makes sense, I've removed it in 1283ebcab2d20bbca4acb3f589f3375f497d1dc2

davidbasalla: Nice, could probably just squash it into the commit where you made the change originally, unless you feel it requires its own commit? (sorry was in a meeting, looking through the rest of it now :) )

app/presenters/place_presenter.rb
```diff
@@ -0,0 +1,29 @@
...
+    place_type
+  ).freeze
+
+  PASS_THROUGH_DETAILS_KEYS.each do |key|
+    define_method key do
+      details[key.to_s] if details
+    end
+  end
+
+  def initialize(content_item, places = nil)
```
davidbasalla: Could places default to an empty array instead? Might save a couple of lines of code...

davidbasalla:
Maybe you've done it already but it would be worth testing https://www.gov.uk/report-child-abuse-to-local-council, since it's a bit of a special case (as far as I remember we use more of the fields for displaying the results on this one)

davidbasalla:
Only a couple of minor comments, looks great!

app/controllers/place_controller.rb
cbaines: Ok, done :)

app/presenters/place_presenter.rb
cbaines: Yep, @emmabeynon and I have changed it to use an empty array, and also moved the code around a bit in the presenter.

cbaines:
> Maybe you've done it already but it would be worth testing https://www.gov.uk/report-child-abuse-to-local-council, since it's a bit of a special case (as far as I remember we use more of the fields for displaying the results on this one)

Checked on integration, looks fine as far as I can see.

cbaines:
(the truncation of the URL has now changed to be in CSS, but this could do with more testing on integration)

app/views/place/show.html.erb
```diff
@@ -26,7 +26,7 @@
              data-track-category="<%= track_category_for_place_results(@publication.places) %>"
              data-track-action="<%= track_action_for_place_results(@publication.places) %>"
              data-track-label="<%= track_label_for_place_results(@publication.places) %>">
-      <% if @publication.places.any? %>
+      <% if @publication.places %>
```
davidbasalla: Should this be reverted to `.any?` now that we're using an array?

cbaines: Yeah, makes sense.

PR #1148, opened 2017-03-02
Only set up AB Test where appropriate

alecgibson:
Prior to this change, we were showing AB Test `meta` tags, and setting `Vary` headers on requests that were to pages other than those included in the AB test.

This change adds tests to check that any pages not involved in the AB test do not receive the same setup.

### Trello

https://trello.com/c/Xehe42eM/464-only-add-a-b-meta-tag-to-pages-in-the-navigation-a-b-test

### DIFF
--- a/app/controllers/concerns/education_navigation_ab_testable.rb
+++ b/app/controllers/concerns/education_navigation_ab_testable.rb
@@ -2,7 +2,7 @@ module EducationNavigationABTestable
   EDUCATION_NAVIGATION_DIMENSION = 41
 
   def self.included(base)
-    base.helper_method :education_navigation_variant
+    base.helper_method :education_navigation_variant, :ab_test_applies?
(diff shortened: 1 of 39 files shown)


tijmenb:
Is there a way to put `education_navigation_variant.analytics_meta_tag.html_safe` in the layout?

alecgibson:
@tijmenb I think that would involve moving the `include EducationNavigationABTestable` up from individual controllers into the `ApplicationController` (because you need access to `ab_test_applies?` You'd lose the nice explicitness you get by declaring a controller as being A/B tested, but you'd then be able to move the meta tag into the layout.

Either that or you'd have to do something like check if `ab_test_applies?` has been defined _and_ it's `true`, but that feels a little dirty to me.

Happy for you to make a call either way.

tijmenb:
@alecgibson thanks, this is indeed better than the alternative then.

PR #1151, opened 2017-03-03
Link from the homepage to the Education taxon in the B variant

suzannehamilton:
In the new navigation, which is the B variant of the education navigation A/B test, link directly to the Education taxon rather than the Education browse page, and use the title and description of the taxon.

The layout of the homepage has been changed slightly to accommodate the longer taxon description: 'Housing and local services' has been moved from the second column into the third. This matches the prototype that has been tested in user research.

The variants are cached separately using the `Vary` header, but unlike on other A/B-tested pages, the `meta` tag is not added. This is because we do not want to include *all* journeys through the homepage in the A/B test analytics; we are only concerned with user journeys which include education content.

This requires a release of govuk_ab_testing before the tests will pass.

@carvil - I can't remove the feature flag yet because it relies on the shared `EducationNavigationAbTestHelper`.

Trello: https://trello.com/c/FOUYzmzX/477-change-homepage-topic-and-description-for-education

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,7 @@ gem "therubyracer", "~> 0.12.0"
 gem 'uglifier'
 gem 'uk_postcode', '~> 2.1.0'
 gem 'unicorn', '~> 4.9.0' # version 5 is available
-gem 'govuk_ab_testing', '1.0.0'
+gem 'govuk_ab_testing', '1.0.3'
 
 group :development do
   gem 'better_errors'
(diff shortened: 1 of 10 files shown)


carvil:
@suzannehamilton ok, that makes sense.

Could you please add a screenshot of how it looks now?

suzannehamilton:
### A

screen shot 2017-03-03 at 17 27 18

### B

screen shot 2017-03-03 at 17 25 52

PR #1157, opened 2017-03-07
Configure nav helpers with statsd

carvil:
This PR allows us to track search metrics for the taxonomy sidebar in statsd.

Trello: https://trello.com/c/S72dddmO/488-configure-navigation-helpers-with-statsd

### Counter

screen shot 2017-03-07 at 12 40 54

### Timer

screen shot 2017-03-07 at 12 47 07


tijmenb:
Would it be better to add statsd to a `Services` module? Then we don't have to worry about initialisation order.

carvil:
@tijmenb I've gone with your approach, good shout.

PR #1159, opened 2017-03-07
Show old related links for some mainstream pages

carvil:
Until we have the ability to create curated related links for the new
taxonomy, we need to still show the old sidebar because the links shown
in there are quite important.

This is a temporary fix that allows us to show the old sidebar on the
new navigation for a number of mainstream pages.

When `content-tagger` allows to do this, we should revert this commit.

Trello: https://trello.com/c/m3ziS0lh/503-show-old-curated-links-for-a-defined-list-of-content-before-new-curation-is-ready

### Page affected

screen shot 2017-03-07 at 17 25 10

### Page not affected

screen shot 2017-03-07 at 17 25 19


config/mainstream_content.json
```diff
@@ -0,0 +1,24 @@
+[
```
alecgibson: Is this _all_ mainstream content? Or just stuff meant to have the old sidebar? If it's the latter, could we please have a more descriptive filename/folder structure to avoid confusion (eg `mainstream_content_with_curated_sidebar.json` or similar). It might also be nice to reflect that in the `MainstreamContentFetcher` class, which sounds like it's fetching all mainstream content (which it isn't).

carvil: This isn't all mainstream content. I will rename the file

app/services/mainstream_content_fetcher.rb
```diff
@@ -0,0 +1,12 @@
+class MainstreamContentFetcher
```
alecgibson: Could you please also rename this class - it's actually fetching a subset of Mainstream Content.

carvil: As discussed IRL, I've renamed the method instead.

PR #1166, opened 2017-03-10
[DO NOT MERGE] Render travel advice index from expanded links.

danielroseman:
Travel-advice-publisher will soon stop sending the countries data
explicitly in the details hash; it will be expanded automatically by
publishing-api, which means they will be in the "children" element of
the details hash, and have a slightly different structure.

This should not be merged until the data is fully populated and publishing-api is expanding the links correctly.

### DIFF
--- a/app/presenters/travel_advice_index_presenter.rb
+++ b/app/presenters/travel_advice_index_presenter.rb
@@ -5,7 +5,7 @@ class TravelAdviceIndexPresenter
 
   def initialize(attributes)
     details = attributes.fetch("details")
-    country_data = details.fetch("countries")
+    country_data = attributes.dig("links", "children")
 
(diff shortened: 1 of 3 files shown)


cbaines:
Now that the schema changes have been merged and deployed, the frontend test suite is failing (except on this branch) as the test data from the schemas is not compatible with the previous presenter.

This should have been caught at the point the govuk-content-schemas change was in review, but the frontend tests simply were not run (fixed by https://github.com/alphagov/govuk-content-schemas/pull/564 and https://github.com/alphagov/frontend/pull/1167 ).

One way to get the test suite passing again would be to make the presenter and tests backwards compatible, I've done this here https://github.com/alphagov/frontend/pull/1168 . This might also be better for deployment, as it means that frontend will automatically switch when appropriate, rather than the deploys having to be carefully timed.

PR #1167, opened 2017-03-10
Pass env.SCHEMA_BRANCH to contentSchemaDependency

cbaines:
Otherwise, setting the environment value doesn't change anything.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,7 +36,6 @@ node('mongodb-2.4') {
   try {
     govuk.initializeParameters([
       'IS_SCHEMA_TEST': 'false',
-      'SCHEMA_BRANCH': DEFAULT_SCHEMA_BRANCH,
     ])
 
     if (!govuk.isAllowedBranchBuild(env.BRANCH_NAME)) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -60,7 +60,7 @@ node('mongodb-2.4') {
     }
 
     stage("Set up content schema dependency") {
-      govuk.contentSchemaDependency()
+      govuk.contentSchemaDependency(env.SCHEMA_BRANCH)
```
suzannehamilton: This looks good, but I discovered the other day that you can use `params.SCHEMA_BRANCH` instead. That means you can also delete the `initializeParameters` call on line 37.

cbaines: Ok, I've updated this branch accordingly. I didn't remove the call, but the inclusion of SCHEMA_BRANCH in 9d521bb , can the entire call be removed?

suzannehamilton: Yeah, it's safe to remove it entirely. The jenkinslib in govuk-puppet used to call `env.IS_SCHEMA_TEST` but it also uses `params` now.

cbaines: Great, I've removed it in  f85e4c8

cbaines:
The tests for frontend are failing at the moment, I've proposed https://github.com/alphagov/frontend/pull/1168 to fix this.

PR #1168, opened 2017-03-13
[DO NOT MERGE] Travel advice index links with backwards compatibility

cbaines:


cbaines:
This is an optional replacement for https://github.com/alphagov/frontend/pull/1166 with backwards compatibility. As this is backwards compatible, merging and deploying these changes won't break frontend with respect to the travel advice index, and will also make the test suite pass for other branches that are now failing due to the changes in the govuk-content-schemas.

danielroseman:
Thanks. I did originally try something like this, but it still needs us to be fairly careful of the sequence of events - as soon as https://github.com/alphagov/travel-advice-publisher/pull/175 is deployed, TAP will stop updating the countries hash - so there is the possibility that the links get updated but the countries don't, leading to frontend displaying the wrong data.

cbaines:
> Thanks. I did originally try something like this, but it still needs us to be fairly careful of the sequence of events - as soon as alphagov/travel-advice-publisher#175 is deployed, TAP will stop updating the countries hash - so there is the possibility that the links get updated but the countries don't, leading to frontend displaying the wrong data.

I'm not sure what you mean by "countries don't"?

I'm not familiar with this area, but looking at the publishing_api rake task in Travel Advice Publisher, it patches the links for the index page prior to publishing it. So I would hope that as the old data would be missing, and the new links published prior to the document being published, the content item would then hit the live content store without the old data, and with the correct new links?

PR #1171, opened 2017-03-13
Add support for displaying a banner on the GOV.UK homepage

boffbowsh:
This allows us to deploy a partial containing copy to be placed below the search bar on the GOV.UK homepage. The example at `app/views/homepage/_promo_banner_example.html.erb` should be copied to `app/views/homepage/_promo_banner.html.erb` and deployed. `app/views/homepage/_promo_banner.html.erb` should be left empty if the banner is not in use.

Example: 
screen shot 2017-03-13 at 16 58 15

### DIFF
--- a/app/views/homepage/_promo_banner_example.html.erb
+++ b/app/views/homepage/_promo_banner_example.html.erb
@@ -1,0 +1,6 @@
+<div id="homepage-promo-banner">
+  <div class="banner-message">
+    <p><strong>Some Title</strong> On some date something has happened.
+    <a href="https://something.gov.uk" rel="external nofollow">More&nbsp;information<span class="visuallyhidden"> about things</span></a></p>
+  </div>
+</div>
--- a/test/functional/homepage_controller_test.rb
+++ b/test/functional/homepage_controller_test.rb
@@ -11,5 +11,38 @@ class HomepageControllerTest < ActionController::TestCase
       get :index
       assert_equal "max-age=1800, public", response.headers["Cache-Control"]
     end
+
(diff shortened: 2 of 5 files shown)


app/views/homepage/_promo_banner_example.html.erb
```diff
@@ -0,0 +1,6 @@
+<div id="homepage-promo-banner">
+  <div class="banner-message">
+    <p><strong>Some Title</strong> On some date something has happened.
+    <a href="https://something.gov.uk" rel="external nofollow">More&nbsp;information</a></p>
```
fofr: This should include the `visuallyhidden` part that the global banner has so that extra context is added for accessibility

test/functional/homepage_controller_test.rb
```diff
@@ -11,5 +11,38 @@ class HomepageControllerTest < ActionController::TestCase
...
+        assert_template "homepage/_promo_banner"
+        assert @response.body.include?("homepage-promo-banner")
+        assert @response.body.include?("Some Title")
+      end
+    end
+  end
+
+  def stub_template(hash)
+    require 'action_view/testing/resolvers'
+    @controller.view_paths.unshift(ActionView::FixtureResolver.new(hash))
```
fofr: Nicely done 👍

app/views/homepage/_promo_banner_example.html.erb
boffbowsh: Done. Thanks for pointing that out, I'll make sure we have copy for the hidden part

PR #1173, opened 2017-03-15
Remove video format rendering

emmabeynon:
The video format is being retired so we remove all code that renders the format along with associated tests.

This PR is related to https://github.com/alphagov/publisher/pull/586

Trello card: https://trello.com/c/tTF5fAxc/684-retire-video-format-3

Paired with @cbaines


davidbasalla:
Great, so much red! :) We can remove 'video' from [artefact_retriever.rb](https://github.com/alphagov/frontend/blob/master/lib/artefact_retriever.rb#L13-L15) too I think

davidbasalla:
There's a reference to `video?` in application_test_helper.rb... not quite sure what's supposed to be doing, maybe we can remove/replace it?

emmabeynon:
@davidbasalla those references are now removed ⚡️

PR #1174, opened 2017-03-15
Send correct url to feedex

brenetic:
When this document type was migrated we incorrectly changed the view to use the slug of the publication for the field used for feedback forms.

This returns a slug of the form `done/a-completed-transaction` which is then munged for sending to feedback as `www.gov.ukdone/a-completed-transaction`. 

We now use the base_path which looks like `/done/a-completed-transaction` which should now send the correct url to the feedback app.

https://trello.com/c/8xALJ6OT/687-fix-done-page-feedex-problem

cc @h-lame

### DIFF
--- a/app/views/completed_transaction/show.html.erb
+++ b/app/views/completed_transaction/show.html.erb
@@ -47,7 +47,7 @@
 
       <form class="contact-form" action="/contact/govuk/service-feedback" method="post" id="completed-transaction-form">
         <input type="hidden" id="service_slug" name="service_feedback[slug]" value="<%= @publication.slug.gsub("done/", "") %>" />
-        <input type="hidden" id="service_done_page_url" name="service_feedback[url]" value="<%= @publication.slug %>" />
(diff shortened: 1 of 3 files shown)


app/views/completed_transaction/show.html.erb
```diff
@@ -47,7 +47,7 @@
 
       <form class="contact-form" action="/contact/govuk/service-feedback" method="post" id="completed-transaction-form">
         <input type="hidden" id="service_slug" name="service_feedback[slug]" value="<%= @publication.slug.gsub("done/", "") %>" />
-        <input type="hidden" id="service_done_page_url" name="service_feedback[url]" value="<%= @publication.slug %>" />
+        <input type="hidden" id="service_done_page_url" name="service_feedback[url]" value="<%= @publication.base_path %>" />
```
h-lame: Is it possible to send the full url with domain and protocol instead of just the base path?  This would avoid any future unpleasantness with munging things; particularly as it's not obvious why `slug` and `base_path` would be different, so the value for either of these inputs could be changed in the future.

brenetic: Coolio, have updated to send that

PR #1176, opened 2017-03-15
Remove support for countries hash in travel index.

danielroseman:
Now that travel-advice-publisher is no longer sending this data and the page has been republished, we can remove the backwards compatibility.


cbaines:
:+1: :)

PR #1180, opened 2017-03-16
Remove Frontend JSON API

emmabeynon:
In order to remove Frontend's JSON API, we have deleted the ApiRedirectable module and references to it.

https://trello.com/c/7IRxAVoJ/688-retire-frontend-s-json-api

Paired with @cbaines


davidbasalla:
There's [a line in the routes.rb file](https://github.com/alphagov/frontend/blob/master/config/routes.rb#L26) that looks like it could be removed too?

emmabeynon:
> There's a line in the routes.rb file that looks like it could be removed too?

Good spot - now removed 🎉

lib/core-ext/openstruct.rb
```diff
@@ -1,5 +0,0 @@
-class OpenStruct
```
whoojemaflip: this is freaking me out a little. Any idea what it was for?

app/presenters/publication_presenter.rb
```diff
@@ -48,10 +48,15 @@ def updated_at
     DateTime.parse(date) if date
   end
 
-  def to_json
-    {
-      places: places
-    }.to_json
+  def video_embed_url
```
whoojemaflip: not convinced that this belongs in this PR

lib/core-ext/openstruct.rb
emmabeynon: Looking at the commit history it was created for use in the root controller, which no longer exists.  We couldn't find any other use of it in the rest of the app.

app/presenters/publication_presenter.rb
emmabeynon: We can move this into a separate PR if you'd like.

emmabeynon: Sorry at first I thought you were referring to the `to_json` method, now I realise it's about the `video_embed_url` method.  This has been removed in another PR, which is why I think this looks like it's being added.  Now removed it again 👍

brenetic:
do we still need the `help.json`,  `licence-generic.json` and the `pay-bear-tax.json` fixtures?

emmabeynon:
@brenetic 
> do we still need the help.json, licence-generic.json and the pay-bear-tax.json fixtures?

I tried deleting `help.json` but it broke tests related to rendering the help page.  `pay-bear-tax.json` and `licence-generic.json` did not break tests however so those are now deleted 👍

PR #1190, opened 2017-03-22
Test special route publishing

suzannehamilton:
Add tests to check that all special routes are valid content items.

This was part of #1186, but it requires some changes to the content schemas and it would be good to have these tests in place _before_ making those changes.

https://trello.com/c/u1H1rgDA/525-some-finding-pages-are-still-being-tracked-as-thing

### DIFF
--- a/lib/special_route_publisher.rb
+++ b/lib/special_route_publisher.rb
@@ -1,0 +1,98 @@
+require 'gds_api/publishing_api/special_route_publisher'
+
+class SpecialRoutePublisher
+  def initialize(publisher_options)
+    @publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new(publisher_options)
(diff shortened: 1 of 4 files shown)


suzannehamilton:
@cbaines: I could split it if you like - it's basically two changes: refactoring to make it more testable, and then adding the test. I thought the refactoring didn't make much sense to commit on its own, but I can change it if you think it's clearer.

cbaines:
> @cbaines: I could split it if you like - it's basically two changes: refactoring to make it more testable, and then adding the test. I thought the refactoring didn't make much sense to commit on its own, but I can change it if you think it's clearer.

I prefer that style with more smaller commits, but I realise it is a personal preference. I'm quite happy for this to be merged regardless.

PR #1197, opened 2017-03-28
Fix English heading on Welsh Transaction pages

emmabeynon:
A bug was discovered whereby "More information" was displaying on Welsh
Transaction pages instead of the Welsh translation for "Before you start".
`/register-to-vote` was also displaying "More Information" instead of in the
house style lower case "information". This fixes those bugs.

https://trello.com/c/4zLU8Ptn/707-english-heading-on-welsh-start-pages

### DIFF
--- a/app/views/transaction/_additional_information_single.html.erb
+++ b/app/views/transaction/_additional_information_single.html.erb
@@ -1,6 +1,6 @@
 <% if transaction.more_information.present? %>
   <div id="before-you-start">
-    <h2><%= t 'formats.transaction.more_information' %></h2>
+    <h2><%= t 'formats.transaction.before_you_start' %></h2>
     <%= transaction.more_information.try(:html_safe) %>
   </div>
 <% end %>


binaryberry:
Would updating https://github.com/alphagov/frontend/blob/master/config/locales/cy.yml#L20 to "more_information" and its translation in Welsh be a better alternative?

emmabeynon:
> Would updating https://github.com/alphagov/frontend/blob/master/config/locales/cy.yml#L20 to "more_information" and its translation in Welsh be a better alternative?

Mark had a chat about this with the content team, and they're happy to keep it as it was before i.e. "Before you start".  However it's probably worth investigating this in a bit more detail to be sure it's the right choice, so I'm going to write up a separate card for the backlog so this can be looked at in future.

nickcolley:
This also fixes an issue on smaller viewports where some badly scoped CSS will override padding.

image

image

image

PR #1198, opened 2017-03-28
Use the central build script when building on CI

suzannehamilton:
This will speed up the builds because the central script improves job locking, and will let this project pick up future improvements.

The Jenkins build won't pass until the new parameters have been added to the jenkinslib: alphagov/govuk-puppet#5676.


Jenkinsfile
```diff
@@ -1,96 +1,21 @@
 #!/usr/bin/env groovy
 
-REPOSITORY = 'frontend'
-DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
-
 node('mongodb-2.4') {
```
tijmenb: I think this might be added here by mistake previously, frontend doesn't use mongo.

Jenkinsfile
```diff
@@ -1,96 +1,21 @@
...
+      stage("Test setup") {
+        govuk.setEnvar("RACK_ENV", "")
       }
-
-      stage("Deploy to integration") {
-        govuk.deployIntegration(REPOSITORY, env.BRANCH_NAME, 'release', 'deploy')
+    },
+    testTask: {
+      stage("Run tests") {
+        govuk.runRakeTask("ci:setup:rspec default")
```
tijmenb: This actually looks like it could run the normal `default` rake task. I think I confused this repo with `static`, where we can't run the test tasks as part of `default`:

https://github.com/alphagov/static/blob/master/Jenkinsfile#L65-L66

Tried it in https://github.com/alphagov/static/pull/954.

Jenkinsfile
```diff
@@ -1,96 +1,21 @@
...
-    if (env.BRANCH_NAME == "master") {
-      stage("Push release tag") {
-        govuk.pushTag(REPOSITORY, env.BRANCH_NAME, "release_${env.BUILD_NUMBER}")
+  govuk.buildProject(
+    sassLint: false,
+    beforeTest: {
+      // TODO: Make this the default in Jenkinslib and override in individual
+      // projects that need `env.RACK_ENV`
+      stage("Test setup") {
+        govuk.setEnvar("RACK_ENV", "")
```
tijmenb: What's this needed for?

Jenkinsfile
suzannehamilton: Deleted. 👍

Jenkinsfile
suzannehamilton: Great, I've deleted the `testTask` parameter and it seems to do the same thing.

I did originally try it and the tests failed, but that turned out to be caused by the `RACK_ENV` issue and I didn't revisit it until now.

Jenkinsfile
suzannehamilton: I did a bit of digging, and it turns out that there was some invalid HTML in one of the frontend templates. Setting `RACK_ENV` to `test` (which is what the shared script does) causes slimmer to run stricter HTML validation, which caused tests to fail.

I'd assumed that if the frontend Jenkins file didn't set `RACK_ENV` then it wasn't necessary, but that was a bad assumption and I think we should leave the `RACK_ENV` line in the shared Jenkinsfile.

I've added a commit to fix the HTML. 👍

suzannehamilton:
Jenkins has picked up the govuk-puppet changes, so the build is now passing. 🎉

PR #1208, opened 2017-04-11
Use redis to determine promo banner display state

bevanloon:
## Context
The emergency banner implementation is currently being re-written such that instead of using a fabric script to add content to some partials in Frontend and Static, we will populate a hash in Redis with values needed for the emergency banner.

This will mean that should we need to deploy an emergency banner, Static and Frontend deploys are no longer blocked.

This PR depends on c8db3a2ba86acc71753dd0b4f7a79275104abb28

## This PR
In doing this work, we have also identified that the actual display of the emergency banner can be consolidated into Frontend (see the [Trello card](https://trello.com/c/3eq0AeHA/30-consolidate-emergency-banner-templates-in-static-and-frontend)). However, we are no consolidating the display of the promo banner at this point, so we still need to selectively display the promo banner on the homepage if there is no emergency banner to display.

The promo banner must only display if no emergency banner is enabled. We are querying Redis to discover this. This PR exposes the fields we need from Redis and ensures that the promo banner will only display if the emergency banner is not set in Redis.

* Add the `redis` gem.

* Create a minimal EmergencyBanner class which we can use to determine if we need to display an emergency banner by querying the required fields (`heading` and `campaign_class`)

* Pass an extra boolean variable to the `index.html.erb` from `homepage_controller` to indicate the state of the emergency banner (enabled or disabled).

* Add a condition in `index.html.erb` to selectively display the promo banner based on this new local variable.

* Amend a couple of tests for the promo banner.

## Future work
Future work will remove the current implementation of the emergency banner code which relies on empty partials that are populated by a fabric task.

### DIFF
--- a/test/functional/homepage_controller_test.rb
+++ b/test/functional/homepage_controller_test.rb
@@ -27,7 +27,7 @@ class HomepageControllerTest < ActionController::TestCase
     end
 
     context "with a promo banner partial" do
-      should "render promo banner" do
+      setup do
         stub_template "homepage/_promo_banner.html.erb" => <<-EOF
         <div id="homepage-promo-banner">
           <div class="banner-message">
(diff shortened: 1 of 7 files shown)


test/functional/homepage_controller_test.rb
```diff
@@ -36,11 +36,23 @@ class HomepageControllerTest < ActionController::TestCase
...
+        should "not render the promo banner" do
+          EmergencyBanner.any_instance.stubs(:enabled?).returns(true)
+          get :index
+          refute @response.body.include?("homepage-promo-banner")
+        end
+      end
+
+      context "with an emergency banner not enabled" do
+        should "render promo banner" do
+          get :index
```
leenagupte: Shouldn't the EmergencyBanner be explicitly set to `false`? Similar to the set up of the previous test?

PR #1209, opened 2017-04-12
Use either lgil_code or lgil override

sihugh:
We're in the process of making lgil_code mandatory and retiring lgil_override.
In the meantime, we'd like to be able to use either.

https://trello.com/c/RnJC1d7a/5-assign-lgil-codes-to-all-local-transactions


sihugh:
Yep, I'd deliberately left a test in to use the old one.  I'll update that when we remove `lgil_override` in due course.

PR #1220, opened 2017-05-05
[DO NOT MERGE] Add Search Match Length AB test

brenetic:
We are about to start testing changes to match length for search results. This needs to be tracked by GA as well.

- [ ] Merge after https://github.com/alphagov/govuk-cdn-config/pull/51
- [x] Merge after https://github.com/alphagov/govuk_ab_testing/pull/37
- [x] Merge after https://github.com/alphagov/govuk_ab_testing/pull/38
- [x] Merge after https://github.com/alphagov/rummager/pull/775

https://trello.com/c/43eBQfwf/80-prepare-the-infrastructure-for-the-minimum-should-match-a-b-test

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,8 +1,12 @@
 class SearchController < ApplicationController
   before_filter :set_expiry
   before_filter :remove_search_box
+  helper_method :search_match_length_variant
+  after_filter :set_search_match_length_response_header
 
   rescue_from GdsApi::BaseError, with: :error_503
+
+  SEARCH_MATCH_LENGTH_DIMENSION = 42
 
   def index
     search_params = SearchParameters.new(params)
@@ -12,7 +16,8 @@ class SearchController < ApplicationController
     if search_params.no_search? && params[:format] != "json"
       render action: 'no_search_term' and return
     end
-    search_response = SearchAPI.new(search_params).search
+    variant = search_match_length_variant.variant_name
+    search_response = SearchAPI.new(search_params, SearchMatchLength: variant).search
 
     @search_term = search_params.search_term
 
@@ -31,6 +36,22 @@ class SearchController < ApplicationController
       format.html { render locals: { full_width: true } }
       format.json { render json: @results }
     end
+  end
+
+  def search_match_length_ab_test
+    GovukAbTesting::AbTest.new(
+      "SearchMatchLength",
+      dimension: SEARCH_MATCH_LENGTH_DIMENSION
+    )
+  end
+
+  def search_match_length_variant
+    @_search_match_length_variant ||=
+      search_match_length_ab_test.requested_variant(request.headers)
+  end
+
+  def set_search_match_length_response_header
+    search_match_length_variant.configure_response(response)
   end
 
 protected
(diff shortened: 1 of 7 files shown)


test/functional/search_controller_test.rb
```diff
@@ -317,6 +317,19 @@ def no_results
     assert_equal "1",       @response.headers["X-Slimmer-Result-Count"]
   end
 
+  test "should pass the search variant through when one is present" do
+    Services.rummager.expects(:search).with(has_entry(:match_length_variant, "B")).returns(response([]))
+    SearchAPI.unstub(:new)
+    @request.headers["HTTP_GOVUK_ABTEST_SEARCHMATCHLENGTH"] = "B"
```
suzannehamilton: You can use the `with_variant` test helper instead of adding this header directly. It also confirms that the Vary header and meta tag have been added correctly.

app/controllers/concerns/search_match_length_ab_testable.rb
```diff
@@ -0,0 +1,25 @@
+module SearchMatchLengthABTestable
```
suzannehamilton: Thinking out loud here: I wonder if we should remove this module and move the logic into the controller? Unlike the navigation A/B test, the search A/B test only affects a single controller. And we might end up adding multiple A/B tests, which will probably push us to extract this code in a different way.

app/controllers/search_controller.rb
```diff
@@ -12,7 +12,8 @@ def index
     if search_params.no_search? && params[:format] != "json"
       render action: 'no_search_term' and return
     end
-    search_response = SearchAPI.new(search_params).search
+    variant = search_match_length_variant.variant_name
+    search_response = SearchAPI.new(search_params, match_length_variant: variant).search
```
suzannehamilton: I wonder whether it would be worth making this parameter name match the A/B test name? It would give a simple mapping between the A/B cookie, header and API parameter. It would probably be easier to understand when debugging.

MatMoore: Assuming this param gets passed to Rummager, we should make sure rummager supports it before merging this PR, as it will 400 otherwise.

test/functional/search_controller_test.rb
```diff
@@ -317,6 +319,18 @@ def no_results
     assert_equal "1",       @response.headers["X-Slimmer-Result-Count"]
   end
 
+  test "should pass the search variant through when one is present" do
+    with_variant SearchMatchLength: "B" do
+      get :index, q: "Variant B"
+    end
```
dwhenry: This is not clear when the test is or that there even is one. 

I assume a method named `with_variant` would not be responsible for setting a test condition.

brenetic: `with_variant` checks whether the test name and the variant are in the response header. Should we assert something additional to that?

dwhenry: I would assume `with_variant` was only responsible for putting the variant into the request header. Maybe renaming it to something like `with_variant_in_response` would make it clearer that it actually adds an expectation.

Happy either way.

dwhenry:
This has been tested on Integration with https://github.com/alphagov/rummager/pull/775 and correct passed the ab_test parameter as expected

PR #1227, opened 2017-05-11
Remove references to the Content API

alecgibson:
The [`content_api`](https://github.com/alphagov/govuk_content_api) has
been retired, and all content has been migrated to the
[`content_store`](https://github.com/alphagov/content-store).

Therefore, we can remove all references to the Content API, and rely
solely on the Content Store.

### Trello

https://trello.com/c/2BFMkQvB/124-stop-using-the-content-api-in-the-frontend-app


test/fixtures/child-tax-credit.json
```diff
@@ -4,7 +4,7 @@
     },
     "title": "Get Child Tax Credit",
```
tijmenb: This file looks like a content-api artefact, so we should be able to remove it too.

fofr:
What were Publications in the context of frontend? I only recall us moving Whitehall Publications to government-frontend.

davidbasalla:
@fofr I think `PublicationPresenter` in the context of Frontend was the base presenter which was used for all formats before Custom Team gave each format a dedicated presenter. So I think removing it should be fine, nothing should be using it anymore... The wording in the PR about Publications being rendered by government-frontend probably isn't 100% correct, the term Publication in Frontend has been a bit misleading...

alecgibson:
@davidbasalla I didn't know that history thanks - I'll update the commit message.

alecgibson:
@tijmenb I've removed that and other unused text fixtures.

alecgibson:
@tijmenb / @fofr / @davidbasalla are you happy with this change?

README.md
```diff
@@ -49,12 +47,10 @@ To run the application standalone, run
...
 
 ```
-PLEK_SERVICE_STATIC_URI=http://127.0.0.1:3013 PLEK_SERVICE_CONTENTAPI_URI=https://www.gov.uk/api ./startup.sh
+PLEK_SERVICE_STATIC_URI=http://127.0.0.1:3013 ./startup.sh
 ```
 
-which uses the production
-[content API](https://github.com/alphagov/govuk_content_api) and a local copy of
-static.
+which uses the production a local copy of static.
```
fofr: Copy here doesn't read correctly

davidbasalla:
I'll have a look through!

davidbasalla:
You can probably remove `require 'gds_api/content_api'` from:
https://github.com/alphagov/frontend/blob/master/app/controllers/application_controller.rb#L2
https://github.com/alphagov/frontend/blob/master/config/initializers/gds_api.rb#L2

davidbasalla:
https://github.com/alphagov/frontend/blob/master/test/unit/helpers/application_helper_test.rb#L37 still refers to `artefact`. Artefacts were the old world concept for Documents within this app, so could be updated to 'ContentItem' probably

davidbasalla:
Looks like you can remove `def error_404; error 404; end` in the ApplicationController as well, that's not being used anymore

davidbasalla:
Looks good! Only a few minor comments ^

alecgibson:
@fofr / @davidbasalla I've addressed your comments - could you please re-review?

PR #1230, opened 2017-05-15
Don't index search result pages

MatMoore:
Search result pages are excluded from robots.txt, but search engines can still
index the page.

This means it's possible to see results with the query prefilled with spam from
other websites.

screen shot 2017-05-15 at 16 22 55

A meta tag should do a better job at preventing indexing. See https://support.google.com/webmasters/answer/93710?hl=en

Trello: https://trello.com/c/AMj9IOuV/92-don-t-let-search-engines-index-site-search-results


fofr:
Will this block the search page when there is no search query? Is that desirable?
Should this tag be included only if there is a query?

MatMoore:
@fofr when there's no query a different template is rendered, which doesn't have the meta tag.

PR #1233, opened 2017-05-17
Add sri

h-lame:
For: https://trello.com/c/b3FnLalV/148-enable-subresource-integrity-sri-on-frontend-s

This only adds SRI attributes to the assets delivered as part of frontend itself; JS and CSS delivered by templates included from static require SRI to be added by static itself (see: https://github.com/alphagov/static/pull/1008 for work in this area).

Local testing can be achieved by setting `config.assets.debug = false` in `config/development.rb` as sprockets-rails doesn't do integrity calculations in dev mode otherwise.


h-lame:
Flagging as DNM while we investigate an issue with older versions of firefox (see https://github.com/alphagov/government-frontend#368))

h-lame:
See https://github.com/alphagov/govuk_template/pull/308 and https://github.com/alphagov/govuk_template/pull/301 for discussion on impact.  We're holding off on SRI until there are fewer users using the version of Firefox that has an issue with calculating SRI.

h-lame:
Removed DNM and rebased, this is ready to go now as we've solved the issue with SRI on firefox (see https://trello.com/c/UcJJBH0k/183-investigate-bom-issue-with-asset-compilation) and are slowly rolling SRI out across our apps.

PR #1236, opened 2017-05-22
Switch Content API link with Content Store

carvil:
Pages have a `rel=alternate` link linking to the JSON version of the
page. That link was pointing to the Content API, but should point to the
Content Store instead. This commit makes sure we link to the Content
Store version of the page.

Trello: https://trello.com/c/Uo9tfmSE/139-investigate-who-is-using-travel-advice-endpoints-on-content-api-in-production

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,9 +25,7 @@ module ApplicationHelper
     html_classes.join(' ')
   end
 
-  def publication_api_path(publication, opts = {})
-    path = "/api/#{publication.slug}.json"
-    path += "?edition=#{opts[:edition]}" if opts[:edition]
-    path
+  def publication_api_path(publication)
+    "/api/content/#{publication.slug}"
   end
 end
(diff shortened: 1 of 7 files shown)


fofr:
Is the link necessary at all?

carvil:
I'm not sure. If there aren't any real cases for this I'm happy to delete it. Do you know of anything that would use this link @fofr ?

davidbasalla:
I vaguely remember looking at this on Custom Formats team. I think we said it could be removed, since we don't officially support these APIs for the public. See https://trello.com/c/7IRxAVoJ/688-retire-frontend-s-json-api and https://github.com/alphagov/frontend/pull/1180, they seem related...

fofr:
@carvil We omitted for Help, Answers and Guides when we switched to government-frontend. We couldn't identify a need for those formats. I also agree with @davidbasalla about the APIs not having  public support.

carvil:
@fofr / @davidbasalla thanks for your comments. I've opened a new PR where I remove this functionality. Could you please have a look and let me know if that makes sense? https://github.com/alphagov/frontend/pull/1237

PR #1245, opened 2017-06-07
Update to startup script

andysellick:
* Changed to follow structure of same script in government-frontend


fofr:
Is this the pattern that allows the vars to be overridden?

nickcolley:
Yeah see: https://github.com/alphagov/government-frontend/commit/465915c856c8062eee8aa4e69b06f20aea339c5f

PR #1249, opened 2017-06-13
A/B test the DVLA start button

issyl0:
- Given we're testing two things on the same page, we made the
  `DvlaTitleABTestable` module into a generic `DvlaABTestable`, so we
  didn't have to create `DvlaButtonABTestable` too.
- There are two meta tags on the page, because we're running two tests
  at the same time on the same page (`/contact-the-dvla`). We set
  `assert_meta_tag: false` in the tests, because the `govuk_ab_testing`
  gem asserts that there's only one meta tag per page. It's a known
  limitation.

https://trello.com/c/R6kWqXF5/150-dvla-button-develop-and-deploy-a-b-test


app/controllers/application_controller.rb
```diff
@@ -6,7 +6,7 @@ class ApplicationController < ActionController::Base
   include Slimmer::Headers
   include Slimmer::Template
   include Slimmer::GovukComponents
-  include BenchmarkingContactDvlaTitleABTestable
+  include BenchmarkingContactDvlaABTestable
```
issyl0: Yes, because without things in here, the tests were failing all over the place, because we check for `is_benchmarking_variant?` in partials, and if that is just included in SimpleSmartAnswersController, then that won't work.

For example:

```
471 runs, 711 assertions, 0 failures, 149 errors, 0 skips
[...]
149) Error:
TransactionControllerTest#test_: loading the jobsearch page should set correct expiry headers. :
ActionView::Template::Error: undefined method `should_show_benchmarking_variant?' for #<#<Class:0x007fec9894b100>:0x007fec9cb19668>
    app/views/shared/_base_page.html.erb:6:in `_app_views_shared__base_page_html_erb__1737781897475647101_70327080888160'```

PR #1250, opened 2017-06-13
Remove DVLA title A/B test and make DVLA button A/B test non-lab

issyl0:
- We're done with the A/B test for the DVLA title. We've also tested the DVLA button in the lab, and want to make it into a "proper" test.

This is a branch from a branch - #1249 - and so needs rebasing when that's merged. We don't want this merging until Thursday anyway when the lab is over with.

https://trello.com/c/R6kWqXF5/150-dvla-button-develop-and-deploy-a-b-test


sihugh:
Needs a rebase after yesterday's fix

PR #1253, opened 2017-06-16
Revert the title from the last AB test

sihugh:
Because we reused the AB variant name for the currently running test, this was still being
shown as part of the B variant.


issyl0:
Assuming you meant "test" as the last word.

PR #1257, opened 2017-06-26
Remove search format boosting A/B test

suzannehamilton:
The A/B test is successful and statistically significant. 🎉 

One commit also removes all the A/B test parameters from the search API since we're not planning to run another A/B test immediately. It should be easy to revert when we want to start a new one.

https://trello.com/c/Z5KURpnB/147-deploy-format-boosting-%F0%9F%9A%80

Paired with @brenetic.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,12 +1,8 @@
 class SearchController < ApplicationController
   before_filter :set_expiry
   before_filter :remove_search_box
-  helper_method :search_ab_test_variant
-  after_filter :set_search_ab_test_response_header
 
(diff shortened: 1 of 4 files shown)


app/controllers/search_controller.rb
```diff
@@ -38,22 +35,6 @@ def index
     end
   end
 
-  def search_ab_test
-    GovukAbTesting::AbTest.new(
-      "SearchFormatBoosting",
-      dimension: SEARCH_FORMAT_BOOSTING_DIMENSION
```
dwhenry: have we killed this constant?

suzannehamilton: Good spot. Now deleted.

PR #1270, opened 2017-08-11
Remove search sections for world locations

rubenarakelyan:
This commit removes the “section” links below world location results in search. These sections no longer exist in the new world location navigation pages.

screen shot 2017-08-11 at 16 08 05


davidbasalla:
Looks good 👍 Do we need to republish the WorldLocations for this to take effect?

rubenarakelyan:
Nope I think this happens straight away.

PR #1271, opened 2017-08-16
Add Childcare and parenting links to the AB-test

koetsier:
Add Childcare-and-parenting links to the AB-test


tijmenb:
This is changing more things than Childcare and parenting. Can you do that in a different commit/PR and/or update the PR description? Can you also add some screenshots?

cbaines:
homepage

This is what I see when I test locally, It's only the link that has been changed.

PR #1275, opened 2017-09-09
Add promo link to Hurricane Irma advice

MatMoore:
Before:
screen shot 2017-09-09 at 11 17 58

After:
screen shot 2017-09-09 at 11 17 49


fofr:
@MatMoore This is a 113kb image, I think that could be optimised.

MatMoore:
@fofr what's the best way to do that? Is there anyone else who can pick this up today?

I've just used the source jpg directly from FCO here, I haven't converted it in any way (this was raised out of hours).

fofr:
@MatMoore An online JPEG optimiser should do the trick

Here's a 64kb version at 80% quality:
ezgif-3-14ab42b260

PR #1278, opened 2017-09-18
Update Hurricane information to include Maria

nickcolley:
"
HI folks,

could the Hurricane Irma homepage promo please be updated as soon as possible so that it reads:

Title: Hurricanes Irma and Maria

Text: Advice for British nationals affected by Hurricanes Irma and Maria.

Because: FCO has updated the news article this links to similarly to include information on Maria. 

Many thanks

best

Gavan
"

https://govuk.zendesk.com/agent/tickets/2414424

### DIFF
--- a/app/views/homepage/index.html.erb
+++ b/app/views/homepage/index.html.erb
@@ -85,9 +85,9 @@
           <div class="departments-promo-sections">
             <div class="promo-image">
               <div class="promo-content">
-                <a href="/government/news/hurricane-irma-advice-for-british-nationals"><%= image_tag 'homepage/hurricane-irma-advice.jpg', alt: 'Advice for British nationals affected by Hurricane Irma' %></a>
-                <h3>Hurricane Irma</h3>
-                <p><a href="/government/news/hurricane-irma-advice-for-british-nationals">Advice for British nationals</a> affected by Hurricane Irma.</p>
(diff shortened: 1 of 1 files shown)


app/views/homepage/index.html.erb
```diff
@@ -85,9 +85,9 @@
           <div class="departments-promo-sections">
             <div class="promo-image">
               <div class="promo-content">
-                <a href="/government/news/hurricane-irma-advice-for-british-nationals"><%= image_tag 'homepage/hurricane-irma-advice.jpg', alt: 'Advice for British nationals affected by Hurricane Irma' %></a>
-                <h3>Hurricane Irma</h3>
-                <p><a href="/government/news/hurricane-irma-advice-for-british-nationals">Advice for British nationals</a> affected by Hurricane Irma.</p>
+                <a href="/government/news/hurricane-irma-advice-for-british-nationals"><%= image_tag 'homepage/hurricane-irma-advice.jpg', alt: 'Advice for British nationals affected by Hurricanes Irma and Maria' %></a>
```
fofr: Image alt tag refers to Maria, image doesn't show Maria. This seems like an incorrect use of the alt attribute.

nickcolley: I think we'd need to source a new image, given the time scales around this, seems more important to get this shipped.

In general this pattern isn't very good, this issue comes up often and should be addressed separately.

PR #1280, opened 2017-09-20
Use Sentry for error reporting

tijmenb:
Via https://docs.publishing.service.gov.uk/manual/upgrade-to-sentry.html

https://trello.com/c/vkjdBYmt/52-update-frontend


config/initializers/govuk_navigation_helpers.rb
```diff
@@ -1,4 +1,4 @@
 GovukNavigationHelpers.configure do |config|
-  config.error_handler = Airbrake
-  config.statsd = Services.statsd
+  config.error_handler = GovukError
+  config.statsd = GovukStatsd
```
h-lame: I think this might need to be `GovukStatsd.client` because `GovukStatsd` only exposes `increment` but `GovukNavigationHelpers` uses `time` as well (see: https://github.com/alphagov/govuk_navigation_helpers/blob/13e385b891e21aa1d5423cdea740f1407a8471c1/lib/govuk_navigation_helpers/rummager_taxonomy_sidebar_links.rb#L33)

PR #1287, opened 2017-10-16
Add Dockerfile for Frontend

surminus:
I want to test Frontend in a container. I nicked the Dockerfile from government-frontend.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM ruby:2.3.1
+RUN apt-get update -qq && apt-get upgrade -y
+
+RUN apt-get install -y build-essential nodejs && apt-get clean
+
+ENV GOVUK_APP_NAME frontend
+ENV GOVUK_ASSET_ROOT http://assets-origin.dev.gov.uk
+ENV PORT 3005
+ENV RAILS_ENV development
+
+ENV APP_HOME /app
+RUN mkdir $APP_HOME
+
+WORKDIR $APP_HOME
+ADD Gemfile* $APP_HOME/
+ADD .ruby-version $APP_HOME/
+RUN bundle install
+
+ADD . $APP_HOME
+
+CMD bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p $PORT -b '0.0.0.0'"


Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ruby:2.3.1
+RUN apt-get update -qq && apt-get upgrade -y
+
+RUN apt-get install -y build-essential nodejs && apt-get clean
```
nickcolley: What environment is this run in, what version of Node.js will be installed?

surminus: Whatever the base `ruby:2.3.1` image uses. I'll investigate.

surminus: From what I can gather the Ruby image is built from Debian Jessie, which should have the later versions of Node.js.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ENV PORT 3005
+ENV RAILS_ENV development
+
+ENV APP_HOME /app
+RUN mkdir $APP_HOME
+
+WORKDIR $APP_HOME
+ADD Gemfile* $APP_HOME/
+ADD .ruby-version $APP_HOME/
+RUN bundle install
```
tuzz: Is this so that a Docker snapshot is created after bundling so that the container need not be re-created if the gems have remained the same?

I was puzzled why these files are picked out specifically rather than `ADD . $APP_HOME` then bundling?

surminus: To be honest, I just pinched this from whoever did government-frontend. If it were me, I probably would just have done the below, but I assumed whoever built the government-frontend one did it a reason:

```
COPY . .
RUN bundle install
```

thomasleese: I'm pretty sure this was been done to support caching, so if the `Gemfile*` or `.ruby-version` hasn't changed, it won't need to run `bundle install` again.

surminus: Sure, makes sense.

surminus:
Is there anything else that needs to be set so that we can get a running container for Frontend? I was playing with getting all the frontend apps running in containers, although that's gone to the back burner for a bit now.

surminus:
Thanks!

PR #1289, opened 2017-10-25
Disable education B variant on mainstream pages

vanitabarrett:
This PR:

- Removes all of the Education AB test logic
- Shows original (A variant) layout on pages tagged to mainstream browse (and not worldwide taxonomy)
- Show taxonomy navigation (B variant) layout on pages tagged to worldwide taxonomy or any other taxonomy.

Decision tree for logic: https://gov-uk.atlassian.net/wiki/spaces/GFED/pages/187924634/When+we+show+what+on+the+content+pages

**Pages showing taxonomy navigation**
There are currently only 3 pages in Frontend which are not tagged to mainstream browse and should therefore show the new taxonomy navigation. These are:

- [/independent-school-registration](http://www.gov.uk/independent-school-registration)
- [/contact-ukvi-inside-outside-uk](http://www.gov.uk/contact-ukvi-inside-outside-uk)
- [/special-needs-school-approval-wales](http://www.gov.uk/special-needs-school-approval-wales)

screen shot 2017-11-02 at 10 21 01

Part of:
https://trello.com/c/ZQNGFeUa/65-remove-ab-test-from-frontend-and-government-frontend


app/controllers/help_controller.rb
```diff
@@ -10,12 +10,6 @@ def tour
     render locals: { full_width: true }
   end
 
-  def ab_testing
```
tijmenb: This page can stay, it's a permanent example of an A/B test used by the [smoke tests][] to verify that we've configured everything correctly:

https://www.gov.uk/help/ab-testing

More context in https://github.com/alphagov/frontend/pull/1099.

Probably also means we can keep the gem in the gemfile and keep some more tests.

[smoke tests]: https://github.com/alphagov/smokey/blob/master/features/ab_testing.feature

app/models/navigation_type.rb
```diff
@@ -0,0 +1,27 @@
+class NavigationType
+  def initialize(content_item)
+    @content_item = content_item
+  end
+
+  def should_present_taxonomy_navigation?
```
alecgibson: Possible alternative way of writing this (although it's obviously a matter of personal preference):

```ruby
def should_present_taxonomy_navigation?
  return true if tagged_to_world_wide_taxonomy?
  return false if content_is_tagged_to_browse_pages?
  
  content_is_tagged_to_a_taxon?
end
```

app/views/layouts/application.html.erb
```diff
@@ -25,25 +25,9 @@
...
-                </a>
-                MESSAGE
-              }
-            )
-          %>
-        <% end %>
         <%= render partial: 'govuk_component/breadcrumbs',
           locals: {
-            breadcrumbs: breadcrumbs[:breadcrumbs],
-            collapse_on_mobile: present_new_navigation?
```
alecgibson: Was a design decision made to actively remove this feature?

fofr: It's still in the component, and we are considering whether it is useful everywhere (if it is we'd make it the default).

But we may also redesign aspects of the breadcrumb as part of our multivariate test.

This change is consistent with what we've done on government-frontend.

fofr: This should still be collapsed on mobile when the taxonomy is shown. Good catch.

app/views/layouts/application.html.erb
```diff
@@ -25,25 +25,9 @@
   <body class="mainstream <%= yield :body_classes %>">
     <div id="wrapper" class="<%= wrapper_class(@publication || @presenter) %>">
       <% unless current_page?(root_path) %>
-        <% if present_new_navigation? %>
```
fofr: Spoke to Imeh, Mia and Alex about what we should do regarding the banner. Have concluded that we will remove this beta banner and survey link (as you have done here, and will do in a new PR in government-frontend), and use a new survey on our new tests.

sihugh:
This looks OK to me 👍

app/views/layouts/application.html.erb
```diff
@@ -25,10 +25,18 @@
   <body class="mainstream <%= yield :body_classes %>">
     <div id="wrapper" class="<%= wrapper_class(@publication || @presenter) %>">
       <% unless current_page?(root_path) %>
-        <%= render partial: 'govuk_component/breadcrumbs',
-          locals: {
-            breadcrumbs: breadcrumbs[:breadcrumbs]
-          } %>
+        <% if @navigation.should_present_taxonomy_navigation? %>
```
alecgibson: Is it weird that we show breadcrumbs differently between the two styles?

app/controllers/concerns/taxonomy_navigation.rb
```diff
@@ -1,23 +1,25 @@
 module TaxonomyNavigation
-  def should_present_taxonomy_navigation?
-    tagged_to_world_wide_taxonomy? ||
-      (!content_is_tagged_to_browse_pages? &&
-      content_is_tagged_to_a_taxon?)
+  def should_present_taxonomy_navigation?(content_item)
+    if content_item.present?
```
alecgibson: I find it a bit weird to nest a boolean logic chain inside an `if` statement. As I commented before, I personally prefer the line-by-line style of writing this, otherwise if you insist on using operators like `||` and `&&`, shouldn't you be consistent and lump this on as an `&&` condition?

app/views/layouts/application.html.erb
fofr: Yep. I want us to  look into this. Hoping that collapse on mobile can be the default too.

alecgibson: Sure, so long as this is the intended behaviour for now.

test/functional/taxonomy_navigation_test.rb
```diff
@@ -0,0 +1,89 @@
+require "test_helper"
+
+class FakeController < ApplicationController
```
maxgds: Would it make sense to call this FakeTaxonomyNavigationController? Someone else might need a fake controller in future for similar purposes.

PR #1290, opened 2017-10-26
Update invalid_utf8_rejector to v0.0.4

danielburnley:


Gemfile.lock
```diff
@@ -351,4 +351,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.15.1
+   1.15.4
```
tijmenb: We seem to be using `1.15.1` on the VMs, so maybe skip this change. I think we'll keep changing it back & forth otherwise?

https://github.com/alphagov/govuk-puppet/search?q=bundler_version

PR #1293, opened 2017-10-27
Remove site search

dwhenry:
This page has been moved to finder-frontend and is no longer required to be served from here.

https://trello.com/c/4CscCsGa/377-remove-old-site-search-code-form-frontend

must be run after: https://github.com/alphagov/rummager/pull/956


brenetic:
@dwhenry We should remove the special route registrations for `search` and `search.json`: https://github.com/alphagov/frontend/blob/remove-site-search/lib/special_route_publisher.rb#L59

We can update the rummager PR(https://github.com/alphagov/rummager/pull/956) with the correct path . . .

config/routes.rb
```diff
@@ -3,9 +3,6 @@
 Frontend::Application.routes.draw do
   get "/homepage" => redirect("/")
 
-  get "/search" => "search#index", as: :search
-  get "/search/opensearch" => "search#opensearch"
```
suzannehamilton: Has this endpoint been moved to finder-frontend? Just checking because I don't recognise it.

dwhenry: Yes, this is now being render from finder-frontend

PR #1295, opened 2017-10-31
Add task list sidebar with AB test

sihugh:
This is for the tasklist public beta.  The first theme is learning to drive a
car.  This commit adds a tasklist sidebar to content served by Frontend that
is listed in app/controllers/concerns/tasklist_ab_testable.rb.

The content in Frontend that is relevant is limited to transactions and
simple_smart_answers at present.

The tasklist is configured using the file at config/tasklists/learn-to-drive-a-car.json

Testing of the component itself is limited as testing components from static is
not currently possible (as noted in https://github.com/alphagov/slimmer#testing-components)

This limitation is subject to an RFC at present.  What we can do is test the
json that Slimmer emits in the test - which is what we've done.

We need to get a design review on this before merging.

Screenshots:
### Desktop
www-origin integration publishing service gov uk_change-theory-test
### iPhone 6
www-origin integration publishing service gov uk_change-theory-test iphone 6

### DIFF
--- a/app/assets/stylesheets/helpers/_task-list-title.scss
+++ b/app/assets/stylesheets/helpers/_task-list-title.scss
@@ -1,0 +1,18 @@
+.task-list-title {
+  @include bold-24;
+
+  margin-bottom: 0;
+  margin-top: 30px;
+
+  @include media(tablet) {
+    margin-top: 180px;
+  }
+
+  a {
+    text-decoration: none;
+
+    &:hover {
+      text-decoration: underline;
+    }
+  }
+}
--- a/app/controllers/concerns/tasklist_ab_testable.rb
+++ b/app/controllers/concerns/tasklist_ab_testable.rb
@@ -11,6 +11,7 @@ module TasklistABTestable
     /change-theory-test
     /check-driving-test
     /check-theory-test
+    /driving-eyesight-rules
     /driving-lessons-learning-to-drive/
     /driving-test/what-to-take
     /find-driving-schools-and-lessons
@@ -28,7 +29,6 @@ module TasklistABTestable
     /apply-for-your-full-driving-licence
     /automatic-driving-licence-to-manual
     /complain-about-a-driving-instructor
-    /driving-eyesight-rules
     /driving-licence-fees
     /driving-test-cost
     /dvlaforms
--- a/app/controllers/concerns/tasklist_ab_testable.rb
+++ b/app/controllers/concerns/tasklist_ab_testable.rb
@@ -1,0 +1,86 @@
+module TasklistABTestable
+  TASKLIST_DIMENSION = 66
+
+  TASKLIST_PRIMARY_PAGES = %w(
+    /apply-first-provisional-driving-licence
+    /book-driving-test
+    /book-theory-test
+    /cancel-driving-test
+    /cancel-theory-test
+    /change-driving-test
+    /change-theory-test
+    /check-driving-test
+    /check-theory-test
+    /driving-lessons-learning-to-drive/
+    /driving-test/what-to-take
+    /find-driving-schools-and-lessons
+    /government/publications/car-show-me-tell-me-vehicle-safety-questions
+    /guidance/the-highway-code
+    /legal-obligations-drivers-riders
+    /pass-plus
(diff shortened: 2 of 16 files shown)


app/controllers/concerns/tasklist_ab_testable.rb
```diff
@@ -0,0 +1,86 @@
...
+        dimension: TASKLIST_DIMENSION
+      )
+  end
+
+  def tasklist_ab_test_applies?
+    page_is_included_in_test?
+  end
+
+  def should_show_tasklist_sidebar?
+    tasklist_ab_test_applies? && tasklist_variant.variant_b?
```
Davidslv: does `variant_b?` still works? I though it was `variant?('B')`

sihugh: We're running an older version of the gem in Frontend.  I was weighing up whether to update it but figured we could wait until the Education AB test has been removed.

app/views/shared/_tasklist_sidebar.html.erb
```diff
@@ -0,0 +1,6 @@
+<aside class="govuk-tasklist-sidebar">
```
nickcolley: What is this class for?

app/assets/stylesheets/helpers/_task-list-title.scss
```diff
@@ -0,0 +1,18 @@
+.task-list-title {
+  @include bold-24;
+
+  margin-bottom: 0;
+  margin-top: 30px;
+
+  @include media(tablet) {
+    margin-top: 180px;
```
nickcolley: This looks like a magic number, is this robust against changing content?

app/views/shared/_tasklist_sidebar.html.erb
sihugh: I was using it to scope local styles, however now we're resetting within the component it's a little redundant.  We are using it within the tests to verify the behaviour of the ab test.

app/assets/stylesheets/helpers/_task-list-title.scss
andysellick: It's a magic number. I know. The idea is that the task list in the sidebar shouldn't be right at the top, because people tend to ignore it, but if it's a little bit lower, it works better. Unfortunately as you point out, every page has a different layout so it's hard to position this sensibly.

app/views/shared/_tasklist_sidebar.html.erb
nickcolley: I'd recommend changing this to something that indicates it's used by smokey to test GOV.UK.

This ensures developers don't use this class to add styling, and properly communicates why it exists.

We do something similar for binding to components with JavaScript '`.js-tasklist` 

for this we could do something like

`.smokey-tasklist-sidebar` or `.qs-tasklist-sidebar` etc

See https://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/ for more context

sihugh: That's nice, I'll do that, thanks.

app/assets/stylesheets/helpers/_task-list-title.scss
nickcolley: Should we be reorganising the grid so that we don't make this assumption?

andysellick: Wouldn't that involve restructuring a ton of page templates?

app/views/shared/_tasklist_sidebar.html.erb
```diff
@@ -0,0 +1,6 @@
+<aside class="qa-tasklist-sidebar">
+  <h2 class="task-list-title">
+    <a href="<%= tasklist[:base_path] %>"><%= tasklist[:title] %></a>
+  </h2>
+  <%= render "govuk_component/task_list", tasklist[:tasklist] %>
```
deborahchua: I think you missing the `small: true` attribute here which is used for the sidebar.

sihugh: It's in the json config (which is therefore slightly different from the collections one) Also, heading_level: 3

PR #1296, opened 2017-11-01
Fix Welsh start button translation bug

emmabeynon:
For [Trello](https://trello.com/c/uQuBPfYU/153-fix-welsh-translation-issues-for-start-now-button)

A bug has occurred where Welsh content displays “Start now” as start button
text instead of the Welsh translation “Dechrau nawr” (see Zendesk tickets [1](https://govuk.zendesk.com/agent/tickets/2430403) and [2](https://govuk.zendesk.com/agent/tickets/2498433)).

screen shot 2017-11-01 at 15 22 28

While investigating this issue we also found that publishers are able to select "Sign in" as start button text, but there was not a Welsh translation available, so this has been added in as well.

## Start now
screen shot 2017-11-01 at 15 21 30

## Sign in
screen shot 2017-11-01 at 15 20 36

### DIFF
--- a/test/integration/transaction_test.rb
+++ b/test/integration/transaction_test.rb
@@ -116,4 +116,35 @@ class TransactionTest < ActionDispatch::IntegrationTest
       assert page.has_no_selector?("#transaction_cross_domain_analytics", visible: :all)
     end
   end
+
+  context "locale is 'cy'" do
+    setup do
+      @payload = {
+        base_path: "/cymraeg",
+        content_id: "d6d6caaf-77db-47e1-8206-30cd4f3d0e3f",
+        document_type: "transaction",
+        locale: "cy",
+        publishing_app: "publisher",
+        rendering_app: "frontend",
+        schema_name: "transaction",
+        title: "Cymraeg",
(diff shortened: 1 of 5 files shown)


test/integration/transaction_test.rb
```diff
@@ -116,4 +116,35 @@ class TransactionTest < ActionDispatch::IntegrationTest
       assert page.has_no_selector?("#transaction_cross_domain_analytics", visible: :all)
     end
   end
+
+  context "locale is 'cy'" do
```
nickcolley: Do we need tests for both 'sign in' and 'start now'?

emmabeynon: I was wondering about that, then I thought it might be overkill as an integration test?  Couldn't figure out a way to test it as a unit test.

nickcolley: On government-frontend we do presenter level tests which would be more appropriate here I guess, not sure if this application has that concept though.

emmabeynon: I had a brainwave about presenter unit tests - now pushed 🚀

PR #1298, opened 2017-11-01
Remove open in new window feature for transactions

nickcolley:
Proposal to remove this so we don't need to add it as a feature when consuming the [button component](https://govuk-static.herokuapp.com/component-guide/button/start_now_button_with_info_text)

It's not clear why this was added, but it's wildly inconsistent since we don't seem to be updating it.

Be good to get any input in on if this is a good idea to remove.

Originally added in https://github.com/alphagov/frontend/pull/43

### DIFF
--- a/app/presenters/transaction_presenter.rb
+++ b/app/presenters/transaction_presenter.rb
@@ -16,26 +16,8 @@ class TransactionPresenter < ContentItemPresenter
     end
   end
 
-  NEW_WINDOW_TRANSACTIONS = %w(
-    apply-blue-badge
-    claim-state-pension-online
-    pension-credit-calculator
-    report-benefit-fraud
-    report-extremism
-    pay-court-fine-online
-    send-vat-return
-    use-construction-industry-scheme-online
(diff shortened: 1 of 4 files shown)


fofr:
Got a failing test:
```
1) Failure:
TransactionTest#test_: a transaction which should open in a new window should have the correct target attribute set on the 'Start now' link for a new window.
/test/integration/transaction_test.rb:84:
Expected false to be truthy.
```

fofr:
I approve of this change. It's weird when links that look the same do different things.

alextea:
From what I remember these were transactions on external sites – old directgov or local council transactions.

nickcolley:
Given that we do not add to this behaviour for new external sites, I think this makes sense to remove.

This'll mean the start now buttons have a consistent behaviour and avoid any potential accessibility issues associated with opening links in a new window.

keithiopia:
Like @alextea says - I've got a vague recollection of this being done to cater for legacy services, but I see no reason to keep it.

PR #1304, opened 2017-11-06
Tasklist header ab test

deborahchua:
Trello card: ttps://trello.com/c/tcB8YTxA/

The tasklist header logic has been set so that it replaces the breadcrumbs and beta label when it is displayed.

screen shot 2017-11-06 at 16 37 04

### DIFF
--- a/app/views/shared/_tasklist_header.html.erb
+++ b/app/views/shared/_tasklist_header.html.erb
@@ -1,0 +1,3 @@
+<%= render "govuk_component/task_list_header",
+  title: tasklist[:title],
+  path: tasklist[:base_path] %>
--- a/test/functional/simple_smart_answers_controller_test.rb
+++ b/test/functional/simple_smart_answers_controller_test.rb
@@ -41,6 +41,30 @@ class SimpleSmartAnswersControllerTest < ActionController::TestCase
       end
     end
 
+    context "tasklist header A/B testing" do
+      setup do
+        content_store_has_random_item(base_path: '/learn-to-drive-dangerously', schema: 'simple_smart_answer')
+
+        @controller.stubs(:tasklist_header_ab_test_applies?).returns(true)
+      end
+
+      should "not show the tasklist header by default" do
+        with_variant TaskListHeader: "A" do
+          get :show, slug: "learn-to-drive-dangerously"
+
+          assert_template partial: "_tasklist_header", count: 0
+        end
+      end
+
+      should "show the tasklist header for the 'B' version" do
+        with_variant TaskListHeader: "B" do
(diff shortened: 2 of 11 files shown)


app/views/shared/_tasklist_header.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render "govuk_component/task_list_header",
+  title: tasklist[:title],
+  path: tasklist[:base_path] %>
```
sihugh: Can we add a skip link to this please (but only if the task list sidebar is rendered)?  https://govuk-static.herokuapp.com/component-guide/task_list_header/with_a_skip_link_with_custom_text  

We'll need to add something to the task list sidebar so that we have a target for the skip link.  It might be worth chatting to Andy about what we should link to precisely as I think he spoke to the accessibility team about it.

test/functional/simple_smart_answers_controller_test.rb
```diff
@@ -41,6 +41,30 @@ def simple_smart_answer_content_item_with_start_button_text
...
+      should "not show the tasklist header by default" do
+        with_variant TaskListHeader: "A" do
+          get :show, slug: "learn-to-drive-dangerously"
+
+          assert_template partial: "_tasklist_header", count: 0
+        end
+      end
+
+      should "show the tasklist header for the 'B' version" do
+        with_variant TaskListHeader: "B" do
```
sihugh: Could we have a test to make sure that the header doesn't get rendered (in the B variant) if `tasklist_header_ab_test_applies?` is false?

test/functional/transaction_controller_test.rb
```diff
@@ -58,6 +58,31 @@ class TransactionControllerTest < ActionController::TestCase
...
+      should "not show the tasklist header by default" do
+        with_variant TaskListHeader: "A" do
+          get :show, slug: "learn-to-drive-miss-daisy"
+
+          assert_template partial: "_tasklist_header", count: 0
+        end
+      end
+
+      should "show the tasklist header for the 'B' version" do
+        with_variant TaskListHeader: "B" do
```
sihugh: Could we have a test to make sure that the header doesn't get rendered (in the B variant) if `tasklist_header_ab_test_applies?` is false?

deborahchua: Thanks @sihugh , I have added this in now.

test/functional/simple_smart_answers_controller_test.rb
deborahchua: Thanks @sihugh , I have added this in now.

PR #1309, opened 2017-11-08
Upgrade to Rails 5.1.4

danielburnley:
Main changes required revolved around the changes to how params are
passed within tests when calling get. Rather than passing key/values
when calling get, you now need to pass them within a params hash.

In addition to this, the params hash within controllers is now an
object which has to explictly allow parameters.

Finally, we upgraded anything that was required to be upgraded to be
compatible with Rails 5


Gemfile
```diff
@@ -4,26 +4,27 @@ gem 'addressable'
...
 gem 'govuk_navigation_helpers', '~> 6.3'
 gem 'htmlentities', '~> 4.3.0'
 gem 'invalid_utf8_rejector', '~> 0.0.0'
 gem 'logstasher', '~> 1.1.0'
 gem 'plek', '~> 1.12.0'
-gem 'rack_strip_client_ip', '~> 0.0.0'
-gem 'rails', '4.2.10' # version 5 is available
-gem 'rails-i18n', '~> 4.0.0'
+gem 'rack_strip_client_ip', '~> 0.0.2'
+gem 'rails', '5.1.4' # version 5 is available
```
h-lame: We probably don't need the `# version 5 is available` comment any more.

config/initializers/cookies_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# Be sure to restart your server when you modify this file.
+
+# Specify a serializer for the signed and encrypted cookie jars.
+# Valid options are :json, :marshal, and :hybrid.
+Rails.application.config.action_dispatch.cookies_serializer = :marshal
```
RoryMacDonald: This should be :hybrid

config/spring.rb
```diff
@@ -0,0 +1,6 @@
+%w(
+  .ruby-version
+  .rbenv-vars
+  tmp/restart.txt
+  tmp/caching-dev.txt
```
RoryMacDonald: I don't think we need the spring config (not seeing it used)

config/puma.rb
```diff
@@ -0,0 +1,56 @@
...
+# option, you will want to use this block to reconnect to any threads
+# or connections that may have been created at application boot, as Ruby
+# cannot share connections between processes.
+#
+# on_worker_boot do
+#   ActiveRecord::Base.establish_connection if defined?(ActiveRecord)
+# end
+#
+
+# Allow puma to be restarted by `rails restart` command.
```
RoryMacDonald: We've been removing the puma configs, as unicorn seems to be running.

config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
```
RoryMacDonald: I don't think we need actioncable - can remove this file

PR #1311, opened 2017-11-10
Don't show new nav for world wide mainstream pages

tijmenb:
Same as https://github.com/alphagov/government-frontend/pull/533.

We now determine which navigation to show according to a complex decision tree:

image

When we rolled it out we thought that it would be good to always show the new navigation if the content is tagged to the world wide taxonomy (https://github.com/alphagov/government-frontend/pull/510), because it would be closest to what the B-variant was.

This had unintended effects of showing the world wide navigation on some mainstream pages, because they're tagged to the "generic" taxons in the world wide taxonomy.

## Before

screen shot 2017-11-10 at 09 21 46

## After

screen shot 2017-11-10 at 09 21 30

https://trello.com/c/TqnWwFif

### DIFF
--- a/app/controllers/concerns/taxonomy_navigation.rb
+++ b/app/controllers/concerns/taxonomy_navigation.rb
@@ -1,17 +1,11 @@
 module TaxonomyNavigation
   def should_present_taxonomy_navigation?(content_item)
-    content_item.present? && (tagged_to_world_wide_taxonomy?(content_item) ||
-        (!content_is_tagged_to_browse_pages?(content_item) &&
-        content_is_tagged_to_a_taxon?(content_item)))
+    content_item.present? &&
+      !content_is_tagged_to_browse_pages?(content_item) &&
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/taxonomy_navigation.rb
```diff
@@ -1,18 +1,12 @@
 module TaxonomyNavigation
   def should_present_taxonomy_navigation?(content_item)
-    content_item.present? && (tagged_to_world_wide_taxonomy?(content_item) ||
```
steventux: @tijmenb we've employed similar logic in smart-answers, are they going to suffer from a similar problem?

tijmenb: Yes, we should remove the special casing there too. I've added it to  https://trello.com/c/TqnWwFif/101-mop-up-issues-with-the-world-wide-taxonomy.

Issue #1059, opened 2016-11-22
Homepage: Promo headings get repeated


selfthinker:
The promo headings on the homepage ([source](https://github.com/alphagov/frontend/blob/master/app/views/homepage/index.html.erb#L163-L189])) get repeated on Assistive Technology because the decorative images have the very same `alt` parameter as the `h3` directly underneath. Ideally the `alt` parameter should be empty but cannot be as it's surrounded by a link.

Removing the link and emptying the `alt` parameter would fix that. Having a link around the image seems to be a deliberate decision, so another fix could be to add the `h3` to the link. So, `<a><img /><h3 /></a>`. There might be other solutions.

fofr:
+1 for moving heading into link

tijmenb:
I think https://github.com/alphagov/frontend/pull/1313 has fixed this issue. Please reopen if that's not the case.

PR #1312, opened 2017-11-10
[Merge + Deploy on Monday 13th November] Update homepage promo slot to remind users about MOT

nickcolley:
Should go live November 13th as per https://trello.com/c/qndJ8bos/170-change-feature-promo-slot-on-govuk-homepage

Optimized the JPG image

I intend to do a follow up PR, that may need a longer discussion, to try and resolve the
on going accessibility issues with these promo slots as discussed here: https://github.com/alphagov/frontend/issues/1059

screen shot 2017-11-10 at 12 37 41


selfthinker:
I don't think it makes sense to have anything in a commit message which isn't about the actual change itself and anything directly to do with it which helps understand the change.

What you intend to do later doesn't have anything directly to do with this particular change. It can make sense in the PR description, though.
Although having a link to the original request (here: to Trello) makes absolute sense in the PR, a commit should be self-contained and have everything important to it in it .
And when this should go live is also less relevant for the change itself, but very relevant for the PR description.

nickcolley:
I wrote that in the commit so people would know why I separated these changes, due to the time constraints.

selfthinker:
I don't understand your comment. It should have been in two separates commits anyway.

PR #1313, opened 2017-11-10
Fix accessibility issue with promo slots

nickcolley:
Follow on from https://github.com/alphagov/frontend/pull/1312

This has been a long standing accessibility issue that we duplicate content in our promo slots.

This change reduces the amount of time we're linking to one, and keeps the thumbnail clickable.

I've intentionally left the heading looking like a link to make it obvious to all users that they can click it.

This also allows us to remove the duplicate link from the copy below.

See https://github.com/alphagov/frontend/issues/1059 for original discussion

Before

screen shot 2017-11-10 at 12 37 41

After

screen shot 2017-11-10 at 12 34 14

screen shot 2017-11-10 at 12 36 28


nickcolley:
Asked the content community on Slack:

"Looks good to me. Link text looks easier to scan and is more meaningful.' - Andy Harsant

"I think it's a good idea. But I thought you weren't meant to put headings inside links - is that allowed now?" - Ben Mortimer

I double checked Ben's question, and yes with HTML5 this is fine.

https://www.w3.org/TR/html5/text-level-semantics.html#the-a-element

> The a element may be wrapped around entire paragraphs, lists, tables, and so forth, even entire sections, so long as there is no interactive content within (e.g. buttons or other links). This example shows how this can be used to make an entire advertising block into a link:

miaallers:
Looks good to me!

nickcolley:
Thanks everyone for the comments. 👍

accessiblewebuk:
Looks good to me too.

PR #1316, opened 2017-11-13
Upgrade to Rails 5.1.4

danielburnley:
This PR is to put the Rails 5 upgrade back into master - it was reverted as a result of forms not being able to POST correctly. This issue is fixed in #1317 which will be merged into this branch if it is an acceptable solution


nickcolley:
Hey Daniel, can you explain why this is being reverted?

danielburnley:
This is to put the Rails upgrade back into master - as it was reverted as a result of the PR you linked above. The title could do with editing to make it clear though

PR #1317, opened 2017-11-13
[WIP] Remove CSRF Token validation for Rails 5 upgrade

danielburnley:
When migrating to Rails 5 - posting forms was broken to cause all post requests to be unable to validate the authenticity token, despite the token being passed through with the params correctly.

As the application has no sense of users and/or sessions and all POST requests are from unauthenticated users, there is no benefit from CSRF token protection and in removing it we fix the issue introduced in the Rails 5 upgrade

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,7 +1,6 @@
 require 'gds_api/helpers'
 
 class ApplicationController < ActionController::Base
-  protect_from_forgery
   include GdsApi::Helpers
   include Slimmer::Headers
   include Slimmer::Template
--- a/config/application.rb
+++ b/config/application.rb
@@ -60,5 +60,9 @@ module Frontend
     config.action_dispatch.default_headers = {
       'X-Frame-Options' => 'ALLOWALL'
     }
+
+    config.middleware.delete ActionDispatch::Cookies
+    config.middleware.delete ActionDispatch::Session::CookieStore
+    config.action_controller.allow_forgery_protection = false
   end
 end


tijmenb:
Is there an explanation of why this stops working between Rails 4/5? 

Since this app can't set cookies, I don't really understand why it CSRF has ever worked.

danielburnley:
After digging into the Rails code I discovered that in Rails 4 - when a token couldn't be verified it only set environment variables and set the session to an empty hash: 

```ruby
def handle_unverified_request
    request = @controller.request
    request.session = NullSessionHash.new(request.env)
    request.env['action_dispatch.request.flash_hash'] = nil
    request.env['rack.session.options'] = { skip: true }
    request.env['action_dispatch.cookies'] = NullCookieJar.build(request)
end
```

In Rails 5 however, this has been changed to use the flash middleware, which we didn't have enabled which threw an exception. 
 
```ruby
def handle_unverified_request
    request = @controller.request
    request.session = NullSessionHash.new(request)
    request.flash = nil
    request.session_options = { skip: true }
    request.cookie_jar = NullCookieJar.build(request, {})
end
```

If you change `protect_from_forgery` in the Rails 4 application controller to `  protect_from_forgery with: :exception`, you can see it actually throws the exception. Before now it was just putting a warning in the log which said: 

```
Processing by FindLocalCouncilController#find as HTML                                                                                                                                                                
  Parameters: {"postcode"=>"[FILTERED]"}                                                                                                                                                                             
Can't verify CSRF token authenticity  
```

The only reason it throws an exception now is because we don't have the flash middleware enabled to see this failure, the token itself was always invalid

PR #1318, opened 2017-11-14
Update link to statistics

frankieroberto:
https://www.gov.uk/government/publications?publication_filter_option=statistics redirects to https://www.gov.uk/government/statistics, so this just updates the link to remove the redirect request, and also make the visited style appear correctly if you've visited that section.


nickcolley:
Hey @frankieroberto !

Thanks so much for the contribution, it's looking good, but I need to do some poking to get the build to pass since you're an external contributor.

nickcolley:
Nice thanks, we will deploy this soon. So you should see this fix live then! 👍

frankieroberto:
@nickcolley no probs, and thanks!

PR #1324, opened 2017-11-23
Update Ruby & bundle update

danielburnley:
We have upgraded Ruby alongside running a general bundle update to reduce the number of PRs that dependabot will create when being activated on this repository.

There are new deprecation warnings which are generated from json-schema, which is a dependency of govuk-content-schema-test-helpers. To fix these, it will need to 
be fixed in that repository first, and then the version bumped in this repository

https://trello.com/c/vX8qwh2X/46-ruby-upgrade-of-all-apps


Gemfile
```diff
@@ -43,7 +43,7 @@ group :test do
   gem 'capybara'
   gem 'ci_reporter'
   gem 'govuk-content-schema-test-helpers'
-  gem 'govuk_schemas'
+  gem 'govuk_schemas', '2.3.0'
```
cbaines: What's the reasoning behind adding this restriction?

danielburnley: There is a breaking change in version 3.0 - to avoid doing too much inside this PR I've locked it to the last version before 3.0

PR #1340, opened 2017-12-11
Fix some dep warnings & lint offenses

h-lame:
I spotted while doing #1334 that there were a bunch of warnings in the console output so I thought I'd fix them.  I also decided to remove all the custom exclusions from `.rubocop.yml` and fix any outstanding lint issues.  I've done these in lots of small commits.

Mostly they are entirely harmless but 3 lint fixes in particular made actuall changes to the app that bear close inspection:

1. `When setting LOWEST_TIER_AREA_TYPES don't mutate UNITARY_AREA_TYPES` - ee671ccb9e8540b96f7134fd34e4a55e473a699f - changes what we consider a unitary council type
2. `Fix outstanding Date & Time lint offenses` - bec379039eb776e70d77fc908f70e559c34d259f - changes the output format of times in the travel advice atom feed (it should be functionally identical though)
3. `Remove wait_until and with/without_javascript helpers` - a5d420b8c576647b71553949daee243707658bd3 - removes some integration test helpers relating to javascript


h-lame:
@boffbowsh - In theory, it should be ok because `+00:00` and `Z` refer to the same thing, but you never really know what some external parser is going to do, so I think you're right to be wary.  I did a little more digging and found out two things:

0. We don't really have much control over the `updated` element that the rails atom builder will produce.  It's going to assume it gets an object it can call `.xmlschema` on so it's not like we can provide it with a pre-formatted string instead of a calendrical object.
1. If we call `.to_datetime` on the `Time.zone.parse` result we get a `DateTime` instance back and that means we keep the original `xmlschema` behaviour and output `+00:00` instead of `Z`. Happily, rubocop doesn't complain that we are using `DateTime` anymore either so I guess we're respecting the law of the cop, if not the spirit.
2. The reason we get `Z` from `Time.zone.parse` is that this app has `config.time_zone` set to `UTC` (that is, it hasn't explicitly made a choice about timezone).  If, like whitehall and government-frontend, we set the time_zone of the app to `London` then we'd get `+00:00` now and `+01:00` in summer when we're in BST.

I think that setting the time_zone of the app to `London` is the right thing to do here (along with the rest of the use Time not DateTime changes).

h-lame:
FWIW it's pretty [hit and miss if our apps have config.time_zone set to anything](https://github.com/search?q=org%3Aalphagov+config.time_zone&type=Code), which feels a bit unusual to me and might explain any strange BST behaviour we get.

boffbowsh:
I'm in favour of setting time zone to London for this. I'd rather not risk alerting people to things that happened in the past. While I agree an RSS reader _shouldn't_ see a difference, I can quite imagine something somewhere is doing `hash(updatedAt)` or something.

h-lame:
@boffbowsh - thanks!  I've re-written the [`Fix outstanding Date & Time lint offenses`](https://github.com/alphagov/frontend/pull/1340/commits/da1cf17b02d5acc356d89609dbc24624046887a1) commit to change the time zone in config.

PR #1341, opened 2017-12-11
Fix cop name in rubocop.yml

tijmenb:
Prevents this deprecation message on CI:

```
bundle exec govuk-lint-ruby --rails --diff --format clang app test lib config
.rubocop.yml: Style/AccessModifierIndentation has the wrong namespace - should be Layout
```


h-lame:
I've got that in here #1340 (among other things).

tijmenb:
Ah thanks!

PR #1358, opened 2017-12-21
Update number of agencies and public bodies

emmabeynon:
Trello: https://trello.com/c/YRSePZfr/266-update-the-number-of-depts-and-orgs-on-the-homepage

This updates the number of "Other agencies and public bodies" on the homepage
to 385, which matches the number on "/government/organisations".


emmabeynon:
@boffbowsh might be an idea 😂  Thanks for the review!

PR #1367, opened 2018-01-04
Bump govuk_app_config

sihugh:
This includes unicorn and logstasher so we don't need to reference them in the
Gemfile any more.

We also remove logstasher config that is already set up in [govuk_app_config](https://github.com/alphagov/govuk_app_config/blob/d1b537fee1aae18ca98aaf1be1574eafcdfb62d4/lib/govuk_app_config/govuk_logging.rb)


sihugh:
I think this problem is to do with the jasmine test runner.  I'll have a poke at it and see if we can't update it a bit

PR #1371, opened 2018-01-05
Remove the conditional for precompiling assets in Dockerfile

RyanMacG:
As part of the work to use the built images pushed up to Docker Hub for E2E testing, where we run the tests in the production RAILS_ENV, we will need the image pushed up to contain the compiled assets.

We can't see a reason to keep the conditional for now as any performance hit will be experienced only as part of CI.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,7 @@ RUN bundle install
 
 ADD . $APP_HOME
 
-ARG COMPILE_ASSETS=false
-RUN if [ "$COMPILE_ASSETS" = "true" ] ; then bundle exec rails assets:precompile ; fi
+RUN GOVUK_WEBSITE_ROOT=https://www.gov.uk GOVUK_APP_DOMAIN=www.gov.uk RAILS_ENV=production bundle exec rails assets:precompile
 
 HEALTHCHECK CMD curl --silent --fail localhost:$PORT || exit 1
 


Dockerfile
```diff
@@ -18,8 +18,7 @@ RUN bundle install
 
 ADD . $APP_HOME
 
-ARG COMPILE_ASSETS=false
-RUN if [ "$COMPILE_ASSETS" = "true" ] ; then bundle exec rails assets:precompile ; fi
+RUN GOVUK_WEBSITE_ROOT=http://www.dev.gov.uk GOVUK_APP_DOMAIN=www.gov.uk RAILS_ENV=production bundle exec rails assets:precompile
```
kevindew: I guess for consistency it'd be good if the GOVUK_WEBSITE_ROOT matched GOVUK_APP_DOMAIN even though we don't expect them to be used.

Dockerfile
```diff
@@ -18,8 +18,7 @@ RUN bundle install
 
 ADD . $APP_HOME
 
-ARG COMPILE_ASSETS=false
-RUN if [ "$COMPILE_ASSETS" = "true" ] ; then bundle exec rails assets:precompile ; fi
+RUN GOVUK_WEBSITE_ROOT=http://www.dev.gov.uk GOVUK_APP_DOMAIN=dev.gov.uk RAILS_ENV=production bundle exec rails assets:precompile
```
kevindew: Sorry nitpicking again. For consistency with all the other repos can this be production: `GOVUK_WEBSITE_ROOT=https://www.gov.uk GOVUK_APP_DOMAIN=www.gov.uk`

PR #1402, opened 2018-02-15
Use button component

steventux:
https://trello.com/c/3VwTvDrj/256-use-button-component-in-frontend-formats

Depends on https://github.com/alphagov/static/pull/1289

Use button component in Frontend formats. This will help with consistency and moving formats to other apps in future.

### DIFF
--- a/app/assets/stylesheets/views/_simple-smart-answer.scss
+++ b/app/assets/stylesheets/views/_simple-smart-answer.scss
@@ -1,0 +1,8 @@
+.step {
+  .next-question {
+    button {
+      font-size: 19px;
+      line-height: 1.3684210526;
+    }
+  }
+}
--- a/app/views/completed_transaction/show.html.erb
+++ b/app/views/completed_transaction/show.html.erb
@@ -15,8 +15,11 @@
         <p>If you needed an organ transplant would you have one? If so please help others.</p>
         <p>If you live in Wales and want to be an organ donor, you don’t need to do anything. Find out about your choices for <a href="https://www.organdonation.nhs.uk/supporting-my-decision/welsh-legislation-what-it-means-for-me/" rel="external">organ donation in Wales</a>.</p>
         <p>
-          <%= link_to 'Join', @publication.promotion['url'],
-                title: "Register to become an organ donor", rel: "external", class: "button", role: "button" %>
+          <%= render "govuk_component/button",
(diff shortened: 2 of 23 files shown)


test/integration/transaction_test.rb
```diff
@@ -102,9 +108,9 @@ class TransactionTest < ActionDispatch::IntegrationTest
       visit "/foo"
```
maxgds: This doesn't seem to test the statement in the context any more.

steventux: `assert_has_button_component` [performs an equality test on all button attributes](https://github.com/alphagov/frontend/pull/1402/files/53c9bb105bd37cc57e21d8185dceb4f78b3d0c25#diff-fe8c1aa14e9f516646c2efbae788a973R13) (after extracting `:text`) so this does assert attributes passed to the button do not contain data tracking attrs as in the above test.

app/views/completed_transaction/show.html.erb
```diff
@@ -15,8 +15,11 @@
         <p>If you needed an organ transplant would you have one? If so please help others.</p>
         <p>If you live in Wales and want to be an organ donor, you don’t need to do anything. Find out about your choices for <a href="https://www.organdonation.nhs.uk/supporting-my-decision/welsh-legislation-what-it-means-for-me/" rel="external">organ donation in Wales</a>.</p>
         <p>
-          <%= link_to 'Join', @publication.promotion['url'],
-                title: "Register to become an organ donor", rel: "external", class: "button", role: "button" %>
+          <%= render "govuk_component/button",
+                      text: "Join",
+                      href: @publication.promotion['url'],
+                      rel: "external",
+                      extra_attrs: { title: "Register to become an organ donor" } %>
```
steventux: I just spotted this, it's a problem because [the button component only accepts data attributes](https://github.com/alphagov/static/blob/master/app/views/govuk_component/button.raw.html.erb#L16)

app/assets/stylesheets/views/_simple-smart-answer.scss
```diff
@@ -0,0 +1,8 @@
+.step {
+  .next-question {
+    button {
+      font-size: 19px;
```
nickcolley: @steventux does this need to be fixed in the button component?

PR #1403, opened 2018-02-15
Drop number of procs

sihugh:
Since guides, answers and help pages are no longer served by frontend it has
much less traffic than it used to.  We're going to drop a unicorn worker process
and add it to government-frontend.

This should be deployed before https://github.com/alphagov/govuk-puppet/pull/7255
is merged.  We'll monitor it for a bit to make sure it's OK.

Once the puppet PR is merged, the ENV var should take precedence over the value
set in unicorn.rb.

### DIFF
--- a/config/unicorn.rb
+++ b/config/unicorn.rb
@@ -4,4 +4,4 @@ end
 load_file_if_exists(self, "/etc/govuk/unicorn.rb")
 
 working_directory File.dirname(File.dirname(__FILE__))
-worker_processes 4
+worker_processes Integer(ENV.fetch("UNICORN_WORKER_PROCESSES", 3))


config/unicorn.rb
```diff
@@ -4,4 +4,4 @@ def load_file_if_exists(config, file)
 load_file_if_exists(self, "/etc/govuk/unicorn.rb")
 
 working_directory File.dirname(File.dirname(__FILE__))
-worker_processes 4
+worker_processes Integer(ENV.fetch("UNICORN_WORKER_PROCESSES", 3))
```
cbaines: @kevindew has been making some changes in this area recently, and for consistency, it might be good to switch to using the Unicorn configuration in newer versions of `govuk_app_config` [1]?

1: https://github.com/alphagov/govuk_app_config#unicorn

The effect would be different though, as the default is 2, rather than 3.

sihugh: Yeah, there's a bit more work involved than just bumping the gem.  The jasmine tests break when you bump to the new logstasher config, somewhere around the repointing of `stdout`.  I was hoping to avoid that here :-)

kevindew: ahh interesting! I was tinkering with frontend earlier didn't realise there was some pain with the bump of govuk_app_config and this. I'll be trying to sort it anyway for a docker change I'm trying to make.

sihugh: It's in https://github.com/alphagov/govuk-puppet/pull/7255 but I didn't want to merge it yet because I want to make sure frontend is OK first

kevindew: 🎉 🎉 🌮

boffbowsh:
I'm not convinced that giving one unicorn worker to government-frontend and taking it away from frontend will help much. The boxes only have two cores and they're around ~60% utilised as-is. There's not much RAM pressure on the machine, so we don't actually need to remove a frontend worker.

We can test this out in production temporarily by sending signals to the Unicorns to change the number of workers on the fly: https://bogomips.org/unicorn/SIGNALS.html
Send `TTOU` to frontend then `TTIN` to government-frontend to achieve this effect. Then we can see  if there's any changes to latency etc.

sihugh:
[Kevin's PR](https://github.com/alphagov/frontend/pull/1414) means this can all be done in puppet now so I'll bow out gracefully.

PR #1428, opened 2018-02-27
Revert "Use button component"

steventux:
Reverts alphagov/frontend#1402
Reverting due to styling differences, the component is needs to include the `.button` css class styles
Component:
screenshot from 2018-02-27 16-24-21

Production:
screenshot from 2018-02-27 16-24-37


nickcolley:
@steventux looks like you need to make sure you're using the 'start' variant of the component http://govuk-static.herokuapp.com/component-guide/button/start_now_button

PR #1430, opened 2018-02-28
Use button component

steventux:
https://trello.com/c/3VwTvDrj/256-use-button-component-in-frontend-formats

Another attempt at https://github.com/alphagov/frontend/pull/1402 adds the start button styling missed in the previous PR.

### DIFF
--- a/app/assets/stylesheets/views/_simple-smart-answer.scss
+++ b/app/assets/stylesheets/views/_simple-smart-answer.scss
@@ -2,7 +2,7 @@
   .next-question {
     button {
       font-size: 19px;
-      line-height: 1.3684210526;
+      line-height: 1.25;
     }
   }
 }
--- a/app/assets/stylesheets/views/_simple-smart-answer.scss
+++ b/app/assets/stylesheets/views/_simple-smart-answer.scss
@@ -1,0 +1,8 @@
+.step {
+  .next-question {
+    button {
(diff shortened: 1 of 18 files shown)


app/assets/stylesheets/views/_simple-smart-answer.scss
```diff
@@ -0,0 +1,8 @@
+.step {
+  .next-question {
+    button {
+      font-size: 19px;
+      line-height: 1.3684210526;
```
maxgds: This is a weird number - where we see this in a browser it's usually a result of something being given as a % of something else, it doesn't look like something we should be hard coding.

vanitabarrett: Agreed, the previous button seemed to use `line-height: 1.25` which I think we can use to have the same effect.

I'd also ask whether we need these styles here - it'd be better not to be overriding component styles for specific formats.

steventux: I've amended the line height to `1.25` here https://github.com/alphagov/frontend/pull/1430/commits/0f129bbc6b4f76f2e95babacf347baad3e721ad2
The button font size was controlled by the `big` css class, we don't make provision for passing additional css classes to the button component so either an override is necessary or we accept the stylistic differences.

PR #1432, opened 2018-03-01
Makes use of usesLicensify API param

elenatanasoiu:
For: https://trello.com/c/QswdEVKO/69-2-changes-to-licensing-controller-for-licensify

It was pointed out that some licenses are showing up incorrectly as available in licensify,
with the user being offered the option to start applying for it, although they cannot continue
further than the first step. This is frustrating because the user gets the feeling of being stuck
when in fact they should've never been offered the option in the first place.

The API response already tells us whether the licence should be using Licensify, in a param called
`usesLicensify` so we now check it before showing a start page to the user.

If the param is set to false, we now display a "License unavailable" page instead.

### DIFF
--- a/app/presenters/licence_details_presenter.rb
+++ b/app/presenters/licence_details_presenter.rb
@@ -19,6 +19,20 @@ class LicenceDetailsPresenter
     !licence_details["location_specific"]
   end
 
+  def has_any_actions?
+    authority && authority["actions"].present?
+  end
+
+  def uses_licensify(chosen_action = action)
+    return false unless authority
+    chosen_action_info = authority.dig("actions", chosen_action)
+    if chosen_action_info.present?
+      chosen_action_info.any? { |link| link && link['uses_licensify'] }
+    else
+      false
+    end
+  end
+
   def action
     return nil unless interaction
     raise RecordNotFound unless authority["actions"].keys.include?(interaction)
@@ -78,7 +92,8 @@ private
                 'url' => link['url'],
                 'introduction' => link['introductionText'],
                 'description' => link['description'],
-                'payment' => link['payment']
+                'payment' => link['payment'],
+                'uses_licensify' => (link['usesLicensify'] == true)
               }
             }
             actions
--- a/app/views/licence/_action.html.erb
+++ b/app/views/licence/_action.html.erb
@@ -18,11 +18,16 @@
 
     <%= simple_format link['introduction'] %>
 
-    <p>
-      <%= render "govuk_component/button",
-        text: "Apply online",
-        start: true,
-        href: link['url'] %>
-    </p>
+    <% if link['uses_licensify'] %>
+      <p>
+        <%= render "govuk_component/button",
+                   text: "Apply online",
+                   start: true,
+                   href: link['url'] %>
+      </p>
+    <% else %>
+      <%= render partial: 'licensify_unavailable' %>
+    <% end %>
   </section>
+
 <% end %>
--- a/app/views/licence/_authority_details.html.erb
+++ b/app/views/licence/_authority_details.html.erb
@@ -24,7 +24,11 @@
 <article role="article" class="content-block group">
   <div id="overview" class="inner">
     <% if @licence_details.action.present? %>
-      <%= render :partial => "action", :locals => {:action => @licence_details.action } %>
+      <% if @licence_details.uses_licensify %>
+        <%= render :partial => "action", :locals => {:action => @licence_details.action } %>
+      <% else %>
+        <%= render partial: 'licensify_unavailable' %>
+      <% end %>
     <% elsif @publication.licence_overview.present? %>
       <header><h1>Overview</h1></header>
       <%= raw @publication.licence_overview %>
--- a/test/integration/licence_test.rb
+++ b/test/integration/licence_test.rb
@@ -117,20 +117,23 @@ class LicenceTest < ActionDispatch::IntegrationTest
                     "url" => "/licence-to-kill/westminster/apply-1",
                     "description" => "Apply for your licence to kill",
                     "payment" => "none",
-                    "introduction" => "This licence is issued shaken, not stirred."
+                    "introduction" => "This licence is issued shaken, not stirred.",
+                    "usesLicensify" => true
                   }, {
                     "url" => "/licence-to-kill/westminster/apply-2",
                     "description" => "Apply for your licence to hold gadgets",
                     "payment" => "none",
-                    "introduction" => "Q-approval required."
-                  }
+                    "introduction" => "Q-approval required.",
+                    "usesLicensify" => true
+                }
                 ],
                 "renew" => [
                   {
                     "url" => "/licence-to-kill/westminster/renew-1",
                     "description" => "Renew your licence to kill",
                     "payment" => "none",
-                    "introduction" => ""
+                    "introduction" => "",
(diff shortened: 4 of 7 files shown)


app/presenters/licence_details_presenter.rb
```diff
@@ -19,6 +19,12 @@ def licence_authority_specific?
     !licence_details["location_specific"]
   end
 
+  def uses_licensify?
+    return false unless interaction
```
h-lame: Should this be `unless action`?  It's not 100% clear to me why we guard against a missing `interaction` when we don't then use it later in the method.

elenatanasoiu: interaction is what action is initialized with

```
  def action
    return nil unless interaction
    raise RecordNotFound unless authority["actions"].keys.include?(interaction)

    interaction
  end
```

It is better to just check for action so it's clearer so I'll do that. I was trying to avoid some unnecessary checks.

app/presenters/licence_details_presenter.rb
```diff
@@ -18,7 +18,20 @@ def local_authority_specific?
...
+  end
+  
+  def uses_licensify(chosen_action)
+    return false unless chosen_action
+    authority["actions"] && authority["actions"][chosen_action][0]['uses_licensify']
+  end
+  
+  def uses_licensify?
+    uses_licensify(action)
+  end
```
h-lame: I think it might be more consistent if we combine the "takes an argument" and "doesn't take an argument" versions here into:
```ruby
def user_licensify?(chosen_action = action)
  return false unless chosen_action
  authority["actions"] && authority["actions"][chosen_action][0]['uses_licensify']
end
```

I found it a little jarring down in the views where sometimes it was `uses_licensify(something)` and sometimes `uses_licensify?`.

app/presenters/licence_details_presenter.rb
```diff
@@ -18,7 +18,20 @@ def local_authority_specific?
...
   end
-
+  
+  def any_action_uses_licensify?
+    authority["actions"] && authority["actions"].keys.uniq.any?{|chosen_action| uses_licensify(chosen_action) == true }
+  end
+  
+  def uses_licensify(chosen_action)
+    return false unless chosen_action
+    authority["actions"] && authority["actions"][chosen_action][0]['uses_licensify']
```
h-lame: Do we need to check _all_ the possible links for a given action, rather than the first one?  I'm not 100% sure of the structure of the response but it sounds like: there's 1 authority, it can have zero or more actions, and each of those can have one or more links (not sure about these names but they'll do for now).  In the view when rendering the info about each link we check specifically the `uses_licensify` details of each link before rendering it, but we also guard the navigation that would take us to the page where we render those links by using this method and it only checks the first link for each action.

app/views/licence/_action.html.erb
```diff
@@ -2,22 +2,26 @@
...
+      <h2><a id="form-<%= i+1 %>"></a><%= i+1 %>. <%= link['description'] %></h2>
 
-    <%= simple_format link['introduction'] %>
+      <%= simple_format link['introduction'] %>
 
-    <p><%= link_to "Apply online", link['url'], :class => "button" %></p>
-  </section>
+      <p><%= link_to "Apply online", link['url'], :class => "button" %></p>
+    </section>
+  <% end %>
```
h-lame: I don't know enough about the contents of the api responses to know if this question makes sense, but instead of rendering nothing if a given link doesn't use licensify, could we render the information we do have about the link and put up the "licensify_unavailable" partial for it?

Is it useful for someone to know that there are several links for a given action for a license and some of them can be applied for online via licensify, but others can't?  They're presumably still things that you might need to know about or do, so maybe we should provide the user that info even if they can't apply online?

app/views/licence/_authority_details.html.erb
```diff
@@ -10,9 +10,9 @@
           <li class="active">Overview</li>
         <% end %>
         <% @licence_details.authority['actions'].keys.uniq.each do |action| %>
-          <% if @licence_details.action == action %>
+          <% if @licence_details.action == action && @licence_details.uses_licensify(action) %>
             <li class="active"><%= "How to #{action}" %></li>
-          <% else %>
+          <% elsif @licence_details.uses_licensify(action) %>
```
h-lame: Are we sure that the right thing to do is not render any navigational links to the "how to `whatever`" action/link if it doesn't use licensify?  If I'm reading the code correctly, we can still render the navigational link, but instead of showing a big green "do it on licensing" button we'd instead render the "licensing unavailable" page which suggests they contact their council directly.

I don't know if that's correct or not based on the contents of the api response and what it actually means though.

This is sort of the same as my comment where an action has multiple links - is it better to show nothing if they don't use licensify, or show something and point people onwards to the council for more help?

test/integration/licence_test.rb
```diff
@@ -746,4 +746,279 @@ class LicenceTest < ActionDispatch::IntegrationTest
...
+      should "not display authority" do
+        refute page.has_content? 'Ministry of Love'
+        refute page.has_button? 'Get started'
+      end
+
+      should "display the licence unavailable message" do
+        assert page.has_content?("You can't apply for this licence online")
+        assert page.has_content?("Contact your local council")
+      end
+    end
```
h-lame: These integration tests are super useful.  What we don't have is a test covering the (possibly apocryphal?) scenario where a single action has multiple links, and what we do in those cases (currently hide the ones without `usesLicensify` or with `usesLicensify` set to `false`).

app/presenters/licence_details_presenter.rb
elenatanasoiu: Combined

app/presenters/licence_details_presenter.rb
elenatanasoiu: Checking all links

app/views/licence/_action.html.erb
elenatanasoiu: Done

app/views/licence/_authority_details.html.erb
elenatanasoiu: Removed, this is handled in the `_action.html.erb` instead.

test/integration/licence_test.rb
elenatanasoiu: Added tests for multiple link scenario.

test/support/component_helpers.rb
```diff
@@ -16,14 +16,27 @@ def assert_has_component_title(title, context = nil)
...
   end
 
+  def refute_has_button_component(text, attrs = {})
+    all(shared_component_selector('button')).each do |button|
+      data = JSON.parse(button.text).symbolize_keys
+      next unless text == data.delete(:text)
+      data.delete(:data_attributes) if data[:data_attributes].blank?
+      return refute_equal attrs, data
+    end
+  end
```
h-lame: These are nice!

app/presenters/licence_details_presenter.rb
```diff
@@ -54,6 +68,10 @@ def authority
 
 private
 
+  def has_actions?
+    authority.present? && authority["actions"].present?
+  end
```
h-lame: Can we drop this?  There's `has_any_actions?` defined above which has basically the same method body.

elenatanasoiu: Done

PR #1436, opened 2018-03-06
Use content store for step by step navigation

sihugh:
Step by step navigation helpers have [moved to govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/pull/196) as they were so closely coupled to the components themselves.  Our approach has changed a little, as the navigation information is now obtained from the content item's links rather than a hardcoded
file in the navigation helpers gem.

We can remove some tests because we are not responsible for things that the
component does, we just need to ensure that it renders in a somewhat sensible fashion.

https://trello.com/c/hMYAcWL8/444-frontend-renders-step-by-step-navigation-from-content-item


app/views/layouts/application.html.erb
```diff
@@ -26,8 +26,8 @@
   <body class="mainstream <%= yield :body_classes %>">
     <div id="wrapper" class="<%= wrapper_class(@publication || @presenter) %>">
       <% unless current_page?(root_path) %>
-        <% if show_step_nav? %>
-          <%= render 'shared/step_nav_header', step_nav_content: current_step_nav %>
+        <% if step_nav_helper.show_header? %>
```
Davidslv: This will not work if `step_nav_helper` is nil

PR #1478, opened 2018-04-05
Use search component from gem, not static

andysellick:
Search component has moved into the components gem.

Related to this PR to [add the search component to the gem](https://github.com/alphagov/govuk_publishing_components/pull/267).


app/assets/javascripts/all.js
```diff
@@ -5,3 +5,4 @@
 //
 //= require_tree ./
 //= require_tree ./views
+//= require govuk_publishing_components/all_components
```
tijmenb: What does this file do compared to `frontend.js`?

andysellick: An excellent question. I have no idea. But I've just checked, and this line doesn't need to be here, but it does need to be in frontend.js.

PR #1488, opened 2018-04-17
[DO NOT MERGE] Add 'MOT reminder' done page promotion

deborahchua:
A new promotion `MOT reminder` needs to be added to the `completed_transaction` done pages. If `mot_reminder` has been selected as the promotion option in `publisher` the appropriate text will be shown. Some placeholder text has been inserted for now as we are waiting for DVSA to provide the exact content.

Related PRs:
alphagov/govuk-content-schemas#747
https://github.com/alphagov/publisher/pull/830

Trello card: Request for new option in Completed Transaction 'promotion' slot

"Promote MOT Reminders" option selected in `Publisher:
publisher_promotion_options_after

Frontend view: 
frontend_completed_transaction_mot_reminder


app/assets/stylesheets/views/_completed-transaction.scss
```diff
@@ -1,5 +1,5 @@
 .transaction-done {
-  #organ-donor-registration-promotion {
+  #organ-donor-registration-promotion, #mot-reminder-promotion {
```
thomasleese: I wonder if we should make this a “promotion” class rather than using IDs in the CSS?

barrucadu: I guess it depends on how likely we are to add more in the future.

thomasleese: Yeah, I can just see this comment coming up every time we add a new one and it might be worth doing it now 😄

deborahchua: Thanks @thomasleese , I've updated this now. I'll change the other promotions to classes rather than ID's in a separate PR.

thomasleese: Ah, I actually meant to call the class `promotion` and then on https://github.com/alphagov/frontend/pull/1488/files#diff-6e10a2c10820051cc48f8764b75ff654R13 and the new MOT reminder `div`, changing it to `id="..." class="promotion"`

deborahchua: Ah right sorry! I've changed this now.

PR #1494, opened 2018-04-24
Use button component from gem

andysellick:
Button component has moved from static to the gem (https://github.com/alphagov/govuk_publishing_components/pull/271)

This PR updates frontend to use that gem component.

Includes a significant rewrite of the component_helpers to do the testing, previously did some slimmer magic to get at the static components.


test/support/component_helpers.rb
```diff
@@ -7,54 +7,38 @@ def assert_has_component_title(title, context = nil)
...
-      else
-        next unless text == data.delete(:text)
-      end
-      data.delete(:data_attributes) if data[:data_attributes].blank?
-      return refute_equal attrs, data
-    end
+    assert page.has_no_css?(process_button_attributes(attrs), text: text)
   end
 
   def click_button_component(text)
```
tijmenb: I think instead of changing the functionality here we can search/replace `click_button_component` with `click_on`?

It's used here already: https://github.com/alphagov/frontend/blob/8c115a328797ff3b8b81912f1b14dec73609f2f8/test/integration/licence_test.rb#L550

test/support/component_helpers.rb
```diff
@@ -7,54 +7,38 @@ def assert_has_component_title(title, context = nil)
...
-      data = JSON.parse(button.text).symbolize_keys
-      if attrs[:href] && data[:href]
-        next unless text == data.delete(:text) && attrs[:href] == data[:href]
-      else
-        next unless text == data.delete(:text)
-      end
-      data.delete(:data_attributes) if data[:data_attributes].blank?
-      return assert_equal attrs, data
-    end
+  def assert_has_button_component(text)
```
tijmenb: Can we make this `assert_has_button`. The tests shouldn't have to care about how the button is rendered (with a component or not).

test/support/component_helpers.rb
```diff
@@ -7,54 +7,38 @@ def assert_has_component_title(title, context = nil)
...
-      data = JSON.parse(button.text).symbolize_keys
+    if page.has_css?("a.gem-c-button")
+      click_link text
+    else
+      click_button text
+    end
+  end
 
-      next if data[:text] != text
+  def process_button_attributes(attrs)
```
tijmenb: This couples the tests really tightly to the CSS/HTML of the component. Would we be able to test whatever we want to test at a higher level? For example, by clicking on the button to check that something happens?

test/support/component_helpers.rb
andysellick: Great, will change.

test/support/component_helpers.rb
andysellick: Yeah, that's cleaner.

test/support/component_helpers.rb
andysellick: I was hoping to achieve this but didn't know how. I figured that at least the only instances of this were in one place so could easily be changed.

There is this one test that is very specific about this, which is partly why I did this...

https://github.com/alphagov/frontend/blob/master/test/integration/licence_test.rb#L1195-L1212

andysellick: @tijmenb this thread has been marked as outdated even though it isn't, I think it's the only issue blocking this PR. Shall we catch up offline?

PR #1506, opened 2018-05-08
[Do not merge] Remove some jobsearch specific styles

steventux:
Missed in https://github.com/alphagov/frontend/pull/1503


kevindew:
I'm guessing we moved on and didn't actually need this.

PR #1509, opened 2018-05-09
Use component for radio buttons in simple smart answers

tijmenb:
This replaces the old-style radio button blocks with the component, which uses the [new pattern for radio buttons][blog].

## Examples

https://govuk-frontend-app-pr-1509.herokuapp.com/sold-bought-vehicle/y
https://govuk-frontend-app-pr-1509.herokuapp.com/settle-in-the-uk/y
https://govuk-frontend-app-pr-1509.herokuapp.com/exchange-foreign-driving-licence/y

## Screenshots

| Before | After |
| --- | --- |
| screen shot 2018-05-08 at 21 19 31 | screen shot 2018-05-08 at 21 19 33 |
| screen shot 2018-05-08 at 21 19 48 | screen shot 2018-05-08 at 21 19 50 | 

[blog]: https://designnotes.blog.gov.uk/2016/11/30/weve-updated-the-radios-and-checkboxes-on-gov-uk

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -54,3 +54,9 @@
 
 // components
 @import 'govuk_publishing_components/all_components';
+
+// Delete after https://github.com/alphagov/govuk_publishing_components/pull/301
+// is merged and used in this application.
+.gem-c-radio__input {
+  font-size: inherit;
+}
--- a/app/views/simple_smart_answers/_current_question.html.erb
+++ b/app/views/simple_smart_answers/_current_question.html.erb
@@ -11,17 +11,11 @@
           <p class="error-message" role="alert">Please answer this question</p>
         <% end %>
         <%= hidden_field_tag 'response', '' %>
-        <ul class="options">
-          <% question.options.each do |option| %>
-            <li>
-              <label for="response_<%= option.slug %>" class="selectable">
-                <%= radio_button_tag "response", option.slug, (option.slug == params[:previous_response]), :id => "response_#{ option.slug }" %>
-                <%= option.label %>
-              </label>
-            </li>
-          <% end %>
-        </ul>
 
+        <%= render "govuk_publishing_components/components/radio", {
+          name: "response",
+          items: question.options.map { |option| { text: option.label, value: option.slug, checked: option.slug == params[:previous_response] } }
+        } %>
       </div>
 
       <%= render partial: 'draft_fields' %>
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -54,3 +54,8 @@
 
 // components
 @import 'govuk_publishing_components/all_components';
+
+.gem-c-radio__input {
```
andysellick: Why are these styles here? It looks like the size of these elements is already defined in the component (where it should be) and disabling them [doesn't change the appearance](https://govuk-frontend-app-pr-1509.herokuapp.com/sold-bought-vehicle/y)

app/views/simple_smart_answers/_current_question.html.erb
```diff
@@ -11,17 +11,11 @@
           <p class="error-message" role="alert">Please answer this question</p>
         <% end %>
         <%= hidden_field_tag 'response', '' %>
-        <ul class="options">
```
andysellick: Is this class used elsewhere or can it be removed from the CSS?

app/assets/stylesheets/application.scss
tijmenb: Yes! Wanted to ask you about this. The `em` thing doesn't seem to work in this application so the clickable area is about half the size:

screen shot 2018-05-09 at 09 58 42

app/views/simple_smart_answers/_current_question.html.erb
tijmenb: No, it doesn't seem to be used here or in static. The `selectable` class [comes from static](https://github.com/alphagov/static/blob/9e8795b262c948a65b798029864cbe04f193262a/app/assets/stylesheets/helpers/_selectable.scss).

app/assets/stylesheets/application.scss
andysellick: Ah, I see the problem now. Seems to be that the component is inadvertently relying upon some core styles for inputs, which includes font-size: inherit. As a test, if you replace these two lines with that, the input comes out at the right size. Suggest we make an adjustment to the component to define an explicit font size so this works in isolation - might be able to move the font size declaration for the label higher up so it applies to the whole component.

As a side point, I'm surprised that the radio button works this way. I would have thought it would be much easier to simply make the label full width and play with left padding and positioning of the circle, so even though it looks like you're clicking on the radio button you'd in fact always be clicking on the label. But that would involve rebuilding the component, which is way outside of this scope.

PR #1544, opened 2018-06-26
Remove previous universal jobmatch

thomasleese:
This hasn't been the case for a while now so we don't need to include this text.

[Trello Card](https://trello.com/c/bTQqoWTm/511-change-link-text-on-the-homepage)

### DIFF
--- a/app/views/homepage/index.html.erb
+++ b/app/views/homepage/index.html.erb
@@ -26,7 +26,7 @@
             <div class="floated-inner-block content-links-inner">
               <h2>Popular on GOV.UK</h2>
               <ul>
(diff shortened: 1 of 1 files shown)


app/views/homepage/index.html.erb
```diff
@@ -26,7 +26,8 @@
             <div class="floated-inner-block content-links-inner">
               <h2>Popular on GOV.UK</h2>
               <ul>
-                <li><a href="/jobsearch">Find a job (previously Universal Jobmatch)</a></li>
+                <li><a href="/jobsearch">
+                  </a></li>
```
barrucadu: ?

thomasleese: oops

PR #1554, opened 2018-07-13
HUB-194: AB test for DVLA View Driving Licence

jakubmiarka:
We're running an AB test where Verify is the primary option and NINO is in the 'Other ways'
This commit creates a new B variant which is hardcoded and served for user with B cookie.
This test is due to be started on the morning of Tuesday 17th and initially run for 3 days.
The cookie name is 'ABTest-ViewDrivingLicence'.

Co-Authored-By: Rachel Smith <rachel.smith@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/controllers/transaction_controller.rb
+++ b/app/controllers/transaction_controller.rb
@@ -8,7 +8,7 @@ class TransactionController < ApplicationController
   def show
     if request.params['slug'] == 'view-driving-licence'
       ab_test = GovukAbTesting::AbTest.new(
-        'ABTest-ViewDrivingLicence',
+        'ViewDrivingLicence',
         dimension: 68,
         allowed_variants: %w(A B),
         control_variant: 'A'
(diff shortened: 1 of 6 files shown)


app/views/transaction/show.html.erb
```diff
@@ -2,6 +2,7 @@
   <%= render 'govuk_publishing_components/components/machine_readable_metadata',
     schema: :article,
     content_item: @content_item %>
+  <%= @requested_variant.analytics_meta_tag.html_safe if @ab_test_a_variant %>
```
rubenarakelyan: This should always be included rather than a conditional

app/controllers/transaction_controller.rb
```diff
@@ -6,6 +6,23 @@ class TransactionController < ApplicationController
...
+        allowed_variants: %w(A B),
+        control_variant: 'A'
+      )
+      @requested_variant = ab_test.requested_variant(request.headers)
+      @requested_variant.configure_response(response)
+
+      if @requested_variant.variant?('B')
+        render "transaction_variant/show"
+      else
+        @ab_test_a_variant = true
```
rubenarakelyan: This can be removed - see other comment

PR #1645, opened 2018-10-05
[DO NOT MERGE] Add security.txt file

rubenarakelyan:
This commit adds a basic initial `security.txt` file which contains a link to be used by security researchers to contact GOV.UK if they find security issues. It also clarifies that researchers don’t have permission to carry out security testing against GOV.UK.

`security.txt` (https://securitytxt.org/) is a draft RFC that defines a standard file template at a standard location that provides information for security researchers.

We should subsequently add more information to this file, such as a PGP key that researchers can use to encrypt potentially sensitive disclosures that they send to us.

The file is served under `.well-known`, which is the path specified in the RFC, but also at the root domain to match where other files like `robots.txt` are placed, since many researchers expect to find it there.

### DIFF
--- a/app/controllers/securitytxt_controller.rb
+++ b/app/controllers/securitytxt_controller.rb
@@ -1,0 +1,16 @@
+class SecuritytxtController < ApplicationController
+  include Cacheable
+
+  def index
+    render plain: securitytxt_content
+  end
+
+private
+
+  def securitytxt_content
+    <<~SECURITYTXT
+      Contact: https://www.gov.uk/contact/govuk
+      Permission: none
+    SECURITYTXT
+  end
+end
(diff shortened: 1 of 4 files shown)


andreagrandi:
@rubenarakelyan I can only comment on the content here, so: do you think we should provide a PGP key where people can report things securely? I'm not sure if we have one and honestly I don't know how we could manage a shared one. Ideas?

rubenarakelyan:
@andreagrandi Yes we should but I wanted to get the basic file out there first just to make sure that works while we work out how we handle the shared PGP key. Most probably I'll generate one and store it with our other shared secrets. We then just need to make sure that 1st line know what to do if they receive a PGP-encrypted message.

rubenarakelyan:
Waiting for some work to be done on processing these requests before we make the file available.

PR #1694, opened 2018-12-05
HUB-384: Add Variant Support to transaction renderer

chrisclayson:
Update TransactionPresenter and TransactionController to support variants.
Update routes to include a optional `variant` slug on the transaction route.

### DIFF
--- a/app/presenters/transaction_presenter.rb
+++ b/app/presenters/transaction_presenter.rb
@@ -1,6 +1,5 @@
 class TransactionPresenter < ContentItemPresenter
-  attr_writer :variant_slug
-  attr_reader :variant_slug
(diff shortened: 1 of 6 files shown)


app/presenters/transaction_presenter.rb
```diff
@@ -1,4 +1,7 @@
 class TransactionPresenter < ContentItemPresenter
+  attr_writer :variant_slug
+  attr_reader :variant_slug
```
rubenarakelyan: You can use `attr_accessor : variant_slug` here rather than the separate `attr_writer` and `attr_reader`.

rubenarakelyan:
Can you also please put in a short description in the PR

PR #1726, opened 2019-01-14
Replace jQuery-based tabs with component

alex-ju:
After [updating the calendars frontend](https://github.com/alphagov/calendars/pull/483) to use [the new tabs component from the gem](https://content-publisher.integration.publishing.service.gov.uk/component-guide/tabs) this is the only other place where we use the jQuery-based tabs.

This PR replaces the jQuery-based tabs with the tabs component in an effort to [remove legacy scripts from static](https://github.com/alphagov/static/tree/master/app/assets/javascripts) (in this case `jquery.tabs.js` and `jquery.history.js` with a cumulative potential save of ~16kb).

### Before
apply-renew-passport-before

### After
apply-renew-passport-after

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -63,3 +63,22 @@
 .gem-c-radio__input {
   font-size: inherit;
 }
+
+// overwrites
+
+// `content-block` leaks styles into the `govuk-tabs` component so we need to
+// overwrite a few attributes to defend the component and preserve these values
+.content-block {
+  .govuk-tabs {
+    ul {
+      padding: 0;
+      margin: 0;
+    }
+
+    li {
+      list-style: none;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -63,3 +63,22 @@
...
+// overwrite a few attributes to defend the component and preserve these values
+.content-block {
+  .govuk-tabs {
+    ul {
+      padding: 0;
+      margin: 0;
+    }
+
+    li {
+      list-style: none;
```
thomasleese: I think this might have affected the content on certain pages.

For example https://www.gov.uk/view-driving-licence#other-ways-to-apply on the section under the `## Apply by post` there is a list `<li>` that isn't showing the bullet points because of this style (found by @keithiopia).

PR #1728, opened 2019-01-16
Remove PLEK_SERVICE_RUMMAGER_URI

barrucadu:
This app does not use rummager.


cbaines:
I think it does actually https://github.com/alphagov/frontend/blob/master/app/controllers/random_controller.rb#L5 but it calls it search, so this config won't have any effect.

PR #1753, opened 2019-02-11
Remove updated date information

sihugh:
"Do a thing" content is our best content and is aimed at the majority of GOV.UK users. When it's updated though, it is rarely a major update which means the "last updated" date on the content is often something like 2013.  This does not add anything to the user's comprehension or confidence that they're looking at something that is current and relevant.

We're about to publish a "Services" finder which allows users to find this sort of content more easily, however we won't be adding a date sort to the finder because it's just not relevant for this sort of content.

cc @keithiopia


keithiopia:
Background: the content team wasn't selecting the major update flag, so the dates we're displaying are incorrect.
Separately, old dates may lead users to think content isn't up to date anymore when it is.
Recent dates trigger questions from departments when they haven't requested a change; date changes have been caused by backend changes in the past, sometimes they were just typo fixes.

PR #1754, opened 2019-02-11
[WIP] Use a 405 response for requests that aren't matched

sihugh:
404s should be handled by the router.

Using a 405 will enable us to categorise requests to pages that are using an
inappropriate verb.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,6 +5,7 @@ Rails.application.routes.draw do
 
   root to: 'homepage#index', via: :get
   get "/homepage" => redirect("/")
+  match "/", to: proc { |env| [405, {}, ["Not Allowed"]] }, via: :all
 
   get "/random" => "random#random_page"
 
@@ -66,5 +67,5 @@ Rails.application.routes.draw do
     get "*any", to: "error#handler"
   end
 
-  root to: 'homepage#index', via: :get
+  match "*any", to: proc { |env| [405, {}, ["Not Allowed"]] }, via: :all
 end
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -5,6 +5,7 @@
 
   root to: 'homepage#index', via: :get
   get "/homepage" => redirect("/")
+  match "/", to: proc { |env| [405, {}, ["Not Allowed"]] }, via: :all
```
tijmenb: Is this bit necessary, if we already have the `match "*any"`?

config/routes.rb
```diff
@@ -65,5 +67,5 @@
     get "*any", to: "error#handler"
   end
 
-  root to: 'homepage#index', via: :get
+  match "*any", to: proc { |env| [405, {}, ["Not Allowed"]] }, via: :all
```
tijmenb: Hmm, on 2nd thought, this will also return 405 for GET requests for pages that don't exist, so 405 isn't the correct status code.

> The HyperText Transfer Protocol (HTTP) 405 Method Not Allowed response status code indicates that the request method is known by the server but is not supported by the target resource.

https://httpstatusdogs.com/405-method-not-allowed

config/routes.rb
sihugh: The `any*` doesn't seem to match `/` unfortunately

config/routes.rb
sihugh: But those shouldn't get to frontend, so is that a problem?

sihugh: I thought about listing the verbs individually, so happy to do that.  Some posts are legit too though, so 🤷‍♂️

PR #1771, opened 2019-03-01
Remove unneeded gems

sihugh:
We don't need to included therubyracer as we have nodejs installed everywhere.  This was removed from a bunch of the other frontend apps ages ago.

I don't think the other gems are adding anything.  There's nothing particularly fancy about frontend, so we should minimise the number of gems that it uses that aren't elsewhere on the GOV.UK stack.


Gemfile
```diff
@@ -17,14 +17,12 @@ gem 'sass', '~> 3.7.3'
...
 gem 'slimmer', '~> 13.1.0'
 gem 'sprockets-rails', '~> 3.2.0'
-gem 'therubyracer', '~> 0.12.0'
 gem 'uglifier'
 gem 'uk_postcode', '~> 2.1.3'
 gem 'asset_bom_removal-rails', '~> 1.0.0'
 
 group :development do
   gem 'better_errors'
-  gem 'binding_of_caller'
```
cbaines: This is an optional dependency of `better_errors`. So removing this will change the available functionality.

sihugh: Thanks, I've left that in now

PR #1808, opened 2019-04-15
Remove related links test setup

karlbaker02:
This PR removes the related links test setup, which were concluded at the end of Q4.


sihugh:
@karlbaker02 shall we merge this?

PR #1836, opened 2019-05-20
Homepage: patch_links with GDS as organisation

peteglondon:
Currently the homepage does not have a `primary_publishing_organisation`
or any other organisations.

We want to be able to filter on `primary_publishing_organisation`
in `content-data-api' so want this to be populated.

This commit adds GDS to `organisations` and `primary_publishing_organisation`
in `links` with a call to `patch_links` when the homepage is updated.

The existing `publishing_api:publish_special_routes` rake task will
make this happen.


peteglondon:
[Add primary org to Homepage](https://trello.com/c/axDapHeZ/1394-add-primary-org-to-homepage)

PR #1840, opened 2019-05-22
Update search box

sihugh:
This sends keyword searches from the homepage to the all content finder
and also sets the order to relevance.

We're going to talk about whether setting the relevance order makes more sense in finder-frontend, but it might prove clunky, so we'll do it here first.


sihugh:
Closing this as going via the redirect allows us to hit the search page `/search` if you search without a query.  

I'll revert https://github.com/alphagov/static/pull/1692 so that it behaves the same as this again.  We'll have to plan in future work if we want to avoid the redirect because we'll still want the search boxes to navigate differently with different inputs, so not doing it now is the easiest option.

PR #1852, opened 2019-06-02
Update homepage 'department and policy' section links

kr8n3r:
The homepage section about government departments has two adjacent links that both link to the same page, but have different text.

**How does this affect users**
Confusing for screen readers and confusing for cognitive issues
WCAG 2.4.4 (A)

To mitigate against this, we should amend both links and the page they point to, so that links anchor to the relevant section.

This work needs to follow https://github.com/alphagov/collections/pull/1134

Side note: shouldn't the "Agencies and other public bodies" number be updated as well?


kr8n3r:
number will be updated in #1853. will rebase once merged

PR #1856, opened 2019-06-04
Add cookie settings page

vanitabarrett:
New page at `/help/cookies` which will allow users to adjust their cookie preferences (opt in/out of the different categories)
 
Screenshot_2019-06-04 Cookie settings on GOV UK

The page needs to be published as a special route using the `publish_special_routes` rake task - this will be done after this PR is merged. 

Page can be viewed here: https://new-cookie-banner.herokuapp.com/help/cookies

### DIFF
--- a/app/assets/javascripts/modules/cookie-settings.js
+++ b/app/assets/javascripts/modules/cookie-settings.js
@@ -1,0 +1,89 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function CookieSettings () { }
+
+  CookieSettings.prototype.start = function ($module) {
+    this.$module = $module[0]
+
+    this.$module.submitSettingsForm = this.submitSettingsForm.bind(this)
+
+    document.querySelector('form[data-module=cookie-settings]').addEventListener('submit', this.$module.submitSettingsForm)
+
+    this.setInitialFormValues()
+  }
+
+  CookieSettings.prototype.setInitialFormValues = function () {
+    if (!window.GOVUK.cookie('cookie_policy')) {
+      window.GOVUK.setDefaultConsentCookie()
+    }
+
+    var currentConsentCookie = window.GOVUK.cookie('cookie_policy')
+    var currentConsentCookieJSON = JSON.parse(currentConsentCookie)
+
+    // We don't need the essential value as this cannot be changed by the user
+    delete currentConsentCookieJSON["essential"]
+
+    for (var cookieType in currentConsentCookieJSON) {
+      var radioButton
+
+      if (currentConsentCookieJSON[cookieType]) {
+        radioButton = document.querySelector('input[name=cookies-' + cookieType + '][value=on]')
+      } else {
+        radioButton = document.querySelector('input[name=cookies-' + cookieType + '][value=off]')
+      }
+
+      radioButton.checked = true
+    }
+  }
+
+  CookieSettings.prototype.submitSettingsForm = function (event) {
+    event.preventDefault()
+
+    var formInputs = event.target.getElementsByTagName("input")
+    var options = {}
+
+    for ( var i = 0; i < formInputs.length; i++ ) {
+      var input = formInputs[i]
+      if (input.checked) {
+        var name = input.name.replace('cookies-', '')
+        var value = input.value === "on" ? true : false
+
+        options[name] = value
+      }
+    }
+
+    window.GOVUK.setConsentCookie(options)
+    if (!window.GOVUK.cookie("seen_cookie_message")) {
+      window.GOVUK.setCookie("seen_cookie_message", true)
+    }
+
+    this.showConfirmationMessage()
+
+    return false
+  }
+
+  CookieSettings.prototype.showConfirmationMessage = function () {
+    var confirmationMessage = document.querySelector('div[data-cookie-confirmation]')
+    var previousPageLink = document.querySelector('.cookie-settings__prev-page')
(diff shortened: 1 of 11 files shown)


vanitabarrett:
[Productionize the new cookie settings page](https://trello.com/c/fADkc6L5/160-productionize-the-new-cookie-settings-page)

spec/javascripts/unit/modules/cookie-settings.spec.js
```diff
@@ -0,0 +1,176 @@
...
+        } else {
+          expect(radioButtons[i].checked).not.toBeTruthy()
+        }
+      }
+    });
+  });
+
+  describe('submitSettingsForm', function() {
+    beforeEach(function() {
+      confirmationContainer = document.createElement('div')
```
chao-xian: We should be able to DRY these up from here and `showConfirmationMessage` into the parent `beforeEach`

app/assets/stylesheets/views/_cookie-settings.scss
```diff
@@ -0,0 +1,4 @@
+.cookie-settings__confirmation {
+  @include core-19;
```
alex-ju: Is there a reason for not using the new mixins? `@include govuk-font(19);`

app/views/help/cookie_settings.html.erb
```diff
@@ -0,0 +1,114 @@
...
+          text: t('cookies.off')
+        }
+      ]
+    } %>
+
+    <%= render "govuk_publishing_components/components/radio", {
+      name: "cookies-campaigns",
+      inline: true,
+      heading: t('cookies.types.campaigns'),
+      hint: "These cookies are set by third party websites and do things like:
```
alex-ju: We could avoid `html_safe` here by capturing this block above, which could improve the markup readability too:

```
<% cookies_campaigns_hint = capture do %>
  <ul>...</ul>
<% end %>
```

app/assets/javascripts/modules/cookie-settings.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  CookieSettings.prototype.showConfirmationMessage = function () {
+    var confirmationMessage = document.querySelector('div[data-cookie-confirmation]')
+    var previousPageLink = document.querySelector('.cookie-settings__prev-page')
+    var referrer = CookieSettings.prototype.getReferrerLink()
+
+    document.body.scrollTop = document.documentElement.scrollTop = 0
+
+    if (referrer && referrer !== '/help/cookies') {
```
alex-ju: Is there a way to read this path from markup (a data attribute for example) so we don't need to keep it updated in two places?

app/assets/javascripts/modules/cookie-settings.js
```diff
@@ -0,0 +1,89 @@
...
+
+    return false
+  }
+
+  CookieSettings.prototype.showConfirmationMessage = function () {
+    var confirmationMessage = document.querySelector('div[data-cookie-confirmation]')
+    var previousPageLink = document.querySelector('.cookie-settings__prev-page')
+    var referrer = CookieSettings.prototype.getReferrerLink()
+
+    document.body.scrollTop = document.documentElement.scrollTop = 0
```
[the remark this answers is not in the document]
vanitabarrett: @alex-ju The confirmation message does have a wrapper around it with `aria-live=assertive`, so it should read out the message when the form is submitted (for those using screen readers).

https://github.com/alphagov/frontend/pull/1856/files#diff-e394cdb6571fb405da3283b4f5d5cfb8R4

app/assets/stylesheets/views/_cookie-settings.scss
vanitabarrett: @alex-ju No, just failing to break out of old habits! Will correct 👍

app/assets/javascripts/modules/cookie-settings.js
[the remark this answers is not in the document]
alex-ju: Ah, I've missed that one. It doesn't seem to do the job without moving the focus there (tested with VoiceOver on Safari and Chrome). Can we do some testing around that, if we haven't already.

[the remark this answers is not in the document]
vanitabarrett: @alex-ju Hmm strange, I think it worked when I tested it before. I'll take a look 🙂

app/assets/javascripts/modules/cookie-settings.js
vanitabarrett: @alex-ju I've changed the logic slightly instead, so it now checks that the referrer isn't equal to `document.location.pathname` (e.g: the current URL).

alex-ju: cool, even better

app/assets/javascripts/modules/cookie-settings.js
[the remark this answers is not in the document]
vanitabarrett: @alex-ju I've got a PR here for adding the ability to add an `aria-live` attribute to the notice component https://github.com/alphagov/govuk_publishing_components/pull/911

vanitabarrett:
I've now added the `aria_live: true` flag to the notice component. https://github.com/alphagov/frontend/pull/1856/files#diff-e394cdb6571fb405da3283b4f5d5cfb8R7

This won't come into effect until `govuk_publishing_components` has been versioned and bumped in this PR, but I'm holding off doing that for now until we're ready to go live as we have more changes in the gem which we need before going live anyway. cc @alex-ju

vanitabarrett:
[GO LIVE](https://trello.com/c/TG1lBZbk/168-go-live)

PR #1857, opened 2019-06-04
Adjust departments list alignment on gov.uk homepage

kr8n3r:
Adjust departments and bodies list alignment using govuk-frontend styling.

IMHO looks better this way. Preview: https://govuk-frontend-app-pr-1857.herokuapp.com/

**Before:** 
Screen Shot 2019-06-04 at 16 16 16

Screen Shot 2019-06-04 at 16 16 35

**After:**

Screen Shot 2019-06-04 at 16 16 03

Screen Shot 2019-06-04 at 16 15 45

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -362,6 +362,11 @@ body.homepage {
         }
       }
     }
+    
+    .large-numbers__list {
+      padding-left: 0;
+    }
+
     .departments-promo-sections {
       @extend %contain-floats;
       clear: both;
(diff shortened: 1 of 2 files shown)


miaallers:
I think this change makes sense, looks good to me!

kr8n3r:
@nickcolley it's a reset

> Makes sense to me 👍
> 
> You might want to consider some custom CSS instead of using the frontend class, just cos this isn't really the use case for that class.

yeah, better. updated.

nickcolley:
@kr8n3r epic nice work 👏

PR #1876, opened 2019-06-20
Reduced image size 'UK leaves the EU'

andysellick:
- homepage JPG size was nearly 100KB, compressed without image quality loss to 19KB

Screen Shot 2019-06-20 at 08 45 24

Thanks to @Nooshu for spotting this and supplying the new image.


andysellick:
The images either side of this one are PNG files, but the smallest PNG I can get of this image is 38KB.

PR #1884, opened 2019-06-25
[WIP] Update text for organ donor promotion

thomasleese:
This updates the text for the organ donor promotion to include the new
separate opt in and opt out links.

It's backwards compatible with the old promotions which have only a
single `url` key, and this is treated as the opt in URL.

Depends on https://github.com/alphagov/publisher/pull/1088 and https://github.com/alphagov/govuk-content-schemas/pull/910.

[Trello Card](https://trello.com/c/HFsNGxTy/1088-investigate-organ-donation-promos-on-done-pages)


thomasleese:
It looks like we won't need this anymore.

PR #1913, opened 2019-07-25
Update Statistics link on homepage

hannako:
Update statistics link on homepage to point to the shiny new finder `search/research-and-statistics`

https://trello.com/c/EEJupWMu/516-replace-stats-link-on-homepage


hannako:
@barrucadu I've got the brakeman error you had earlier https://github.com/alphagov/specialist-publisher/pull/1471. Shall I follow the same steps to ignore it?

barrucadu:
I don't think there are any cookies here we care about breaking, but yes, best to be safe.

PR #1914, opened 2019-07-25
Ignore warning for cookie type

hannako:
Brakeman wants us to use :json and not :marshal or :hybrid, but
switching straight to :json from :marshal breaks existing sessions.

ref: https://github.com/alphagov/specialist-publisher/pull/1471


cbaines:
What cookies is this app setting? I can't think of any cookie based interaction the Rails bit's of frontend would have?

PR #1927, opened 2019-08-19
Upgrade frontend component gem

andysellick:
Upgrade components gem to 20, which includes the breaking changes from V3 of `govuk-frontend`.

Much of the code still needs converting to use components and govuk-frontend instead of custom CSS in the app, but that can be done another day. The objective of this PR was to get the new gem installed without errors and ensure that the homepage looks good.

Some of the pages have inconsistent focus styles (new alongside old). This is due to some of these elements coming from static.

## Visual changes

**Tour page** - complete rebuild. Videos and layout were broken on live. Fixed by using components for the video, details component for transcripts and search component for search box. Using govuk-frontend grid, headings, paragraphs and link styles. Removed all custom CSS and JS for this page.

Before: 

Screen Shot 2019-08-20 at 09 25 46

After:

Screen Shot 2019-08-20 at 09 26 00

**Help** - rebuilt with govuk-frontend layout, some minor visual changes. Before:

Screen Shot 2019-08-20 at 09 45 57

After:

Screen Shot 2019-08-23 at 14 36 57

**Travel advice page** - no visual changes to this page apart from link focus styles. There is a lot more we could do to improve this page (using govuk-frontend and removing custom styles) but I don't have time to do it now and the page renders okay.

**Find your local council** - bit of a rebuild. Used components for the search input, button and error message. Unfortunately the hint text in the input component failed contrast requirements, so I've changed the background of the box as well. 

You might notice on this page that the input is a bit misaligned, this is due to styles coming through from static. As it doesn't cause a huge problem (it's still usable and looks okay) and we'll be updating static fairly soon anyway, I haven't put override styles in to fix this.

Unfortunately I've not been able to get `mapit` up and running locally so I can't access the results page that follows this one. If anyone is able to get that running locally please let me know - the page on live looks okay but eventually it would be good to strip out some more of the custom styles for govuk-frontend.

Before:

Screen Shot 2019-08-22 at 08 56 00

After:

Screen Shot 2019-08-22 at 08 56 08

Pages to sanity check:

- https://govuk-frontend-app-pr-1927.herokuapp.com/settle-in-the-uk (simple smart answer)
- https://govuk-frontend-app-pr-1927.herokuapp.com/register-to-vote (transaction start page)
- https://govuk-frontend-app-pr-1927.herokuapp.com/ (homepage)
- https://govuk-frontend-app-pr-1927.herokuapp.com/tour (tour of GOV.UK)
[...]

PR #1939, opened 2019-08-30
Remove govuk_frontend_toolkit dependency

andysellick:
This PR removes the `frontend` Sass dependency on `govuk_frontend_toolkit`. This has involved significant HTML and CSS restructuring to some parts of the application, but hopefully minimal visual changes (if any).

(was doing this as part of https://github.com/alphagov/frontend/pull/1927 but it's turned into too big a bit of work, so splitting it out to do first)

I want to do more changes to frontend, because there's more to be rebuilt using `govuk-frontend`, but this PR seems big enough for now so I'll do those later (some have been done as part of 1927 already). The main focus of this PR is updating mixins and rebuilding the home page.

## Homepage desktop before

image

## Homepage desktop after

The black column in the header is slightly wider and in a slightly different position. This is due to the fact that this element is absolutely positioned and the containers have changed slightly. I don't think this is worth the effort required to fix it (it's a few pixels)

image

## Homepage mobile before

image

## Homepage mobile after

Should be the same, but please check.

image


app/assets/stylesheets/views/_travel-advice.scss
```diff
@@ -238,64 +230,10 @@
...
-      float: none;
-      text-align: left;
-      width: auto;
-
-      @include media-down(mobile) {
-        padding-left: 1em;
-      }
-    }
+  .page-navigation {
+    border-bottom: 1px solid #bbb;
```
alex-ju: Is there a reason for not using a standard colour for this (e.g. `$govuk-border-colour`)?

andysellick: Updated.

andysellick:
@alex-ju was all ready to merge and then checked it in IE8. We've been having a conversation this week about support for older browsers (some 40,000 people are still on <= IE8) and govuk-frontend's media queries no longer support IE8 and below, meaning that this change will take those users from this:

Screen Shot 2019-09-05 at 11 19 56

to this:

Screen Shot 2019-09-05 at 11 20 10

(media queries are ignored, so they get the mobile version, but without the max-width on the container, so it all collapses left)

Any thoughts?

alex-ju:
@andysellick looks quite bad. I would add `max-width` and `width: auto` (and other specific CSS if needed) in the IE8 stylesheet.

andysellick:
@alex-ju I've added some IE specific styles to make the homepage look more like it should. I've also finally figured out why compiling all this stuff takes so long - see the last commit in this PR (we could adopt this on other repos).

alex-ju:
Good find @andysellick! 👏 Good to be merged when e2e tests are passing.

andysellick:
@alex-ju have also realised that all the `$ie` variables in those stylesheets are now unnecessary, so have removed.

PR #1940, opened 2019-08-30
[DO NOT MERGE] Remove IE specific stylesheets

andysellick:
Removing the IE6 and IE7 stylesheets as we don't support these browsers anymore.

Removing the IE8 stylesheet is more complicated, so I've moved that to [another PR](https://github.com/alphagov/frontend/pull/1941) for now - the aim of these changes is to reduce the local sass compile time for developers.


andysellick:
I'm uncertain about removing support for these browsers on something as significant as the homepage. Closing for now, will discuss.

PR #1959, opened 2019-09-09
Use chevron banner component

vanitabarrett:
## What
Updating the hardcoded chevron banner in Frontend with the chevron banner component in the govuk_publishing_components gem.

## Why
The chevron banner has been hardcoded in Frontend. We should use the component instead.

Screen Shot 2019-09-09 at 10 07 33

https://govuk-frontend-app-pr-1959.herokuapp.com/


vanitabarrett:
@andysellick Yep, the component is set to take up 100% width of the container it's placed in.

PR #1965, opened 2019-09-13
Add basic auth to the review app

sihugh:
This has been lifted from [Collections](https://github.com/alphagov/collections/blob/master/app/controllers/application_controller.rb#L11-L16)

The env variables have been added to the review app already.

This is to prevent users from accidentally landing on a page in the review
app and believing that it's GOV.UK.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,6 +10,13 @@ class ApplicationController < ActionController::Base
   rescue_from GdsApi::InvalidUrl, with: :cacheable_404
   rescue_from RecordNotFound, with: :cacheable_404
 
+  if ENV["BASIC_AUTH_USERNAME"]
+    http_basic_authenticate_with(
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -10,6 +10,13 @@ class ApplicationController < ActionController::Base
   rescue_from GdsApi::InvalidUrl, with: :cacheable_404
   rescue_from RecordNotFound, with: :cacheable_404
 
+  if ENV["BASIC_AUTH_USERNAME"]
```
karlbaker02: Should this be using `ENV["REQUIRE_BASIC_AUTH"]` instead?

sihugh: I've removed the `REQUIRE_BASIC_AUTH` env var from app.json now, it wasn't set anywhere and is not much use if the username and passwords aren't set.  Thanks

PR #1984, opened 2019-09-30
Remove govuk_frontend_toolkit from frontend

andysellick:
Removes the final traces of `govuk_frontend_toolkit` from frontend. Most of the legwork had already been done in previous PRs, this PR replaces a few variables with govuk-frontend variables, removes some unused code, and takes out the toolkit Gemfile dependency.


vanitabarrett:
Should we be replacing the `grid-row` class in `application.html.erb` with `govuk-grid-row`? https://github.com/alphagov/frontend/blob/master/app/views/layouts/application.html.erb#L33

vanitabarrett:
Does this flag still need to be set: `$govuk-compatibility-govukfrontendtoolkit: true;` ?

andysellick:
@vanitabarrett I've addressed those changes.

PR #1988, opened 2019-10-04
Fix layout with govuk-frontend grid

andysellick:
Looks like I missed a few pages when I removed toolkit recently - fortunately this hasn't been deployed to live yet.

- fixes simple smart answers (e.g. /settle-in-the-uk), find local council (/find-local-council) and content pages (e.g. /change-driving-test)

Example of the broken:

Screen Shot 2019-10-04 at 09 13 35

Is fixed in this PR to look like this:

Screen Shot 2019-10-04 at 09 13 48

### DIFF
--- a/app/views/find_local_council/_base_page.html.erb
+++ b/app/views/find_local_council/_base_page.html.erb
@@ -4,13 +4,13 @@
         content="Find your local authority in England, Wales, Scotland and Northern Ireland" />
 <% end %>
 
-<main id="content" role="main">
+<main id="content" role="main" class="govuk-grid-column-two-thirds">
   <header class="page-header">
     <%= render "govuk_publishing_components/components/title", title: "Find your local council" %>
   </header>
   <%= yield %>
 </main>
 
-<div class="related-container">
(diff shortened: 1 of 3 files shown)


app/views/find_local_council/_base_page.html.erb
```diff
@@ -4,13 +4,13 @@
         content="Find your local authority in England, Wales, Scotland and Northern Ireland" />
 <% end %>
 
-<main id="content" role="main">
+<main id="content" role="main" class="govuk-grid-column-two-thirds">
```
36degrees: Do these need to be [inside a `div.govuk-grid-row`](https://design-system.service.gov.uk/styles/layout/#common-layouts), or is that already catered for in the layout?

(It looks right from the screenshots, I just can't see where the row is coming from)

alex-ju: I checked that too. It seems like it is (probably the grid-row class lives in the layout (or whatever outer view we have). Would be good to align the layout for all apps (potentially using https://govuk-frontend-review.herokuapp.com/full-page-examples/start-page as a reference).

PR #2001, opened 2019-10-17
Rescue 403 errors from content-store

leenagupte:
Trello: https://trello.com/c/BFSkOY22
Follows on from: https://github.com/alphagov/collections/pull/1324

## What's changed and why?
At the moment frontend is throwing a 500, when it receives an error code it doesn't recognise
from content-store. That means that users are shown a "Problem has occurred" page.

We need to handle this error properly so that in the case of a 403, the user will be properly routed
to signon and asked to login if there is any form of access limiting on the content item (e.g. hosted on draft stack, access limited to organisation etc)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,6 +8,7 @@ class ApplicationController < ActionController::Base
   rescue_from GdsApi::HTTPGone, with: :error_410
   rescue_from GdsApi::HTTPNotFound, with: :cacheable_404
   rescue_from GdsApi::InvalidUrl, with: :cacheable_404
+  rescue_from GdsApi::HTTPForbidden, with: :error_403
   rescue_from RecordNotFound, with: :cacheable_404
 
   if ENV["BASIC_AUTH_USERNAME"]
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -8,6 +8,7 @@ class ApplicationController < ActionController::Base
   rescue_from GdsApi::HTTPGone, with: :error_410
   rescue_from GdsApi::HTTPNotFound, with: :cacheable_404
   rescue_from GdsApi::InvalidUrl, with: :cacheable_404
+  rescue_from GdsApi::HTTPForbidden, with: :error_403
```
ChrisBAshton: Nitpick - I'd move this to be between lines 8 and 9, so that the status codes are in ascending order.

leenagupte: I considered it, but thought I'd just leave the status codes in descending order, so 503, 410, 404, 403.

ChrisBAshton: Makes sense 👍

ChrisBAshton:
Works!

PR #2002, opened 2019-10-17
Do not select cookie options in setting form until user manually submits choices

### LINKED ISSUE #2007: Cookie page container misaligned
Content on the page seems to be missing a left and right margin - so the content is not the same width as the blue bar at the top of the page.

Adding 15px ([three on the GOV.UK Frontend scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale)) of left and right padding to the `main` element with an id of `content` fixes it.

Screenshot 2019-10-21 at 11 46 12

maxgds:
### Requirements

- Deselect radio button defaults but warn user in text that we
  are saving cookies
- On successful submit (option chosen for each item) save a
  new essential cookie (added in govuk_publishing_components).
  Show an error message if not all radios are selected.
- On detecting the cookie, select the user's choices in the form

Also fixes #2007.

### Before
image

### After
image

### New content

Warning:
image

Error:
image

Success:
image


chao-xian:
@maxgds please note that https://github.com/alphagov/frontend/pull/2002/commits/005125a2e97ed2a5c775d0aad3f29823564512c8 were added to the wrong place, and needs to be included in the form.

vanitabarrett:
Not your change, but I've noticed the page width seems to be larger than our standard page width. The page content is not in line with the blue header border or the feedback survey at the bottom, which other pages normally are. I think this is because we're wrapping the page in a `govuk-grid-row` but not an actual grid-column

app/views/layouts/application.html.erb
```diff
@@ -25,9 +25,7 @@
 
   <body class="mainstream <%= yield :body_classes %>">
     <div id="wrapper" class="<%= wrapper_class(@publication || @presenter) %>">
-      <% unless current_page?(root_path) %>
-        <%= render 'govuk_publishing_components/components/contextual_breadcrumbs', content_item: @content_item %>
-      <% end %>
+
```
maxgds: This was a mistake - only removed for dev purposes, it needs to be restored.

app/views/help/cookie_settings.html.erb
```diff
@@ -1,6 +1,20 @@
 <% content_for :title, "Cookies on GOV.UK" %>
 
 <main id="content" role="main" class="group help-page cookie-settings">
+
```
maxgds: Need to add aria-hidden when appropriate.

app/views/layouts/application.html.erb
injms: Restored ✅

app/views/help/cookie_settings.html.erb
injms: This doesn't seem to be an issue - so, bearing in mind the [first rule of ARIA](https://www.w3.org/TR/using-aria/#rule1), I've held off from adding it. I'll dig into this a bit deeper to, but I don't want this to be a blocker.

app/assets/javascripts/modules/cookie-settings.js
```diff
@@ -43,27 +46,41 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
     }
+    this.hideWarningMessage()
 
+    var errorMessage = document.querySelector('div[data-cookie-error]')
+    if(errorMessage !== null && errorMessage.style.display == "block") {
+      this.hideErrorMessage()
+    }
     this.showConfirmationMessage()
 
+
```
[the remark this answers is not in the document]
injms: Removed ➖

app/assets/javascripts/modules/cookie-settings.js
```diff
@@ -21,20 +21,23 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
     delete currentConsentCookieJSON["essential"]
 
-    for (var cookieType in currentConsentCookieJSON) {
-      var radioButton
-
-      if (currentConsentCookieJSON[cookieType]) {
-        radioButton = document.querySelector('input[name=cookies-' + cookieType + '][value=on]')
-      } else {
-        radioButton = document.querySelector('input[name=cookies-' + cookieType + '][value=off]')
+    // If user has selected options previoiusly re-select them, otherwise they must be left blank
```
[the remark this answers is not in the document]
injms: Corrected 👍

chao-xian:
Once this https://github.com/alphagov/frontend/pull/2031 is merged in, this is good to go.

PR #2011, opened 2019-10-23
Add basic funding form controller

thomasleese:
We will use this controller as the basis for the funding form.


app/views/funding_form/index.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render layout: 'base_page' do %>
```
brucebolt: I think there needs to be a `_base_page.html.erb` partial file in either `app/views/funding_form` or `app/views/application`

thomasleese: Oh yep, I copied the index from a different format which had that file as well... I've amended it without any partials now.

PR #2016, opened 2019-10-23
Change path for EU Funding Registration form

brucebolt:
We used `/funding-form` as a placeholder.  This should be changed to `/brexit-eu-funding`.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -64,7 +64,7 @@ Rails.application.routes.draw do
   # Funding Form pages
   get "/brexit-eu-funding" => "funding_form#index"
   get "/brexit-eu-funding/organisation-name" => "funding_form/organisation_name#show"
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -62,9 +62,9 @@
   end
 
   # Funding Form pages
-  get "/funding-form" => "funding_form#index"
-  get "/funding-form/organisation-name" => "funding_form/organisation_name#show"
-  post "/funding-form/organisation-name" => "funing_form/organisation_name#submit"
+  get "/register-eu-funding-form" => "funding_form#index"
+  get "/register-eu-funding-form/organisation-name" => "funding_form/organisation_name#show"
+  post "/register-eu-funding-form/organisation-name" => "funing_form/organisation_name#submit"
```
thomasleese: Not part of this PR, but:

```suggestion
  post "/register-eu-funding-form/organisation-name" => "funding_form/organisation_name#submit"
```

PR #2018, opened 2019-10-23
Add organisation type page

alex-ju:
Use the existing organisation name template (as it will not be needed)
 - The form posts to the same page - remaining for us to decide what's going to happen after posting.
 - It uses the .yml file for content declarations to avoid repetition and make it easier to change by content designers.
 - Prevents contextual breadcrumb from trying to be rendered in the main layout file (as it required a `@publication` object that we don't currently have).

[Preview on Heroku](https://govuk-frontend-app-pr-2018.herokuapp.com/funding-form/organisation-type) or check it locally at: `/funding-form/organisation-type`

[Trello card](https://trello.com/c/S6Fza3YQ)


config/routes.rb
```diff
@@ -63,8 +63,8 @@
 
   # Funding Form pages
   get "/funding-form" => "funding_form#index"
-  get "/funding-form/organisation-name" => "funding_form/organisation_name#show"
-  post "/funding-form/organisation-name" => "funing_form/organisation_name#submit"
+  get "/funding-form/organisation-type" => "funding_form/organisation_type#show"
+  post "/funding-form/organisation-type" => "funing_form/organisation_type#submit"
```
brucebolt: Looks like my error in a previous PR, but this line has `funing_form` instead of `funding_form`.  Probably a good time to fix that.

app/views/funding_form/organisation_type.html.erb
```diff
@@ -0,0 +1,50 @@
...
+              text: t('funding_form.organisation_type.options.research_and_technology.label'),
+              hint_text: t('funding_form.organisation_type.options.research_and_technology.hint'),
+              bold: true
+            },
+            {
+              value: t('funding_form.organisation_type.options.public_sector_or_charity.label'),
+              text: t('funding_form.organisation_type.options.public_sector_or_charity.label'),
+              hint_text: t('funding_form.organisation_type.options.public_sector_or_charity.hint'),
+              bold: true
+            },
```
alex-ju: I'm missing the 'other' option, with the conditional reveal.

alex-ju:
> Looks great, but when I run the code locally, I'm not seeing the `Other` option :/

Yep, I was missing that - I figured that this morning.

Updated now with a sketch of a feature test which will hopefully make more sense when we have at least two pages and a redirect

PR #2026, opened 2019-10-24
Add FAQ page schema for transaction pages

sihugh:
This is to expose Register to Vote content in a way that can be used more easily by external consumers.  The content has been recently reworked to include more things that users are searching for around this topic.

We can use a slightly more minimal schema on these pages because the Article schema is still present.

This is only to be activated on /register-to-vote initially, though there's no reason to not expand this, should research support that it's effective.

The code to extract questions and answers from HTML in the content item has been lifted from [the FAQPage schema used in guides](https://github.com/alphagov/govuk_publishing_components/blob/a5b0342aa6529c10ce339b9047fb09e7c13f933a/lib/govuk_publishing_components/presenters/machine_readable/faq_page_schema.rb#L45-L76). It has been tweaked from the guide implementation of the FAQPage schema in that we're ignoring preamble text in the page (that is, content that is not preceded by an H2 heading), rather than using a default "question" for that text.

We're generating the schema here so that we can iterate quickly without the need to republish the publishing components gem. 

Testing https://govuk-frontend-app-pr-2026.herokuapp.com/register-to-vote in the tools we have gives valid responses:
 
## Rich results tool

https://search.google.com/test/rich-results?view=search-preview&id=KR0JOqFXBoWK1sdA6MsUng

Screenshot 2019-10-24 10 47 28

## Structured data tool 

Screenshot 2019-10-24 10 46 49


app/presenters/machine_readable/transaction_faq_page_schema.rb
```diff
@@ -0,0 +1,156 @@
...
+
+  def self.build(transaction)
+    if CONTENT_IN_SCOPE.include?(transaction.content_id)
+      new(transaction)
+    end
+  end
+
+  def initialize(transaction)
+    @transaction = transaction
+  end
```
bilbof: You could move methods + constants like `initialize` and `CONTENT_IN_SCOPE` under `private`.

sihugh:
> Do you have any ideas on how to package the content generation logic (e.g. `question_and_answers`), is that something for the publishing components gem?

It could be - we're probably just about to bin it for guides though, so it might be a bit soon to work out where it goes.  Good thought though 👍

app/presenters/machine_readable/transaction_faq_page_schema.rb
sihugh: Good shout, have done so.

PR #2031, opened 2019-10-24
Pin sass-rails to prevent Sass compilation error

injms:
## What
Pin Sass-Rails to a version less than 6.

## Why
Sass-Rails v6 is a wrapper around SassC-Rails. SassC-Rails uses LibSass to compile the Sass files to CSS files. LibSass doesn't support CSS `calc()`, which is used in GOV.UK Frontend. Pinning Sass-Rails to v5 means that CSS `calc()` is supported and won't throw an error.

More information on SassC-Rails and why this throws an error be found on [this issue on `finder-frontend`](https://github.com/alphagov/finder-frontend/issues/1124), and [this issue on `whitehall`](https://github.com/alphagov/whitehall/issues/4724).


nickcolley:
@injms have you raised this issue with the sassc-rails folks?

PR #2032, opened 2019-10-24
Added what programme funding is from form

RobBond-GDS:

Specifically this view is to allow selection of the body that the grant recipient receives their funding from, from a specified list

trello card - https://trello.com/c/bMJh0VOu


andysellick:
@RobBond-GDS can you provide more detail in this PR description please? Linking to a trello card is good, but not everyone has access to that particular board.

PR #2058, opened 2019-10-25
Concatenate address into a single column

thomasleese:
This is currently how they're importing the row into the spreadsheet.


brucebolt:
There's a possibility not all address lines will be populated and there could be additional spaces/commas.  Would this be an issue for the department?

thomasleese:
@brucebolt Good point, I've updated this to only combine the fields which were entered.

PR #2079, opened 2019-10-27
Update count of agencies and public bodies to 407

huwd:
I thought this looked a bit weird,

On the [home page](https://www.gov.uk/) scroll down to the `<section>` `departments-and-policy` and we tell users there are 405 currently.

Screenshot 2019-10-27 at 10 09 09

Click though on that number and immediately you see a different number...

Screenshot 2019-10-27 at 10 10 08

This PR brings them back into alignment.

As an aside, I expect this number changes frequently. I'm not quite sure how best to deal with that. I was thinking of writing an e2e test that stores this number, clicks through and then assets equal with the number on whitehall?

Is there a good reason why this number doesn't dynamically update already?
Happy to open an issue or extend an existing one if this discussion has already been had?


huwd:
@ChrisBAshton ah ha excellent!
I'll check that out

PR #2085, opened 2019-10-31
added back links to relevant pages

RobBond-GDS:

Specifically, adding in of "< back" link navigation with the transactional views on the journey

Trello card - https://trello.com/c/opssAUTV


app/views/funding_form/contact_information.html.erb
```diff
@@ -4,6 +4,9 @@
 <% end %>
 
 <div class="govuk-width-container">
+  <%= render "govuk_publishing_components/components/back_link", {
+    href: "https://www.gov.uk/guidance/register-as-an-organisation-getting-funding-directly-from-the-eu"
```
alex-ju: This looks good, but I'm thinking we should use [Plek](https://github.com/alphagov/plek) to dynamically generate these paths so that they work in each environment instead of hardcoding the production URL. I would also use the same technique for the rest of the "back" links instead of using relative ones.

alex-ju: I've added a commit to replace the hardcoded URLs with locally defined routes - as Plek requests are not necessary within the app.

PR #2087, opened 2019-10-31
Pep content

huwd:


app/views/homepage/index.html.erb
```diff
@@ -34,30 +34,7 @@
...
-            }
-          %>
-        </p>
-      </div>
-    </div>
-
-    <hr class="home-divider" aria-hidden="true" />
-
-    <section class="home-services" aria-labelledby="services-and-information-label" id="services-and-information">
+    <section class="home-services homepage__ready-container" aria-labelledby="services-and-information-label" id="services-and-information">
```
vanitabarrett: I'd suggest not using the `homepage__ready-container` class here, as it's a bit confusing (was previously used for the chevron section, and now we're moving it to something which isn't "getting ready for Brexit"). I'd suggest tweaking the `home-services` class padding instead until it matches what we want.

huwd: amended

PR #1891, opened 2019-07-01
Fix typo in CONTRIBUTING.md

mpsq:

PR #2091, opened 2019-11-05
Fix typo in CONTRIBUTING.md

issyl0:
This is from #1891 but I had to make it a new PR so that CI checks would run.


cbaines:
> This is from #1891 but I had to make it a new PR so that CI checks would run.

I believe you just have to push the commit to this GitHub repository (the one Jenkins looks at), you don't need to create a 2nd pull request.

issyl0:
It needed a rebase and you can't do that with forked branches that don't exist any more - the force-push doesn't show up on the original PR.

PR #2096, opened 2019-11-06
Update home top links title size

alex-ju:
It was initially 14px, then updated to 24px (which is too big), now 19px — which is in line with the rest of the text in the header.

### Before
Screen Shot 2019-11-06 at 10 05 11

### After
Screen Shot 2019-11-06 at 10 06 06


issyl0:
Thanks and sorry for the hassle!

alex-ju:
no worries @issyl0, thank you for cleaning the stale PRs and address issues across alphagov

PR #2105, opened 2019-11-18
Swap FAQ page implementation

sihugh:
We'd originally implemented the FAQPage schema generator to automatically split the page by section and render each heading as an FAQPage `question`, and the following text as an `answer`.

Unfortunately, this doesn't quite work for the way we want to use this. Sometimes the content that provides an `answer` spans a couple of sections.  In this case, we want to be able to include the content from more than one heading in the `answer`.

We also want to occasionally omit some text from the FAQ page schema, because it might be content that helps the page read well, but doesn't add anything when split into smaller sections.

Because of this, we've removed the `transaction_faq_page_schema_presenter` and replaced it with one that reads from a yaml file.  (Hooray! :-/ )

The `yaml_faq_page_schema_presenter.rb` has been lifted from [government-frontend](https://github.com/alphagov/government-frontend/blob/master/app/presenters/machine_readable/yaml_faq_page_schema_presenter.rb) where it's fulfilling a similar job for another piece of content.  After this period is over, we'll evaluate the use of this and see whether we think it's worth leaving as is (unlikely), promoting to a shared library or burning with fire.

The page is https://govuk-frontend-app-pr-2105.herokuapp.com/register-to-vote (this is the only page in this app to use the FAQPage schema at present)

The rich results preview is here: https://search.google.com/test/rich-results?view=search-preview&id=YcpJvBAL5qAeHDOeU82ryQ (This matches the marked up content I've been given)


sihugh:
> Only thing is when I viewed google's rich text preview, only the external links render, ie the gov.uk links don't. Do we need to use the full urls in the yaml rather than relative ones?

That seems to be a "feature" of the preview tool.  They work fine in real life - which is good, as I want to keep the YAML as close to the markup on live as possible otherwise it gets impossible to manage.  Thanks for checking!

PR #2113, opened 2019-11-20
Fix conditional reveal for EU Funding Form

brucebolt:
There were a number of issues around the conditional reveal questions in the EU Funding Form:
- "Yes" was always pre-selected with a blank value in the text field when the user had not previously answered the question
- If the user left the field blank but returned to check their answers, the text field would be populated with the word "Yes"
- If the user selected "No" but returned to check their answers, the free text field would be populated with the word "No"

This resolved the issue by storing the boolean value separately from the free text response.  Additionally, the logic has been fixed for determining whether to check the boxes, and whether to display text value.

Trello card: https://trello.com/c/QmvBxA8k


brucebolt:
Yes, we can.  I've pushed an update.

PR #2120, opened 2019-11-22
[Do not merge] Increase the timeout for licence applications

leenagupte:
Requests to licence_application are timing out in licensify. This means that the licence application forms are not loading
on pages like: `https://www.gov.uk/premises-licence/birmingham`

After some investigation it appears that the request to licensify is taking longer than the default 4 seconds.
This increases the timeout whilst the underlying problem is being investigated.

### DIFF
--- a/app/controllers/licence_controller.rb
+++ b/app/controllers/licence_controller.rb
@@ -48,7 +48,7 @@ private
     return {} if @publication.continuation_link.present?
 
     begin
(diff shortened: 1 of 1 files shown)


leenagupte:
It looks like increasing the number of CPUs on the database instances seems to have fixed the problem. Keeping this PR open for a few days whilst investigations continue.

PR #2125, opened 2019-11-26
New colours on homepage

injms:
## What
Changes to the homepage to make it ready for the new higher contrast colour scheme. This has been checked with the new colours turned on, but since the colours are a global setting across all of Frontend they've been turned off. Once all of the views have been checked with the new colours this setting can be turned on.

Also added in lazy-loading of images, since it's a trivial change to make with a positive impact.

## Why
To allow GOV.UK to use the more accessible, higher contrast colour scheme.

Trello card: https://trello.com/c/wIvM9an9/113-switch-to-new-colours-in-frontend

### Large numbers before:
Screenshot 2019-11-26 at 14 17 52

### Large numbers after:
Screenshot 2019-11-26 at 14 17 44

---

### Promo section before:
Screenshot 2019-11-26 at 14 21 38

### Promo section after:
Screenshot 2019-11-26 at 14 21 48

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,5 +1,7 @@
 $govuk-compatibility-govuktemplate: true;
 
+$govuk-use-legacy-palette: false;
+
 // BASE Stylesheet
 
 // Components from govuk_publishing_components gem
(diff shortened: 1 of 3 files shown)


injms:
> This looks fine to go but I didn't think this kind of work was the scope of stage 2. The focus areas are big and overwhelming in several places, but changing the colour doesn't alter that. I thought our aim was simply to ensure the entire page gets the right colour when the flag is flipped, get the high contrast out asap and then revisit ugly design, hopefully switching in components to fix many examples of it. This is absolutely an improvement, but one to handle later, I thought.

I completely got carried away on this one - I'll do better and stick to the scope in the future.

PR #2127, opened 2019-11-26
New colours on travel advice

injms:
## What
Changes to the foreign travel advice page with the list of countries to make it ready for the new higher contrast colour scheme. 

The biggest visual difference is that the text size for the countries has been increased with the use of `govuk-body-s`. This should make it more readable.

The input and subscription links are now components, as the hardcoded versions before weren't picking up the new focus states.

I ran Standard lint with the `--fix` option for the JavaScript that shows and hides the list of countries, which is a reasonably big diff with no real change. A couple of things, such as a purposeful use of `==` instead of `===`, meant that the linting needed to be turned off for specific lines.

This has been checked with the new colours turned on, but since the colours are a global setting across all of Frontend they've been turned off. Once all of the views have been checked with the new colours this setting can be turned on.

## Why
To allow GOV.UK to use the more accessible, higher contrast colour scheme.

Trello card: https://trello.com/c/wIvM9an9/113-switch-to-new-colours-in-frontend

## Visual changes

Before:
image

After:
image

---

Before:
image

After:
image

---

Before:
image

After:

image

### DIFF
--- a/app/assets/stylesheets/views/_travel-advice.scss
+++ b/app/assets/stylesheets/views/_travel-advice.scss
@@ -1,253 +1,78 @@
 .travel-container {
-  .list {
-    ul {
-      display: inline-block;
-      padding-left: 0;
-      width: 100%;
-      border-bottom: 1px solid #ccc;
-      padding-bottom: 20px;
+  .country-filter-form {
+    display: none;
+
+    .js-enabled & {
+      display: block;
     }
-    ul.countries {
-      li {
-        width: 49%;
-        float: left;
-        margin-right: 1%;
+  }
+
+  .country-filter-form__form-group {
+    margin: 0; // Because fieldset has 2px margins by default.
+  }
+
+  .subscriptions-wrapper {
+    @include govuk-media-query($from: tablet) {
+      .js-enabled & {
+        margin-top: 22px;
+        text-align: right;
       }
     }
   }
 
-  .section-wrap {
-    padding: 0 0 0;
-    display: table-row;
-
-    @include media-down(tablet) {
-      display:block;
-      border-bottom: 1px solid $govuk-border-colour;
-    }
-
-    section {
-      width:50%;
-      vertical-align:bottom;
-      display:table-cell;
-      border-bottom: 1px solid $govuk-border-colour;
-
-      @include media-down(tablet) {
-        display:block;
-        border-bottom: none;
-        width:100%;
-      }
-    }
-
+  .app-full-width-break {
+    clear: both;
+    margin-right: govuk-spacing(3);
+    margin-left: govuk-spacing(3);
   }
 
-  .subscriptions {
-    text-align:right;
-
-    ul li:first-child {
-      margin-right: 0.3em;
-    }
-  }
-}
-
-.travel-advice {
-    .full-width .page-header div {
-      margin-right: 0em;
-    }
-
-    .list {
-      ul {
-        padding-left: 0;
-      }
-      ul.countries {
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/views/_travel-advice.scss
```diff
@@ -1,253 +1,88 @@
...
-  .country-metadata {
-    padding: 0;
-    margin: 0;
-    color: $govuk-secondary-text-colour;
+  .countries-list {
+    border-bottom: 1px solid $govuk-border-colour;
+    display: block;
+    padding-bottom: govuk-spacing(4);
 
     li {
```
andysellick: Can this be a class rather than an element selector?

app/assets/stylesheets/views/_travel-advice.scss
```diff
@@ -1,253 +1,88 @@
...
-    }
-
-    @include media-down(tablet) {
-      text-align: left;
-      width: auto;
-      padding-top: 0;
-
-
+  .countries-wrapper {
+    h2 {
```
andysellick: Can this be a class as well?

app/assets/stylesheets/views/_travel-advice.scss
```diff
@@ -1,253 +1,88 @@
...
-
-        &.email-alerts {
-          background-image: image-url("travel-advice/mail-icon.png");
-          @include govuk-device-pixel-ratio() {
-            background-image: image-url("travel-advice/mail-icon-x2.png");
-            background-size: 20px 14px;
-          }
-        }
-      }
+    h3 {
```
andysellick: Can this be a class as well?

app/assets/stylesheets/views/_travel-advice.scss
```diff
@@ -1,253 +1,88 @@
...
-  .page-navigation {
-    border-bottom: 1px solid $govuk-border-colour;
-    margin: 0 0 2.5em;
-    padding-bottom: 1em;
-  }
+// TODO: This #wrapper.travel-advice turns off the default margin and width,
+// which now isn't needed thanks to the grid in GOV.UK Frontend.
+#wrapper.travel-advice {
+  margin: 0;
+  max-width: none;
```
andysellick: Can you explain why this is here? I had a quick look and turning off these styles doesn't seem to make any difference, not sure the comment is clear enough.

app/views/travel_advice/index.html.erb
```diff
@@ -3,55 +3,71 @@
...
-          <% end %>
-        </div>
-      </section>
-    </div>
+    <section class="js-countries-wrapper countries-wrapper govuk-grid-row">
+      <div class="countries govuk-grid-column-one-third">
+        <h2 class="govuk-heading-m">Countries or territories</h2>
+        <p class="country-count js-country-count">
+          <span class="js-filter-count filter-count "><%= @presenter.countries.length %></span>
+          <span class="visuallyhidden">Countries or territories</span>
```
andysellick: `govuk-visually-hidden` ?

app/assets/stylesheets/views/_travel-advice.scss
injms: Good catch! It's a hangover from a previous step on this branch that I forgot to remove - gone now.

app/views/travel_advice/index.html.erb
injms: There's always one that gets away...

Changed to `govuk-visually-hidden`.

PR #2129, opened 2019-11-27
Update service URL on register to vote schema

sihugh:
This matches the updated service url on [register to vote](https://www.gov.uk/register-to-vote)


sihugh:
It turns out that this URL causes problems with sessions - the current one is correct 👍

PR #2139, opened 2019-12-05
Update local transaction views

injms:
## What

The back link on the local transaction results page was updated to be a component from GOV.UK Publishing Components.

 - [Trello card](https://trello.com/c/YXG9Fe6l/136-frontend-local-transaction-colours-checked)

## Why
The local transaction views needed updating to make sure that the newest focus states were being used on all links on it - the back link was the only link that wasn't picking up the new focus state.

## Where

 - https://www.gov.uk/pay-council-tax/westminster
 - https://www.gov.uk/apply-council-tax-reduction/westminster

### Before
Screenshot 2019-12-05 at 12 00 31

Screenshot 2019-12-05 at 12 01 23

### After
Screenshot 2019-12-05 at 12 00 56

Screenshot 2019-12-05 at 12 01 30

### DIFF
--- a/app/assets/stylesheets/views/_local-transaction.scss
+++ b/app/assets/stylesheets/views/_local-transaction.scss
@@ -1,6 +1,6 @@
 .interaction{
-  margin-top: 0.5em;
-  margin-bottom: 1em;
+  margin-top: govuk-spacing(3);
+  margin-bottom: govuk-spacing(6);
 
   .interaction-match{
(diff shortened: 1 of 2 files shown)


andysellick:
@injms these pages aren't working for some reason in the review app, can you put it onto integration?

andysellick:
Have checked on integration, looks good 👍

PR #2197, opened 2020-01-22
Homepage transition content

huwd:
Trello: https://trello.com/c/ktTNSwAa/398-spike-how-to-update-taxons-so-brexit-becomes-transition

A part of performing tests for potential migrations. Content TBC.

- [x] Updates homepage popular link

**Before**
Screenshot_2020-01-22 Welcome to GOV UK

**After**
Screenshot_2020-01-22 Welcome to GOV UK(1)

- [x] Updates homepage promo slot including image

**Before**
Screenshot_2020-01-22 Welcome to GOV UK(3)

**After**
Screenshot_2020-01-22 Welcome to GOV UK(2)


vanitabarrett:
Is it possible to compress the PNG so the file size isn't so large?

huwd:
Sure, figured I'd wait to see if that's the final png, but can do for this one too!

vanitabarrett:
@huwd No worries, no need to do it for the test, but we should remember for the real thing

PR #2198, opened 2020-01-23
Update simple smart answers

andysellick:
## What

Updates simple smart answers to use govuk-frontend markup for the page and also to use components properly - radio component was in use but error messages, hints and question titles were still being rendered external to the component.

## Why

Removes unnecessary CSS and markup and improves component use. This came about from looking into what sass in `static` was needed and what wasn't.

## Visual changes

- Example simple smart answer: https://www.gov.uk/settle-in-the-uk
- Test URL: https://govuk-fronte-update-sim-kcpuvt.herokuapp.com/settle-in-the-uk

---

**Before:**

Screenshot 2020-01-23 at 12 55 13

Screenshot 2020-01-23 at 12 56 45

---

**After:**

Screenshot 2020-01-23 at 12 55 21

Screenshot 2020-01-23 at 12 56 56

### DIFF
--- a/app/views/simple_smart_answers/_completed_question.html.erb
+++ b/app/views/simple_smart_answers/_completed_question.html.erb
@@ -1,10 +1,11 @@
-<li class="done">
-  <h3 class="question">
-    <span class="question-number"><%= completed_question_counter + 1 %></span><%= completed_question.question.title %>
-  </h3>
-
-  <div class="answer"><%= completed_question.label %></div>
-  <p class="undo">
-    <%= link_to "Change this answer", change_completed_question_path(completed_question_counter), class: "govuk-link" %>
-  </p>
-</li>
+<div class="govuk-summary-list__row">
+  <dt class="govuk-summary-list__key">
+    <%= completed_question_counter + 1 %>. <%= completed_question.question.title %>
+  </dt>
+  <dd class="govuk-summary-list__value">
+    <%= completed_question.label %>
+  </dd>
+  <dd class="govuk-summary-list__actions">
+    <a href="<%= change_completed_question_path(completed_question_counter) %>" class="govuk-link">Change <span class="govuk-visually-hidden">this answer</span></a>
+  </dd>
+</div>
--- a/app/views/simple_smart_answers/_outcome.html.erb
+++ b/app/views/simple_smart_answers/_outcome.html.erb
@@ -1,13 +1,10 @@
-<div class="article-container">
-  <div class="content-block outcome group" data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
-    <div class="inner group">
-      <div class="result-info">
-        <div class="summary"><h2 class="result-title"><%= outcome.title %></h2></div>
+<div data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
+  <%= render "govuk_publishing_components/components/heading", {
+    text: outcome.title,
+    margin_bottom: 4
+  } %>
 
-        <%= render "govuk_publishing_components/components/govspeak", {} do %>
-          <%= raw outcome.body %>
-        <% end %>
-      </div>
-    </div>
-  </div>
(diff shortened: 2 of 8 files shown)


app/views/simple_smart_answers/_completed_question.html.erb
```diff
@@ -1,10 +1,11 @@
-<li class="done">
-  <h3 class="question">
-    <span class="question-number"><%= completed_question_counter + 1 %></span><%= completed_question.question.title %>
```
injms: Would there removal of headings make it less scannable for people using screen readers to pull out the headings in the page?

app/views/simple_smart_answers/_completed_question.html.erb
```diff
@@ -1,10 +1,11 @@
...
-  </h3>
-
-  <div class="answer"><%= completed_question.label %></div>
-  <p class="undo">
-    <%= link_to "Change this answer", change_completed_question_path(completed_question_counter), class: "govuk-link" %>
-  </p>
-</li>
+<div class="govuk-summary-list__row">
```
injms: This description list feels as if it's trying to be a table - so I wonder if a table would be a more semantic way of displaying this information?

app/views/simple_smart_answers/_outcome.html.erb
```diff
@@ -1,13 +1,10 @@
-<div class="article-container">
-  <div class="content-block outcome group" data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
-    <div class="inner group">
-      <div class="result-info">
-        <div class="summary"><h2 class="result-title"><%= outcome.title %></h2></div>
+<div data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
+  <%= render "govuk_publishing_components/components/heading", {
```
injms: Components ftw :taco:

app/views/simple_smart_answers/flow.html.erb
```diff
@@ -3,19 +3,15 @@
...
-      <div class="start-again">
-        <%= link_to "Start again", simple_smart_answer_path(@publication.slug, :edition => @edition), class: "govuk-link" %>
-      </div>
-      <ol>
-        <%= render :partial => "completed_question", :collection => @flow_state.completed_questions %>
-      </ol>
+    <div class="start-over">
+      <%= link_to "Start again", simple_smart_answer_path(@publication.slug, :edition => @edition), class: "govuk-link" %>
     </div>
+    <dl class="govuk-summary-list govuk-!-margin-bottom-9">
```
injms: I thought that we (as in, we the GOV.UK Frontenders) decided that the `!important` override classes shouldn't be used if possible?

app/views/simple_smart_answers/_outcome.html.erb
injms: Also, autocorrect ftw.

app/views/simple_smart_answers/_completed_question.html.erb
andysellick: That's a good point, but I don't think that every question in this bit needs to be a heading. I've added some visually hidden headings for 'your answers' and 'current question', see new commit.

app/views/simple_smart_answers/_completed_question.html.erb
andysellick: That's what I thought too - in fact when I rebuilt `smart-answers` I did the equivalent section using a table. I was going to do the same here but then I found [some Design System guidance](https://design-system.service.gov.uk/patterns/check-answers/) on the subject.

Using a description list is also a bit more responsive - see [a current smart-answer](https://www.gov.uk/maternity-paternity-calculator/y/paternity) for the table use.

app/views/simple_smart_answers/flow.html.erb
andysellick: Ack, sorry, missed that. Will correct.

app/views/simple_smart_answers/_completed_question.html.erb
injms: Super - in that case I'm happy with the use of description list.

PR #2200, opened 2020-01-24
Remove .button class usage from surveys and use components

andysellick:
## What

Removing `.button` class usage on survey pages and updating the forms to use components.

## Why

We're trying to remove this class from static (pulled in from toolkit) and components should be used to make forms like these ones more accessible.

## Changes

See individual commits for specific changes.

Pages changed:

- /done/register-flood-risk-exemption (digital survey - Send feedback button)
- /done/waste-carrier-or-broker-registration (digital survey - Send feedback button)
- /done/register-waste-exemption (digital survey - Send feedback button)
- /done/check-settled-status (standard survey - Send feedback button)
- /done/transaction-finished (Join button)
- /done/driving-transaction-finished (Join button)

## Visual changes (assisted digital satisfaction survey)

There's some form elements that get shown and hidden depending on what options are chosen, which follow after each other in the original page. Using components these options are now nested using the conditional reveal functionality of the radio component.

Note that there's an input element for 'other' which appears unstyled. This is because static is overriding the styles from the component. Since this isn't utterly horrible and we're currently working to remove these styles, I've decided to leave this for now rather than add an override.

Before | After
------- | -------
Screenshot 2020-02-03 at 14 07 16 | Screenshot 2020-02-03 at 14 10 41
Screenshot 2020-02-03 at 14 07 32 | Screenshot 2020-02-03 at 14 10 52
Screenshot 2020-02-03 at 14 08 15 | Screenshot 2020-02-03 at 14 11 10
Screenshot 2020-02-03 at 14 08 22 | Screenshot 2020-02-03 at 14 11 29

## Visual changes (standard survey)

There's no show/hide functionality on this page.

Before | After
------- | -------
Screenshot 2020-02-03 at 14 14 31 | Screenshot 2020-02-03 at 14 15 24


vanitabarrett:
@andysellick 
According to [this file](https://github.com/alphagov/frontend/blob/master/lib/assisted_digital_satisfaction_survey.rb), it's on these pages:

https://govuk-fronte-remove-but-coyfjh.herokuapp.com/done/register-flood-risk-exemption
https://govuk-fronte-remove-but-coyfjh.herokuapp.com/done/waste-carrier-or-broker-registration
https://govuk-fronte-remove-but-coyfjh.herokuapp.com/done/register-waste-exemption

I'm getting a Javascript error when I try to submit the form - not sure if it's been introduced by this PR or not. Perhaps it's worth checking on integration?

andysellick:
@vanitabarrett thanks for the research.

I've tested this on integration and it works okay. I'm a bit confused why this is happening on heroku - the only change as far as the page is concerned is the class on the button.

andysellick:
@vanitabarrett sorry, ignore my previous comment. Also happens on integration. Will look into this. Thanks for the research.

andysellick:
@vanitabarrett have had another look at this and expanded the PR somewhat. Please can I get a review when you have time, thanks.

app/views/completed_transaction/_assistance_received.html.erb
```diff
@@ -0,0 +1,37 @@
...
+      value: 2,
+      text: "Dissatisfied"
+    },
+    {
+      value: 1,
+      text: "Very dissatisfied"
+    },
+  ]
+} %>
+<%= render "govuk_publishing_components/components/textarea", {
```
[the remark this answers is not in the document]
andysellick: That'd be ideal, thanks. The only thing I'd hesitate over is that there doesn't seem to be any backend restriction on this, but can't hurt to add it here. Will do, thanks.

app/assets/javascripts/transactions.js
```diff
@@ -19,7 +19,7 @@ $(document).ready(function () {
     append('<input type="hidden" name="service_feedback[javascript_enabled]" value="true"/>').
     append($('<input type="hidden" name="referrer">').val(document.referrer || "unknown"));
 
-  $('#completed-transaction-form button.button').click(function() {
+  $('#completed-transaction-form button[type="submit"]').click(function() {
```
injms: Is it worth changing this selector for a JavaScript-hook-named class? The class is named to be JS only, and functionality can survive other attribute changes.

```suggestion
  $('.js-completed-transaction-form-button').click(function() {
```

app/assets/stylesheets/views/_completed-transaction.scss
```diff
@@ -1,50 +0,0 @@
-.transaction-done {
```
injms: Hooray for ⌫

app/views/completed_transaction/_assistance_received.html.erb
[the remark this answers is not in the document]
andysellick: Added.

app/views/completed_transaction/_assisted_digital_satisfaction_survey.html.erb
```diff
@@ -1,165 +1,130 @@
...
+  } %>
+
+  <%= render "govuk_publishing_components/components/textarea", {
+    label: {
+      text: "Do you have any ideas for how this service could be improved?",
+      bold: true
+    },
+    name: "service_feedback[improvement_comments]",
+  } do %>
+    <span class="govuk-hint">Limit is 1200 characters</span>
```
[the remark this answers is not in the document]
andysellick: Added.

app/views/completed_transaction/_standard_satisfaction_survey.html.erb
```diff
@@ -1,42 +1,52 @@
...
-    <textarea name="service_feedback[improvement_comments]" class="full-size counted" id="improvement-comments"></textarea>
-     <p id="improvement-commentscounter" class="hint" aria-live="polite" aria-atomic="false">(Limit is 1200 characters)</p>
+  <%= render "govuk_publishing_components/components/textarea", {
+    label: {
+      text: "How could we improve this service?",
+      bold: true
+    },
+    name: "service_feedback[improvement_comments]",
+  } do %>
+    <span class="govuk-hint">Limit is 1200 characters</span>
```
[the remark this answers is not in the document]
andysellick: Added.

app/views/completed_transaction/_assisted_digital_satisfaction_survey.html.erb
```diff
@@ -1,163 +1,130 @@
...
+        conditional: did_you_receive_assistance
+      },
+      {
+        value: "no",
+        text: "No"
+      }
+    ]
+  } %>
+
+  <%= render "govuk_publishing_components/components/radio", {
```
injms: Is this worth using the `assistance_received` partial since the answers are the same?

app/assets/javascripts/transactions.js
andysellick: Agreed, except the element in question is a component. I chose `button[type="submit"]` because it most clearly shows which button the JS acts on and will survive changes as you said. But yeah, there's nothing in the template to indicate that there's JS involved (to be fair, there wasn't before, I only found this by accident).

app/views/completed_transaction/show.html.erb
```diff
@@ -16,20 +15,32 @@
         <p>If you needed an organ transplant would you have one? If so please help others.</p>
         <p>If you live in Wales and want to be an organ donor, you don’t need to do anything. Find out about your choices for <a href="https://www.organdonation.nhs.uk/supporting-my-decision/welsh-legislation-what-it-means-for-me/" rel="external" class="govuk-link">organ donation in Wales</a>.</p>
         <p>
-          <%= link_to 'Join', @publication.promotion['url'],
-                title: "Register to become an organ donor", rel: "external", class: "button", role: "button" %>
+          <%= render "govuk_publishing_components/components/button", {
+            text: "Join",
```
injms: Outside the scope of this pull request - I'll see who I need to flag about making these more explanatory, since "Join", "Sign up", etc don't work on their own.

app/views/completed_transaction/_assisted_digital_satisfaction_survey.html.erb
andysellick: If I understand correctly, you're suggesting repeating the use of the partial that's already used further up? I could modify the partial to accept values for the question text and name attributes for each component (they differ slightly) but the partial was created for a very specific problem, so I'm not sure. Is that what you mean?

app/assets/javascripts/transactions.js
injms: Ah boo. However, I do wonder if we use enough JavaScript for this to be a useful thing?

app/views/completed_transaction/_assisted_digital_satisfaction_survey.html.erb
```diff
@@ -1,163 +1,132 @@
...
-      Very dissatisfied
-    </label>
-  </fieldset>
+  <% other_person = capture do %>
+    <%= render 'assistance_received', locals: { id_prefix: 'other'} %>
+  <% end %>
 
-  <br />
+  <% other_input = capture do %>
+    <%= render "govuk_publishing_components/components/input", {
```
injms: Shame the styles for this input are being overridden by govuk_template in static. Soon it will be gone!

app/views/completed_transaction/_assisted_digital_satisfaction_survey.html.erb
injms: That's what I mean - give the partial the  values for `name` and for the heading text so you can use the partial for each set of satisfied -> unsatisfied radio buttons.

PR #2236, opened 2020-02-21
Upgrade to rails 6

AlanGabbianelli:
See commit messages for more detail.

Trello card: https://trello.com/c/Ik7ulDXQ/1768-3-upgrade-info-frontend-to-rails-6

Co-authored-by: Rebecca Pearce <rebecca.pearce@digital.cabinet-office.gov.uk>


issyl0:
Merging https://github.com/alphagov/frontend/pull/2255 and rebasing this will once again exclude `bin` (but keep `config`) from RuboCop checks, so sorry for the wasted effort. I didn't realise there'd be so many dependencies!

config/sidekiq.yml
```diff
@@ -3,3 +3,5 @@
 :queues:
   - default
   - mailers
+  - active_storage_analysis
+  - active_storage_purge
```
thomasleese: Do we need this? I can't imagine we would need ActiveStorage in this application?

AlanGabbianelli: I suppose we can avoid these two if we're sure we'll never use ActiveStorage. It was to be sure we can safely use all the new rails 6 defaults and specifically:
- `Rails.application.config.active_storage.queues.analysis = :active_storage_analysis`
- `Rails.application.config.active_storage.queues.purge = :active_storage_purge`

thomasleese: I did a quick search and I don't see any mentions of it being used in the app, in fact it doesn't even seem to be `require`d at any point.

AlanGabbianelli: Sounds good. I'll rebase and edit the commit to remove them.

PR #2237, opened 2020-02-21
Add AB test for electric cars

sihugh:
This test compares click through rates of people visiting the two done pages:

 - /done/vehicle-tax
 - /done/check-vehicle-tax

We append the variant name on to the end of the URL that leads to the GoUltralow website so that we they can do some measurement based on what users have seen here.

Measurement on GOV.UK only occurs if users have opted in to cookies (I can't speak for what happens on the destination website). People who haven't opted in will get the control variant which does not add a tracking query parameter to the link.

The config in en.yml will be changing when we get the final list from DfT.

https://trello.com/c/In6XcrQU/1399-set-up-dft-multi-variant-a-b-test-test-dev-work

Also hooks in with https://trello.com/c/EpuznchJ/1398-dft-for-the-multivariate-test-analytics-tasks

### DIFF
--- a/app/views/completed_transaction/show.html.erb
+++ b/app/views/completed_transaction/show.html.erb
@@ -1,5 +1,6 @@
 <% content_for :extra_headers do %>
   <meta name="robots" content="noindex, nofollow" />
+  <%= electric_car_variant.analytics_meta_tag.html_safe if electric_car_testable? %>
 <% end %>
 
 <%= render layout: 'shared/base_page', locals: {
@@ -8,7 +9,18 @@
   edition: @edition,
 } do %>
 
-  <% if @publication.promotion %>
+  <% if electric_car_testable? %>
+    <div class="promotion">
+      <p><%= t electric_car_translation_key("text") %></p>
+      <p>
+        <%= render "govuk_publishing_components/components/button", {
+          text: t(electric_car_translation_key("button_text")),
+          href: t(electric_car_translation_key("link")),
+          rel: "external"
+        } %>
+      </p>
+    </div>
+  <% elsif @publication.promotion %>
     <div class="promotion">
       <% if @publication.promotion['category'] == 'organ_donor' %>
         <p>Please join the NHS Organ Donor Register.</p>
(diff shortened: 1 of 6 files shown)


app/views/completed_transaction/show.html.erb
```diff
@@ -8,7 +9,18 @@
...
 } do %>
 
-  <% if @publication.promotion %>
+  <% if electric_car_testable? %>
+    <div class="promotion">
+      <p><%= t electric_car_translation_key("text") %></p>
+      <p>
+        <%= render "govuk_publishing_components/components/button", {
+          text: t(electric_car_translation_key("button_text")),
+          href: t(electric_car_translation_key("link")),
```
bilbof: This is a neat solution 👍

test/functional/completed_transaction_controller_test.rb
```diff
@@ -22,4 +24,24 @@ class CompletedTransactionControllerTest < ActionController::TestCase
...
+  context "A/B testing" do
+    setup do
+      content_store_has_example_item("/done/vehicle-tax", schema: "completed_transaction")
+      content_store_has_example_item("/done/weee", schema: "completed_transaction")
+    end
+
+    %w[A B C D E F G H Z].each do |variant|
+      should "show the #{variant} variant of the electric car AB test on applicable pages" do
+        with_variant ElectricCarABTest: variant do
+          get :show, params: { slug: "done/vehicle-tax" }
```
bilbof: Is there a way to test that the response has been modified for the AB test here? I saw below you used `assert_response_not_modified_for_ab_test` which I hadn't seen before :)

sihugh: It's a little abstracted (and not particularly well named), but `with_variant` wraps up some assertions on this: https://github.com/alphagov/govuk_ab_testing/blob/master/lib/govuk_ab_testing/abstract_helpers.rb#L7-L20

bilbof: Ah cool, thanks! 👍

sihugh:
Thanks for the review @bilbof!

PR #2261, opened 2020-03-06
Reinstate A/B cookies text, under analytics cookies

barrucadu:
Screenshot 2020-03-06 at 11 08 29

We should probably run this by a content person.

---

[Trello card](https://trello.com/c/2R7HyMj8/1425-add-ab-testing-content-to-cookie-policy-page)


sihugh:
This looks good to me, I vote we ship it

PR #2262, opened 2020-03-06
Add a task for unpublishing the EU funding form routes

thomasleese:
It currently isn't possible to [unpublish these using the special-route-publisher](https://github.com/alphagov/special-route-publisher/blob/0e0bfa8fdf9aef998a86601ac787c87aadc173f0/data/special_routes.yaml#L23-L71) so I've added this temporary Rake task. Once I've run the task, I will delete this code.

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -23,4 +23,66 @@ namespace :publishing_api do
 
     HomepagePublisher.publish!(publishing_api, logger)
   end
+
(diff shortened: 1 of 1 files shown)


thomasleese:
I've run the task in integration and it seems to have worked: https://www-origin.integration.publishing.service.gov.uk/brexit-eu-funding/confirmation

brucebolt:
Looks good.  I'll be pleased to see this form go away!

PR #2264, opened 2020-03-06
Do not report exceptions rescued by ActionDispatch

AlanGabbianelli:
After upgrading to rails 6, we started seeing this error:
`Mime::Type::InvalidMimeType "charset=utf-8" is not a valid MIME type`
Example here:
https://sentry.io/organizations/govuk/issues/1549200925/?end=2020-03-06T12%3A00%3A23&environment=staging&project=202225&query=is%3Aunresolved+url%3Ahttps%3A%2F%2Fwww-origin.staging.govuk.digital%2Freport-an-unregistered-trader-or-business&start=2020-03-05T16%3A07%3A23&utc=true

This is because, even though these errors are rescued by ActionDispatch,
they still reach Sentry. Here's the explanation from Sentry's docs:
> Rails catches exceptions in the ActionDispatch::ShowExceptions or
> ActionDispatch::DebugExceptions middlewares, depending on the
> environment. When rails_report_rescued_exceptions is true (it is by
> default), Raven will report exceptions even when they are rescued by
> these middlewares.

So, in order to stop this behaviour, we can set this option to be false.

Related PRs:
- https://github.com/alphagov/email-alert-api/pull/1138
- https://github.com/alphagov/govuk_app_config/pull/137

Trello card: https://trello.com/c/Ik7ulDXQ/1768-3-upgrade-frontend-to-rails-6

### DIFF
--- a/config/initializers/raven.rb
+++ b/config/initializers/raven.rb
@@ -1,0 +1,3 @@
+Raven.configure do |config|
+  config.rails_report_rescued_exceptions = false
+end


thomasleese:
Do we need both this change here and https://github.com/alphagov/govuk_app_config/pull/137? It seems like we're likely to want to set `rails_report_rescued_exceptions = false` across all our applications, so should we set that in `govuk_app_config` instead. Presumably then we don't need to have so many Rails exceptions in the [`excluded_exceptions`](https://github.com/alphagov/govuk_app_config/blob/6c050a5a6b1c5e804d70a0ca9d6b332d2a3838df/lib/govuk_app_config/configure.rb#L10-L26).

kevindew:
Oh this config option looks neat! 

Toms suggestion to set this in govuk_app_config sounds great.

AlanGabbianelli:
@thomasalanlee both changes have the effect of `Mime::Type::InvalidMimeType` not showing up in Sentry but they are different:
- https://github.com/alphagov/frontend/pull/2264 will act specifically for `frontend` and will stop all exceptions rescued from rails (not only `Mime::Type::InvalidMimeType`) from appearing in Sentry (this app - all errors)
- https://github.com/alphagov/govuk_app_config/pull/137 will act for all apps that use `govuk_app_config` and will stop only `Mime::Type::InvalidMimeType` from appearing in Sentry (all apps - this error)

I think your suggested approach is better. If we take the config that acts on all errors and apply it to the gem that acts on all apps, we can stop all exceptions rescued from rails from appearing in Sentry for all apps that depend on `govuk_app_config`.

I'll close https://github.com/alphagov/frontend/pull/2264 and https://github.com/alphagov/govuk_app_config/pull/137 and open a new PR for that.

AlanGabbianelli:
Closing this in favour of https://github.com/alphagov/govuk_app_config/pull/138

PR #2274, opened 2020-03-12
Optimise the new budget image further using MozJPEG

Nooshu:
Further optimisation to the new budget homepage image using MozJPEG at 75%, and resized to 600x400. 

| Image           | Before | After  | Change |
|-----------------|--------|--------|--------|
| budget-2020.jpg | 131KB  | 21.5KB | -84%   |


injms:
The image is displayed at a max of 600px wide at the breakpoint between mobile and desktop. It's worth resizing it to that instead of keeping it at 960px wide.

Nooshu:
> The image is displayed at a max of 600px wide at the breakpoint between mobile and desktop. It's worth resizing it to that instead of keeping it at 960px wide.

Yes thanks, done and updated PR comment accordingly 👍

PR #2281, opened 2020-03-16
Add Listen gem to developer dependencies

injms:
## What

Adds the Listen gem to the developer dependencies.

## Why

The Listen gem was required when this app was upgraded to use the Rails 6
defaults - [see the diff showing this change](https://github.com/alphagov/frontend/commit/d83ce6ff613b6861269a95612b0c8baf4f52256b#diff-a2f09e1db7a47eb67c5849478a97f3d7R51) - but the gem wasn't added to the Gemfile. 

## Visual differences

None.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem "uk_postcode", "~> 2.1.5"
 group :development do
   gem "better_errors"
   gem "binding_of_caller"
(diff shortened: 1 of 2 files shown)


DilwoarH:
ah too late - fix tested and working 👍

PR #2289, opened 2020-03-20
Update homepage popular links to point to new page

sihugh:
The longer title wraps - is this something we want?

https://trello.com/c/GP3fDVtc/48-update-popular-links-on-the-govuk-homepage


sihugh:
Mia's approved the longer title.  We could increase spacing between links in the future

DilwoarH:
> Mia's approved the longer title. We could increase spacing between links in the future

Probably not a good idea as we don't have much space

PR #2298, opened 2020-03-23
Temporarily rate limit users accessing the coronavirus-vulnerable-people service

alex-ju:
There are concerns that the demand would drive too much traffic to the "Get coronavirus support as an extremely vulnerable person" service, resulting in the service being overloaded. If we need to deploy this, the change should be reverted – hopefully we don't 🤞

The effect of this code was to put each new visitor to the service start page into one of two buckets based on a random number. 20% of visitors would be put into a bucket where they would see a short message in the place of the service start button, asking them to try again soon. After two minutes, the cookie that identified which bucket the user belongs to would expire, allowing the user to go through the randomised process again, with another 20% chance of being delayed. This amounted to a 4% chance of not being allowed through to the service after 4 minutes.

This solution was chosen as the most pragmatic and low-risk solution after some careful consideration of the risks, pros and cons of the alternatives. We understood that users who were unable to run Javascript would not be rate-limited, and we determined that this was acceptable.

See individual commits for details of the change.

<table>
<tr><th>80%</th><th>20%</th></tr>
<tr><td valign="top">

localhost_3005_coronavirus-extremely-vulnerable

</td><td valign="top">

localhost_3005_coronavirus-extremely-vulnerable (1)

</td></tr>
</table>

### DIFF
--- a/app/assets/javascripts/coronavirus-extremely-vulnerable.js
+++ b/app/assets/javascripts/coronavirus-extremely-vulnerable.js
@@ -1,0 +1,48 @@
+/* eslint-env jquery */
+/* global GOVUK */
+
+$(function () {
+  // x minutes in days: x / (60 minutes * 24 hours)
+  var minutesForCohort = 2
+  var minutesInDay = 1440 // 60 * 24
+
+  if (window.location.href.indexOf('/coronavirus-extremely-vulnerable') > -1) {
+    // Get the current cohort or assign one if it has not been already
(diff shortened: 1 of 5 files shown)


alex-ju:
Got [21.34.0](https://github.com/alphagov/frontend/pull/2299) in and rebased this. Should be ready to be merged in case of emergency.

alex-ju:
@timblair do we still need to have this ready, or shall I close it?

timblair:
No, please close it. Thank you for preparing it just in case.

PR #2313, opened 2020-04-02
Improve performance of homepage images

injms:
## What

 - Set the width and height of the images 
 - Added `srcset` for budget image

## Why

Setting the width and height because some browsers will now use those attributes to calculate the intrinsic size of the image and make the element take up space before it's loaded to help prevent page layout jank.

Adding `srcset` for the budget photograph allows the browser to choose an appropriate image to load. The fallback for browsers that don't support `srcset` is the `src` attribute, which is using the original (now renamed) image.

<table>
<thead>
<tr>
<th>Screen width</th>
<th>Image used size</th>
<th>Reduction from original file size (22KB)</th>
</tr>
</thead>
<tbody>
<tr>
<td>1280px</td>
<td>8KB</td>
<td rowspan=2>64% reduction</td>
</tr>
<tr>
<td>768px</td>
<td>8KB</td>
</tr>
<tr>
<td>700px</td>
<td>5KB</td>
<td>72% reduction</td>
</tr>
<tr>
<td>640px</td>
<td>22KB</td>
<td>0% reduction</td>
</tr>
<tr>
<td>480px</td>
<td>8KB</td>
<td>64% reduction</td>
</tr>
<tr>
<td>320px</td>
<td>15KB</td>
<td>32% reduction</td>
</tr>
</tbody>
</table>

The other two images are in the PNG format and saw barely any reduction in file size when resized because they're single blocks of colour with text in them. (Though they probably could be replaced with an SVG or HTML... 🐇🕳)

## Visual difference

None.


app/views/homepage/index.html.erb
```diff
@@ -115,7 +127,21 @@
 
         <div class="govuk-grid-column-one-third">
           <a href="/government/topical-events/budget-2020" aria-hidden="true" tabindex="-1">
-            <%= image_tag 'homepage/budget-2020.jpg', alt: '', class: 'home-promo__image', loading: "lazy" %>
+            <%= image_tag "homepage/budget-2020--480w.jpg", {
+              alt: "",
+              class: "home-promo__image",
+              height: 400,
+              loading: "lazy",
+              sizes: "(max-width: 640px) 100vw, (max-width: 960px) 33vw, 300px",
```
ChrisBAshton, line 135: This could potentially be 1020px rather than 960px, as the 30px of margin either side means we continue to take up 33vw of the screen until we reach 1020px. But I don't think it'll make a difference in practice.

```suggestion
              sizes: "(max-width: 640px) 100vw, (max-width: 1020px) 33vw, 300px",
```

chao-xian:
There is guidance on our (confluence) wiki for content people sourcing images for the homepage. We'll need to give them a heads up.

PR #2320, opened 2020-04-08
Use individual component sass

andysellick:
Update frontend to import the sass for only the components it is using, instead of all components, as [documented here](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#import-sass-for-individual-components).

## Filesize comparison

Before:

Screenshot 2020-04-08 at 09 57 18

Total **900 KB**

After:

Screenshot 2020-04-08 at 09 58 26

Total **562 KB**

## Additional changes

Also update [foreign travel advice](/foreign-travel-advice/) to use the warning text component. This was previously not possible because the component did not allow for a heading but has been updated to allow this.

This is being done as part of this PR because this page used styles from the component but not the component itself - so I noticed it had broken when I made the Sass change (because the warning text component is not used anywhere, it was not part of the 'suggested sass' output. It now is).

Screenshot 2020-04-08 at 13 40 55

Trello card: https://trello.com/c/GWttlBxs/228-upgrade-applications-to-use-individual-component-sass

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -4,7 +4,35 @@ $govuk-use-legacy-palette: false;
 // BASE Stylesheet
 
 // Components from govuk_publishing_components gem
-@import "govuk_publishing_components/all_components";
+@import 'govuk_publishing_components/govuk_frontend_support';
+@import 'govuk_publishing_components/component_support';
+@import 'govuk_publishing_components/components/back-link';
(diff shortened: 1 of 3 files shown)


andysellick:
@alex-ju is both of us switching to the warning component going to cause you a merge conflict, or are we doing exactly the same thing?

alex-ju:
yeah, I'll have a conflict, but will rebase, don't worry.

PR #2341, opened 2020-04-30
Implement calendars

barrucadu:
Does what it says on the tin.  I've tried to commit this in sensible chunks which each pass all the tests; but each commit is huge so I'm not sure how best to review this.  Good luck!

Before merging this we should deploy it to integration, update the path reservations in publishing-api, and check that the calendars can be published by frontend.

---

[Trello card](https://trello.com/c/Y7SJdPT9/1945-5-move-when-do-the-clocks-change-and-uk-bank-holidays-services-from-calendars-to-frontend-%F0%9F%8D%90)


lib/tasks/publishing_api.rake
```diff
@@ -23,4 +23,12 @@ namespace :publishing_api do
 
     HomepagePublisher.publish!(publishing_api, logger)
   end
+
+  desc "Send pages to the publishing-api"
+  task publish: [:environment] do
```
edwardkerry, line 28: should this perhaps be renamed to `publish_calendars` now it is in a less specialised app?

barrucadu:
All looks good in staging, I'm able to republish the calendars, and all the pages work.

PR #2412, opened 2020-07-03
Update global travel advice callout to reflect 4th July updates

issyl0:
- As per request from FCO.


thomasleese:
Time to start booking holidays!

PR #2433, opened 2020-07-28
Fix simple smart answers headings

injms:
### [Find out if you can apply to settle in the UK](https://www.gov.uk/settle-in-the-uk/y/you-re-the-family-member-or-partner-of-a-british-citizen/no/partner/you-re-still-with-your-partner)

| Before | After |
|-|-|
| image | image |

### [Tell DVLA you've sold, transferred or bought a vehicle](https://www.gov.uk/sold-bought-vehicle/y)
| Before | After |
|-|-|
| image | image |

### [Find help to get your child back from abroad or arrange contact](https://www.gov.uk/return-or-contact-abducted-child/y)
| Before | After | 
|-|-|
| image | image |

PR #2435, opened 2020-07-30
Fix simple smart answers headings

injms:
## What 

The headings in the simple smart answers have been updated so that the heading on each page is the question being asked, rather than the title of the simple smart answer itself.

This is a partial replacement pull request for #2433 due to a decrease in scope - this only fixes the headings and not the `<title>` element.

Trello card: https://trello.com/c/HpZBzZAu

## Why

This is to try and meet [WCAG 2.4.2 - Page titles](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html), which asks that pages don't have the same titles on each page.

In this case "title" is understood to mean the main heading on the page rather than the `<title>` element.

If there are pages in a flow that have the same heading on each page, then it isn't immediately obvious that the browser has gone to a new page. Reading the `h1` of the page aloud will give the same result for every step of the process, which can be confusing and lead to a user being unaware that they've successfully continued to the next page.

Since the question itself should be the primary heading for each page, this has meant a rearrangement of the order in the template markup. 

Visually, the summary of answers is displayed first with the question appearing after it. But this isn't possible if the question is the page heading - the question needs to be before the summary because:

 * The summary can't be the page heading, or there will be the same heading for each of the pages again and we're back to the original problem.

 * The summary can't come before the page heading or there will be either two `h1`s (not good) or a `h2` before a `h1` (also not good).

~~So the markup needs to have the question first, then the summary afterwards. This will mean either visual changes (which is fine, but outside the scope this ticket) or visually reordering the layout - which is what this pull request does. The layout is unchanged with the summary first and the question second, but the markup has the question first and the summary second.~~

~~This still passes WCAG Success Criteria 1.3.2 Meaningful Sequence as the guidance states that ["providing a particular linear order is only required where it affects meaning"](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-sequence.html). For example, the instructions in a recipe require a specific (and therefore meaningful) order but the ingredients can be displayed in any order.~~

~~The visual order of the summary and the question doesn't affect the meaningful sequence as they are as easily understood in whichever order they come in.~~

Another solution - change the layout to have the summary after the questions. This also better matches the smart answers layout.

### Find out if you can apply to settle in the UK - [Live site](https://www.gov.uk/settle-in-the-uk/y/you-re-the-family-member-or-partner-of-a-british-citizen/no/partner), [Deployed site](https://govuk-fronte-fix-simple-xr3wkv.herokuapp.com/settle-in-the-uk/y/you-re-the-family-member-or-partner-of-a-british-citizen/no/partner)

| Before | After |
|-|-|
| image | image |

### [Tell DVLA you've sold, transferred or bought a vehicle](https://www.gov.uk/sold-bought-vehicle/y)
| Before | After |
|-|-|
| image | image |

### [Find help to get your child back from abroad or arrange contact](https://www.gov.uk/return-or-contact-abducted-child/y)
| Before | After | 
|-|-|
| image | image |

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -58,6 +58,7 @@ $govuk-use-legacy-palette: false;
 @import "views/travel-advice";
 @import "views/report-child-abuse";
 @import "views/local-authority";
+@import "views/simple-smart-answer";
 
 // exceptional format overrides
 @import "views/jobsearch";
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -58,7 +58,6 @@ $govuk-use-legacy-palette: false;
 @import "views/travel-advice";
 @import "views/report-child-abuse";
 @import "views/local-authority";
-@import "views/simple-smart-answer";
 
 // exceptional format overrides
 @import "views/jobsearch";
--- a/app/assets/stylesheets/views/_simple-smart-answer.scss
+++ b/app/assets/stylesheets/views/_simple-smart-answer.scss
@@ -1,4 +1,0 @@
-.start-over {
-  @include govuk-font(19);
-  margin-bottom: govuk-spacing(2);
-}
--- a/app/assets/stylesheets/views/_simple-smart-answer.scss
+++ b/app/assets/stylesheets/views/_simple-smart-answer.scss
@@ -1,0 +1,4 @@
+.simple-smart-answer__question-and-outcome {
+  padding-top: govuk-spacing(8);
+  padding-bottom: govuk-spacing(6)
+}
--- a/app/views/simple_smart_answers/_current_question.html.erb
+++ b/app/views/simple_smart_answers/_current_question.html.erb
@@ -6,11 +6,9 @@
       error_message = t('formats.simple_smart_answer.please_answer')
     end
   %>
-  <%
-    description = render "govuk_publishing_components/components/govspeak", {
-      content: sanitize(question.body),
-    }
-  %>
+  <% description = render "govuk_publishing_components/components/govspeak", {
+    content: sanitize(question.body),
+  } if question.body %>
 
   <%= render "govuk_publishing_components/components/radio", {
     description: description,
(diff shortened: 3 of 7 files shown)


maxgds:
Technically this is an elegant solution, I can't see anything wrong with it and have looked at it in a few browsers. A couple of things, though. 1) I've invited Mia to sign it off as it feels like a reasonably big design change and I couldn't see any evidence of her involvement to date. 2) You state "a h2 before a h1 (not allowed)" - what do you mean by not allowed? WCAG or GDS ruling? While it's discouraged, Anika assures me it's not a WCAG fail, and it's already done across the site, the most prominent example being the cookie banner having a h2 that appears in the flow before every h1. We may need to be more specific with that guidance, wherever it lives, and consider whether it's ok in a page like this one.

injms:
> Technically this is an elegant solution, I can't see anything wrong with it and have looked at it in a few browsers. A couple of things, though. 1) I've invited Mia to sign it off as it feels like a reasonably big design change and I couldn't see any evidence of her involvement to date. 2) You state "a h2 before a h1 (not allowed)" - what do you mean by not allowed? WCAG or GDS ruling? While it's discouraged, Anika assures me it's not a WCAG fail, and it's already done across the site, the most prominent example being the cookie banner having a h2 that appears in the flow before every h1. We may need to be more specific with that guidance, wherever it lives, and consider whether it's ok in a page like this one.

Thanks for asking Mia - I'll add the extra spacing below the question that she requested and share with her.

I probably should've said "h2 before a h1 (not good)" rather than "not allowed". You're absolutely correct - there's nothing technically that says it shouldn't be done. I'll amend the comment and commit message.

The [MDN usage notes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#Usage_notes) and [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#Accessibility_concerns) recommends starting a page at `h1` and going in sequence. This seems to match the spirit of [2.4.6 - headings and labels](https://www.w3.org/WAI/WCAG21/Understanding/headings-and-labels.html) and [1.3.2 - meaningful sequence](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-sequence.html) - however neither of these explicitly mention that an `h2` shouldn't come before an `h1`.

miaallers:
Happy with the answers sitting below the questions, looks good! Approved from my point of view

PR #2436, opened 2020-07-31
Make app consistent with upgrades

cdccollins:
This is to make Frontend consistent with the other Rails 6 upgrades we've been doing. 

It also fixes a warning that was appearing on master as well about passing unsanitised HTML into the govspeak component, which will become deprecated in a future version.


thehenster:
`config/new_framework_defaults_5_1.rb` can be deleted. There aren't any uses of `#form_with` in the codebase so no need to worry about that `Rails.application.config.action_view.form_with_generates_remote_forms = false` in there.

`bin/setup` can use the new and trendy `db:prepare` task. https://github.com/rails/rails/blob/v6.0.3.2/railties/lib/rails/generators/rails/app/templates/bin/setup.tt#L31

`script/rails` is old school. The `script` dir can go.

Rhian (approved):
Updates as per the previous review comments look fine to me.

PR #2438, opened 2020-07-31
Changes page `<title>` logic to fix simple smart answers

injms:
## What
This pull request flips the logic used to choose how to display a page's `<title>` - `:title` is used if present, otherwise the publication title helper is used, and finally a fallback of "GOV.UK". 

## Why

This change allows the `<title>` of a page to be set from with the template, even if the page is a publication - which means we can fix the `<title>` for the simple smart answers.

The problem - simple smart answers were displaying the same text in each step's `<title>`, which means that the pages don't meet the [success criteria 2.4.2 of the WCAG](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html).

The page title has an conditional statment where if the page is a publication the `<title>` is set by a helper. This means that using `content_for :title` doesn't work for publications because the title text is generated by the helper - and simple smart answers are considered a publication.

## Visual changes

The `<title>` of a simple smart answer should be:
    
 - "Error" (if an error is present)
 - The question being asked
 - The name of the simple smart answer
 - "GOV.UK"

Before: https://www.gov.uk/settle-in-the-uk/y/you-re-the-family-member-or-partner-of-a-british-citizen/no/partner
After: https://govuk-fronte-fix-simple-ov4eqp.herokuapp.com/settle-in-the-uk/y/you-re-the-family-member-or-partner-of-a-british-citizen/no/partner

Before: https://www.gov.uk/return-or-contact-abducted-child/y
After: https://govuk-fronte-fix-simple-ov4eqp.herokuapp.com/return-or-contact-abducted-child/y

Before: https://www.gov.uk/sold-bought-vehicle/y
After: https://govuk-fronte-fix-simple-ov4eqp.herokuapp.com/sold-bought-vehicle/y

### DIFF
--- a/app/views/simple_smart_answers/_outcome.html.erb
+++ b/app/views/simple_smart_answers/_outcome.html.erb
@@ -1,14 +1,8 @@
 <%
-  page_title = []
-
-  page_title << outcome.title
-  page_title << @publication.title
-  page_title << "GOV.UK"
-
   content_for :title do
-    page_title.join(' - ')
+    simple_smart_answer_page_title(outcome.title)
   end
- %>
+%>
  <div data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
   <%= render "govuk_publishing_components/components/title", {
     context: @publication.title,
--- a/app/views/simple_smart_answers/_outcome.html.erb
+++ b/app/views/simple_smart_answers/_outcome.html.erb
@@ -1,4 +1,15 @@
-<div data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
+<%
+  page_title = []
+
+  page_title << outcome.title
+  page_title << @publication.title
+  page_title << "GOV.UK"
+
+  content_for :title do
+    page_title.join(' - ')
+  end
+ %>
+ <div data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
   <%= render "govuk_publishing_components/components/title", {
     context: @publication.title,
     font_size: "l",
--- a/app/controllers/simple_smart_answers_controller.rb
+++ b/app/controllers/simple_smart_answers_controller.rb
@@ -22,9 +22,10 @@ class SimpleSmartAnswersController < ApplicationController
 private
 
   helper_method(
-    :smart_answer_path_for_responses,
-    :change_completed_question_path,
     :answer_summary_data,
+    :change_completed_question_path,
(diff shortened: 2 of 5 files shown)


danacotoran:
@injms I'm having a look at this! for now I have some questions: 
1. Should we add a test to check that the title is correctly formatted for the error case? 
2. On the outcome page, should we use the same pattern as for the questions, and have `Outcome message - Simple smart answer name - GOV.UK` (currently it's  just `Simple smart answer name - GOV.UK`)?

injms:
> @injms I'm having a look at this! for now I have some questions:
> 
> 1. Should we add a test to check that the title is correctly formatted for the error case?
> 2. On the outcome page, should we use the same pattern as for the questions, and have `Outcome message - Simple smart answer name - GOV.UK` (currently it's  just `Simple smart answer name - GOV.UK`)?

Good ideas - thank you.

I've added a test to check that the title is okay when a validation error is present; and I've updated the outcome page title (plus a test to make sure that's okay too).

app/views/simple_smart_answers/_outcome.html.erb
```diff
@@ -1,4 +1,15 @@
-<div data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
+<%
+  page_title = []
```
danacotoran, line 2: I wonder if it would be worth creating a helper function that takes in a list of all the various strings that are included in the page title (error, outcome/question title, simple smart answer title) and spits out the formatted title? 

I just think the code would read a bit cleaner if we moved that logic outside the template, and instead of this:

```
<%
  page_title = []
  page_title << outcome.title
  page_title << @publication.title
  page_title << "GOV.UK"

  content_for :title do
    page_title.join(' - ')
  end
 %>
```
...we had something like this: 
```
<%
  content_for :title do
    some_page_title_builder_function(outcome.title, @publication.title)
  end
 %>
```

injms: Yes! Good idea - done 👍

app/views/simple_smart_answers/_outcome.html.erb
```diff
@@ -1,4 +1,9 @@
-<div data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
+<%
+  content_for :title do
+    simple_smart_answer_page_title(outcome.title)
+  end
+%>
+ <div data-module="track-smart-answer" data-smart-answer-node-type="outcome" data-smart-answer-slug="<%= outcome.slug %>">
```
danacotoran, line 6: Nitpick, indentation here doesn't look right

PR #2444, opened 2020-08-07
Update Govspeak usage to stop unsafe HTML warning

injms:
## What

Removes use of `content` attribute when the component is directly outputting HTML into the page.

So 

```erb
<%= render "govuk_publishing_components/components/govspeak", {
  content: sanitize(some_content),  
} %>
```

was replaced with

```erb
<%= render "govuk_publishing_components/components/govspeak", {} do %>
  <%= sanitize(some_content) %>
<% end %>
```

## Why

To stop the warnings that unsafe HTML was being passed into the Govspeak component.

## Visual differences

None.

### DIFF
--- a/app/views/local_transaction/results.html.erb
+++ b/app/views/local_transaction/results.html.erb
@@ -9,12 +9,13 @@
         We've matched this postcode to <span class="local-authority"><%= @local_authority.name %></span>.
       </p>
       <p id="get-started" class="get-started group">
-        <%= render "govuk_publishing_components/components/button",
-                    text: "Go to their website",
-                    rel: "external",
-                    start: true,
-                    margin_bottom: true,
-                    href: @interaction_details['local_interaction']['url'] %>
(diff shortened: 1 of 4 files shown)


app/views/local_transaction/results.html.erb
```diff
@@ -48,14 +50,14 @@
...
-      href: local_transaction_search_path(@publication.slug)
+      href: local_transaction_search_path(@publication.slug),
     } %>
   </div>
   <section class="more">
     <div class="more">
-      <%= render "govuk_publishing_components/components/govspeak", {
-        content: sanitize(@publication.more_information)
-      } %>
+      <%= render "govuk_publishing_components/components/govspeak", {} do %>
```
andysellick, line 58: Is the `{}` needed?

injms: Nope, it's not needed. It renders the same with or without the `{}` - I left it in as that's what the examples in the component gem docs do.

PR #2452, opened 2020-08-25
Ensure assistive tech can consume the Cookie settings success message

maxgds:
https://trello.com/c/Bl66U7iq/297-cookies-page-status-message-cannot-be-programmatically-determined

This switches the Notice component for the more semantically appropriate Success Alert component as the code is indicating to the user that their cookies were successfully updated. 

If the user is on the cookie settings page with a success message already present and chooses to resubmit the form with different preferences then they are scrolled to the top of the screen where the success message exists. This didn't give enough indication for screenreaders that something had changed, so it now ensures the message is hidden before it is presented again after the scroll and the alert message is sounded.

The only visual change is a difference in colour between the success component and the notice component. The hide/show behaviour is too fast to note.


app/views/help/cookie_settings.html.erb
```diff
@@ -1,15 +1,11 @@
...
-      <p><%= t('cookies.confirmation_message') %></p>
-      <a class="govuk-link cookie-settings__prev-page" href='#' data-module="track-click" data-track-category="cookieSettings" data-track-action="Back to previous page">
-        <%= t('cookies.confirmation_previous_referrer_link') %>
-      </a>
-    <% end %>
+    <div class="cookie-settings__confirmation" data-cookie-confirmation="true">
+      <%= render "govuk_publishing_components/components/success_alert", {
+        message: t('cookies.confirmation_title'),
+        description: "<p>#{t('cookies.confirmation_message')}</p>
+        <a class='govuk-link cookie-settings__prev-page' href='#' data-module='track-click' data-track-category='cookieSettings' data-track-action='Back to previous page'>#{t('cookies.confirmation_previous_referrer_link')}</a>".html_safe
```
owenatgov, line 7: The component rendering in the gem doesn't include any sanitisation. This html includes dynamic content so I'd suggest adding a `sanitize` to this.

owenatgov (changes requested):

owenatgov (approved):

PR #2453, opened 2020-08-26
Fix HTML render issues

gclssvglx:
## Issue
HTML was not being rendered correctly in `govspeak` blocks.

## Solution
It was found that the `sanitize` method was being used. This method strips all HTML tags and attributes that aren’t whitelisted and so the content being passed into the `govspeak` blocks was plain text.

We need to ensure the HTML presented is safe, so applying the `html_safe` method to the content.  Unfortunately, some fields are not mandatory so we need guard clauses to stop `nil` exceptions on all optional fields.

[Trello card](https://trello.com/c/s6Cs6X9t/971-fix-html-rendering-in-simple-smart-answers)

### DIFF
--- a/app/views/local_transaction/results.html.erb
+++ b/app/views/local_transaction/results.html.erb
@@ -48,16 +48,20 @@
       <% end %>
     </div>
   <% end %>
+
   <div class="search-again">
     <%= render "govuk_publishing_components/components/back_link", {
       href: local_transaction_search_path(@publication.slug),
     } %>
   </div>
-  <section class="more">
-    <div class="more">
-      <%= render "govuk_publishing_components/components/govspeak", {} do %>
-        <%= sanitize(@publication.more_information) %>
-      <% end %>
-    </div>
-  </section>
+
+  <% if @publication.more_information.present? %>
+    <section class="more">
+      <div class="more">
+        <%= render "govuk_publishing_components/components/govspeak", {
+          content: @publication.more_information.html_safe
+        } %>
+      </div>
+    </section>
+  <% end %>
 <% end %>
--- a/app/views/local_transaction/search.html.erb
+++ b/app/views/local_transaction/search.html.erb
@@ -9,13 +9,17 @@
   publication: @publication,
   edition: @edition,
 } do %>
-  <section class="intro">
-    <div class="get-started-intro">
-      <%= render "govuk_publishing_components/components/govspeak", {} do %>
-        <%= sanitize(@publication.introduction) %>
-      <% end %>
-    </div>
-  </section>
+
+  <% if @publication.introduction.present? %>
+    <section class="intro">
+      <div class="get-started-intro">
+        <%= render "govuk_publishing_components/components/govspeak", {
+          content: @publication.introduction.html_safe
+        } %>
+      </div>
+    </section>
+  <% end %>
+
   <%= render partial: 'location_form',
              locals: {
                method: 'post',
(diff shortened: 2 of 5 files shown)


app/views/local_transaction/results.html.erb
```diff
@@ -55,9 +55,9 @@
...
   <section class="more">
     <div class="more">
-      <%= render "govuk_publishing_components/components/govspeak", {} do %>
-        <%= sanitize(@publication.more_information) %>
-      <% end %>
+      <%= render "govuk_publishing_components/components/govspeak", {
+        content: @publication.more_information.html_safe
+      } %>
     </div>
   </section>
```
kevindew, line 62: Looking at the [schema](https://github.com/alphagov/govuk-content-schemas/blob/7d63f4f503c161e7756a867c02b58af4938dde1b/formats/local_transaction.jsonnet#L7-L10) this more_information field does not seem to be a required field. It's probably prudent to put this whole `<section>` element within a conditional to reflect that.

app/views/local_transaction/search.html.erb
```diff
@@ -29,14 +29,14 @@
...
   <% end %>
     <div class="more">
-      <%= render "govuk_publishing_components/components/govspeak", {} do %>
-        <%= sanitize(@publication.more_information) %>
-      <% end %>
+      <%= render "govuk_publishing_components/components/govspeak", {
+        content: @publication.more_information.html_safe
+      } %>
     </div>
   </section>
```
kevindew, line 41: As previous comment, none of these seem to be required fields in https://github.com/alphagov/govuk-content-schemas/blob/7d63f4f503c161e7756a867c02b58af4938dde1b/formats/local_transaction.jsonnet#L7-L10 so should probably put their outer containers within conditionals (this may be a theme for this review)

app/views/place/show.html.erb
```diff
@@ -45,12 +45,12 @@
...
-        <% end %>
+        <%= render "govuk_publishing_components/components/govspeak", {
+          content: @publication.need_to_know.html_safe
+        } %>
+        <%= render "govuk_publishing_components/components/govspeak", {
+          content: @publication.more_information.html_safe
+        } %>
       </div>
     </section>
   <% end %>
```
kevindew, line 56: Similar story here with the place schema, where seemingly none of the fields are marked as required: https://github.com/alphagov/govuk-content-schemas/blob/7d63f4f503c161e7756a867c02b58af4938dde1b/formats/place.jsonnet#L4-L6 which is really unhelpful.

kevindew (commented):
Thanks Graham - looks great. I think the commit could use a bit more explanation about what sanitize does and why it was the wrong choice. Since this is something that has been reverted back it'd be good for the documentation to be clear as to why.

Looks like a few amends are needed for local-transaction and place templates. The schemas for those seem disappointingly short on required fields which probably means they don't accurately represent the data that is required in mainstream publisher which is annoying. But because the schemas lack that info we should conform to them as it's likely that test failures are due to randomly generated content items based on those schemas.

app/views/local_transaction/results.html.erb
gclssvglx: Thanks @kevindew - will fix.

app/views/local_transaction/search.html.erb
gclssvglx: Thanks @kevindew - will fix.

app/views/place/show.html.erb
gclssvglx: Thanks @kevindew - will fix.

kevindew (approved):
Nice one looks good to me

PR #2471, opened 2020-09-18
[DO NOT MERGE] Performance platform page

ChrisBAshton:
Awaiting content. Opening PR early for review. If all is good, we should just be able to add a final commit swapping in the confirmed content, then merge.

Trello: https://trello.com/c/em2mpzVT/2144-3-create-static-page-and-route-for-https-wwwgovuk-performance

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -27,6 +27,7 @@ Rails.application.routes.draw do
 
   # Performance platform pages
   get "/performance" => "performance#index"
+  get "/performance/*other" => "performance#redirect_to_index"
 
   # Done pages
   constraints FormatRoutingConstraint.new("completed_transaction") do
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -25,6 +25,9 @@ Rails.application.routes.draw do
   get "/tour", to: "help#tour"
   get "/help/cookies", to: "help#cookie_settings"
 
+  # Performance platform pages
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -25,6 +25,10 @@
   get "/tour", to: "help#tour"
   get "/help/cookies", to: "help#cookie_settings"
 
+  # Performance platform pages
+  get "/performance" => "performance#index"
+  get "/performance/*other" => "performance#redirect_to_index"
```
thomasleese, line 30: An alternative to this might be to publish two special routes. One exact route to`/performance` and one prefix route to `/performance` which redirects to `/performance`. This means the redirect will be handled by router, so wouldn't need to go so far into our stack, but I'm not sure how router would handle two routes at `/performance` one exact and one prefix. In some ways I like that the logic is contained in this app rather than "hidden" away in router, so I don't really mind which way we decide to go, just wanted to present it as an option.

thomasleese (approved):

config/routes.rb
ChrisBAshton: Thanks - hadn't thought of that. I like the idea of it being slightly more efficient, but code cohesion probably better here 👍

ChrisBAshton:
https://www.gov.uk/performance/ is retired and redirects to a page on data.gov.uk.

PR #2472, opened 2020-09-21
Fix not running half the tests

benthorner:
Please see the commits for more details.


Rakefile
```diff
@@ -5,10 +5,19 @@ require File.expand_path("config/application", __dir__)
 require "rake"
 require "ci/reporter/rake/test_unit" if Rails.env.development? || Rails.env.test?
 require "ci/reporter/rake/rspec" if Rails.env.development? || Rails.env.test?
+require "rake/testtask"
 
 Frontend::Application.load_tasks
 
+# RSpec goes and clobbers the normal :test task without asking, so we have to rename it.
```
ChrisBAshton, line 12: It's not hugely clear why we're not just running the RSpec `:test` task. Why isn't it serving its purpose?

Is it because we have files under `test/` rather than `spec/`? A better approach might be to create a `.rspec` configuration file with contents `--pattern test/**/*_test.rb`.

benthorner: Aha, so the surprising (/WTF) thing about this repo is that it uses two, incompatible test frameworks for its unit tests: RSpec (in `spec/`) and Minitest (`test/`). You can easily see this by trying to run `bundle exec rspec test` and watch it error away. The problem is that these frameworks don't play well together in a Rails app, which has resulted in RSpec clobbering the Minitest ones, so they don't actually run as part of `rake default`.

Does that help? I could tweak the commit/comment to emphasise just how surprising this repo is!

ChrisBAshton: Thanks, that helps, but raises more questions!

So I think `:spec` and `:test` were originally supposed to run the RSpec and Minitest tests respectively? But running them individually, I see that `rake spec` does nothing, and `rake test` runs RSpec, making 922 assertions.

Have just tried out this branch and it looks like:

- `rake spec` still does nothing
-  I can't run `rake test_unit` directly ("Don't know how to build task 'test_unit'")
- `rake` doesn't combine the coverage ("Coverage report generated for Integration Tests, RSpec to /govuk/frontend/coverage. 1543 / 2633 LOC (58.6%) covered.", then several lines later: "Coverage report generated for RSpec to /govuk/frontend/coverage. 906 / 1528 LOC (59.29%) covered.").

And I'm still a bit confused about whether the Minitest tests are even running 🤔

benthorner: Yeah, it's all kinds of confusing. I've figured out some things:

- I've amended the first commit (`default` was being cleared correctly after all).
- I don't know how `rake test` runs RSpec, but it makes sense if RSpec assumes it's "THE framework".
- The `test` task gets removed from `default` by rspec-rails, so you'd never see RSpec running twice.
- `rake spec` was just broken. I've added another commit to fix it.
- `rake test_unit` works for me. `rake` does weird caching sometimes - you may want to wipe any containers.
- `rake` does combine coverage, but you may need to delete the `coverage` folder first, to wipe the corrupt state.

You should be convinced both sets of tests are running when you see the two sets of test output: both dots, but the Minitest dots are ~3 more numerous. Hopefully a bit less confusing now (not that this repo makes much sense!)?

Rakefile
```diff
@@ -20,4 +20,7 @@ end
 # RSpec shoves itself into the default task without asking, which confuses the ordering.
 # https://github.com/rspec/rspec-rails/blob/eb3377bca425f0d74b9f510dbb53b2a161080016/lib/rspec/rails/tasks/rspec.rake#L6
 Rake::Task["default"].clear
-task default: [:lint, :spec, :test_unit, "jasmine:ci"]
+
+# Combined coverage only works if RSpec tests come last. Having two test frameworks sucks.
```
ChrisBAshton, line 24:
```suggestion
# Combined coverage only works if RSpec tests come _after_ Minitest. Having two test frameworks sucks.
```

(`jasmine:ci` is the last task here so the comment is a little confusing!)

ChrisBAshton (approved):
Good bit of investigation, there! Thanks Ben. ⭐

Rakefile
benthorner: Sure, I've amended that 👍

benthorner:
Thanks for taking another look @ChrisBAshton! I've had to add another commit on to the end, so this can go in - would you mind checking it's OK?

ChrisBAshton (approved):

PR #2473, opened 2020-09-21
Add postal code autocomplete attribute

owenatgov:
## What
Add the `postal-code` autocomplete attribute to the postcode form partial. [Example form usage](https://www.gov.uk/find-local-council).

## Why
Without the autocomplete attribute, this fails WCAG 1.3.5 because users with cognitive or motor impairments will find it more difficult to fill in forms.

No visual changes.

**Card:** https://trello.com/c/aQzSeo2n/408-postcode-form-missing-autocomplete-attribute


injms (approved):
Super 👍

PR #2477, opened 2020-09-23
Remove code related to sending emails

thomasleese:
See individual commits for more details, but we no longer send emails in this application so we don't need this code anymore.

I did this work because recently we noticed this app was using both RSpec and Minitest, and we're trying to stick with just one testing framework. Rather than convert the tests for the notify delivery method into Minitest, I thought we should just remove the code.

This was missed when we removed the code that did use Notify: https://github.com/alphagov/frontend/pull/2263


benthorner (commented):
Hooray, I noticed this when doing my audit of our apps, but decided it was too much of a distraction to go and remove all the code. I think we can remove a couple of gems, too?

benthorner (commented):
Hooray, I noticed this when doing my audit of our apps, but decided it was too much of a distraction to go and remove all the code. I think we can remove a couple of gems, too?

thomasleese:
@benthorner Yep, good point! I've added a commit which removes the Gem.

Gemfile
```diff
@@ -10,7 +10,6 @@ gem "govuk_publishing_components"
 gem "govuk_sidekiq"
```
benthorner, line 10: Thanks @thomasleese - I realise I should have explicitly mentioned this one can go too.

thomasleese: Good point... that's gone too now!

benthorner (approved):

PR #2478, opened 2020-09-23
Convert Calendar tests to Minitest

thomasleese:
This effectively reverts https://github.com/alphagov/calendars/pull/852 and converts all the tests for the calendars pages from RSpec back to Minitest (sorry @barrucadu).

The reason for doing this is because currently this application has both Minitest and RSpec which is confusing for developers. It can also cause problems where CI only runs half of the tests: https://github.com/alphagov/frontend/pull/2472

Once this and #2477 are merged we can remove RSpec from this repo.


test/functional/calendar_controller_test.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  context "GET 'calendar'" do
+    setup do
+      Calendar.stubs(:find).returns(Calendar.new("something", "title" => "Brilliant holidays!", "divisions" => []))
+      stub_content_store_has_item("/bank-holidays")
+      stub_content_store_has_item("/when-do-the-clocks-change")
+    end
+
+    context "HTML request (no format)" do
+      should "load the calendar and shows it" do
```
barrucadu, line 15: `show` not `shows`

test/unit/presenters/calendar_content_item_test.rb
```diff
@@ -0,0 +1,37 @@
+require "test_helper"
+require "govuk-content-schema-test-helpers/test_unit"
+
+class CalendarContentItemTest < ActiveSupport::TestCase
```
barrucadu, line 4: Should this file (introduced in commit 770219ab194d2d61e99c00b728ed5c0b79a319d4) have more tests?  It's missing `it "contains the correct locale"` and `context "localised content items"`

thomasleese:
@barrucadu Thanks, I think your comments should be resolved now!

barrucadu (approved):

PR #2486, opened 2020-10-12
Remove RSpec

thomasleese:
RSpec was added in https://github.com/alphagov/frontend/pull/2010 and subsequently used for two test suits which have since [been removed](https://github.com/alphagov/frontend/pull/2477) or [converted to Minitest](https://github.com/alphagov/frontend/pull/2478). Removing RSpec avoids potential issues where we have two test frameworks in the same repo and [only some of the tests run in CI](https://github.com/alphagov/frontend/pull/2472).


barrucadu (approved):

Rakefile
```diff
@@ -4,23 +4,7 @@
...
-  t.warning = false
-end
-
-# RSpec shoves itself into the default task without asking, which confuses the ordering.
-# https://github.com/rspec/rspec-rails/blob/eb3377bca425f0d74b9f510dbb53b2a161080016/lib/rspec/rails/tasks/rspec.rake#L6
-Rake::Task["default"].clear
-
-# Combined coverage only works if RSpec come after Minitest. Using two frameworks sucks.
-# https://github.com/simplecov-ruby/simplecov#merging-results
-task default: [:lint, :test_unit, :spec, "jasmine:ci"]
```
benthorner, line 26: I think we need to keep `:lint` and `:jasmine:ci`, else they won't run?

thomasleese: They're still there, and I've tested that they run: https://github.com/alphagov/frontend/blob/a8afbdf89f66d7a81be151dbcb9c7782d788d8b9/Rakefile#L10

benthorner: Ahh, sorry - I missed the "green" line in the diff. Thanks for checking again 👍.

PR #2492, opened 2020-10-16
Add new promo image to the homepage

sihugh:
I've compressed it, though it doesn't seem to want to go that small, possibly because of the diagonal lines

The image is 600 x 400

## Before 

Screenshot 2020-10-16 at 18 30 52

## After
Screenshot 2020-10-16 at 18 31 28

https://trello.com/c/mh5Po2iB/530-add-new-promo-box-image-on-homepage


hannako (approved):
Pretty snaz

PR #2493, opened 2020-10-19
Update new promo image to use better compression and responsive sizes.

Nooshu:
Optimised the new "Time is running out image" and adopted the responsive image strategy. 63% reduction in file size at 600 width.

image size

Graph shows the image size on the homepage over the past 7 days.


andysellick (approved):
Looks good. Just curious - why is `480w` set as the default?

Nooshu:
> Looks good. Just curious - why is `480w` set as the default?

Great question. I was just following the example already set on the live page. For mobile first it would make sense for `600w` since the image tops out at 592px before switching. Maybe @injms will know more? It may just be a mistake.

I've also just noticed I've missed the `sizes` attribute, so will add that in.

sihugh:
Thanks @Nooshu, this was on my list for today 🙌

injms:
> > Looks good. Just curious - why is `480w` set as the default?
> 
> Great question. I was following the example already set on the live page. For mobile first it would make sense for `600w` since the image tops out at 595px before switching. Maybe @injms will know more? It may just be a mistake.
> 
> I've also just noticed I've missed the `sizes` attribute, so will add that in.

I originally chose 480 wide since I thought it was the right size to get a good balance larger and smaller screens - not too much downsizing for desktop, and not too much upscaling for mobile. 

Looking at browser stats from GA and the [Can I Use information for `srcset`](https://caniuse.com/srcset), browsers that can't use `srcset` are mostly desktop browsers (although of course this doesn't indicate a users window size) - so it'd probably be a safe bet to use the `320w` image instead.

PR #2495, opened 2020-10-20
Remove previous clock changes

chao-xian:
### Before
www gov uk_when-do-the-clocks-change(iPad Pro)

### After
govuk-fronte-remove-pre-w7qhym herokuapp com_when-do-the-clocks-change(iPad Pro)


injms (approved):
🕑 👍

hannako (approved):

PR #2508, opened 2020-11-03
Update Govspeak component usage

injms:
## What

Updates any Govspeak component that was using the `content` parameter to use a `do` block.

## Why
So things are fixed. See https://github.com/alphagov/govuk_publishing_components/pull/1632 for the cause of this change.

## Visual changes

None.

### DIFF
--- a/app/views/local_transaction/search.html.erb
+++ b/app/views/local_transaction/search.html.erb
@@ -13,9 +13,9 @@
   <% if @publication.introduction.present? %>
     <section class="intro">
       <div class="get-started-intro">
-        <%= render "govuk_publishing_components/components/govspeak", {
-          content: @publication.introduction.html_safe
-        } %>
+        <%= render "govuk_publishing_components/components/govspeak", {} do %>
+          <%= raw @publication.introduction %>
+        <% end %>
       </div>
     </section>
   <% end %>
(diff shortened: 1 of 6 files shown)


app/views/local_transaction/search.html.erb
```diff
@@ -13,9 +13,9 @@
   <% if @publication.introduction.present? %>
     <section class="intro">
       <div class="get-started-intro">
-        <%= render "govuk_publishing_components/components/govspeak", {
-          content: @publication.introduction.html_safe
-        } %>
+        <%= render "govuk_publishing_components/components/govspeak", {} do %>
+          <%= raw @publication.introduction.html_safe %>
```
maxgds, line 17: is this effectively doing .html_safe twice at the moment? I think you can drop it inside the raw

app/views/simple_smart_answers/_current_question.html.erb
```diff
@@ -8,7 +8,7 @@
     end %>
 
   <% description = render "govuk_publishing_components/components/govspeak", {
-    content: question.body.html_safe
+    content: question.body.html_safe,
```
maxgds, line 11: not sure what's going on here? no do block and still using content param

maxgds (changes requested):
Not sure if this was maybe not ready for review but looks like it needs some tweaks.

app/views/local_transaction/search.html.erb
injms: Fixed

app/views/simple_smart_answers/_current_question.html.erb
injms: Good catch - thank you.

maxgds (approved):
🚢

PR #2513, opened 2020-11-09
Update the Wales link on travel advice call out

leenagupte:
Trello: https://trello.com/c/iS4WvL1j

Updates the link to Wales restrictions on https://www.gov.uk/foreign-travel-advice

We changed the link for wales information to:
https://gov.wales/coronavirus-firebreak-frequently-asked-questions

We should change it back to https://gov.wales/local-lockdown


JonathanHallam (approved):
Looks good to me!

PR #2520, opened 2020-11-12
Add 2022 dates

benjamineskola:
Standard generated bank hliday dates plus special case for royal occasion.

https://trello.com/c/tzHLWwhE/2227-add-2022-bank-holidays-and-clock-change-dates


AlanGabbianelli (approved):
🎉

PR #2536, opened 2020-12-04
Add cookieless test

karlbaker02:
This PR adds a new Cookieless A/A test, which sends hit-only data to analytics up until cookies have been accepted by the user. This is necessary to understand whether we can rely on the approach offered by our CDN to reliably assign users to a variant of an A/B test and maintain that variant, such that we'd expect to see an even distribution of users across all active variants of a test (or according to any prescribed percentages of variants).

**This PR should not be merged until we're ready to turn on the A/A test.**

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -28,6 +28,10 @@
     <% if @meta_section %>
       <meta name="govuk:section" content="<%= @meta_section %>">
     <% end %>
+    <%= cookieless_variant.analytics_meta_tag.html_safe %>
+    <% unless cookieless_variant.variant?('Z') %>
+      <meta name="Cookieless-Variant" content="<%= cookieless_variant.variant_name %>" data-module="track-variant">
+    <% end %>
   </head>
 
   <body class="mainstream <%= yield :body_classes %>">
--- a/app/assets/javascripts/modules/cookieless-tracker.js
+++ b/app/assets/javascripts/modules/cookieless-tracker.js
@@ -1,0 +1,116 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  var CookielessTracker = function (trackingId, fieldsObject) {
+    var trackerName = fieldsObject.name + '.'
+
+    function configureProfile () {
+      // https://developers.google.com/analytics/devguides/collection/analyticsjs/command-queue-reference#create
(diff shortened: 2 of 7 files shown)


app/views/layouts/application.html.erb
```diff
@@ -28,10 +28,15 @@
     <% if @meta_section %>
       <meta name="govuk:section" content="<%= @meta_section %>">
     <% end %>
+    <%= cookieless_variant.analytics_meta_tag.html_safe %>
```
DilwoarH, line 31: Where is this coming from? I wasn't able to trace it

karlbaker02: Comes from the `cookieless_testable` concern

app/views/layouts/application.html.erb
```diff
@@ -28,10 +28,15 @@
...
   </head>
 
   <body class="mainstream <%= yield :body_classes %>">
     <div id="wrapper" class="<%= wrapper_class(@publication || @presenter) %>">
+      <% unless cookieless_variant.variant?('Z') %>
+        <meta data-variant="<%= cookieless_variant.variant_name %>" data-module="track-variant">
```
DilwoarH, line 37: Is there a reason why this meta tag isn't in the head too?

karlbaker02: Good point - it makes me think actually that the previous analytics tag above is redundant. Let me check and get back to you

karlbaker02: So I've double checked and we do need both of the `meta` tags in this commit.

The first on line 31 is going to be picked up by the analytics code in Static and will be used to include this test as a custom dimension when any existing events are sent to GA. This is necessary when the existing hit event is sent to GA if users accept cookies, as we want to include the variant of this test that the user was in when they accepted.

The meta tag on line 37 is used by the track-variant.js file to get the variant of the test that the user is seeing. I included this as a separate meta tag specifically to be able to name the tag and easily get the value out - this isn't possible with the meta tab defined on line 31, which is of the form:

```
<meta name="govuk:ab-test" content="EducationNav:A" data-analytics-dimension="207" data-allowed-variants="A,B">
```

DilwoarH:
A Review date has been put in for 27th of Jan 2021 to decide on how we will approach this going forward.

DilwoarH (approved):

DilwoarH:
A Review date has been put in for 27th of Jan 2021 to decide on how we will approach this going forward.

PR #2537, opened 2020-12-04
Change aria-live region to be less verbose

injms:
## What
Change the aria-live region on the foreign travel advice page - the region is now only wraps the number of countries and territories, not the entire list of countries and territories.

## Why
Previously the entire list of countries was being read out as you typed into the search input. This is too verbose and led to the page being nearly unusable when using JAWS.

For example, if you were searching for "Australia" all the countries with "A" in their name would be read out (all 198 of them), then all the countries with "Au" (9), then the two with "Aus" in them.

This has been changed to only read out the number of countries or territories that found for a search result. This means that searching for "Australia" will cause "198 countries or territories" to be read out, then "9 countries or territories", then "Two countries or territories".

## Before and after

Before - note the `aria-live="polite"` is on the highlighted element that wraps the entire list of countries:

Screenshot 2020-12-04 at 17 18 34

After - the `aria-live` attribute is on the smaller element that wraps the number of countries:

Screenshot 2020-12-04 at 17 18 08

### DIFF
--- a/app/assets/javascripts/views/travel-advice.js
+++ b/app/assets/javascripts/views/travel-advice.js
@@ -1,4 +1,5 @@
 /* global GOVUK */
+
 (function () {
   'use strict'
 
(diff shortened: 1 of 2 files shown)


chris-gds (approved):
The change sounds / behaves fine to me, as you type in search it reads back the number at the next available moment at expected.

PR #2550, opened 2020-12-14
Add Test and Trace Payment Service nation validation

gclssvglx:
## What

The Test and Trace self isolation payment is a service delivered by local councils.  However, it isn't available in Scotland and Northern Ireland. 

User's currently searching with a postcode for those nations are incorrectly told to search their local council site for more information. 

## Why

We'd like to explicitly tell users that the service isn't available in the country to which the postcode they have entered relates, and give them a link to the national COVID page.

[Trello](https://trello.com/c/QSY1CoAB/643-test-and-trace-nation-validation)

### DIFF
--- a/app/controllers/local_transaction_controller.rb
+++ b/app/controllers/local_transaction_controller.rb
@@ -16,12 +16,17 @@ class LocalTransactionController < ApplicationController
   def search
     if request.post?
       @location_error = location_error
+
       if @location_error
         @postcode = postcode
-      elsif mapit_response.location_found? && lgsl == 364 && country_name == "Northern Ireland" ## LGSL code 364 = electoral registration
-        redirect_to local_transaction_results_path(local_authority_slug: "electoral-office-for-northern-ireland")
-      elsif mapit_response.location_found? && local_authority_slug
-        redirect_to local_transaction_results_path(local_authority_slug: local_authority_slug)
+      elsif mapit_response.location_found?
+        slug = if lgsl == 364 && country_name == "Northern Ireland"
+                 "electoral-office-for-northern-ireland"
+               else
+                 local_authority_slug
+               end
+
+        redirect_to local_transaction_results_path(local_authority_slug: slug)
       end
     end
   end
@@ -30,6 +35,13 @@ class LocalTransactionController < ApplicationController
     @postcode = postcode
     @interaction_details = interaction_details
     @local_authority = local_authority
+    @country_name = @local_authority.country_name
+
+    if LocalTransactionServices.instance.unavailable?(lgsl, @country_name)
+      render :unavailable_service
+    else
+      render :results
+    end
   end
 
 private
--- a/app/views/local_transaction/unavailable_service.html.erb
+++ b/app/views/local_transaction/unavailable_service.html.erb
@@ -1,0 +1,40 @@
+<%= render layout: 'shared/base_page', locals: {
+  title: "This service is not available in #{@country_name}",
+  publication: @publication,
+  edition: @edition,
+} do %>
+  <div class="interaction">
+    <p class="govuk-body">
+      We’ve matched the postcode to <span class="local-authority"><%= @local_authority.name %></span>.
+    </p>
+
+    <p class="govuk-body">
+      This service is not available in <span class="country-name"><%= @country_name %></span>. You can find other services on the <span class="local-authority"><%= @local_authority.name %></span> website.
+    </p>
+
+    <p id="get-started" class="get-started group">
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Find other services",
+        rel: "external",
+        margin_bottom: true,
+        href: @local_authority.url,
+      } %>
+    </p>
+  </div>
+
+  <div class="search-again">
+    <%= render "govuk_publishing_components/components/back_link", {
+      href: local_transaction_search_path(@publication.slug),
+    } %>
+  </div>
+
+  <% if @publication.more_information.present? %>
+    <section class="more">
+      <div class="more">
+        <%= render "govuk_publishing_components/components/govspeak", {} do %>
+          <%= raw @publication.more_information %>
+        <% end %>
+      </div>
+    </section>
+  <% end %>
+<% end %>
--- a/app/models/local_transaction_services.rb
+++ b/app/models/local_transaction_services.rb
@@ -1,0 +1,18 @@
+require "singleton"
+
+class LocalTransactionServices
+  include Singleton
+
+  def initialize
+    yaml_file_path = if Rails.env.test?
+                       "test/fixtures/unavailable_services.yml"
+                     else
+                       "config/unavailable_services.yml"
+                     end
+    @config = YAML.safe_load(File.open(Rails.root.join(yaml_file_path)))
+  end
+
+  def unavailable?(lgsl, country_name)
+    @config.dig("services", lgsl).present? && @config.dig("services", lgsl).include?(country_name)
+  end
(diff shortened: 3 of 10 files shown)


app/models/unavailable_service.rb
```diff
@@ -0,0 +1,21 @@
...
+    @country = country
+    @config = config
+    @service_path = ["services", lgsl]
+  end
+
+  def unavailable?
+    config.dig(*@service_path).present? && config.dig(*@service_path).include?(country)
+  end
+
+  def self.load_config
```
theseanything, line 18: Am I right in assuming a UnavailableService object represents a single service? Might be cleaner and easier to understand responsibilities, if we extract the information about all service from the object.

We could follow the pattern in the Brexit checker, where there is a class method that returns an array of objects. But each object is only aware of itself i.e. we don't store the entire config file for all object on each object. This would simplify things, as the config attribute is only for the particular service and you don't need to dig for it and you wouldn't need the derived service_path attribute.

https://github.com/alphagov/finder-frontend/blob/979b344ab47708a5e2519d5e7b16e80eac5b1f0f/app/lib/brexit_checker/action.rb#L46

Also unclear if this object is scoped per country? I don't think it needs to be if we made the `unavailable?` method accept a country name param

theseanything: Also it would be good to memoize the file (like the Brexit checker) so it remain in memory. We don't have to load the file for every request.

app/models/unavailable_service.rb
```diff
@@ -0,0 +1,21 @@
+class UnavailableService
```
theseanything, line 1: I'm worried that this name might be a bit confusing relative to the rest of the codebase. As frontend service many type of pages, a developer might not understand what a "service" is in this context. It's also might be confusing as Service is used as Rails patterning for describing interactions with third parties.

```suggestion
class LocalTransactionService
```

theseanything: Also maybe using the word "Unavailable" could be confusing as we have the method unavailable too? If the class represents a unavailable service, why would we have to check if its unavailable?

app/views/local_transaction/unavailable_service.html.erb
```diff
@@ -0,0 +1,41 @@
...
+        href: @interaction_details["local_interaction"]["url"],
+      } %>
+    </p>
+  </div>
+
+  <div class="search-again">
+    <%= render "govuk_publishing_components/components/back_link", {
+      href: local_transaction_search_path(@publication.slug),
+    } %>
+  </div>
```
theseanything, line 30: Think this should be above the body content on line 6

app/views/local_transaction/unavailable_service.html.erb
```diff
@@ -0,0 +1,41 @@
...
+
+    <p class="govuk-body">
+      This service is not available in <span class="country-name"><%= @country_name %></span>. You can find other services on the <span class="local-authority"><%= @local_authority.name %></span> website.
+    </p>
+
+    <p id="get-started" class="get-started group">
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Find other services",
+        rel: "external",
+        start: true,
```
theseanything, line 19: Think we can remove this - as its not a start button.

theseanything (changes requested):

app/views/local_transaction/unavailable_service.html.erb
gclssvglx: This change would also be required in `https://github.com/alphagov/frontend/blob/master/app/views/local_transaction/results.html.erb` for consistency.  

Are you happy for me to do this as a separate commit as part of this PR, or would you prefer this change to be a completely separate story/PR?

This also applies to https://github.com/alphagov/frontend/pull/2550#discussion_r543416633

app/views/local_transaction/unavailable_service.html.erb
gclssvglx: See https://github.com/alphagov/frontend/pull/2550#discussion_r543416233

app/models/unavailable_service.rb
gclssvglx: Hi @theseanything - thanks for the review and your comments.  I'll definitely be looking into memoizing the file as you suggest and I agree with your suggested change to the `unavailable?` method.

However, I don't believe it does represent a single service - no.  It's more like - all local transaction services that are unavailable in the named countries as defined in the [config](config/unavailable_services.yml) file.  I understand the current name is a bit confusing, and your request below to rename it below.  I agree there must be a better name.  How would you feel about `UnavailableLocationTransactionServices` - bit of a mouthful, but is definitely more illustrative of my intent, its use and purpose?

My reasoning for encapsulating the config into an object like this, is that I wanted change or add the minimal amount of code - especially in the `LocalTransactionController`.

app/models/unavailable_service.rb
gclssvglx: I agree - there MUST be a better name 👍

app/models/unavailable_service.rb
gclssvglx: Actually, the more I think about this class and your suggestion to memoize the file - the more this looks like a `singleton`.

app/views/local_transaction/unavailable_service.html.erb
theseanything: I don't think we are dependent on the results page being correct and we don't need to fix results page as part of the PR. As this is an entirely new page it probably should be correct from the start. It'd be better that we keep consistency with the rest of GOV.UK, rather than consistency with the results page thats going to be fixed anyway.

app/views/local_transaction/unavailable_service.html.erb
gclssvglx: Removed from `unavailable_service` page only

app/views/local_transaction/unavailable_service.html.erb
gclssvglx: Moved on `unavailable_service` page only

app/models/unavailable_service.rb
gclssvglx: Renamed as per suggestion

app/models/unavailable_service.rb
gclssvglx: Class `LocalTransactionService` is now a Singleton.  Config file loaded only once.

app/controllers/local_transaction_controller.rb
```diff
@@ -16,20 +16,32 @@ class LocalTransactionController < ApplicationController
...
+               end
+
+        redirect_to local_transaction_results_path(local_authority_slug: slug)
       end
     end
   end
 
   def results
-    @postcode = postcode
+    @postcode = params[:postcode]
```
theseanything, line 35: why not?
```suggestion
    @postcode = postcode
```

app/models/local_transaction_service.rb
```diff
@@ -0,0 +1,18 @@
+require "singleton"
+
+class LocalTransactionService
```
theseanything, line 3:
```suggestion
class LocalTransactionServices
```

As it can represent multiples service

app/controllers/local_transaction_controller.rb
gclssvglx: Ah, meant to fix that - apologies - fixed

app/models/local_transaction_service.rb
gclssvglx: Renamed as requested

theseanything (approved):

PR #2556, opened 2020-12-19
Foreign travel advice call out update

kevindew:
This is changed to reflect updated guidance from FCDO on December 19th.

## Before

Screenshot 2020-12-19 at 22 28 09

## After

Screenshot 2020-12-19 at 22 29 35


JonathanHallam (approved):
🎉

PR #2558, opened 2020-12-21
Update to use components

chris-gds:
## What?

As part of on-going work, update to `frontend` to use components when appropriate.

## Why?

To improve a11y ensuring consistency of component architecture
[Trello ticket](https://trello.com/c/wk438rpt)

## Visuals

Live Link: https://www.gov.uk/register-to-vote

Before (Left) After (Right)
1-mobile

1-desktop

---

https://www.gov.uk/report-child-abuse-to-local-council

2-mobile

2-desktop

---

https://www.gov.uk/find-local-council/warwick

3-mobile

3-desktop

---

https://www.gov.uk

Note `margin-top` visual change on mobile, this is inline with the [Design system spacing](https://design-system.service.gov.uk/styles/spacing/)
- if preferred desktop can visually update instead. 

4-mobile

4-desktop

## Anything else?

- [GDS Stylelint](https://github.com/alphagov/stylelint-config-gds) installed, config added and run on app.
- update to redirected link within `README`

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -31,11 +31,6 @@ body.homepage {
   position: relative;
 }
 
-.home-top__title {
-  @include govuk-font(48, $weight: bold);
-  padding: 25px 0 15px;
-}
-
 .home-top__intro {
   margin: 0;
 }
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -3,6 +3,7 @@ body.homepage {
   #global-header-bar {
     display: none;
   }
+
   #wrapper,
   #content {
     margin: 0;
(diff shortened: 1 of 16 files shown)


app/assets/stylesheets/views/_homepage.scss
```diff
@@ -139,13 +135,13 @@ body.homepage {
 }
 
 .home-divider {
-  margin: 30px 0 15px 0;
+  margin: 30px 0 15px;
```
owenatgov, line 138: You could use `govuk-spacing` here for those margin values and for the other instances on line 148 to keep things consistent. [Details](https://design-system.service.gov.uk/styles/spacing/#spacing-on-custom-components). I think this would look like...
```suggestion
  margin: govuk-spacing(6) 0 govuk-spacing(3);
```

It's more verbose but protects us from any future (though maybe unlikely) spacing changes.

owenatgov (commented):
One little nitpick but looking good, almost there 👍

app/assets/stylesheets/views/_homepage.scss
chris-gds: Works for me, updated 👍🏼

owenatgov (approved):
🚀

PR #2562, opened 2020-12-30
Update Brexit homepage promo image and text

DilwoarH:
https://trello.com/c/WV7GB49l/752-deploy-31-12-evening-update-homepage-promo-box-6

This is done to make the promotion image and text in line with new post Brexit messaging

## Before

Screenshot 2020-12-30 at 15 06 21

## After

Screenshot 2020-12-30 at 15 05 41


app/views/homepage/index.html.erb
```diff
@@ -109,29 +109,21 @@
...
-            <%= image_tag "homepage/time-is-running-out--480w.jpg", {
+            <%= image_tag "homepage/transition-period.png", {
               alt: "",
               class: "home-promo__image",
               height: 400,
               loading: "lazy",
-              width: 600,
-              sizes: "(max-width: 640px) 100vw, (max-width: 1020px) 33vw, 300px",
-              srcset: {
-                "homepage/time-is-running-out--600w.jpg": "600w",
```
huwd, line 120: no srcset for this one?
Is this less optimised for mobile?

app/views/homepage/index.html.erb
```diff
@@ -109,29 +109,21 @@
 
         <div class="govuk-grid-column-one-third">
           <a href="/transition" aria-hidden="true" tabindex="-1">
-            <%= image_tag "homepage/time-is-running-out--480w.jpg", {
+            <%= image_tag "homepage/transition-period.png", {
```
benthorner, line 112: Note from our discussions: this is still in the repo despite not being used (see [here](https://github.com/alphagov/frontend/blob/master/app/assets/images/homepage/transition-period.png)); we agreed to stick with the single variant, noting it's only ~5KB in size!

benthorner (commented):
Looks good - thanks for your help with this 🥇. Only thing I think we're missing is to remove all the redundant "time-is-running-out" files. Can we do that as part of this?

app/views/homepage/index.html.erb
DilwoarH: Removed it because the image we have is ~5kb - we might add in various sized variants if we need to afterwards

DilwoarH:
> Looks good - thanks for your help with this 🥇. Only thing I think we're missing is to remove all the redundant "time-is-running-out" files. Can we do that as part of this?

Done

benthorner (approved):
👍

benthorner:
Adding "DNM" as this is scheduled to go out tomorrow afternoon.

PR #2563, opened 2020-12-31
Cache content store requests

thomasleese:
This caches the response from `GdsApi.content_store.content_item` for 30 seconds. I've updated the various places where `GdsApi.content_store.content_item` is used to call `CachedContentItem.fetch` instead.

This should help us to handle spikes in traffic to certain pages by avoiding an unnecessary call to the Content Store. On the other hand, it does mean that newly published content won't necessarily be visible immediately, which is why I've set the cache time to a low 30 seconds.

There is a wider architectural question in this app related to when it fetches from the content store. At the moment for each request there will be two calls to the content store. First in the routes from the `ContentFormatInspector` and `FormatRoutingConstraint` and then a second time in the controller which is handling the request. It feels like this could be simplified to only attempt to fetch from the content store once however this is a bigger change as unpicking the bits of code will take time. This new `CachedContentItem` solves the problem by caching, but it doesn't actually stop the code from attempting to retrieve the content item twice.

I based this code off similar code we've got in Finder Frontend: https://github.com/alphagov/finder-frontend/blob/806aa066879c201fdbd9f9c3ab13331be2b4e143/app/lib/services.rb#L10-L16

[Incident report](https://docs.google.com/document/d/1GNk77PBcKBE7TFuUvJUnii_qOFNrifGGWut-z3EpdQY/edit)


lib/services.rb
```diff
@@ -1,17 +0,0 @@
...
-    )
-  end
-
-  def self.publishing_api
-    @publishing_api ||= GdsApi::PublishingApi.new(
-      Plek.new.find("publishing-api"),
-      bearer_token: ENV["PUBLISHING_API_BEARER_TOKEN"] || "example",
-    )
-  end
-end
```
kevindew, line 17: Yay great to see these go! 😁

kevindew (commented):
This looks good, nice speedy work! Naturally there's conversations going on about whether this is suitable elsewhere so I won't dig too much into the implementation. 

Couple of questions and thoughts:

Do we use the content store response for cache headers at all? I'd figure we lose that if we convert it to hash, but I'm guessing we don't given this works - which seems a bit suspect that this app may not be following content store cache headers 😢 

Other thing could be whether we can use middleware to set the content item as part of the request somewhat like: https://stackoverflow.com/questions/5192175/how-to-set-in-a-middleware-a-variable-accessible-in-all-my-application I'm not sure if there's any weird gotchas lurking in that (for example if it breaks a 404 handling).

Hmm life would certainly be simpler if content store cache headers could be trusted all the way through the stack and be low enough to not need manual invalidation.

thomasleese:
> Do we use the content store response for cache headers at all? I'd figure we lose that if we convert it to hash, but I'm guessing we don't given this works - which seems a bit suspect that this app may not be following content store cache headers 😢

Yes, as far as I can tell, the cache headers from the content store are not used anywhere in the app.

> Other thing could be whether we can use middleware to set the content item as part of the request somewhat like: https://stackoverflow.com/questions/5192175/how-to-set-in-a-middleware-a-variable-accessible-in-all-my-application I'm not sure if there's any weird gotchas lurking in that (for example if it breaks a 404 handling).

Yeah, middleware seems like the way to go here. In fact we already set some things in the `env` as part of the routing, but they're not used anywhere else. I'm going to have a go with a middleware solution now and see how far I get.

kevindew:
I'm assuming this is superseded by https://github.com/alphagov/frontend/pull/2564 - feel free to re-open if not and still relevant.

PR #2564, opened 2020-12-31
Only make one request to the content store

thomasleese:
This changes the controllers for most of the routes to use the existing content item which was fetched as part of the routing process to match the content item `schema_name` with the appropriate controller. This avoids making two requests to the content store for each request.

I wasn't able to do this in a middleware because the `slug` of the route needs to be known to fetch the item from the content store, and this only happens once the routes have been parsed. Instead I've slightly modified the `FormatRoutingConstraint` to store the content item in the request environment, and then updated the `Navigable` concern to pull that content item out of the request environment if it's there.

I've also taken the opportunity to rename the keys we're using to store these in the request environment so they're slightly easier to reason about: `:content_item` and `:content_item_error`.

This is an alternative to https://github.com/alphagov/frontend/pull/2563 but doesn't introduce any caching.

[Incident report](https://docs.google.com/document/d/1GNk77PBcKBE7TFuUvJUnii_qOFNrifGGWut-z3EpdQY/edit)


sihugh (approved):
Works for me 👍

lib/format_routing_constraint.rb
```diff
@@ -1,32 +1,27 @@
...
 
   def matches?(request)
-    @request = request
-    content_format == @format
+    content_item = set_content_item(request)
+    @format == content_item&.[]("schema_name")
   end
 
-private
+  def set_content_item(request)
```
theseanything, line 11: Nit: This method name is the same as [ApplicationController#set_content_item](https://github.com/alphagov/frontend/blob/adcddb8f8fff45e21c25b535baf7ba9c4c712e2b/app/controllers/application_controller.rb#L72), and is somewhat confusing as they do different things. (but probably worth renaming the other method instead... or this method could be something like `fetch_content_item` to make it more clear its requesting external data, rather than just setting an attribute)

theseanything (approved):

PR #2570, opened 2021-01-06
Optimise covid image.

Nooshu:
Small optimisation to the `covid-19-promo-stay-home.png` running it through [TinyPNG](https://tinypng.com/). 61% reduction, 14.6KB to 5.7KB.


andysellick (approved):
Brilliant. Glad you're watching this stuff @Nooshu! We should probably be a bit more diligent...

alex-ju:
Thank you, Matt! Will roll this out in production.

PR #2582, opened 2021-01-11
No longer render text as <strong> importance

Rosa-Fox:
The last line should not be `<strong>`, so this has been amended.

## Before

frontend-before

## After

frontend-after


GDSNewt (approved):
:shipit:

PR #2583, opened 2021-01-12
Update COVID call out box on travel advice index

huwd:
[Trello](https://trello.com/c/J01vLuo9/1042-display-different-travel-advice-content-based-on-country)

## What

Updates the index call out box to match both call out box for the
majority of country specific pages (PR in Government Frontend to
follow).

## Looks like
image


JonathanHallam (approved):
Excellent!

PR #2607, opened 2021-01-26
Allow custom content for 'Service unavailable for <country>' page

gclssvglx:
## What

Expand on the existing `unavailable_service` view by moving and abstracting the page content in to the `unavailable_services` YAML configuration file.

## Why
    
We would like to support the ability to customise the text on the 'Service unavailable for <country>' page.

[Trello](https://trello.com/c/vcsLBkOw/748-allow-custom-content-for-service-unavailable-for-country-page)

### DIFF
--- a/config/unavailable_services.yml
+++ b/config/unavailable_services.yml
@@ -1,4 +1,9 @@
 services:
   1826:
-    - Scotland
-    - Northern Ireland
+    countries:
+      unavailable_in:
+        - Scotland
+        - Northern Ireland
+      content: |
+        <p class="govuk-body">We've matched the postcode to <span class="local-authority">%{local_authority_name}</span>.</p>
+        <p class="govuk-body">This service is not available in %{country_name}. You can find other services on the <span class="local-authority">%{local_authority_name}</span> website.</p>
--- a/test/unit/models/local_transaction_services_test.rb
+++ b/test/unit/models/local_transaction_services_test.rb
@@ -10,4 +10,22 @@ class LocalTransactionServicesTest < ActiveSupport::TestCase
       assert_not LocalTransactionServices.instance.unavailable?(461, "Northern Ireland")
     end
   end
+
+  context ".content" do
+    should "return a string with the given country and local authority name in" do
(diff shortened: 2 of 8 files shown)


config/unavailable_services.yml
```diff
@@ -1,4 +1,9 @@
...
   1826:
-    - Scotland
-    - Northern Ireland
+    countries:
+      unavailable_in:
+        - Scotland
+        - Northern Ireland
+      content: |
+        <p class="govuk-body">We’ve matched the postcode to <span class='local-authority'>%{local_authority_name}</span>.</p>
+        <p class="govuk-body">This service is not available in %{country_name}. You can find other services on the <span class='local-authority'>%{local_authority_name}</span> website.</p>
```
theseanything, line 9: I think we have inconsistent quotes:
```suggestion
        <p class="govuk-body">This service is not available in %{country_name}. You can find other services on the <span class="local-authority">%{local_authority_name}</span> website.</p>
```

test/unit/models/local_transaction_services_test.rb
```diff
@@ -10,4 +10,10 @@ class LocalTransactionServicesTest < ActiveSupport::TestCase
       assert_not LocalTransactionServices.instance.unavailable?(461, "Northern Ireland")
     end
   end
+
+  context ".content" do
+    should "return a string with the given country and local authority name in" do
+      assert_equal "This service is unavailable in Dundee, Scotland", LocalTransactionServices.instance.content(461, "Scotland", "Dundee")
+    end
+  end
```
theseanything, line 18: Maybe with having a test for what happens if content is missing from the config. I.e. a test showing how we get an empty string back

theseanything (changes requested):

config/unavailable_services.yml
gclssvglx: Whoops!  Fixed - thanks.

test/unit/models/local_transaction_services_test.rb
gclssvglx: Ah, yes - good point thanks.  We should also account for no content at all, but with the key; and no content key.  Tests added.

theseanything (approved):

PR #2608, opened 2021-01-26
Add page for public GOVUK roadmap

danacotoran:
Create a public facing roadmap page as per the [designs](https://www.figma.com/file/AhvxMRBilODjqSgBbIqRiG/Roadmap?node-id=274%3A922).

The page will be available at gov.uk/roadmap

The roadmap gives on overview of what GOV.UK is working on. We expect it will be linked to from the footer and from a blog post announcing it.


app/views/roadmap/index.html.erb
```diff
@@ -0,0 +1,116 @@
...
+          <%= render "image" %>
+        </div>
+      </div>
+    </section>
+    <section class="govuk-roadmap-section govuk-!-padding-bottom-4 govuk-!-margin-bottom-6">
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <%= render "govuk_publishing_components/components/heading", {
+            text: t('roadmap.numbers_section.heading_large'),
+            margin_bottom: 0,
```
andysellick, line 25: Isn't the default for this component to have zero margin_bottom?

app/views/roadmap/index.html.erb
```diff
@@ -0,0 +1,116 @@
...
+              <%# Deliberate use of image tag as a fallback method https://lynn.ru/examples/svg/en.html %>
+              <image src="<%= asset_path('roadmap/crown.png') %>"></image>
```
andysellick, line 58: When this fallback is used does it need an `alt` tag or does the aria-hidden true still apply? (same for the other image)

app/views/roadmap/index.html.erb
```diff
@@ -0,0 +1,116 @@
...
+          </p>
+        </div>
+        <div class="govuk-grid-column-one-third">
+          <p class="govuk-body govuk-!-font-weight-bold govuk-!-font-size-24">
+            <span class="govuk-!-font-size-80 govuk-!-display-block">
+              <%= t('roadmap.numbers_section.numbers_columns.column_two_count') %></span><%= t('roadmap.numbers_section.numbers_columns.column_two_text') %>
+          </p>
+        </div>
+      </div>
+      <p class="govuk-body-s govuk-!-margin-bottom-0"><span aria-hidden="true">*</span><%= t('roadmap.numbers_section.footnote') %></p>
```
andysellick, line 69: This paragraph is using the spacing override to have no margin bottom, but the wrapping element has the spacing override to add what looks like the equivalent amount of padding. If both these classes were removed I think (?) it would look exactly the same.

app/views/roadmap/index.html.erb
andysellick: (see also line 76)

config/locales/en.yml
```diff
@@ -101,3 +101,129 @@ en:
...
+          items:
+          - text: How to expand and automate data reports
+          - text: Further ways to automate analysis of user feedback so we can identify trends more quickly
+      - heading: Everyone in the UK needs to understand how they are affected by coronavirus and Brexit
+        paragraph: Create tools and guidance that help users find critical and up-to-date information
+        steps:
+        - heading: Recently shipped
+          items:
+          - text: Continuous updates to the coronavirus landing page and tools to help users find guidance, services and goverment support for coronavirus
+          - text: Launch a service for the public to ask ministers questions about coronavirus (gov.uk/ask)
```
andysellick, line 197: Should this be a link? It probably isn't in the design, but it looks odd to me as text.

config/locales/en.yml
```diff
@@ -101,3 +101,129 @@ en:
...
+      heading: GOV.UK roadmap
+    numbers_section:
+      footnote: 'We can only count visits to the GOV.UK website if users give us permission to.'
+      heading_large: GOV.UK in numbers
+      heading_small: From April 2020 to January 2021
+      numbers_columns:
+        column_one_count: 235
+        column_one_text: million users
+        # the special characters are in place to keep this text on one line
+        column_two_text: "billion\u00A0page\u00A0views"
```
andysellick, line 117: This line is wrapped in quotes `"` but a few others are wrapped in `'`. Could this be made consistent throughout the file or is there a reason for this?

app/assets/stylesheets/application.scss
```diff
@@ -114,3 +114,21 @@ $govuk-use-legacy-palette: false;
...
+  border-bottom: 5px solid govuk-colour('black');
+}
+
+.govuk-roadmap-numbers {
+  @include govuk-media-query($from: tablet) {
+    text-align: center;
+  }
+
+  .govuk-header__logotype-crown {
+    display: none;
```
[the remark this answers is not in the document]
andysellick: Sorry, wait, it does matter. At some screen sizes the grid element is still 'visible' and causes the right hand column to overlap the page boundary.

Screenshot 2021-01-28 at 12 45 56

app/views/roadmap/index.html.erb
```diff
@@ -0,0 +1,116 @@
...
+  <main id="content" class="govuk-main-wrapper">
+    <section class=" govuk-!-padding-bottom-6 govuk-!-margin-bottom-6 govuk-roadmap-section">
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <%= render "govuk_publishing_components/components/title", title: t('roadmap.hero.heading'), margin_top: 0 %>
+          <p class="govuk-body"><%= t('roadmap.hero.description') %></p>
+        </div>
+
+        <div class="govuk-grid-column-one-third">
+          <%= render "image" %>
```
andysellick, line 16: Is it worth trying to limit the width of this image on tablet? It gets quite large compared with the rest of the content. This is at 635px wide.

Screenshot 2021-01-28 at 12 48 13

andysellick (commented):
Looks really good, a few questions and comments.

chao-xian:
You'll need to add a [special_route](https://docs.publishing.service.gov.uk/document-types/special_route.html) for this page. [Instructions for that are in the Dev Docs](https://docs.publishing.service.gov.uk/manual/publish-special-routes.html).

config/locales/en.yml
danacotoran: There is a [reason](https://stackoverflow.com/questions/19706925/insert-html-special-characters-in-i18n-yml) for the double quotes in this particular case.

app/views/roadmap/index.html.erb
danacotoran: Good shout, it doesn't look amazing. 
I've limited the width to max 300px, I think that looks okay but let me know if you think otherwise.

app/assets/stylesheets/application.scss
```diff
@@ -114,3 +114,21 @@ $govuk-use-legacy-palette: false;
...
+.govuk-roadmap-section {
+  border-bottom: 5px solid govuk-colour('black');
+}
+
+.govuk-roadmap-numbers {
+  @include govuk-media-query($from: tablet) {
+    text-align: center;
+  }
+
+  .govuk-header__logotype-crown {
```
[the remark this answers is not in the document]
danacotoran: Nope!

app/assets/stylesheets/application.scss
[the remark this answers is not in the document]
danacotoran: I've discussed this with Nikin and after going over some options we decided to go with 50% columns for tablet which I think looks much better.

app/views/roadmap/index.html.erb
```diff
@@ -0,0 +1,116 @@
+<% content_for :title, t('roadmap.page_title') %>
+<% content_for :extra_headers do %>
+  <meta name="description" content="<%= t('roadmap.hero.description') %>" />
+<% end %>
+
+<div class="govuk-width-container">
+  <main id="content" class="govuk-main-wrapper">
+    <section class=" govuk-!-padding-bottom-6 govuk-!-margin-bottom-6 govuk-roadmap-section">
```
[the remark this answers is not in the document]
danacotoran: In this case I was able to remove the margin class as it's the same for every `govuk-roadmap-section`. The padding does vary so I have left the padding override class in place.

app/views/roadmap/index.html.erb
danacotoran: Yeeeep, you are correct, thanks for pointing that out. I'm not sure why I did that in the first place.
Have removed the override.

app/views/roadmap/index.html.erb
danacotoran: That's a good question and my best answer is "I don't know" which I realise is not very reassuring. 
I'm not sure how to test if the `aria-hidden` from the `svg` still applies when a browser doesn't support `svg`.

As for adding an `alt`, the  svg `<image>` tag doesn't accept the same attributes as ``, therefore I don't think adding `alt` would be valid https://svgwg.org/svg2-draft/embedded.html#ImageElement 
However adding aria attributes to `image` is valid, so I've given it `aria-hidden`.

app/views/roadmap/index.html.erb
danacotoran: You are correct. 👍

config/locales/en.yml
danacotoran: Good shout, I asked Keith and he agreed it should be a link.

app/assets/stylesheets/application.scss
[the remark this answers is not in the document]
andysellick: That's a better solution, good stuff.

andysellick (approved):

chao-xian:
@danacotoran See https://github.com/alphagov/frontend/commit/552e3e8963542bf550fd44395e49e0a67606d74e

chao-xian (changes requested):
Sorry @danacotoran it needs a special route added. See https://github.com/alphagov/frontend/commit/552e3e8963542bf550fd44395e49e0a67606d74e

And to generate a `content_id` see [this bit in the Dev Docs on generating one](https://docs.publishing.service.gov.uk/apps/publishing-api/publishing-application-examples.html#drafting-content-from-a-publishing-app).

(Also I'd fixup your later commit into the first one before merging but that one is nbd)

danacotoran:
Thanks for this @chao-xian, I've updated the PR with the addition of the special route, let me know if that looks okay!
(+ good point, I'll `fixup` before merging)

chao-xian (approved):
Nice work! 🎉

PR #2615, opened 2021-01-29
Add the ability to specify country specific content for unavailable page

gclssvglx:
## What

Add the ability to specify content on a per country basis and display that to the user when they have supplied a postcode that is within a country where the service is unavailable

## Why

We want to give a more tailored and specific response to the user depending on the country that their local authority is in so that we can support them better

[Trello](https://trello.com/c/BG1boLpY/756-add-ability-to-specify-content-per-country-for-unavailable-page)

### DIFF
--- a/config/unavailable_services.yml
+++ b/config/unavailable_services.yml
@@ -1,9 +1,8 @@
 services:
   1826:
-    countries:
-      unavailable_in:
-        - Scotland
-        - Northern Ireland
-      content: |
-        <p class="govuk-body">We've matched the postcode to <span class="local-authority">%{local_authority_name}</span>.</p>
-        <p class="govuk-body">This service is not available in %{country_name}. You can find other services on the <span class="local-authority">%{local_authority_name}</span> website.</p>
+    Scotland: |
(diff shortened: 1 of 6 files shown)


config/unavailable_services.yml
```diff
@@ -1,9 +1,8 @@
...
-    countries:
-      unavailable_in:
-        - Scotland
-        - Northern Ireland
-      content: |
-        <p class="govuk-body">We've matched the postcode to <span class="local-authority">%{local_authority_name}</span>.</p>
-        <p class="govuk-body">This service is not available in %{country_name}. You can find other services on the <span class="local-authority">%{local_authority_name}</span> website.</p>
+    Scotland: |
+      <p class="govuk-body">We've matched the postcode to <span class="local-authority">%{local_authority_name}</span>.</p>
+      <p class="govuk-body">This service is not available in %{country_name}. You can find other services on the <span class="local-authority">%{local_authority_name}</span> website.</p>
```
theseanything, line 5: Should we just hardcode the country name in here instead of a variable?

theseanything (approved):

config/unavailable_services.yml
gclssvglx: Makes perfect sense now the content is country specific - have fixed.

PR #2618, opened 2021-02-02
Update tense on Brexit work

keithiopia:
One of the tasks was written before the transition period ended, so update the task description now it's ended.


hannako (approved):
Lovely 👍

PR #2623, opened 2021-02-05
Update unavaliable pages for mass testing local transaction

theseanything:
This PR updates the unavailable pages for local transaction pages and refactors the singleton `LocalTransactionServices` class into the `LocalTransactionService` class. 

This is to simplify the code and make it easier to customise the unavailable page.

We've also added updated configuration for the "Find a COVID-19 rapid lateral flow test site", to display that it is unavailable in Wales, NI and Scotland.


app/views/local_transaction/unavailable_service.html.erb
```diff
@@ -1,17 +1,23 @@
 <%= render layout: 'shared/base_page', locals: {
-  title: "This service is not available in #{@country_name}",
+  title: @service.title,
   publication: @publication,
   edition: @edition,
 } do %>
   <div class="interaction">
-    <%= @content.html_safe %>
+    <p class="govuk-body">
+      We've matched the postcode to <span class="local-authority"><%= @local_authority.name %></span>.
```
gclssvglx, line 8: Should we not add this to the config too?  I guess this page will only ever be used for postcode lookup, but just thinking if anyone might want to change/alter this text.

gclssvglx (approved):
Nice!  Thanks for doing this - looks much better.

app/views/local_transaction/unavailable_service.html.erb
theseanything: Yeah good question, so our currently all our unavailable pages have this on it. It's pretty generic text, which will make sense in most contexts. I've avoided putting this into config, so we don't need to support interpolation and just to keep things simple.

PR #2632, opened 2021-02-15
Make the Devolved Admin links more specific

GDSNewt:
This is to remove the reference to ‘From 15 February’ and also to add a link,
and make the Devolved Administration links more specific to returning from
international travel.


JonathanHallam (approved):
Looks good to me, thanks :)

PR #2643, opened 2021-02-26
Write "Provider" Pact for Bank Holidays API

ChrisBAshton:
Depends on https://github.com/alphagov/gds-api-adapters/pull/1035

Trello: https://trello.com/c/Pvoq4sQY/2370-5-enable-continuous-deployment-for-frontend


config/environments/development.rb
```diff
@@ -47,4 +47,5 @@
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.hosts << "frontend.dev.gov.uk"
+  config.hosts << "example.org" # for pact tests
```
thomasleese, line 50: I'm wondering where this comes from? Is it because Pact is making requests to the server with the `Host` set to `example.org`? Could we change it to use `frontend.dev.gov.uk`?

ChrisBAshton: Yes - it failed without this line, and I remember it being added to Collections. But on a second look, it looks like [it's been removed](https://github.com/alphagov/collections/commit/0a0523fd42ab1787716999938029e23ff7e4bcb7), so I'll apply the same technique here and explain it in a commit.

thomasleese (approved):

test/service_consumers/pact_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def call(env)
+    env["HTTP_HOST"] = "localhost:3003"
+    response = @real_provider_app.call(env)
+    response
+  end
+end
+
+Pact.service_provider "Bank Holidays API" do
+  app { ProxyApp.new(Rails.application) }
```
hannako, line 26: You don't need to use a proxy here. It was needed for the collections pact tests because the results are paginated. See this [commit](https://github.com/alphagov/collections/pull/2230/commits/0a0523fd42ab1787716999938029e23ff7e4bcb7) for details.

So you can lose lines 13 - 23 and line 26

hannako: You might need to add `"example.org"` to the allowed config.hosts in config/environments/development.rb

ChrisBAshton: Yes, we had this discussion [earlier in the PR](https://github.com/alphagov/frontend/pull/2643#discussion_r584565001) and figured that it'd be better to change the Pact tests than to introduce some unexpected configuration. Do you think otherwise?

It's a shame there's not a better way of overriding this in Pact!

hannako: Yes that makes sense. 

We should probably create a proxy for publishing api then, so we can [lose](https://github.com/alphagov/gds-api-adapters/blob/b5686a840fcc3cac4021b6210272daa89501f1d2/test/publishing_api_test.rb#L2294-L2304) all these stubs

thomasleese: I've been starting to think about how we can consolidate and standardise the way we do Pact tests, and I think this is a good example of something we can put in [govuk_test](https://github.com/alphagov/govuk_test). I've got https://trello.com/c/rnfUFP5o/2103-add-common-configuration-for-pact-tests which I'm using to track these sorts of things for now.

config/environments/development.rb
barrucadu: I'm a bit late to this, but if you do `config.hosts.clear` you don't need to specify them all.

test/service_consumers/pact_helper.rb
ChrisBAshton: As a follow-up, we've found we don't need to set the port in the Provider, only in the Consumer: https://github.com/alphagov/frontend/pull/2710

PR #2644, opened 2021-02-26
Run pact tests with gds-api-adapters as part of deployment pipeline

ChrisBAshton:
Depends on https://github.com/alphagov/gds-api-adapters/pull/1035 and #2643  (otherwise there'd be no pact tests to test against, and the build would fail).

Note that this PR takes a _lot_ of inspiration from https://github.com/alphagov/collections/pull/2281.

Trello: https://trello.com/c/Pvoq4sQY/2370-5-enable-continuous-deployment-for-frontend

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,5 +6,17 @@ node() {
   govuk.setEnvar("PUBLISHING_E2E_TESTS_COMMAND", "test-frontend")
   govuk.buildProject(
     publishingE2ETests: true,
+    extraParameters: [
+      stringParam(
+        name: "GDS_API_ADAPTERS_PACT_VERSION",
+        defaultValue: "master",
+        description: "The branch of gds-api-adapters pact tests to run against"
+      ),
+    ],
+    afterTest : {
+      stage("Verify pact with gds-api-adapters") {
+        govuk.runRakeTask("pact:verify:branch[${env.GDS_API_ADAPTERS_PACT_VERSION}]")
+      }
+    }
   )
 }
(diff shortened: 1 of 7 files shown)


brucebolt (approved):
From the perspective of someone who hasn't worked with pact tests before, this looks good to me

ChrisBAshton:
Thanks @brucebolt. I've simplified the PR a bit by removing some unnecessary ENV variables, squishing this into 3 commits in the process. Would you mind re-reviewing to make sure it makes sense?

Jenkinsfile
```diff
@@ -6,5 +6,17 @@ node() {
...
+        name: "GDS_API_ADAPTERS_PACT_VERSION",
+        defaultValue: "master",
+        description: "The branch of gds-api-adapters pact tests to run against"
+      ),
+    ],
+    afterTest : {
+      stage("Verify pact with gds-api-adapters") {
+        govuk.runRakeTask("pact:verify:branch[${env.GDS_API_ADAPTERS_PACT_VERSION}]")
+      }
+    }
```
ChrisBAshton, line 20: EDIT: we've concluded that this extra Jenkins config is unnecessary, as we're already running the pact tests as part of the default Rake task. This code was removed in https://github.com/alphagov/frontend/pull/2648.

lib/tasks/pact.rake
```diff
@@ -1,5 +1,17 @@
 return if Rails.env.production?
 
 require "pact/tasks"
-require "pact_broker/client/tasks"
 require "pact/tasks/task_helper"
+
+desc "Verify that Frontend honours all contracts in the pactfile created by a given branch of gds-api-adapters"
+task "pact:verify:branch", [:branch_name] => :environment do |t, args|
```
ChrisBAshton, line 7: This task also isn't necessary. Removed in https://github.com/alphagov/frontend/pull/2712.

PR #2645, opened 2021-03-02
Enable compatibility mode in print stylesheet

alex-ju:
### What
Enable compatibility mode in print stylesheet (as we do in [`application.scss`](https://github.com/alphagov/frontend/blob/master/app/assets/stylesheets/application.scss#L1-L2)).

### Why
Not setting the compatibility flag with `govuk-template` results in referencing the new `GDS Transport` font which is not available in `static` at the moment as we still rely on `govuk-template` and the old `nta` font.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

Screenshot 2021-03-02 at 12 46 08

</td><td valign="top">

Screenshot 2021-03-02 at 12 45 26

</td></tr>
</table>

### DIFF
--- a/app/assets/stylesheets/print.scss
+++ b/app/assets/stylesheets/print.scss
@@ -1,3 +1,6 @@
+$govuk-compatibility-govuktemplate: true;
+$govuk-use-legacy-palette: false;
+
 @import 'govuk_publishing_components/govuk_frontend_support';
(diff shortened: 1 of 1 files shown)


injms (approved):
🖨️ 👍

PR #2658, opened 2021-03-09
Add duplicate service with correct LGSL code

gclssvglx:
## What

We want to add a duplicate configuration record for the lateral flow mass testing finder service with an LGSL code of `1827`. Once the LGSL code has been changed from `100001` to `1827` in `local-links-manager` - the config for service `100001` will be removed. 

## Why

When the lateral flow mass testing finder service was created, it was created with a dummy code of `100001`.  We now have a real code - `1827` - so would like to use this code instead of the dummy one.

[Trello](https://trello.com/c/MRGj9keG/210-llup-lateral-mass-test-apply-the-proper-assigned-service-code-1827-agreed-with-local-government-authority)


DilwoarH (approved):
LGTM 👍

PR #2659, opened 2021-03-09
Convoluted approach to changing local transaction error message copy

kevindew:
Trello: https://trello.com/c/BgI9a1t3/205-local-lookup-says-no-create-better-were-not-sure-if-this-council-offers-this-service-copy

This PR changes the renderings of errors when a local authority doesn't have a service, while also cleaning up some things to make this simple change possible.

Before:

Screenshot 2021-03-09 at 19 38 28

After:

Screenshot 2021-03-09 at 18 27 06

---

The other things this PR does is:

- It removes the usage of LocationError to handle cases where a local authority is missing a service. This made making changes harder and already seemed a rather uneasy fit as the conditionals were somewhat bent around it
- Removes an ActiveSuport::Notifier event for logging postcode errors, it hadn't worked since 201

More details in the commits

### DIFF
--- a/app/controllers/local_transaction_controller.rb
+++ b/app/controllers/local_transaction_controller.rb
@@ -7,8 +7,6 @@ class LocalTransactionController < ApplicationController
   before_action -> { response.headers["X-Frame-Options"] = "DENY" }
 
   INVALID_POSTCODE = "invalidPostcodeFormat".freeze
-  NO_AUTHORITY_URL = "laMatchNoLinkNoAuthorityUrl".freeze
-  NO_LINK = "laMatchNoLink".freeze
   NO_MAPIT_MATCH = "fullPostcodeNoMapitMatch".freeze
   NO_MATCHING_AUTHORITY = "noLaMatch".freeze
   BANNED_POSTCODES = %w[ENTERPOSTCODE].freeze
@@ -34,7 +32,7 @@ class LocalTransactionController < ApplicationController
   def results
     @postcode = postcode
     @interaction_details = interaction_details
-    @local_authority = local_authority
+    @local_authority = LocalAuthorityPresenter.new(@interaction_details["local_authority"])
     @country_name = @local_authority.country_name
     @service = LocalTransactionService.new(
       @publication.title, lgsl, @country_name, @local_authority.url
@@ -96,21 +94,8 @@ private
     content_item["details"]["lgil_code"] || content_item["details"]["lgil_override"]
   end
 
-  def local_authority
-    if interaction_details["local_authority"]
-      local_authority = LocalAuthorityPresenter.new(interaction_details["local_authority"])
-    end
-
-    unless interaction_details["local_interaction"]
-      @location_error = error_for_missing_interaction(local_authority)
-    end
-
-    local_authority
-  end
-
   def interaction_details
     council = params[:local_authority_slug]
-    return {} unless council
 
     if council == "electoral-office-for-northern-ireland"
       {
@@ -120,10 +105,5 @@ private
     else
       @_interaction ||= Frontend.local_links_manager_api.local_link(council, lgsl, lgil)
     end
-  end
-
-  def error_for_missing_interaction(local_authority)
-    error_code = local_authority.url.present? ? NO_LINK : NO_AUTHORITY_URL
-    LocationError.new(error_code, local_authority_name: local_authority.name)
   end
 end
--- a/app/models/location_error.rb
+++ b/app/models/location_error.rb
@@ -12,23 +12,6 @@ class LocationError
     when "noLaMatch"
       @message = "formats.local_transaction.no_local_authority"
       @sub_message = ""
-    when "laMatchNoLink"
-      @message =
-        if local_authority_name_starts_with_a_the?
-          "formats.local_transaction.local_authority_starts_with_the_no_service_url_html"
-        else
-          "formats.local_transaction.local_authority_no_service_url_html"
-        end
-      @sub_message = "" # not used in the markup for this case
-    when "laMatchNoLinkNoAuthorityUrl"
-      @message =
-        if local_authority_name_starts_with_a_the?
-          "formats.local_transaction.local_authority_starts_with_the_no_service_url_no_authority_link_html"
-        else
-          "formats.local_transaction.local_authority_no_service_url_no_authority_link_html"
-        end
-
-      @sub_message = "" # not used in the markup for this case
     when "validPostcodeNoLocation"
       # This is a find my nearest exception when no location is found
       @message = "formats.find_my_nearest.valid_postcode_no_locations"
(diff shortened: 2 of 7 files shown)


hannahjump:
Content looks great! Thank you!

test/functional/local_transaction_controller_test.rb
```diff
@@ -342,13 +314,8 @@ def subscribe_logstasher_to_postcode_error_notification
...
 
-    should "expose the 'missing interaction' error to the view" do
-      location_error = assigns(:location_error)
-      assert_equal "laMatchNoLink", location_error.postcode_error
-    end
-
-    should "log the 'missing interaction' error to the view" do
-      assert_equal(LogStasher.store["postcode_error_notification"], postcode_error: "laMatchNoLink")
+    should "expose the 'missing interaction' error in the view" do
+      assert_match "laMatchNoLink", response.body
```
Rosa-Fox, line 318: I think this should be:

```
assert_match "We do not know if they offer this service.", response.body
```

At the moment the test passes as `response.body` contains `laMatchNoLink` as its part of the analytics `data-track-action` but it looks like it should be testing for the error message. However `laMatchNoLink` has been removed from `location_error.rb` so its a bit confusing to see where that comes from.

test/functional/local_transaction_controller_test.rb
```diff
@@ -186,56 +180,35 @@ def subscribe_logstasher_to_postcode_error_notification
...
 
-      should "expose the 'missing local authority' error to the view" do
-        location_error = assigns(:location_error)
-        assert_equal "noLaMatch", location_error.postcode_error
-      end
-
-      should "log the 'missing local authority' error to the view" do
-        assert_equal(LogStasher.store["postcode_error_notification"], postcode_error: "noLaMatch")
+      should "expose the 'missing local authority' error in the view" do
+        assert_match "noLaMatch", response.body
```
Rosa-Fox, line 211: I think this is also passing because `noLaMatch` is part of the analytics actions in `response.body`. The error on the view would be something like this:

```
assert_match "We couldn&#39;t find a council for this postcode", response.body
```

I tried to display with `I18n.t` but didn't seem to work...

app/views/local_transaction/results.html.erb
```diff
@@ -19,13 +19,11 @@
...
-  <% elsif @location_error && @location_error.no_location_interaction? %>
-    <div class="interaction">
-      <p class="govuk-body">
-        <%= t(@location_error.message, **@location_error.message_args) %>
-      </p>
+    <% else %>
       <% if @local_authority.url.present? %>
+        <p class="govuk-body">
+          We do not know if they offer this service. Search their website to find out if they do, or what other services they offer.
+        </p>
```
Rosa-Fox, line 26: Should this be in en.yml?

app/models/location_error.rb
```diff
@@ -12,23 +12,6 @@ def initialize(postcode_error = nil, message_args = {})
     when "noLaMatch"
       @message = "formats.local_transaction.no_local_authority"
       @sub_message = ""
-    when "laMatchNoLink"
```
Rosa-Fox, line 15: 👍

app/controllers/local_transaction_controller.rb
```diff
@@ -96,21 +94,8 @@ def lgil
...
-
-    unless interaction_details["local_interaction"]
-      @location_error = error_for_missing_interaction(local_authority)
-    end
-
-    local_authority
-  end
-
   def interaction_details
     council = params[:local_authority_slug]
```
Rosa-Fox, line 98: will it always have a params[:local_authority_slug]? I guess the validation errors in the publishing app mean that it would?

test/unit/models/location_error_test.rb
```diff
@@ -5,22 +5,6 @@ class LocationErrorTest < ActiveSupport::TestCase
...
-      end
-    end
-
-    context "when not given a postcode_error" do
-      should "not send a postcode_error notification" do
-        ActiveSupport::Notifications.expects(:instrument).never
-
-        LocationError.new
-      end
-    end
```
Rosa-Fox, line 22: I noticed that the `"fullPostcodeNoMapitMatch"` case isn't tested in here, which I guess isn't related to these changes as it wasn't tested before... but could be something we add separately?

app/controllers/local_transaction_controller.rb
kevindew: This interaction details method is only called by the results action which needs a `local_authority_slug` param

app/views/local_transaction/results.html.erb
kevindew: The precedence for this file seems to be hardcoding text (I actually removed the only usage of `t` from it)

test/functional/local_transaction_controller_test.rb
kevindew: Ah I did this to maintain it testing the code aspect since it's used for analytics. On reflection I think your suggestion is better and my analytics tests are rather opaque.

test/unit/models/location_error_test.rb
kevindew: Thanks for pointing out. I've refactored this code a bit so every branch doesn't need testing. It was getting rather closer to adding unit tests to test configuration.

kevindew:
Thanks for the review @Rosa-Fox, I've addressed the comments.

app/views/local_transaction/results.html.erb
Rosa-Fox: 👍

test/functional/local_transaction_controller_test.rb
```diff
@@ -186,56 +180,38 @@ def subscribe_logstasher_to_postcode_error_notification
...
-      should "expose the 'invalid postcode format' error to the view" do
-        location_error = assigns(:location_error)
-        assert_equal location_error.postcode_error, "invalidPostcodeFormat"
-      end
-
-      should "log the 'invalid postcode format' error to the view" do
-        assert_equal(LogStasher.store["postcode_error_notification"], postcode_error: "invalidPostcodeFormat")
+      should "responsds successfully with an error message" do
+        assert response.ok?
+        assert_match CGI.escapeHTML(I18n.t!("formats.local_transaction.invalid_postcode")), response.body
```
Rosa-Fox, line 188: Nice!

test/functional/local_transaction_controller_test.rb
```diff
@@ -186,56 +180,38 @@ def subscribe_logstasher_to_postcode_error_notification
...
       end
 
-      should "expose the 'invalid postcode format' error to the view" do
-        location_error = assigns(:location_error)
-        assert_equal location_error.postcode_error, "invalidPostcodeFormat"
-      end
-
-      should "log the 'invalid postcode format' error to the view" do
-        assert_equal(LogStasher.store["postcode_error_notification"], postcode_error: "invalidPostcodeFormat")
+      should "responsds successfully with an error message" do
```
Rosa-Fox, line 186: One last little thing which is that there a typo in 'responsds`a couple of times in this file @kevindew

test/unit/models/location_error_test.rb
Rosa-Fox: 👍

test/functional/local_transaction_controller_test.rb
Rosa-Fox: 👍

test/functional/local_transaction_controller_test.rb
kevindew: Oops - thanks for catching. And to think I c+p'd that 3 times.

Rosa-Fox: hehe 👍

Rosa-Fox (approved):

PR #2668, opened 2021-03-16
Add GOV.UK sessions

huwd:
[A part of this trello ticket](https://trello.com/c/RaI017p6/675-add-new-public-facing-oauth-endpoints-to-frontend)

What
===

A part of the work to broaden the transition checker login prototype to
the whole of GOV.UK. As outlined in [RFC#134][1] we are moving sign in
and sign out paths from being nested in finder-frontend (which would be
strange for a GOV.UK-wide login!) onto their own paths.

[1]: https://github.com/alphagov/govuk-rfcs/pull/134

### DIFF
--- a/app/controllers/concerns/account_concern.rb
+++ b/app/controllers/concerns/account_concern.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+module AccountConcern
+  extend ActiveSupport::Concern
+
+  ACCOUNT_SESSION_REQUEST_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION"
+  ACCOUNT_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-Session"
+  ACCOUNT_END_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-End-Session"
+  ACCOUNT_SESSION_DEV_COOKIE_NAME = "govuk_account_session"
+
+  included do
+    before_action :fetch_account_session_header
+    helper_method :logged_in?
+
+    attr_accessor :account_session_header
+  end
+
+  def logged_in?
+    account_session_header.present?
+  end
+
+  def fetch_account_session_header
+    @account_session_header =
+      if request.headers[ACCOUNT_SESSION_REQUEST_HEADER_NAME]
+        request.headers[ACCOUNT_SESSION_REQUEST_HEADER_NAME]
+      elsif Rails.env.development?
+        cookies[ACCOUNT_SESSION_DEV_COOKIE_NAME]
+      end
+  end
+
+  def set_account_session_header(govuk_account_session = nil)
+    @account_session_header = govuk_account_session if govuk_account_session
+    response.headers[ACCOUNT_SESSION_RESPONSE_HEADER_NAME] = @account_session_header
+
+    if Rails.env.development?
+      cookies[ACCOUNT_SESSION_DEV_COOKIE_NAME] = {
+        value: @account_session_header,
+        domain: "dev.gov.uk",
+      }
+    end
+  end
+
+  def logout!
+    response.headers[ACCOUNT_END_SESSION_RESPONSE_HEADER_NAME] = "1"
+    @account_session_header = nil
+
+    if Rails.env.development?
+      cookies[ACCOUNT_SESSION_DEV_COOKIE_NAME] = {
+        value: "",
+        domain: "dev.gov.uk",
+        expires: 1.second.ago,
+      }
+    end
+  end
+end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,0 +1,64 @@
+class SessionsController < ApplicationController
+  include AccountConcern
+
+  before_action :set_no_cache_headers
+
+  def create
+    redirect_with_ga account_manager_url and return if logged_in?
+
+    redirect_with_ga GdsApi.account_api.get_sign_in_url(
+      redirect_path: params[:redirect_path],
+      state_id: params[:state_id],
+    ).to_h["auth_uri"]
+  end
+
+  def callback
+    redirect_to Plek.new.website_root and return unless params[:code] && params[:state]
+
+    callback = GdsApi.account_api.validate_auth_response(
+      code: params.require(:code),
+      state: params.require(:state),
+    ).to_h
+
+    set_account_session_header(callback["govuk_account_session"])
+
+    redirect_with_ga(callback["redirect_path"] || account_manager_url, callback["ga_client_id"])
+  rescue GdsApi::HTTPUnauthorized
+    head :bad_request
+  end
+
+  def delete
+    logout!
+    if params[:continue]
+      redirect_with_ga "#{account_manager_url}/sign-out?done=#{params[:continue]}"
+    elsif params[:done]
+      redirect_with_ga Plek.new.website_root
+    else
+      redirect_with_ga "#{account_manager_url}/sign-out?continue=1"
+    end
+  end
+
+protected
+
+  def set_no_cache_headers
+    response.headers["Cache-Control"] = "no-store"
+  end
+
+  def account_manager_url
+    Plek.find("account-manager")
+  end
+
+  def redirect_with_ga(url, ga_client_id = nil)
+    ga_client_id ||= params[:_ga]
+    if ga_client_id
+      url =
+        if url.include? "?"
+          "#{url}&_ga=#{ga_client_id}"
+        else
+          "#{url}?_ga=#{ga_client_id}"
+        end
+    end
+
+    redirect_to url
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -21,6 +21,11 @@ Rails.application.routes.draw do
 
   get "/foreign-travel-advice", to: "travel_advice#index", as: :travel_advice
 
+  # Accounts
+  get "/sign-in", to: "sessions#create", as: :new_govuk_session
+  get "/sign-in/callback", to: "sessions#callback", as: :new_govuk_session_callback
+  get "/sign-out", to: "sessions#delete", as: :end_govuk_session
+
   # Help pages
   get "/help", to: "help#index"
   get "/help/ab-testing", to: "help#ab_testing"
--- a/test/functional/session_controller_test.rb
+++ b/test/functional/session_controller_test.rb
@@ -1,0 +1,114 @@
+require "test_helper"
+require "gds_api/test_helpers/account_api"
+
+class SessionsControllerTest < ActionController::TestCase
+  include GdsApi::TestHelpers::AccountApi
+  include SessionHelpers
+
+  context "GET sign-in" do
+    setup do
+      stub_account_api_get_sign_in_url(redirect_path: "/bank-holiday", state_id: "state123")
+    end
+
+    should "redirect the user to the GOV.UK Account service domain" do
+      get :create, params: { redirect_path: "/bank-holiday", state_id: "state123" }
+      assert_response :redirect
+      assert_equal @response.headers["Location"], "http://auth/provider"
+    end
+
+    should "preserve the _ga tracking parameter if provided" do
+      get :create, params: { redirect_path: "/bank-holiday", state_id: "state123", _ga: "ga123" }
+      assert_equal @response.headers["Location"], "http://auth/provider?_ga=ga123"
+    end
+
+    context "when already logged in" do
+      setup { mock_logged_in_session }
+
+      should "redirect the user to the account manager if the user is already logged in" do
+        get :create, params: { redirect_path: "/bank-holiday", state_id: "state123" }
+        assert_redirected_to Plek.find("account-manager")
+      end
+    end
+  end
+
+  context "GET sign-in/callback" do
+    setup do
+      @root_path = Plek.new.website_root
+    end
+
+    should "redirect to the root path if the :code param is not present" do
+      get :callback, params: { state: "state123" }
+
+      assert_response :redirect
+      assert_equal @response.redirect_url, @root_path
+    end
+
+    should "redirect to the root path if the :state param is not present" do
+      get :callback, params: { code: "code123" }
+
+      assert_response :redirect
+      assert_equal @response.redirect_url, @root_path
+    end
+
+    should "set the 'GOVUK-Account-Session' header" do
+      stub_account_api_validates_auth_response(govuk_account_session: "placeholder")
+      get :callback, params: { code: "code123", state: "state123" }
+
+      assert_equal @response.headers["GOVUK-Account-Session"], "placeholder"
+    end
+
+    should "redirect to the redirect path if a redirect path is in the account-api response" do
+      stub_account_api_validates_auth_response(redirect_path: "/bank-holiday")
+      get :callback, params: { code: "code123", state: "state123" }
+
+      assert_response :redirect
+      assert_includes @response.redirect_url, "/bank-holiday"
+    end
+
+    should "redirect to the account manager url if no redirect path is in the account-api response" do
+      stub_account_api_validates_auth_response(redirect_path: nil, ga_client_id: nil)
+      get :callback, params: { code: "code123", state: "state123" }
+
+      assert_response :redirect
+      assert_equal @response.redirect_url, Plek.find("account-manager")
+    end
+
+    should "redirect with the specified :ga_client_id of the api response if present" do
+      ga_client_id = "ga_client_id"
+      stub_account_api_validates_auth_response(ga_client_id: ga_client_id)
+      get :callback, params: { code: "code123", state: "state123" }
+
+      assert_response :redirect
+      assert_includes @response.redirect_url, ga_client_id
+    end
+
+    should "redirect with the specified :_ga param if :ga_client_id of the api response is not present" do
+      underscore_ga = "underscore_ga"
+      stub_account_api_validates_auth_response(ga_client_id: nil)
+      get :callback, params: { code: "code123", state: "state123", _ga: underscore_ga }
+
+      assert_response :redirect
+      assert_includes @response.redirect_url, underscore_ga
+    end
+
+    should "respond with :bad_request if :code or :state are invalid" do
+      stub_account_api_rejects_auth_response
+      get :callback, params: { code: "code123", state: "state123" }
+
+      assert_response :bad_request
+    end
+  end
+
+  context "GET /sign-out" do
+    context "when the user is logged in" do
+      setup do
+        mock_logged_in_session
+      end
+
+      should "set the 'GOVUK-Account-End-Session' header to 1" do
+        get :delete
+        assert @response.headers["GOVUK-Account-End-Session"].present?
+      end
+    end
+  end
+end
--- a/test/integration/sessions_test.rb
+++ b/test/integration/sessions_test.rb
@@ -1,0 +1,10 @@
+require "integration_test_helper"
+
+class SessionsTest < ActionDispatch::IntegrationTest
+  context "when logged in" do
+    should "redirect the user to the account manager URL if they visit /sign-up" do
+      get "/sign-in", headers: { "GOVUK-Account-Session" => "placeholder" }
+      assert_redirected_to Plek.find("account-manager")
+    end
+  end
+end
--- a/test/support/session_helpers.rb
+++ b/test/support/session_helpers.rb
@@ -1,0 +1,5 @@
+module SessionHelpers
+  def mock_logged_in_session
+    request.headers["GOVUK-Account-Session"] = "placeholder"
+  end
+end


app/controllers/concerns/account_concern.rb
```diff
@@ -0,0 +1,73 @@
...
+    attr_accessor :account_session_header
+  end
+
+  def logged_in?
+    account_session_header.present?
+  end
+
+  def fetch_account_session_header
+    @account_session_header =
+      if request.headers[ACCOUNT_SESSION_HEADER_INTERNAL_NAME]
```
huwd, line 27: We've not tested this path and probably should.
@barrucadu can I find out a bit more about the use case here?

I wonder if we can simplify the block at all if this is no longer needed?

config/routes.rb
```diff
@@ -81,6 +81,10 @@
     get ":scope/:division", to: "calendar#division", as: :division
   end
 
+  get "/sign-in", to: "sessions#create", as: :new_govuk_session
+  get "/sign-in/callback", to: "sessions#callback", as: :new__govuk_session_callback
```
barrucadu, line 85: Should this be `new_govuk_session_callback` ?

app/controllers/concerns/account_concern.rb
```diff
@@ -0,0 +1,73 @@
...
+      elsif request.headers.to_h[ACCOUNT_SESSION_HEADER_NAME]
+        request.headers.to_h[ACCOUNT_SESSION_HEADER_NAME]
+      elsif Rails.env.development?
+        cookies[ACCOUNT_SESSION_DEV_COOKIE_NAME]
+      end
+  end
+
+  def show_signed_in_header?
+    account_session_header.present?
+  end
```
barrucadu, line 38: This is unused (the session controller doesn't return HTML), so it can be removed

app/controllers/concerns/account_concern.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def set_account_variant
+    response.headers["Vary"] = [response.headers["Vary"], ACCOUNT_SESSION_HEADER_NAME].compact.join(", ")
+
+    set_slimmer_headers(
+      remove_search: true,
+      show_accounts: show_signed_in_header? ? "signed-in" : "signed-out",
+    )
+  end
```
barrucadu, line 47: I think we should replace this with just setting no-cache headers for the sessions controller

test/functional/session_controller_test.rb
```diff
@@ -0,0 +1,118 @@
...
+    should "redirect the user to the GOV.UK Account service domain" do
+      get :create, params: { redirect_path: "/bank-holiday", state_id: "abc123" }
+      assert_response :redirect
+      assert_equal @response.headers["Location"], "http://auth/provider"
+    end
+
+    should "preserve the _ga tracking parameter if provided" do
+      get :create, params: { redirect_path: "/bank-holiday", state_id: "abc123", _ga: "abc123" }
+      assert_equal @response.headers["Location"], "http://auth/provider?_ga=abc123"
+    end
```
barrucadu, line 22: Nitpick: if you use different values for the `state_id` and the `_ga`, it'll be clearer that it's really the `_ga` which gets preserved (as this current test would pass if the controller did `"?_ga=#{params[:state_id]}"`_

test/support/session_helpers.rb
```diff
@@ -0,0 +1,8 @@
+module SessionHelpers
+  def mock_logged_in_session
+    request.headers.merge({
+      "Vary" => AccountConcern::ACCOUNT_SESSION_HEADER_NAME,
```
barrucadu, line 4: This request header isn't needed

test/functional/session_controller_test.rb
```diff
@@ -0,0 +1,118 @@
...
+      should "redirect the user to the account manager if the user is already logged in" do
+        get :create, params: { redirect_path: "/bank-holiday", state_id: "abc123" }
+        assert_redirected_to Plek.find("account-manager")
+      end
+    end
+  end
+
+  context "GET sign-in/callback" do
+    setup do
+      @govuk_account_session = "govuk_account_session"
```
barrucadu, line 36: This instance variable is only used in a single test, how about inlining the variable there?

test/functional/session_controller_test.rb
```diff
@@ -0,0 +1,118 @@
...
+  context "GET /sign-out" do
+    context "when the user is logged in" do
+      setup do
+        mock_logged_in_session
+        @account_end_session_header_name = AccountConcern::ACCOUNT_END_SESSION_HEADER_NAME
+      end
+
+      should "set ACCOUNT_END_SESSION_HEADER_NAME response header to 1" do
+        get :delete
+        assert_includes @response.headers[@account_end_session_header_name], "1"
```
barrucadu, line 114: Use `AccountConcern::ACCOUNT_END_SESSION_HEADER_NAME` directly here, instead of assigning it to an instance variable first.

barrucadu: Also the actual value of the header is unimportant, only that there is one.  So this test could just check that the header is present.

app/controllers/concerns/account_concern.rb
barrucadu: The issue I had in finder-frontend is that, with the controller tests, the custom header shows up as `HTTP_GOVUK_ACCOUNT_SESSION`, but with the feature specs it shows up as `GOVUK-Account-Session`.  Or maybe it was the other way around.

The only way I could get both sets of tests passing is by checking for both header names, which is a bit unfortunate.

huwd: Ok I have simplifed this as we don't seem to be getting the feature test problem here.

app/controllers/concerns/account_concern.rb
huwd: great thanks, removed

test/functional/session_controller_test.rb
huwd: that's a great shout, removed.

test/support/session_helpers.rb
huwd: Cool, Vary is gone and we've simplifed this

test/support/session_helpers.rb
```diff
@@ -0,0 +1,8 @@
...
+      "GOVUK-Account-Session" => AccountConcern::ACCOUNT_SESSION_HEADER_NAME,
```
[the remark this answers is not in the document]
huwd: yeah you are right, also done.

app/controllers/concerns/account_concern.rb
barrucadu: Can we add a feature spec, or something else which goes through the full request stack, exercising this code?  It would be bad to write a feature which only works in controller tests and breaks when proper HTTP requests are involved.

test/functional/session_controller_test.rb
huwd: ace

app/controllers/concerns/account_concern.rb
huwd: Is that `Cache-Control: no-cache`?

barrucadu: Apparently [no-cache means that it can be cached, but has to be validated before being served](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control).  So I think we want a minimum of `no-cache, no-store`.

[This SO answer](https://stackoverflow.com/a/748646/1058828) suggests:

```
response.headers["Cache-Control"] = "no-cache, no-store"
response.headers["Pragma"] = "no-cache"
response.headers["Expires"] = "Mon, 01 Jan 1990 00:00:00 GMT"
```

app/controllers/concerns/account_concern.rb
barrucadu: I think so, yes.  Those at least support a `headers:` field on the request methods.

app/controllers/concerns/account_concern.rb
huwd: Yeah we were looking at that one too. Erred a bit as it was so old it was suggesting a `before_filter`.
But we'll stick that in the next commit for a try!

app/controllers/concerns/account_concern.rb
```diff
@@ -0,0 +1,55 @@
...
+      if request.headers.to_h[ACCOUNT_SESSION_HEADER_NAME]
+        request.headers.to_h[ACCOUNT_SESSION_HEADER_NAME]
+      elsif Rails.env.development?
+        cookies[ACCOUNT_SESSION_DEV_COOKIE_NAME]
+      end
+  end
+
+  def set_account_session_header(govuk_account_session = nil)
+    @account_session_header = govuk_account_session if govuk_account_session
+    response.headers[ACCOUNT_SESSION_HEADER_NAME] = @account_session_header
```
barrucadu, line 33: I think it's just the request headers that use the `HTTP_GOVUK_ACCOUNT_SESSION` format, and the response headers use the `GOVUK-Account-Session` format.  See [this bit of govuk_ab_testing](https://github.com/alphagov/govuk_ab_testing/blob/master/lib/govuk_ab_testing/ab_test.rb#L29-L36).

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    redirect_to Plek.new.website_root and return unless params[:code] && params[:state]
+
+    callback = GdsApi.account_api.validate_auth_response(
+      code: params.require(:code),
+      state: params.require(:state),
+    ).to_h
+
+    set_account_session_header(callback["govuk_account_session"])
+
+    redirect_with_ga(callback["redirect_path"] || account_manager_url, callback["ga_client_id"])
```
barrucadu, line 27: I know this is how it currently is in finder-frontend, but this could be

```ruby
redirect_with_ga(callback.fetch("redirect_path", account_manager_url), callback["ga_client_id"])
```

Up to you.

test/functional/session_controller_test.rb
```diff
@@ -0,0 +1,116 @@
...
+  include GdsApi::TestHelpers::AccountApi
+  include SessionHelpers
+
+  context "GET sign-in" do
+    setup do
+      stub_account_api_get_sign_in_url(redirect_path: "/bank-holiday")
+    end
+
+    should "redirect the user to the GOV.UK Account service domain" do
+      get :create, params: { redirect_path: "/bank-holiday", state_id: "state123" }
```
barrucadu, line 14: How does this `:create`, `:callback`, and `:delete` work?  Is it inferring the controller from the name of the class?

test/functional/session_controller_test.rb
```diff
@@ -0,0 +1,116 @@
...
+    end
+  end
+
+  context "GET /sign-out" do
+    context "when the user is logged in" do
+      setup do
+        mock_logged_in_session
+      end
+
+      should "set ACCOUNT_END_SESSION_HEADER_NAME response header to 1" do
```
barrucadu, line 110: The name of this test is now slightly misleading.  How about `set the 'GOVUK-Account-End-Session' header`?

test/integration/sessions_test.rb
```diff
@@ -0,0 +1,10 @@
+require "integration_test_helper"
+
+class SessionsTest < ActionDispatch::IntegrationTest
+  context "when logged in" do
+    should "redirect the user to the account manager URL if they visit /sign-up" do
+      get "/sign-in", headers: { "HTTP_GOVUK_ACCOUNT_SESSION" => "placeholder" }
```
barrucadu, line 6: Can we use the external name of the header, `GOVUK-Account-Session` here?  The point of this test is to check that Rails is actually renaming the header internally.

barrucadu:
Here's a diff making the header situation really explicit: `REQUEST_HEADER_NAME`s and `RESPONSE_HEADER_NAME`s in the code, and the plain external names in the tests:


```diff
diff --git a/app/controllers/concerns/account_concern.rb b/app/controllers/concerns/account_concern.rb
index da45ad81..6edc4bf7 100644
--- a/app/controllers/concerns/account_concern.rb
+++ b/app/controllers/concerns/account_concern.rb
@@ -3,8 +3,9 @@
 module AccountConcern
   extend ActiveSupport::Concern
 
-  ACCOUNT_SESSION_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION"
-  ACCOUNT_END_SESSION_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_END_SESSION"
+  ACCOUNT_SESSION_REQUEST_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION"
+  ACCOUNT_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-Session"
+  ACCOUNT_END_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-End-Session"
   ACCOUNT_SESSION_DEV_COOKIE_NAME = "govuk_account_session"
 
   included do
@@ -21,8 +22,8 @@ module AccountConcern
 
   def fetch_account_session_header
     @account_session_header =
-      if request.headers.to_h[ACCOUNT_SESSION_HEADER_NAME]
-        request.headers.to_h[ACCOUNT_SESSION_HEADER_NAME]
+      if request.headers[ACCOUNT_SESSION_REQUEST_HEADER_NAME]
+        request.headers[ACCOUNT_SESSION_REQUEST_HEADER_NAME]
       elsif Rails.env.development?
         cookies[ACCOUNT_SESSION_DEV_COOKIE_NAME]
       end
@@ -30,7 +31,7 @@ module AccountConcern
 
   def set_account_session_header(govuk_account_session = nil)
     @account_session_header = govuk_account_session if govuk_account_session
-    response.headers[ACCOUNT_SESSION_HEADER_NAME] = @account_session_header
+    response.headers[ACCOUNT_SESSION_RESPONSE_HEADER_NAME] = @account_session_header
 
     if Rails.env.development?
       cookies[ACCOUNT_SESSION_DEV_COOKIE_NAME] = {
@@ -41,7 +42,7 @@ module AccountConcern
   end
 
   def logout!
-    response.headers[ACCOUNT_END_SESSION_HEADER_NAME] = "1"
+    response.headers[ACCOUNT_END_SESSION_RESPONSE_HEADER_NAME] = "1"
     @account_session_header = nil
 
     if Rails.env.development?
diff --git a/test/functional/session_controller_test.rb b/test/functional/session_controller_test.rb
index a7b0c7c5..e0002952 100644
--- a/test/functional/session_controller_test.rb
+++ b/test/functional/session_controller_test.rb
@@ -56,7 +56,7 @@ class SessionsControllerTest < ActionController::TestCase
       stub_account_api_validates_auth_response(govuk_account_session: "placeholder")
       get :callback, params: { code: "code123", state: "state123" }
 
-      assert_equal @response.headers[AccountConcern::ACCOUNT_SESSION_HEADER_NAME], "placeholder"
+      assert_equal @response.headers["GOVUK-Account-Session"], "placeholder"
     end
 
     should "redirect to the redirect path if a redirect path is in the account-api response" do
@@ -109,7 +109,7 @@ class SessionsControllerTest < ActionController::TestCase
 
       should "set ACCOUNT_END_SESSION_HEADER_NAME response header to 1" do
         get :delete
-        assert @response.headers[AccountConcern::ACCOUNT_END_SESSION_HEADER_NAME].present?
+        assert @response.headers["GOVUK-Account-End-Session"].present?
       end
     end
   end
diff --git a/test/integration/sessions_test.rb b/test/integration/sessions_test.rb
index 1fcc3460..9185d8ec 100644
--- a/test/integration/sessions_test.rb
+++ b/test/integration/sessions_test.rb
@@ -3,7 +3,7 @@ require "integration_test_helper"
 class SessionsTest < ActionDispatch::IntegrationTest
   context "when logged in" do
     should "redirect the user to the account manager URL if they visit /sign-up" do
-      get "/sign-in", headers: { "HTTP_GOVUK_ACCOUNT_SESSION" => "placeholder" }
+      get "/sign-in", headers: { "GOVUK-Account-Session" => "placeholder" }
       assert_redirected_to Plek.find("account-manager")
     end
   end
diff --git a/test/support/session_helpers.rb b/test/support/session_helpers.rb
index 689d064d..1430339b 100644
--- a/test/support/session_helpers.rb
+++ b/test/support/session_helpers.rb
@@ -1,5 +1,5 @@
 module SessionHelpers
   def mock_logged_in_session
-    request.headers[AccountConcern::ACCOUNT_SESSION_HEADER_NAME] = "placeholder"
+    request.headers["GOVUK-Account-Session"] = "placeholder"
   end
 end
```

The tests pass with this applied, so it looks like the behaviour is that request headers get renamed to `"HTTP_" + header_name.gsub(/-/,"_").upcase`, whereas response headers are passed through unmangled.

app/controllers/sessions_controller.rb
huwd: yeah great, i prefer the fetch default, thanks!

huwd: So i tried this there's a catch. The callback is returning:
`{"govuk_account_session"=>"govuk-account-session", "redirect_path"=>nil, "ga_client_id"=>nil} `

This is because we're passing it no values here in the mock:
[stub_account_api_validates_auth_response(redirect_path: nil, ga_client_id: nil)](https://github.com/alphagov/frontend/pull/2668/files#diff-e9270928c40e1eef08f8c8576f6c7d9f774a7089b3c5ac515da590f2952bca02R71)

Which we decided to do here, because there is a [sensible default in the GDS API Adaptor test helper](https://github.com/alphagov/gds-api-adapters/blob/e0ed4c8070a2b37009be2cdcb735564f82d7b7c7/lib/gds_api/test_helpers/account_api.rb#L17).

I think our options are:
1. leave this unchanged and not use fetch. Accept but ignore the crumminess in the tests
2. call `.compact` [in the test helper ](https://github.com/alphagov/gds-api-adapters/blob/e0ed4c8070a2b37009be2cdcb735564f82d7b7c7/lib/gds_api/test_helpers/account_api.rb#L22). This seems more elegant but I'd want to think through if there are any scenarios where it would be valid to have a key with no string value... I believe there are some header scenraios where that's possible?
3. I could also put this in the controller code, but that seems to have all the issues of above plus we're changing our code to suit our tests. would prefer to avoid this one.


I can call `.compact` on it, but the

huwd: For now, it's unchanged

test/functional/session_controller_test.rb
huwd: 🤔 so we're inheriting here from `ActionController::TestCase`.
The rails docs for this[ show use of symbols](https://api.rubyonrails.org/classes/ActionController/TestCase.html) that all match the method names of a controller.

My assumption is there's something in `ActionController::TestCase` that matches to a controller file based on the test name... and then uses the symbols to call something like `send(:create)`.

As it's a controller test for function i'd guess this skips the routes alltogether... 

Not really sure let me go take a look around in ActionController::TestCase...

huwd: The top of the docs do say:
> Rails discourages the use of functional tests in favor of integration tests (use ActionDispatch::IntegrationTest).

But when I was floating Rspec rewrites here the general consensus seemed to be that it could be more confusing having tests breaking from the norm here...

So i'm not sure if these controller tests should exist because... that's what we've done so far in frontend... or if because of their weakness around things like routes we shouldn't be writing any more, even in minitest and we should migrate this over to the integration/request tests..?

barrucadu: > So i'm not sure if these controller tests should exist because... that's what we've done so far in frontend...

I think that's a good reason to add them.  Maybe we should migrate frontend over to using the integration tests, as a partial improvement for this repo, but that can come later.

huwd: Looks like it might get the[ path from here](https://github.com/rails/rails/blob/5aaaa1630ae9a71b3c3ecc4dc46074d678c08d67/actionpack/lib/action_controller/test_case.rb#L499-L501)...
But then where it tries to define the `@controller`... there's a load of `klass` milarky going on in there too.

app/controllers/sessions_controller.rb
barrucadu: Ok, let's leave it unchanged

app/controllers/concerns/account_concern.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def logged_in?
+    account_session_header.present?
+  end
+
+  def fetch_account_session_header
+    @account_session_header =
+      if request.headers.to_h[ACCOUNT_SESSION_REQUEST_HEADER_NAME]
+        request.headers.to_h[ACCOUNT_SESSION_REQUEST_HEADER_NAME]
```
barrucadu, line 26: Don't need the `to_h` any more

app/controllers/concerns/account_concern.rb
```diff
@@ -0,0 +1,56 @@
...
+    if Rails.env.development?
+      cookies[ACCOUNT_SESSION_DEV_COOKIE_NAME] = {
+        value: @account_session_header,
+        domain: "dev.gov.uk",
+      }
+    end
+  end
+
+  def logout!
+    response.headers[ACCOUNT_END_SESSION_HEADER_NAME] = "1"
```
barrucadu, line 45: `ACCOUNT_END_SESSION_RESPONSE_HEADER_NAME`

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  include AccountConcern
+
+  before_action :set_no_cache_headers
+
+  def create
+    redirect_with_ga account_manager_url and return if logged_in?
+
+    redirect_with_ga GdsApi.account_api.get_sign_in_url(
+      redirect_path: params[:redirect_path],
+      state_id: params[:state],
```
barrucadu, line 13: [Use `state_id` for the param name too](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-134-govuk-wide-session-cookie-and-login.md#extend-frontend-to-manage-the-auth-process).

test/functional/session_controller_test.rb
```diff
@@ -0,0 +1,116 @@
...
+    end
+  end
+
+  context "GET sign-in/callback" do
+    setup do
+      @root_path = Plek.new.website_root
+    end
+
+    should "redirect to the root path if the :code param is not present" do
+      stub_account_api_validates_auth_response
```
barrucadu, line 40: `stub_account_api_validates_auth_response` isn't needed in this test, because the redirect happens before the `GdsApi.account_api.validate_auth_response`

test/functional/session_controller_test.rb
```diff
@@ -0,0 +1,116 @@
...
+      get :callback, params: { state: "state123" }
+
+      assert_response :redirect
+      assert_equal @response.redirect_url, @root_path
+    end
+
+    should "redirect to the root path if the :state param is not present" do
+      stub_account_api_validates_auth_response
```
barrucadu, line 48: Same here

test/functional/session_controller_test.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  context "GET /sign-out" do
+    context "when the user is logged in" do
+      setup do
+        mock_logged_in_session
+      end
+
+      should "set the 'GOVUK-Account-End-Session' header to 1" do
+        get :delete
+        assert @response.headers[AccountConcern::ACCOUNT_END_SESSION_HEADER_NAME].present?
```
barrucadu, line 112: `AccountConcern::ACCOUNT_END_SESSION_RESPONSE_HEADER_NAME`

app/controllers/concerns/account_concern.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module AccountConcern
+  extend ActiveSupport::Concern
+
+  ACCOUNT_SESSION_REQUEST_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION"
+  ACCOUNT_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-Session"
+  ACCOUNT_END_SESSION_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_END_SESSION"
```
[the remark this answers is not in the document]
huwd: Done here: https://github.com/alphagov/frontend/pull/2668/commits/964be49a4c9a9aa47a4dab5293a610c5eff31aa0

app/controllers/concerns/account_concern.rb
huwd: Done here: https://github.com/alphagov/frontend/pull/2668/commits/964be49a4c9a9aa47a4dab5293a610c5eff31aa0

app/controllers/concerns/account_concern.rb
huwd: done here: https://github.com/alphagov/frontend/pull/2668/commits/205008d03ba25a5f727f330c48247c0ccd055d13

test/functional/session_controller_test.rb
huwd: Done here: https://github.com/alphagov/frontend/pull/2668/commits/964be49a4c9a9aa47a4dab5293a610c5eff31aa0

test/functional/session_controller_test.rb
huwd: ah great spot thanks

test/functional/session_controller_test.rb
huwd: and here: https://github.com/alphagov/frontend/pull/2668/commits/4c01553d45069b6b07b1130b5b2ffd7a5a481540

test/functional/session_controller_test.rb
huwd: done here: https://github.com/alphagov/frontend/pull/2668/commits/4c01553d45069b6b07b1130b5b2ffd7a5a481540

app/controllers/sessions_controller.rb
huwd: Done here: https://github.com/alphagov/frontend/pull/2668/commits/ebb7aa9d44df4229db2aa27904d65f999166731d
Thanks!

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,66 @@
+require "gds_api/account_api"
```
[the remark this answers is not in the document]
huwd: very odd, not sure why that made it in there.. works without the line and there's no precident for that in the other controllers!

app/controllers/concerns/account_concern.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module AccountConcern
+  extend ActiveSupport::Concern
+
+  ACCOUNT_SESSION_REQUEST_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION"
+  ACCOUNT_SESSION_RESPONSE_HEADER_NAME = "GOVUK-Account-Session"
+  ACCOUNT_END_SESSION_RESPONSE_HEADER_NAME = "HTTP_GOVUK_ACCOUNT_END_SESSION"
```
barrucadu, line 8: The response header name is `GOVUK-Account-End-Session`, not `HTTP_GOVUK_ACCOUNT_END_SESSION`

app/controllers/concerns/account_concern.rb
```diff
@@ -0,0 +1,56 @@
...
+  ACCOUNT_SESSION_DEV_COOKIE_NAME = "govuk_account_session"
+
+  included do
+    before_action :fetch_account_session_header
+    before_action :set_account_session_header
```
barrucadu, line 13: I don't think we need the `before_action :set_account_session_header` any more

huwd: removed

barrucadu (approved):

PR #2677, opened 2021-03-24
Add fetch_http_referer header

huwd:
[Trello](https://trello.com/c/43FM4pwv/673-switch-to-the-new-sign-in-out-endpoints)

What
====

Requests to `/sign-in` should be sent here from buttons or links a user
has clicked somewhere on GOV.UK. With those requets should be an
HTTP_REFERER header that tells us where they were when they clicked.

Here we grab that header and use it to populate the redirect_path param
we send to the API.

Why
===

Whenever the user logs in, they are mostly likely trying to get
something done. The login flow appears as in interruption - they were
trying to find out about something, and GOV.UK took them away to log in
when the took an action.

So when they are authenticated we want to return the user to where they
started so they can continue with their journey.

Other options would be to send them to a fixed location, perhaps the
account dashboard, homepage or an account explainer page. But that
assumes that logging in would be the feature the user was seeking to
engage with, when most often they would be logging in so they could
achieve something else on the site.

How
===

Referer is User input and so cannot be fully trusted. We default to not
sending it if anything looks suspicious. In that case the user will be
taken to a fixed location like their account dashboard. We test for:

- If the referer header is present
- If the referer header starts with the Website root from plek
(https://gov.uk on production)
- If the result is a well formatted URI that can be parsed and have a
path returned.

If all is well there, we will populate the redirect_path param we send
to the API.

On line 9, we also accept a param[:redirect_path] and prefer this to
fetching HTTP headers. Meaning we can still add params to sign-in links
if we want to.

These are preferred to the header option as they are more explicit and
transparent to the user, so we will try for those first and only try for
headers after that.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -7,7 +7,7 @@ class SessionsController < ApplicationController
     redirect_with_ga account_manager_url and return if logged_in?
 
     redirect_with_ga GdsApi.account_api.get_sign_in_url(
-      redirect_path: params[:redirect_path],
+      redirect_path: params[:redirect_path] || fetch_http_referrer,
       state_id: params[:state_id],
     ).to_h["auth_uri"]
   end
@@ -46,6 +46,14 @@ protected
 
   def account_manager_url
     Plek.find("account-manager")
+  end
+
+  def fetch_http_referrer
+    http_referrer = request.headers["HTTP_REFERER"]
+
+    return nil unless http_referrer&.start_with?(Plek.new.website_root)
+
+    URI.parse(http_referrer).path
   end
 
   def redirect_with_ga(url, ga_client_id = nil)
--- a/test/integration/sessions_test.rb
+++ b/test/integration/sessions_test.rb
@@ -1,10 +1,43 @@
 require "integration_test_helper"
+require "gds_api/test_helpers/account_api"
 
 class SessionsTest < ActionDispatch::IntegrationTest
+  include GdsApi::TestHelpers::AccountApi
+
   context "when logged in" do
-    should "redirect the user to the account manager URL if they visit /sign-up" do
+    should "redirect the user to the account manager URL if they visit /sign-in" do
       get "/sign-in", headers: { "GOVUK-Account-Session" => "placeholder" }
       assert_redirected_to Plek.find("account-manager")
+    end
+  end
+
+  context "when logged out" do
+    should "prefer the redirect_path over the HTTP Referer" do
+      stub = stub_account_api_get_sign_in_url(redirect_path: "/from-param")
+
+      get "/sign-in", headers: { "Referer" => "#{Plek.new.website_root}/from-referer" }, params: { redirect_path: "/from-param" }
+
+      assert_response :redirect
+      assert_requested stub
+    end
+
+    should "add a redirect_path param to /sign-in from the HTTP Referer header" do
+      stub = stub_account_api_get_sign_in_url(redirect_path: "/from-referer")
+
+      get "/sign-in", headers: { "Referer" => "#{Plek.new.website_root}/from-referer" }
+
+      assert_response :redirect
+      assert_requested stub
+    end
+
+    should "not add an invalid path as the redirect param" do
+      stub_account_api_get_sign_in_url
+      stub_evil = stub_account_api_get_sign_in_url(redirect_path: "/from-referer")
+
+      get "/sign-in", headers: { "Referer" => "//evil/from-referer" }
+
+      assert_response :redirect
+      assert_not_requested stub_evil
     end
   end
 end


test/integration/sessions_test.rb
```diff
@@ -1,10 +1,22 @@
...
-    should "redirect the user to the account manager URL if they visit /sign-up" do
+    should "redirect the user to the account manager URL if they visit /sign-in" do
       get "/sign-in", headers: { "GOVUK-Account-Session" => "placeholder" }
       assert_redirected_to Plek.find("account-manager")
     end
   end
+
+  context "when logged out" do
+    should "add a redirect_path param to /sign-in from the HTTP REFERRER header" do
+      stub_account_api_get_sign_in_url(redirect_path: "/bank-holiday")
```
huwd, line 16: I tried to think of a more explicit test here, perhaps by snooping on the `request`.
But in this case the test fails because webmock is unhappy with the path if the referer header and redirect path in the mock do not align.

barrucadu:
A comment on the motivation:

> The login flow appears as in interruption - they were
> trying to find out about something, and GOV.UK took them away to log in
> when the took an action.

&

> Meaning we can still add params to sign-in links
> if we want to.

We will always set the `redirect_path` explicitly when logging the user in to do something.  Explicit is better than implicit and, anyway, I'm sure there are browser extensions which disable the header.  It will only be unset if the user clicks the "Sign in" link in the header or visits `gov.uk/sign-in` directly (and in that case we won't have a `Referer` header).  This isn't a feature we'll rely on to keep GOV.UK working (eg, we're not now going to go drop the `redirect_path`s from the Transition Checker).

Could the commit message be updated to reflect this?

app/controllers/sessions_controller.rb
```diff
@@ -48,6 +48,16 @@ def account_manager_url
     Plek.find("account-manager")
   end
 
+  def fetch_http_referrer
+    http_referrer = request.headers["HTTP_REFERER"]
+
+    return nil unless http_referrer
+
+    return nil unless http_referrer.start_with?(Plek.new.website_root)
```
barrucadu, line 56: `return nil unless http_referrer&.start_with?(Plek.new.website_root)`

test/integration/sessions_test.rb
```diff
@@ -1,10 +1,22 @@
...
+
+      get "/sign-in", headers: { "HTTP_REFERER" => "#{Plek.new.website_root}/bank-holiday" }
```
barrucadu, line 18: The request header to send in the test is `Referer`, not `HTTP_REFERER`, isn't it?

barrucadu: It would be good to have a couple more tests:

1. Set both a `redirect_path` and a `Referer` - check that the `redirect_path` wins
2. Set a `Referer` going to some other domain - check that it's not used

To be very explicit, the test could even assert that the stub has been called (making it more likely to fail if a bug gets introduced later):

```ruby
stub = stub_account_api_get_sign_in_url(redirect_path: "/from-referer")

get "/sign-in", headers: { Referer: "#{Plek.new.website_root}/from-referer" }

assert_response :redirect
assert_requested stub
```

```ruby
stub_params = stub_account_api_get_sign_in_url(redirect_path: "/from-params")
stub_referer = stub_account_api_get_sign_in_url(redirect_path: "/from-referer")

get "/sign-in", headers: { Referer: "#{Plek.new.website_root}/from-referer" }, params: { redirect_path: "/from-params" }

assert_response :redirect
assert_requested stub_params
assert_not_requested stub_referer
```


```ruby
stub_account_api_get_sign_in_url
stub = stub_account_api_get_sign_in_url(redirect_path: "/from-referer")

get "/sign-in", headers: { Referer: "//evil/from-referer" }

assert_response :redirect
assert_not_requested stub
```

huwd: you are right using `/from-referer` is much clearer than the bank holiday path.
amended those.

I like the evil path test too, added that.
Just checking on the Referer header... These name changes within rails are still confusing me I think.

huwd: Thanks these are amended

app/controllers/sessions_controller.rb
```diff
@@ -48,6 +48,16 @@ def account_manager_url
     Plek.find("account-manager")
   end
 
+  def fetch_http_referrer
+    http_referrer = request.headers["HTTP_REFERER"]
+
+    return nil unless http_referrer
+
+    return nil unless http_referrer&.start_with?(Plek.new.website_root)
```
barrucadu, line 56: With the `&.` in this conditional, we don't need the `return nil unless http_referrer` above.  Also `return nil` is the same as `return`, so these three lines can be

```ruby
return unless http_referrer&.start_with?(Plek.new.website_root)
```

test/integration/sessions_test.rb
```diff
@@ -1,10 +1,34 @@
...
+
   context "when logged in" do
-    should "redirect the user to the account manager URL if they visit /sign-up" do
+    should "redirect the user to the account manager URL if they visit /sign-in" do
       get "/sign-in", headers: { "GOVUK-Account-Session" => "placeholder" }
       assert_redirected_to Plek.find("account-manager")
     end
   end
+
+  context "when logged out" do
```
barrucadu, line 14: Can you add a test like

```ruby
    should "prefer the redirect_path over the HTTP Referer" do
      stub = stub_account_api_get_sign_in_url(redirect_path: "/from-param")

      get "/sign-in", headers: { "Referer" => "#{Plek.new.website_root}/from-referer" }

      assert_response :redirect
      assert_requested stub
    end
```

app/controllers/sessions_controller.rb
huwd: ah great point, removed.

barrucadu (approved):

PR #2691, opened 2021-03-30
Adds country notice to the business support smart answer

DilwoarH:
## WHAT
We get the blue banner up (and down) on the business support smart answer for two weeks for the purpose of the test.

## WHY
We are doing a before and after test and the banner is the key bit!

## Before

Screenshot 2021-03-30 at 20 56 21

screenshot-www gov uk-2021 03 30-20_55_58

## After

Screenshot 2021-03-30 at 20 55 35

screenshot-127 0 0 1_3005-2021 03 30-20_53_18

https://trello.com/c/Py8XjzRQ/275-implement-blue-banner-for-before-and-after-test

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -60,6 +60,7 @@ $govuk-use-legacy-palette: false;
 @import "views/report-child-abuse";
 @import "views/local-authority";
 @import "views/simple-smart-answer";
+@import "views/transaction";
 
 // exceptional format overrides
 @import "views/jobsearch";
--- a/app/presenters/transaction_presenter.rb
+++ b/app/presenters/transaction_presenter.rb
@@ -48,6 +48,10 @@ class TransactionPresenter < ContentItemPresenter
     end
   end
 
+  def show_experimental_country_notice?
+    @content_item["content_id"] == "89edffd2-3046-40bd-810c-cc1a13c05b6a"
+  end
+
 private
 
   def variant_value(key)
--- a/app/presenters/transaction_presenter.rb
+++ b/app/presenters/transaction_presenter.rb
@@ -49,6 +49,8 @@ class TransactionPresenter < ContentItemPresenter
   end
 
   def show_experimental_country_notice?
+    return false if Time.zone.today > Date.new(2021, 4, 20)
+
     @content_item["content_id"] == "89edffd2-3046-40bd-810c-cc1a13c05b6a"
   end
 
--- a/app/views/transaction/show.html.erb
+++ b/app/views/transaction/show.html.erb
@@ -24,6 +24,9 @@
   edition: @edition
 } do %>
   <section class="intro">
+    <%# Temporary experiment to test if applicability banner increases devolved administration usage %>
+    <%= render 'transaction/country_notice' if @publication.show_experimental_country_notice? %>
+
     <div class="get-started-intro">
       <%= render "govuk_publishing_components/components/govspeak", { } do %>
         <%= @publication.introductory_paragraph.try(:html_safe) %>
--- a/app/assets/stylesheets/views/_transaction.scss
+++ b/app/assets/stylesheets/views/_transaction.scss
@@ -1,0 +1,21 @@
+.country-notice {
+  @include govuk-font(19);
+
+  background-color: govuk-colour('blue');
+  color: govuk-colour('white');
+  padding: govuk-spacing(4);
+  margin-bottom: govuk-spacing(4);
+
+  @include govuk-media-query($from: tablet) {
+    padding: govuk-spacing(4) govuk-spacing(6);
+  }
+}
+
+.country-notice__pretext {
+  float: left;
+  padding-right: govuk-spacing(1);
+}
+
+.country-notice__text {
+  @include govuk-font(19, $weight: bold);
+}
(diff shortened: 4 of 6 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -59,6 +59,7 @@ $govuk-use-legacy-palette: false;
 @import "views/report-child-abuse";
 @import "views/local-authority";
 @import "views/simple-smart-answer";
+@import "views/country-notice";
```
kevindew, line 62: The organisation of this file doesn't seem quite right to me, since the country notice is a sub view of transaction rather than a view in its own right. Though looking at the rest of the file the organisation of CSS in this app looks a bit of a car crash 🙈 .

I imagine the simplest thing would be to include this styling on a views/transaction stylesheet that maps to any additional transaction styling.

app/presenters/transaction_presenter.rb
```diff
@@ -1,3 +1,5 @@
+require 'date'
```
kevindew, line 1: I don't think you need this.

app/views/transaction/show.html.erb
```diff
@@ -24,6 +24,8 @@
   edition: @edition
 } do %>
   <section class="intro">
+    <%= render 'transaction/country_notice' if @publication.show_country_notice? %>
```
kevindew, line 27: I feel like there's perhaps not enough about this code that screams this is a temporary experiment. I'd wonder if it's simple to just put all the code in this view so it's all compartmentalised - especially since the HTML is short.

Something like:

```suggestion
    <%# Temporary experiment to test if applicability banner increases devolved administration usage %>
    <% if @publication.base_path == "/business-coronavirus-support-finder" && Date.today < Date.new(2021, 4, 17) %>
        <div class="country-notice">
          <p class="country-notice__text">
            Applies to: <strong>England, Scotland, Wales and Northern&nbsp;Ireland</strong>
          </p>
        </div>
    <% end %>
```

app/presenters/transaction_presenter.rb
```diff
@@ -48,6 +50,12 @@ def start_button_text
     end
   end
 
+  def show_country_notice?
+    return false if Date.today < (Date.new(2020,04,01) + 14.days)
```
kevindew, line 54: Eeep wrong year 🤦

app/presenters/transaction_presenter.rb
```diff
@@ -48,6 +50,12 @@ def start_button_text
...
+
+    @content_item["content_id"] == "89edffd2-3046-40bd-810c-cc1a13c05b6a"
```
kevindew, line 56: I wonder, given this is a short term experiment, whether it's best to use the base_path, having this makes it a lot easier to know which page it's used on.

For something longer term, content_id is best though so I can't fault your decision to use it.

kevindew (commented):
This is looking great, I've got a few comments about wider consistency and logic that need ironing out but hopefully this isn't far off being ready to go.

app/assets/stylesheets/views/_country-notice.scss
```diff
@@ -0,0 +1,15 @@
...
+  @include govuk-font(19);
+
+  background-color: govuk-colour('blue');
+  padding: govuk-spacing(4) govuk-spacing(4) govuk-spacing(4);
+  margin-bottom: govuk-spacing(4);
+
+  @include govuk-media-query($from: tablet) {
+    padding: govuk-spacing(5) govuk-spacing(5) govuk-spacing(5);
+  }
+}
```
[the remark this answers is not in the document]
DilwoarH: The font size needed to be different as start pages seem to be font-size 19 where it would otherwise be 16 for the metadata

app/presenters/transaction_presenter.rb
DilwoarH: epppp... no wonder it was being funny ......

DilwoarH: good spot, thank you

app/presenters/transaction_presenter.rb
DilwoarH: Yes definitely, I looked at similar implementations where we needed it for a specific content item and we have always used content id rather than base path

app/views/transaction/show.html.erb
DilwoarH: good point - I'll add your suggestion in

DilwoarH: I wanted to keep the stuff in the presenter to keep it a bit cleaner, I'll add the comment in though

DilwoarH:
@kevindew - this is now ready

DilwoarH:
@kevindew - all agreed changes have been added

kevindew (approved):
Looks good 👍  We should set a due date on Trello of when to review/remove it once the 2 weeks are up.

alex-ju:
Was there a reason for not using the `app-c-important-metadata` component we have on [`government-frontend`](https://www.gov.uk/guidance/covid-19-coronavirus-restrictions-what-you-can-and-cannot-do)?

alex-ju:
I assume because of the temporary aspect of this experiment. Just a heads up, I [ported this](https://github.com/alphagov/smart-answers/pull/5286) as part of the work to [render 'landing'/'start' pages with smart-answers](https://github.com/alphagov/smart-answers/pull/5277) – for when you need to remove it.

Issue #2542, opened 2020-12-08
Organisation numbers on the home page need manual updating


huwd:
## Problem

Whilst department numbers tend to stay fairly consistent. "Agencies and other public bodies" vary quite frequently. See:

- #2494
- #2494
- #2446
- #2419
- #2280
- #2202
- #2183
- #2079

There is often quite a bit of drift and lag. I started doing this becuase I noticed it before I joined GOV.UK and then realised I could fix it after I joined.

I think it generally makes us look a bit silly to have two different numbers on such prominent pages.

## What to do about it

We can make this dynamic, the numbers here are accessible via the public organisation api.
- See organisations content [item here](https://www.gov.uk/api/content/government/organisations).

if you parsed the JSON something like:

```
  def ministerial_departments_count
    organisations_json.dig("details", "ordered_ministerial_departments").count
  end

  def other_agencies_public_bodies_count
    organisations_json.dig("details", "ordered_agencies_and_other_public_bodies").count
  end
``` 

Would work.

## Issues

The main thing that stopped me getting further with this was a fairly old testing stack on this repo. It doesn't appear to be using rspec currently so I was thinking of adding that alongside minitest here so we could at least ensure new tests follow current GOV.UK convention.

Also there's a good point that before we pour effort into this we should probably stop and ask if this section still brings value?
I understand it arrived quite early during the major transitions of other sites onto a single GOV.UK. That time seems quite far in the past now. I wonder if this section still meets a user need today?

Perhaps this is a good opportunity to check and even free up some homepage real estate before we begin making this dynamic?

huwd:
There is some potential work to revise some parts of the home page.
We should leave it for a little while longer to see how that work plays out.
If this page is going to continue for a bit, I'll automate this.

If it's about to be revised I'll leave it to be folded into that piece of work.

huwd:
This is now resolved 🎉

owenatgov:
@huwd If this is resolved, can we close the issue? 👀

PR #2692, opened 2021-04-01
Automate org number updates

huwd:
## What
The home page... 
https://www.gov.uk/

image

has fallen out of sync with the number of organisations again.
https://www.gov.uk/government/organisations#agencies_and_other_public_bodies

image

[This happens a lot](https://github.com/alphagov/frontend/issues/2542).

I'm finally going to automate this job away beacuse it drives me insane.

## What does it look like after?

Tested with govuk-docker frontend-app-live stack and seems to be pulling through OK!
image

### DIFF
--- a/app/helpers/homepage_helper.rb
+++ b/app/helpers/homepage_helper.rb
@@ -1,0 +1,19 @@
+module HomepageHelper
+  def organisations_json
+    GdsApi.content_store.content_item("/government/organisations")
+  end
+
+  def ministerial_departments_count
+    organisations_json
+      .dig("details", "ordered_ministerial_departments")
+      .filter { |org| org["govuk_status"] != "joining" }
+      .count
+  end
+
+  def other_agencies_public_bodies_count
+    organisations_json
+      .dig("details", "ordered_agencies_and_other_public_bodies")
+      .filter { |org| org["govuk_status"] != "joining" }
+      .count
+  end
+end
--- a/app/views/homepage/index.html.erb
+++ b/app/views/homepage/index.html.erb
@@ -57,13 +57,13 @@
           <ul class="home-numbers">
             <li>
               <a href="/government/organisations#ministerial_departments" class="home-numbers__link home-numbers__link--first govuk-link">
-                <strong class="home-numbers__large">23</strong>
+                <strong class="home-numbers__large"><%= ministerial_departments_count %></strong>
                 Ministerial departments
               </a>
             </li>
             <li>
               <a href="/government/organisations#agencies_and_other_public_bodies" class="home-numbers__link govuk-link">
-                <strong class="home-numbers__large">410</strong>
+                <strong class="home-numbers__large"><%= other_agencies_public_bodies_count %></strong>
                 Other agencies and public&nbsp;bodies
               </a>
             </li>
--- a/test/functional/homepage_controller_test.rb
+++ b/test/functional/homepage_controller_test.rb
@@ -5,7 +5,14 @@ class HomepageControllerTest < ActionController::TestCase
 
   context "loading the homepage" do
     setup do
+      stub_orgs = {
+        details: {
+          ordered_ministerial_departments: Array.new(1, {}),
+          ordered_agencies_and_other_public_bodies: Array.new(1, {}),
+        },
+      }
       stub_content_store_has_item("/", schema: "special_route")
+      stub_content_store_has_item("/government/organisations", stub_orgs)
     end
 
     should "respond with success" do
--- a/test/integration/homepage_test.rb
+++ b/test/integration/homepage_test.rb
@@ -2,6 +2,20 @@ require "integration_test_helper"
 
 class HomepageTest < ActionDispatch::IntegrationTest
   setup do
+    @organisations_payload = {
+      organisation_payload: "/government/organisations",
+      document_type: "finder",
+      schema_name: "organisations_homepage",
+      title: "Departments, agencies and public bodies",
+      public_updated_at: "2021-04-01T11:09:50.000+00:00",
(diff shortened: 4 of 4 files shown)


app/helpers/homepage_helper.rb
```diff
@@ -0,0 +1,23 @@
+module HomepageHelper
+  def organisations_path
+    Plek.find("content-store") + "/content/government/organisations"
```
huwd, line 3: Anyone know, is this the correct way to be using Plek?

I'm not sure if I need a method like:
```
  def base_path
    Rails.env.production? ? Plek.new.website_root : Plek.find("content-store")
  end
```

Plek seems to change based on an environment variable, do we usually have those set to `/` or `https://gov.uk/` in prod?

app/helpers/homepage_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def organisations_json
+    JSON.parse(RestClient.get(organisations_path))
+  end
+
+  def ministerial_departments_count
+    organisations_json
+      .dig("details", "ordered_ministerial_departments")
+      .filter { |org| org["govuk_status"] != "joining" }
```
huwd, line 13: This is here because there seems to be a presenter in [collections](https://github.com/alphagov/collections/blob/53189e4b723d67d27a50cedb57981f39dface7f9/app/presenters/organisations/index_presenter.rb#L51-L55) that filters by this before we render on: https://www.gov.uk/government/organisations#agencies_and_other_public_bodies

🤔 I'm not sure why this would be the case though. Anyone know the history of this?

huwd:
Woops, missed one more test, will put it in a separate commit

app/helpers/homepage_helper.rb
```diff
@@ -0,0 +1,23 @@
+module HomepageHelper
+  def organisations_path
+    Plek.find("content-store") + "/content/government/organisations"
+  end
+
+  def organisations_json
+    JSON.parse(RestClient.get(organisations_path))
```
thomasleese, line 7: I'd probably [use gds-api-adapters for this](https://github.com/alphagov/gds-api-adapters/blob/main/lib/gds_api/content_store.rb#L13-L17), then you don't need to worry about using Plek either.

```suggestion
    GdsApi.content_store.content_item("/organisations")
```

thomasleese: I also wonder if there's a case for using the Search API here instead, since this means a single request to the homepage requires two requests to the content store (the ideal architecture is that you'd only need to make one call to the content store for each page. However, in this case the homepage is hard coded content anyway so it's a bit of a special case.

app/helpers/homepage_helper.rb
thomasleese: Joining organisations aren't yet public on GOV.UK if I remember correctly, so I think this is why they're not shown in the list yet.

app/helpers/homepage_helper.rb
barrucadu: It looks like there's a `GdsApi.organisations` you can use for this

app/helpers/homepage_helper.rb
huwd: ahh I see, thanks @thomasleese!

app/helpers/homepage_helper.rb
kevindew: This is going to play havoc with the intro to tech architecture session 😢

thomasleese: An alternative to using the Search API would be to link the homepage content item with the `/organisations` content item, and then [use link expansion](https://github.com/alphagov/publishing-api/blob/master/docs/link-expansion.md#introduction) on the parts of the `details` you need from the homepage to avoid the second request to the content store.

thomasleese: (Although I'm a bit confused why the organisations at the moment are in the `details` of `/government/organisations` content item and not in the `links`...)

huwd:
ace thanks @barrucadu and @thomasleese I've taken those two suggestions, that seems to work and we're still getting the right numbers on a manual check too:
image

@thomasleese I might just drop into a slack thread to ask about that search API suggestion?

app/helpers/homepage_helper.rb
huwd: I think it's "government/organisations", but that works!

app/helpers/homepage_helper.rb
```diff
@@ -0,0 +1,19 @@
+module HomepageHelper
+  def organisations_json
+    GdsApi.content_store.content_item("/government/organisations")
```
kevindew, line 3: It's probably worth considering how this could fail and handling it. We have exception handling for loading the homepage item, we'd not want this to break the homepage.

It does seem very strange to use a different content item to load the data for this page. I'd definitely think about if it can be done with a different approach. If we need to use a content item link expansion seems the right way rather than pulling straight from an unrelated URL (this /government/organisations path seems rather whacky since it doesn't even have a page on GOV.UK https://www.gov.uk/government/organisations 🤔 )

kevindew: By the way - for reference this is a link the to the underlying architectural principle: https://docs.google.com/document/d/1Oft4akc6dZfhhOjosNPbFpcLUOUjz7YG7QPcVZi8hww/edit#heading=h.h8e5mqu62b4b

huwd: @kevindew you and @thomasleese have convinced me the content item is not the right thing for this.
I'll keep this open but give it a bit more thought and try a few strategies with these tests.

Thanks very much for the review!

barrucadu:
I'll close this now so it stops popping up in the seal every day.

PR #2705, opened 2021-04-07
[DO NOT MERGE] Update roadmap

andysellick:

**DO NOT MERGE** awaiting final copy

## What / why

Update the roadmap page. Mostly content changes, with a few layout changes, specifically:

- moves the visit counter to the bottom of the page
- adds three links to other pages with images at the top of the page
- adds sharing links at the bottom

## Visual changes

Before | After
------ | ------
screencapture-gov-uk-roadmap-2021-04-07-11_34_58 | screencapture-localhost-3005-roadmap-2021-04-07-11_35_35


andysellick:
When you review, please also check that the sharing links at the bottom of the page work as expected.

danacotoran:
@andysellick if this is to be shared on social media I think we should use metadata to improve the way it would be previewed.
As it is currently, if shared on Facebook it would look a little something like this:
Screenshot 2021-04-20 at 10 15 43
Which is fine, but perhaps a little bit generic? I wonder if we should be more deliberate with it.

If shared on Twitter it would simply appear as a link, as none of the metadata to generate a twitter share card is there. 

This definitely needs some custom Open Graph meta tags (my bad for missing this in the original implementation of this page)!

andysellick:
@danacotoran I've updated the copy today and added the new links for sharing. I don't have either of these accounts (twitter, linkedin) so can you test the links work properly please?

> This definitely needs some custom Open Graph meta tags (my bad for missing this in the original implementation of this page)!

Have added some of these as well.

danacotoran:
Checked the social links and this looks good to me now, but it might be good to get a content designer to check the copy updates properly as I'm not the best at spotting issues with that. 
One thing that stood out to me visually was that on some breakpoints the text in only one of the  "Programme updates" columns wraps onto the next line which does not look great but also not the end of the world if it's not an easy thing to fix.
Screenshot 2021-05-14 at 11 49 46

config/locales/en.yml
```diff
@@ -102,126 +102,136 @@ en:
...
-        - heading: Up next
+          - text: Made use of schema.org standards to integrate local election content and services into search engine results
+        - heading: Working on now
           items:
-          - text: Test running GOV.UK on a different type of hosting platform, to identify and resolve areas of highest risk
-          - text: Fix non-English language pages so all page elements are properly translated
-          - text: Improve the way we manage subdomains of GOV.UK for the rest of government
-          - text: Ongoing work on cookies
-          - text: Ongoing work on accessibility
+          - text: Continue work to use <a href="https://schema.org/" class="govuk-link">schema.org</a> standards on GOV.UK
```
keithiopia, line 222: This should be:
```suggestion
          - text: Continue work to use <a href="https://dataingovernment.blog.gov.uk/2020/04/30/using-new-updates-to-schema-org-to-assist-our-response-to-the-coronavirus-pandemic/?utm_medium=referral%20&utm_source=roadmap_gov&utm_campaign=roadmap_update" class="govuk-link">schema.org standards on GOV.UK</a>
```

andysellick: Thanks Keith, have updated.

config/locales/en.yml
```diff
@@ -111,179 +111,139 @@ en:
...
+    page_title: Roadmap - GOV.UK
+    hero:
+      description: GOV.UK is the trusted source of government information, so we’re always working to better meet user needs. This is our current roadmap, which takes us through to late 2021.
+      heading: GOV.UK roadmap
+    updates_section:
+      heading: Programme updates
+      items:
+      - href: "https://gds.blog.gov.uk/2021/03/29/whats-next-for-gov-uk-in-2021-to-2022/"
+        image_src: "roadmap/updates-goals.jpg"
+        image_alt: "alt text"
```
danacotoran, line 123: These images are decorative. Can we please leave alt text blank so that screen readers skip over them?

config/locales/en.yml
```diff
@@ -111,179 +111,139 @@ en:
...
-              trends more quickly
-      - heading: Everyone in the UK needs to understand how they are affected by coronavirus
-          and Brexit
-        paragraph: Create tools and guidance that help users find critical and up-to-date
-          information
+          - text: Find more ways to automate analysis so government can identify and respond to trends more quickly
+          - text: Experiment with audience analysis across sections of GOV.UK to understand page usage vs reading time to support product and content iteration
+          - text: Simplify how feedback is analysed and gathered, so we can better measure our impact and identify issues
+      - heading: Provide a personalised and proactive service
+        paragraph: Develop sitewide account features that offer users relevant information and more helpful notifications
```
danacotoran, line 204: We spell `site-wide` without the hyphen in some places, and with the hyphen in other places, I would think it should be consistent throughout the page @keithiopia?

app/views/roadmap/index.html.erb
```diff
@@ -9,63 +13,42 @@
...
       </div>
     </section>
-    <section class="govuk-roadmap-section">
+
+    <section class="govuk-roadmap-section govuk-!-padding-bottom-6">
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
           <%= render "govuk_publishing_components/components/heading", {
-            text: t('roadmap.numbers_section.heading_large'),
+            text: "Programme updates",
```
danacotoran, line 29: Should this text live in the locale file?

danacotoran (commented):
I made a few observations, but other than that I think it is looking good 👍

config/locales/en.yml
andysellick: I've gone with `site-wide`.

app/views/roadmap/index.html.erb
andysellick: Yes. In fact it's already in there. Odd. Corrected.

app/views/roadmap/index.html.erb
```diff
@@ -9,63 +13,42 @@
...
-            <span class="govuk-!-font-size-80 govuk-!-display-block">
-              <%= t('roadmap.numbers_section.numbers_columns.column_two_count') %></span><%= t('roadmap.numbers_section.numbers_columns.column_two_text') %>
-          </p>
-        </div>
+        <% t('roadmap.updates_section.items').each do |item| %>
+          <div class="govuk-grid-column-one-third">
+            <%= render "govuk_publishing_components/components/image_card", {
+              href: item[:href],
+              image_src: asset_path(item[:image_src]),
+              image_alt: item[:image_alt],
```
danacotoran, line 41: We don't need to specify `image_alt` because the `image_card` component already [defaults to blank alt text if alt text is not specified](https://github.com/alphagov/govuk_publishing_components/blob/0e23f397270ffc3fc4fd811a84190c308ef9838c/lib/govuk_publishing_components/presenters/image_card_helper.rb#L14).
So I reckon we can remove this line as well as the `image_alt` keys from the locale file.

danacotoran (approved):
Sorry, just had one more comment.
But other than that I think it's good to go for Monday so I am approving 🥇

app/views/roadmap/index.html.erb
andysellick: Fair point, removed.

config/locales/en.yml
keithiopia: Grammatically speaking, they both work. Good choice.

PR #2708, opened 2021-04-08
Use memcached as the default cache store.

hannako:
## What

We are going to add a new election lookup service to Frontend. Using a [shared](https://github.com/alphagov/govuk-aws/blob/c993f31604b7023dc0292e8bdf79eb8e2cced8b7/terraform/projects/app-frontend/main.tf#L193-L205) memcached server will help performance. 

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 gem "rails", "6.1.3.1"
 
 gem "addressable"
+gem "dalli"
 gem "gds-api-adapters"
 gem "govuk_ab_testing"
 gem "govuk_app_config"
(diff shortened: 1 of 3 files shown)


sihugh (approved):
Looks good to me. Frontend nodes (the machine type, not the app) already have a shared [local memcached on each of them](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/node/s_frontend.pp#L17-L20), so we could merge this before https://github.com/alphagov/govuk-puppet/pull/11032 if you'd like?

hannako:
@sihugh Ah, I thought they were only on production machines 🤔 So do we need alphagov/govuk-puppet#11032 at all?

sihugh:
Yep, because that lets us use a shared one between all the nodes to increase our cache hit rate

hannako:
Ah yep - we need to set the ENV var in staging for frontend to write to memchached in staging. I remember now.

PR #2710, opened 2021-04-12
Remove port

ChrisBAshton:
It doesn't seem to be necessary to provide a port number.

### DIFF
--- a/test/service_consumers/pact_helper.rb
+++ b/test/service_consumers/pact_helper.rb
@@ -19,8 +19,7 @@ class ProxyApp
 
   def call(env)
     env["HTTP_HOST"] = "localhost"
-    response = @real_provider_app.call(env)
-    response
+    @real_provider_app.call(env)
   end
 end
 
--- a/test/service_consumers/pact_helper.rb
+++ b/test/service_consumers/pact_helper.rb
@@ -16,7 +16,7 @@ class ProxyApp
   end
 
   def call(env)
-    env["HTTP_HOST"] = "localhost:3003"
+    env["HTTP_HOST"] = "localhost"
     response = @real_provider_app.call(env)
     response
   end
--- a/test/service_consumers/pact_helper.rb
+++ b/test/service_consumers/pact_helper.rb
@@ -10,6 +10,8 @@ Pact.configure do |config|
   config.include GdsApi::TestHelpers::ContentStore
 end
 
+WebMock.allow_net_connect!
+
 class ProxyApp
   def initialize(real_provider_app)
     @real_provider_app = real_provider_app


test/service_consumers/pact_helper.rb
```diff
@@ -10,13 +10,15 @@
...
 class ProxyApp
   def initialize(real_provider_app)
     @real_provider_app = real_provider_app
   end
 
   def call(env)
-    env["HTTP_HOST"] = "localhost:3003"
+    env["HTTP_HOST"] = "localhost"
     response = @real_provider_app.call(env)
     response
```
thomasleese, line 23:
```suggestion
    @real_provider_app.call(env)
```

Not related to this PR, but I did notice this could be a bit tidier.

thomasleese (approved):

test/service_consumers/pact_helper.rb
```diff
@@ -10,15 +10,16 @@
...
 class ProxyApp
   def initialize(real_provider_app)
     @real_provider_app = real_provider_app
   end
 
   def call(env)
-    env["HTTP_HOST"] = "localhost:3003"
-    response = @real_provider_app.call(env)
-    response
+    env["HTTP_HOST"] = "localhost"
```
kevindew, line 21: Awesome that this is working, my comment is probably not needed as it looks like things have moved forward.

Do we know quite what this is used for? I'm a bit suspicious of it since it doesn't have a port. I'm thinking that it's actually only used for redirect testing, which makes it have very little affect in this repo as the tests don't use redirects (it just gets us around:  https://github.com/alphagov/frontend/blob/5023f6f5aaae13891d7b29aad37020c854173940/config/environments/development.rb#L49)

I'm expecting that, for a test suite that features absolute redirects, that it doesn't matter what this value is but the service consumer needs to match whatever the provider generated the pacts with. So I'm imagining we'll need the provider to be able to specify an env var to override this.

ChrisBAshton: Beginning to suspect we _do_ need to sync the port between the provider and the consumer.
See this test run: https://ci.integration.publishing.service.gov.uk/job/gds-api-adapters/job/simplify-pact-verify/6/console

Will need to investigate when I get back. (Have to leave for a couple of hours).

kevindew: Ah yes, the absolute URL's. In my previous comment I focused on just redirects but the same would apply for any absolute URL's in a response. 

It's not that we do need the port specifically but there needs to be agreement on the host (since port forms part of it). 

Happy to have a chat with you about it, I suspect there are some good ways to solve this we've not yet considered.

PR #2719, opened 2021-04-16
Consolidate presentation of addresses

sihugh:
This should make it easier to keep the address presentation consistent, and removes some use of `html_safe` which we should minimise when calling from external sources.


hannako (approved):
A beautiful refactor!

PR #2721, opened 2021-04-16
Add some validation/sanitisation of inputs

sihugh:
I've also inverted the controller flow, as it worked out easier to have the default at the end.

The second commit actually disappears in the end, but it helped me work out where I was going.


hannako (approved):
👍

PR #2730, opened 2021-04-22
Handle remote errors for elections lookup

sihugh:
In this PR we're handling the known 400 and 404 errors as they'll indicate something the user might be able to address.

Other errors will just fall through at present because it either reflects something that we've done wrong (403 would be an incorrect API key) or a problem with the remote service. Either way, the user would be unable to do anything, so we'd need to display an error.

We're considering implementing retries for 5xx errors, but not in this PR.

We add some short-lived caching to help handle spiky loads, and add some metrics around API request duration and logging of error types

### DIFF
--- a/app/views/electoral/results.html.erb
+++ b/app/views/electoral/results.html.erb
@@ -1,47 +1,45 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <% content_for :title,   @content_item[:title] %>
+<div class="govuk-grid-column-two-thirds">
+  <% content_for :title, "#{@content_item[:title]} - GOV.UK" %>
 
-    <div>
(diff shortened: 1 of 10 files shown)


hannako (approved):
Looks great! Are the data attributes fine? Can we just let Ash know?

app/views/electoral/results.html.erb
```diff
@@ -1,47 +1,49 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <% content_for :title,   @content_item[:title] %>
+<% content_for :extra_headers do %>
+  <meta name="robots" content="noindex">
+<% end %>
```
hannako, line 3: Ah well remembered! You had that on the prototype and I didn't copy it over.

PR #2738, opened 2021-04-26
Add lang attributes to bank holidays page

danacotoran:
Currently, the [Welsh version of the bank holidays page](https://www.gov.uk/gwyliau-banc) has `lang='en'` even though most of the main content of the page is in Welsh.

These changes add a `lang` attribute to the main landmark on the page, so that the content is correctly labelled as Welsh when viewing the Welsh version of this page.
Additionally, when viewing the Welsh version, add a `lang=en` attribute to the bank holidays experiment panel as we do not have translations for most of this content. This is to ensure that assistive technologies can read this content in the correct language, which is a WCAG requirement.

**NOTE:** I also hard-coded the bank holidays experiment panel to always use the English language version of the nation names. This is because all the rest of the copy is hard coded in English, and we do not have the means to get it all translated with ease right now. It is more cohesive if all of the text is in one language (_in my opinion_), rather than bits in one language interspersed with bits in another language. 
Ideally it would all be translated to Welsh of course.

## Before: Welsh content is labelled with `lang="en"`
Screenshot 2021-04-26 at 16 29 10

## After: Welsh content is labelled with `lang="cy"`, only English content is labelled with `lang="en"`

Screenshot 2021-04-26 at 16 27 57

### DIFF
--- a/app/views/calendar/_save_location.html.erb
+++ b/app/views/calendar/_save_location.html.erb
@@ -1,6 +1,7 @@
 <%
   data_nation = nation.gsub(/\s/,'_')
 
+  section_lang = "lang=en" unless I18n.locale.to_s == "en"
   save_description = "Do you want to save # as your location for bank holidays?"
   save_button_aria_label = "Save # as your location for bank holidays"
   save_button_text = "Save"
@@ -19,6 +20,7 @@
   data-undo-description="<%= undo_description %>"
   data-undo-button-aria-label="<%= undo_button_aria_label %>"
   data-undo-button-text="<%= undo_button_text %>"
+  <%= section_lang %>
 >
   <h2 class="govuk-heading-m js-nation-description" aria-live="polite">
     <%= sanitize(save_description) %>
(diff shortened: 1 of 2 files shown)


app/views/calendar/_save_location.html.erb
```diff
@@ -19,6 +20,7 @@
   data-undo-description="<%= undo_description %>"
   data-undo-button-aria-label="<%= undo_button_aria_label %>"
   data-undo-button-text="<%= undo_button_text %>"
+  <%= section_lang %>
```
AlanGabbianelli, line 23: I'm not sure I understand the logic of this. `<%= section_lang %>` will be either `nil` when the locale is "en" or `"lang=en"` when the locale is not "en". Why do we want it to be `nil` when the locale is "en"?

danacotoran: Good question @AlanGabbianelli

When a parent of the element already has `lang="en"` there's no added benefit in adding that to the child element – every child of that `lang="en"` container will be assumed to be in English unless it is specified otherwise.

On the other hand, since the element inherits `lang` from higher up the tree, when a parent of the element has `lang="cy"`, we need to add `lang="en"` to the element to ensure that screen readers will correctly read it in English. 

That is why I set that `section_lang` to nil when the locale is en –  in this instance, `section_lang` is only needed when the locale is not en.

AlanGabbianelli (approved):

PR #2748, opened 2021-05-05
Remove the register to vote custom schema once voting closes

sihugh:
Voting closes at 10pm on 6 May. We want to not show the custom FAQ schema after that point, as it'll be out of kilter from the published content. I've added an expiry date to the yaml which might be something we want to reuse in the future.

The publication of the content update will invalidate the cache for this page, so it'll be pretty up to date.


hannako (approved):
LGTM 👍

PR #2770, opened 2021-05-19
Fix JS crossbrowser compatibility issue

danacotoran:
# What
On the bank holidays page save nation script, swap the `replaceAll()` string method for the `replace()` method with a global flag.

# Why
The `replaceAll()` method is [not supported in any version of IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility) which is causing an error. 
As a consequence, on Internet Explorer the save nation panel currently reads "Do you want to save # as your location for bank
holidays" instead of e.g "Do you want to save Scotland as your location for bank holidays".
## Before 
Screenshot 2021-05-19 at 17 52 28

## After 
Screenshot 2021-05-19 at 17 51 24

----- 


andysellick (approved):
Good spot, good work.

PR #2767, opened 2021-05-18
Move hardcoded text into translation files

MuriloDalRi:
Move hardcoded content into the English translation file and adds the same keys to the Welsh one for future translation.

https://trello.com/c/HXNiLKQx/2494-add-keys-which-arent-yet-in-any-locale-files-timebox-1-day-per-app

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #2780, opened 2021-05-26
Mark locale keys containing links as HTML safe

brucebolt:
In #2767, strings containing HTML were moved out of the views and into locale files.

Some have the `raw` method applied, but others do not, so the HTML will be escaped and output to the user.
    
Changing the locale key suffix to contain `_html` will [mark these as HTML Safe](https://api.rubyonrails.org/classes/ActionView/Helpers/TranslationHelper.html#method-i-translate).

[Trello card](https://trello.com/c/HXNiLKQx)

### DIFF
--- a/app/views/calendar/bank_holidays.html.erb
+++ b/app/views/calendar/bank_holidays.html.erb
@@ -43,9 +43,9 @@
             } %>
           <% end %>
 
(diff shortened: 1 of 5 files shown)


1pretz1 (approved):
Looks good, thanks for picking this up! ⭐

thomasleese (approved):

PR #2781, opened 2021-05-27
Add save page routes

huwd:

[Trello](https://trello.com/c/FkH145rh/773-add-endpoints-to-frontend-to-save-and-unsave-a-page)

For an unauthenticated session:
- redirect the user to try and sign in
- on success bring them back to try and save a page

For an authenticated session:
- recieve a get request to `/account/saved-pages/add?page_path=/foo`
- save the page through a PUT request to the account-api
- redirect the user to that page

Then also a remove route:
- recieve a get request to `/account/saved-pages/remove?page_path=/foo`
- save the page through a DELETE request to the account-api
- redirect the user to that page

See also: https://github.com/alphagov/special-route-publisher/pull/165/commits/74e4abe1440e5b01cd99eafadd7560628286ae16

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -100,4 +100,19 @@ private
   def cache
     params[:cache] ? { cache: params[:cache] } : {}
   end
+
+  def set_no_cache_headers
+    response.headers["Cache-Control"] = "no-store"
+  end
+
+  def redirect_with_ga(url, ga_client_id = nil)
+    ga_client_id ||= params[:_ga]
+    if ga_client_id
+      uri = Addressable::URI.parse(url)
+      uri.query_values = (uri.query_values || {}).merge("_ga" => ga_client_id)
+      url = uri.to_s
+    end
+
+    redirect_to url
+  end
 end
--- a/app/controllers/save_pages_controller.rb
+++ b/app/controllers/save_pages_controller.rb
@@ -16,6 +16,19 @@ class SavePagesController < ApplicationController
     redirect_to page_path + "?personalisation=page_not_saved"
   end
 
+  def destroy
+    GdsApi.account_api.delete_saved_page(
+      page_path: page_path,
+      govuk_account_session: account_session_header,
+    )
+
+    redirect_to page_path + "?personalisation=page_removed"
+  rescue GdsApi::HTTPUnauthorized
+    authenticate(remove_saved_page_path(page_path: page_path))
+  rescue GdsApi::HTTPNotFound
+    redirect_to page_path + "?personalisation=page_removed"
+  end
+
 private
 
   def page_path
--- a/app/controllers/save_pages_controller.rb
+++ b/app/controllers/save_pages_controller.rb
@@ -1,0 +1,37 @@
+class SavePagesController < ApplicationController
+  include AccountConcern
+
+  before_action :set_no_cache_headers
+  before_action { head :not_found unless feature_flag_enabled? }
+
+  def create
+    GdsApi.account_api.save_page(
+      page_path: page_path,
+      govuk_account_session: account_session_header,
+    )
+    redirect_to page_path + "?personalisation=page_saved"
+  rescue GdsApi::HTTPUnauthorized
+    authenticate(save_page_path(page_path: page_path))
+  rescue GdsApi::HTTPUnprocessableEntity
+    redirect_to page_path + "?personalisation=page_not_saved"
+  end
+
+private
+
+  def page_path
+    @page_path ||= Addressable::URI.parse(params[:page_path]).path
+  end
+
+  def authenticate(redirect_path)
+    logout!
+
+    redirect_with_ga GdsApi.account_api.get_sign_in_url(
+      redirect_path: redirect_path,
+      level_of_authentication: "level0",
+    ).to_h["auth_uri"]
+  end
+
+  def feature_flag_enabled?
+    ENV["FEATURE_FLAG_SAVE_A_PAGE"] == "enabled"
+  end
+end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -44,10 +44,6 @@ class SessionsController < ApplicationController
 
 protected
 
-  def set_no_cache_headers
-    response.headers["Cache-Control"] = "no-store"
-  end
-
   def account_manager_url
     Plek.find("account-manager")
   end
@@ -58,16 +54,5 @@ protected
     return nil unless http_referrer&.start_with?(Plek.new.website_root)
 
     http_referrer.delete_prefix Plek.new.website_root
-  end
-
-  def redirect_with_ga(url, ga_client_id = nil)
-    ga_client_id ||= params[:_ga]
-    if ga_client_id
-      uri = Addressable::URI.parse(url)
-      uri.query_values = (uri.query_values || {}).merge("_ga" => ga_client_id)
-      url = uri.to_s
-    end
-
-    redirect_to url
   end
 end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -63,12 +63,9 @@ protected
   def redirect_with_ga(url, ga_client_id = nil)
     ga_client_id ||= params[:_ga]
     if ga_client_id
-      url =
-        if url.include? "?"
-          "#{url}&_ga=#{ga_client_id}"
-        else
-          "#{url}?_ga=#{ga_client_id}"
-        end
+      uri = Addressable::URI.parse(url)
+      uri.query_values = (uri.query_values || {}).merge("_ga" => ga_client_id)
+      url = uri.to_s
     end
 
     redirect_to url
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -27,6 +27,10 @@ Rails.application.routes.draw do
   get "/sign-in/callback", to: "sessions#callback", as: :new_govuk_session_callback
   get "/sign-out", to: "sessions#delete", as: :end_govuk_session
 
+  scope "/account" do
+    get "/saved-pages/add", to: "save_pages#create", as: :save_page
+  end
+
   # Help pages
   get "/help", to: "help#index"
   get "/help/ab-testing", to: "help#ab_testing"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -29,6 +29,7 @@ Rails.application.routes.draw do
 
   scope "/account" do
     get "/saved-pages/add", to: "save_pages#create", as: :save_page
+    get "/saved-pages/remove", to: "save_pages#destroy", as: :remove_saved_page
   end
 
   # Help pages
--- a/test/functional/save_pages_controller_test.rb
+++ b/test/functional/save_pages_controller_test.rb
@@ -1,0 +1,108 @@
+require "test_helper"
+require "gds_api/test_helpers/account_api"
+
+class SavePagesControllerTest < ActionController::TestCase
+  include GdsApi::TestHelpers::AccountApi
+  include SessionHelpers
+
+  context "when FEATURE_FLAG_SAVE_A_PAGE environment variable is not 'enabled'" do
+    context "GET /account/saved-pages/add" do
+      should "return 404" do
+        get :create, params: { page_path: ministry_of_magic_path }
+        assert_response :not_found
+      end
+    end
+  end
+
+  context "when FEATURE_FLAG_SAVE_A_PAGE environment variable is set to 'enabled'" do
+    context "GET /account/saved-pages/add" do
+      context "when logged out" do
+        setup do
+          stub_account_api_unauthorized_save_page(page_path: ministry_of_magic_path)
+          @stub = stub_account_api_get_sign_in_url(
+            level_of_authentication: "level0",
+            redirect_path: save_page_path(
+              page_path: ministry_of_magic_path,
+            ),
+          )
+        end
+
+        should "redirect the user to the Auth Provider" do
+          with_feature_flag_enabled do
+            get :create, params: { page_path: ministry_of_magic_path }
+
+            assert_response :redirect
+            assert_requested @stub
+          end
+        end
+      end
+
+      context "when logged in" do
+        setup do
+          mock_logged_in_session
+          stub_account_api_save_page(page_path: ministry_of_magic_path, new_govuk_account_session: "placeholder")
+        end
+
+        should "create a saved page and redirect the user to the saved page_path" do
+          with_feature_flag_enabled do
+            get :create, params: { page_path: ministry_of_magic_path }
+            assert_response :redirect
+
+            path = Addressable::URI.parse(@response.headers["Location"]).path
+            assert_equal path, ministry_of_magic_path
+          end
+        end
+
+        should "report a successful save to the requesting app with a query parameter" do
+          with_feature_flag_enabled do
+            get :create, params: { page_path: ministry_of_magic_path }
+            assert_response :redirect
+
+            query = Addressable::URI.parse(@response.headers["Location"]).query
+            assert_equal query, "personalisation=page_saved"
+          end
+        end
+
+        should "report an unsuccessful save to the requesting app with a query parameter" do
+          with_feature_flag_enabled do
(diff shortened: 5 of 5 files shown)


app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,21 @@
+class SavePagesController < ApplicationController
+  include AccountConcern
+
+  before_action :set_no_cache_headers
+  before_action :authenticate
+
+  def create
```
huwd, line 7: fill this out in the next commit

huwd: will likely squash these too after review?

test/functional/save_pages_controller_test.rb
```diff
@@ -0,0 +1,42 @@
+require "test_helper"
+require "gds_api/test_helpers/account_api"
+require "pry"
+
+class SavePagesControllerTest < ActionController::TestCase
```
huwd, line 4: I tried adding some integration tests too, but they ended up doing pretty much the same thing.
So either i need to broaden what I'm testing there or I think this stuff will do?, it's a fairly simple controller for now

app/controllers/sessions_controller.rb
```diff
@@ -44,33 +44,7 @@ def delete
...
 
-  def set_no_cache_headers
-    response.headers["Cache-Control"] = "no-store"
-  end
-
   def account_manager_url
     Plek.find("account-manager")
   end
-
-  def fetch_http_referrer
```
barrucadu, line 55: We only use this method in this controller, so it can stay here

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  rescue GdsApi::HTTPUnauthorized
+    @account_session_header = nil
+    authenticate
+  end
+
+private
+
+  def authenticate
+    unless logged_in?
+      redirect_to new_govuk_session_path(
```
barrucadu, line 21: This will have to be `redirect_with_ga` to pass along the `_ga` param if present.

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  include AccountConcern
+
+  before_action :set_no_cache_headers
+  before_action :authenticate
+
+  def create
+    redirect_to GdsApi.account_api.save_page(
+      page_path: params[:page_path],
+      govuk_account_session: account_session_header,
+    ).to_h["saved_page"]["page_path"]
```
barrucadu, line 11: I think we'll want to pass some indication of success back to the page, so we can display a success box.

We have a similar need in the Transition Checker (for showing the success box when you save your results), but there we do it with the slightly dodgy approach of checking if the timestamp of the latest results is only a short while ago.  Maybe we should standardise on a query parameter to use for passing personalisation-related flash notices?

So maybe this could be something like:

```ruby
GdsApi.account_api.save_page(page_path: params[:page_path], govuk_account_session: account_session_header)
redirect_to params[:page_path] + "?personalisation=page_saved"
```

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  rescue GdsApi::HTTPUnauthorized
+    @account_session_header = nil
+    authenticate
+  end
```
barrucadu, line 15: We'll need to handle the 422 case here, as currently that'll propagate upwards and the user will get a 422 response.

If we go for the query param approach, maybe we could do

```ruby
redirect_to params[:page_path] + "?personalisation=page_not_saved"
```

And display an error box

config/routes.rb
```diff
@@ -26,6 +26,7 @@
   get "/sign-in", to: "sessions#create", as: :new_govuk_session
   get "/sign-in/callback", to: "sessions#callback", as: :new_govuk_session_callback
   get "/sign-out", to: "sessions#delete", as: :end_govuk_session
+  get "/save-a-page", to: "save_pages#create", as: :save_page
```
barrucadu, line 29: Let's hold off on merging this until the discussion about paths in Slack has an answer.  If `/account/saved-pages` is the index, then we could have `/account/saved-pages/add` and `/account/saved-pages/remove` for manipulation.

app/controllers/application_controller.rb
```diff
@@ -100,4 +100,30 @@ def token
...
+
+  def redirect_with_ga(url, ga_client_id = nil)
+    ga_client_id ||= params[:_ga]
+    if ga_client_id
+      url =
+        if url.include? "?"
+          "#{url}&_ga=#{ga_client_id}"
+        else
+          "#{url}?_ga=#{ga_client_id}"
+        end
```
AlanGabbianelli, line 124: We already have Addressable as a dependency so we could use it here
```suggestion
      uri = Addressable::URI.parse(url)
      uri.query_values = (uri.query_values || {}).merge("_ga" => ga_client_id)
      url = uri.to_s
```

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,28 @@
+class SavePagesController < ApplicationController
```
barrucadu, line 1: Can you feature-flag this controller?  Something like

```ruby
before_action do
  head :not_found unless ENV["FEATURE_FLAG_SAVE_A_PAGE"] == "enabled"
end
```

Then we'll be able to make this available in staging but not production, to try it out for real.

config/routes.rb
barrucadu: So it sounds like `/account/saved-pages` is going to be the index, so how about making the add / delete routes subpaths of that?

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+private
+
+  def authenticate
+    if feature_flag_enabled?
+      unless logged_in?
+        redirect_with_ga new_govuk_session_path(
+          redirect_path: "/save-a-page?page_path=#{params[:page_path]}",
+          state_id: params[:state_id],
```
barrucadu, line 25: That `state_id` is only used to make the Transition Checker kind-of-not-a-JWT-but-also-is-definitely-a-JWT journey work, it's only needed in the `SessionsController`, not here.

huwd: oh great! I wasn't sure so I thought i'd include it.
Will clear that out now.

barrucadu: Also, we could eliminate the redirect via `/sign-in` by using `GdsApi.account_api.get_sign_in_url` directly here:

```ruby
  def authenticate
    return if !feature_flag_enabled? || logged_in?

    redirect_with_ga GdsApi.account_api.get_sign_in_url(redirect_path: "...")
  end
```

Removing one redirect from the chain is only a slight improvement, but feels like a nice one.

app/controllers/application_controller.rb
AlanGabbianelli: We use this in [other apps](https://github.com/alphagov/collections-publisher/blob/main/app/presenters/coronavirus/sub_section_json_presenter.rb#L45-L53) too (thank you @leenagupte for showing me this) so I think extracting it would be good. I'm not sure `govuk_personalisation` is the right place though... 🤔

huwd: Ace, ok if it's place is not in the gem then here sounds good.
I've put it in its own prior commit, that is separate from the move into `ApplicationController`

See here:
https://github.com/alphagov/frontend/pull/2781/commits/ac726d7509520a95378f7020912d502caff7d529

Thanks both @AlanGabbianelli and @leenagupte !

config/routes.rb
huwd: done!

app/controllers/save_pages_controller.rb
huwd: Cool, saves us one more hop - I figured we might need to remember to change this if we ever adjusted /sign-in, but that seems like a manageable thing.

Done!

config/routes.rb
```diff
@@ -27,6 +27,11 @@
   get "/sign-in/callback", to: "sessions#callback", as: :new_govuk_session_callback
   get "/sign-out", to: "sessions#delete", as: :end_govuk_session
 
+  scope "/account" do
+    get "/save-a-page/new", to: "save_pages#create", as: :save_page
+    get "/save-a-page/remove", to: "save_pages#destroy", as: :remove_saved_page
```
barrucadu, line 32: I think it should be `/saved-pages`, then we have

- `/account` - the root
- `/account/saved-pages` - the feature
- `/account/saved-pages/new` - the action

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    GdsApi.account_api.save_page(
+      page_path: ensure_is_path(params[:page_path]),
+      govuk_account_session: account_session_header,
+    )
+    redirect_to ensure_is_path(params[:page_path]) + "?personalisation=page_saved"
+  rescue GdsApi::HTTPUnauthorized
+    @account_session_header = nil
+    authenticate
+  rescue GdsApi::HTTPUnprocessableEntity
+    redirect_to ensure_is_path(params[:page_path]) + "?personalisation=page_not_saved"
```
barrucadu, line 18: Since there are three instances of `ensure_is_path(params[:page_path])`, how about sticking it in a variable?

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    return if !feature_flag_enabled? || logged_in?
+
+    redirect_with_ga GdsApi.account_api.get_sign_in_url(
+      redirect_path: sign_in_redirect_path,
+      level_of_authentication: "level0",
+    ).to_h["auth_uri"]
+  end
+
+  def sign_in_redirect_path
+    if request.env["PATH_INFO"] == save_page_path
```
barrucadu, line 42: What's this conditional for?

test/functional/save_pages_controller_test.rb
```diff
@@ -0,0 +1,108 @@
...
+            ),
+          )
+        end
+
+        should "redirect the user to the Auth Provider" do
+          with_feature_flag_enabled do
+            get :create, params: { page_path: ministry_of_magic_path }
+            assert_response :redirect
+
+            assert_response :redirect
```
barrucadu, line 41: You've got this `assert_response :redirect` twice

test/functional/save_pages_controller_test.rb
```diff
@@ -0,0 +1,108 @@
...
+            assert_requested @stub
+          end
+        end
+      end
+
+      context "when logged in" do
+        setup do
+          mock_logged_in_session
+          stub_account_api_save_page(page_path: "/government/organisations/ministry-of-magic", new_govuk_account_session: "placeholder")
```
barrucadu, line 50: Re-use `ministry_of_magic_path` here

test/functional/save_pages_controller_test.rb
```diff
@@ -0,0 +1,108 @@
...
+          end
+        end
+
+        should "report a successful save to the requesting app with a query parameter" do
+          with_feature_flag_enabled do
+            get :create, params: { page_path: ministry_of_magic_path }
+            assert_response :redirect
+
+            query = Addressable::URI.parse(@response.headers["Location"]).query
+            assert_includes query, "personalisation=page_saved"
```
barrucadu, line 69: Use `assert_equal` here, for consistency with the test for `personalisation=page_not_saved`

app/controllers/save_pages_controller.rb
```diff
@@ -18,6 +18,15 @@ def create
     redirect_to ensure_is_path(params[:page_path]) + "?personalisation=page_not_saved"
   end
 
+  def destroy
```
barrucadu, line 21: Don't forget the unauthorized case here

barrucadu: A user might have a session header (so they pass the `authenticate` `before_action`) but that header is invalid (so there's a 401 response).

barrucadu: In fact, you could just drop the `before_action`.

app/controllers/save_pages_controller.rb
huwd: Do you mean like as a secnd param?

Like `ensure_is_path(params[:page_path], query_params)`

Where query params is a hash or something like:
```
{
   personalisation: "page_saved"
}
```

(I think addressable can do a better query param merge?)

test/functional/save_pages_controller_test.rb
huwd: doh, good spot, amended will push shortly.

app/controllers/save_pages_controller.rb
barrucadu: No, like

```ruby
  def create
    page_path = ensure_is_path(params[:page_path])
    GdsApi.account_api.save_page(
      page_path: page_path,
      ...
```

huwd: Ok I see, sure.

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    authenticate(save_page_path(page_path: page_path))
+  rescue GdsApi::HTTPUnprocessableEntity
+    redirect_to page_path + "?personalisation=page_not_saved"
+  end
+
+  def destroy
+    GdsApi.account_api.delete_saved_page(
+      page_path: page_path,
+      govuk_account_session: account_session_header,
+    )
```
barrucadu, line 23: This can also throw a 404 error, which we should treat as success, because it means the page has already been deleted (or wasn't saved)

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    redirect_to page_path + "?personalisation=page_removed"
+  rescue GdsApi::HTTPUnauthorized
+    authenticate(remove_saved_page_path(page_path: page_path))
+  end
+
+private
+
+  def page_path
+    ensure_is_path(params[:page_path])
+  end
```
barrucadu, line 34: You could merge `ensure_is_path` into this as it's only used by this method:

```ruby
  def page_path
    @page_path ||= Addressable::URI.parse(params[:page_path]).path
  end
```

app/controllers/save_pages_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def authenticate(redirect_path)
+    @account_session_header = nil
```
barrucadu, line 37: Use `logout!` here, just setting the `@account_session_header` to `nil` won't set the end-session response header

app/controllers/save_pages_controller.rb
huwd: we could reuse `personalisation=page_removed`
or do `personalisation=page_already_removed`

And let the frontend handle that?

test/functional/save_pages_controller_test.rb
```diff
@@ -0,0 +1,168 @@
...
+          with_feature_flag_enabled do
+            get :destroy, params: { page_path: ministry_of_magic_path }
+
+            assert_response :redirect
+            assert_requested @stub
+          end
+        end
+      end
+
+      context "when logged in" do
```
barrucadu, line 129: I think it'd be good to add a test for the 404 case here - the page not being saved in the first place is a success here

app/controllers/save_pages_controller.rb
barrucadu: Keep it simple for now, go for `page_removed` until we have a need to treat the two cases differently

test/functional/save_pages_controller_test.rb
```diff
@@ -0,0 +1,180 @@
...
+          with_feature_flag_enabled do
+            get :destroy, params: { page_path: ministry_of_magic_path }
+            assert_response :redirect
+
+            query = Addressable::URI.parse(@response.headers["Location"]).query
+            assert_equal query, "personalisation=page_removed"
+          end
+        end
+
+        should "report back a 404 error as page_already_removed " do
```
barrucadu, line 158: Should be `page_removed`, not `page_already_removed`

barrucadu (approved):

barrucadu:
@huwd When this is deployed, could you make a puppet change to enable the feature flag on **staging**?

PR #2774, opened 2021-05-25
Swap gem layout

chris-gds:
## What?

Update `frontend` app to use `gem_layout` template from `static`.

## Why?

As part of on-going a11y work to align architecture & use Components to ensure consistency and future proof to allow a11y improvements to cascade across the site.

## Visuals

Slight changes, some manual spacing adjustments have been made to mimic current design.

### Selection of visuals:

| Dev URL      |  Before / After - Mobile  |  Desktop |
| ----------- | ----------- | ----------- |
| http://frontend.dev.gov.uk/ | Screenshot 2021-05-31 at 10 52 48       | Screenshot 2021-05-31 at 10 46 06  |
| http://frontend.dev.gov.uk/foreign-travel-advice |  Screenshot 2021-05-31 at 12 40 13  |   Screenshot 2021-05-31 at 12 40 39   |
| http://frontend.dev.gov.uk/contact-electoral-registration-office | Screenshot 2021-05-31 at 13 45 50    | Screenshot 2021-05-31 at 13 45 16       |
| http://frontend.dev.gov.uk/help/cookies |  Screenshot 2021-05-31 at 14 15 38  | Screenshot 2021-05-31 at 14 14 59  |
| http://frontend.dev.gov.uk/settle-in-the-uk |  Screenshot 2021-05-31 at 14 19 05     | Screenshot 2021-05-31 at 14 18 19  |
| http://frontend.dev.gov.uk/settle-in-the-uk/y/you-re-an-eu-eea-or-swiss-citizen |  Screenshot 2021-05-31 at 16 46 31  | Screenshot 2021-05-31 at 16 45 45  | Screenshot 2021-05-31 at 16 09 27 | http://frontend.dev.gov.uk/settle-in-the-uk/y/you-re-an-eu-eea-or-swiss-citizen |  Screenshot 2021-05-31 at 16 47 00 |   |

### Related PRs

Enhancement PR:
[Remove duplicate `<headers>`](https://github.com/alphagov/frontend/pull/2775)

Resolve e2e PR
https://github.com/alphagov/govuk_publishing_components/pull/2101

Adding additional URLs to README for ref.
https://github.com/alphagov/frontend/pull/2793

----
Converting the layout to the gem version resulted in the lookup pattern not inheriting the relevant styling correctly.  Upon inspection, this is due to elements from the gem like [`fieldset`](https://components.publishing.service.gov.uk/component-guide/fieldset) not being utilised within `frontend` and therefore not using the shared classes - this has been manually adjusted.

[...]

PR #2793, opened 2021-06-02
Update README

chris-gds:
## What?

Adding an additional list of URLs rendered by `frontend`

## Why?

As part of [this PR](https://github.com/alphagov/frontend/pull/2774). In order to action more in depth testing to increase confidence that updates work as expected across the repo paths (based on traffic) were taken from [Kibana](https://docs.publishing.service.gov.uk/manual/kibana.html#header) - adding to the repo for future reference. 

## Visuals

The markdown table is hard to visualise so rendered output below:
image

## Anything else?

Some of the links might suffer from link rot and need to be removed / updated as time goes on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,14 +23,12 @@ Transaction start pages:
 * https://www.gov.uk/help/ab-testing
 * https://www.gov.uk/foreign-travel-advice (travel advice index page)
 * https://www.gov.uk/find-local-council
-* https://www.gov.uk/passport-interview-office
 * https://www.gov.uk/roadmap (GOV.UK public facing roadmap)
 
 ### Licence finders
 
 Some examples:
 
-* https://www.gov.uk/guidance/hire-out-horses-licence-england
 * https://www.gov.uk/premises-licence
 * https://www.gov.uk/temporary-events-notice
 * https://www.gov.uk/apply-skip-permit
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,6 +66,31 @@ Assisted digital satisfaction surveys:
 * https://www.gov.uk/bank-holidays
 * https://www.gov.uk/when-do-the-clocks-change
 
+### Misc
+
+* http://www.gov.uk/school-term-holiday-dates
```
injms, line 71: If we know it, is it worth mentioning which view the pages use?

chris-gds: Added ✅  

Partially explored how to add outputs to a Dev environment so comments easily identify which view is in use:
`<!-- <%= File.basename(__FILE__) %> -->`

README.md
```diff
@@ -66,6 +66,40 @@ Assisted digital satisfaction surveys:
 * https://www.gov.uk/bank-holidays
 * https://www.gov.uk/when-do-the-clocks-change
 
+### Misc
+
+| URL                                                                                                                                                          |  | Related Views             |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------------ |  | ------------------------- |
```
injms, line 72: This table isn't rendering as a table, just as the raw markdown; I think this is because of the whitespace between two pipes on line 72.

I think that this should be two columns to make it easier to navigate, unless there was some missing content in the middle column?

```suggestion
| URL | Related Views |
| - | - | 
```

chris-gds: 🙈  I've adjusted this entirely - this [previewer](https://markdownlivepreview.com/) seemingly doesn't use the same renderer.

injms: Argh, how annoying! Looks good now though 👍

injms (approved):
Super! Thanks for adding this 👍

README.md
maxgds: Is this valuable? If you run it locally/dev env you can watch what templates are called in the terminal already. The problem lies in the opposite direction when you need to know what urls a template generates and this doesn't support that.

injms: I think there is still value in this as you don't need to spin up the application to see what views a URL uses. I agree there's still work to be done to find out what URLs use which view - Owen's investigation into using Kibana looks to be promising for this.

chris-gds: >Is this valuable? 

So the use case I found this useful was opening 30+ URLs at once to review and wanting to easily see what page I was looking at.   Is it easier to discover this in terminal in this use case?

> know what urls a template generates and this doesn't support that.

This was just an quick experiment - it would be nice to have this. 
(Wondering how that might work now)

maxgds: ah, I read "on a dev environment" as meaning you need to be running it anyway. This might make sense to scope to just dev/integration and implement - avoid giving millions of users useless bytes to collect. It's a nice idea.

chris-gds:
@injms do the merge honours?

PR #2795, opened 2021-06-02
Add alt to images with text

chris-gds:
## What

Updating the homepage images with alt copy.

## Why

These images aren't just decorative as they contain text that's not visible AT

### DIFF
--- a/app/views/homepage/index.html.erb
+++ b/app/views/homepage/index.html.erb
@@ -89,7 +89,7 @@
         <div class="govuk-grid-column-one-third">
           <a href="/coronavirus" aria-hidden="true" tabindex="-1">
             <%= image_tag "homepage/covid-19-promo-hands-face-space.png", {
-              alt: "",
+              alt: "Hands, Face, Space. Fresh Air",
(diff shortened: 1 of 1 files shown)


app/views/homepage/index.html.erb
```diff
@@ -129,7 +129,7 @@
         <div class="govuk-grid-column-one-third">
           <a href="https://www.nhs.uk/conditions/coronavirus-covid-19/testing/get-tested-for-coronavirus/" aria-hidden="true" tabindex="-1">
```
chris-gds, line 130: `hidden` removes these but they're still visible.

injms:
Previously the text in the image has been added to the body text below the image. Could be worth asking a content designer if they think that this is possible? As a bonus, this would better align with the [advice for describing images].

[advice for describing images]: https://www.gov.uk/guidance/content-design/images#describing-images

chris-gds:
Advised being raised directly with Coronavirus Content to update.

chris-gds:
Closing as has been resolved:

image

PR #2800, opened 2021-06-04
Resolve missing .govuk-grid-row for calendar views

kevindew:
These views were using the .govuk-grid-column-* classes without them
being inside a .govuk-grid-row which meant that the column floats
weren't cleared.

This caused the following page component, the feedback prompt, to share
the same layout as the columns. This problem is barely noticeable when
Javascript is enabled - with just some minor spacing problems at the
footer. However without javascript it caused styling to spill out making
the page look broken.

## Before (without JS)

Screenshot 2021-06-04 at 10 02 42

## After (without JS)

Screenshot 2021-06-04 at 10 09 07

### DIFF
--- a/app/views/calendar/bank_holidays.html.erb
+++ b/app/views/calendar/bank_holidays.html.erb
@@ -4,95 +4,97 @@
 
 <%= render :partial => "calendar_head" %>
 
-<div class="govuk-grid-column-two-thirds">
-  <% if @calendar.show_bunting? %>
-    <div class="app-o-epic-bunting">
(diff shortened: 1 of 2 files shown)


kevindew:
This was fixed by the revert in https://github.com/alphagov/frontend/pull/2801 - which seems to imply that the govuk-grid-row class should be set at a higher level than these views.

I'm going to close this now and let @injms know about the problem for when he consider re-applying the change.

Thanks for having a look @danacotoran

PR #2801, opened 2021-06-04
Revert "Swap gem layout"

injms:
Reverting as this is breaking smokey tests and blocking deployment.


sihugh (approved):

kevindew:
@injms for when you consider re-applying this, there seemed to be a problem applied to frontend pages (bank-holiday as an example) where there was an expectation of a `.govuk-grid-row` class wrapping content. I tried fixing with https://github.com/alphagov/frontend/pull/2800 not realising it was a part of this, but then @danacotoran noticed that this revert fixed the problem.

injms:
@kevindew @danacotoran super, thanks for letting us know about this - we'll add it to the new version of this pull request in #2802 👍

PR #2812, opened 2021-06-09
Fix exposed HTML from translation

brucebolt:
The non-breaking space in a translation is being exposed to users (albeit visually hidden, so many users don't see it).  Wrapping in `raw` as we know the string is safe.

[Trello card](https://trello.com/c/WcUqwDjJ)

### DIFF
--- a/app/views/homepage/index.html.erb
+++ b/app/views/homepage/index.html.erb
@@ -52,7 +52,7 @@
 
     <section class="departments-and-policy" aria-labelledby="departments-and-policy-label">
       <div class="govuk-grid-row home-numbers-and-info">
-        <h2 id="departments-and-policy-label" class="govuk-visually-hidden"><%= t('homepage.index.departments_and_policy') %></h2>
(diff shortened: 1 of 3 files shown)


MuriloDalRi (approved):

app/views/homepage/index.html.erb
```diff
@@ -52,7 +52,7 @@
 
     <section class="departments-and-policy" aria-labelledby="departments-and-policy-label">
       <div class="govuk-grid-row home-numbers-and-info">
-        <h2 id="departments-and-policy-label" class="govuk-visually-hidden"><%= t('homepage.index.departments_and_policy') %></h2>
+        <h2 id="departments-and-policy-label" class="govuk-visually-hidden"><%= raw(t('homepage.index.departments_and_policy')) %></h2>
```
1pretz1, line 55: Instead of calling `raw`, could we use the `*_html` key suffix instead?

brucebolt: Yes, that's a good point.  We should probably standardise our use of this sometime too.

1pretz1 (approved):
LGTM - thanks for addressing my comment! 🍏

PR #2818, opened 2021-06-11
Render results based on devolved administration availability from the content item

Rosa-Fox:

[Trello](https://trello.com/c/ahXDKAYp/402-update-frontend-for-local-transactions-with-devolved-administration-data).

# What?

This task is part of an epic to give publishers control over specifying devolved administration availability of a service. As a result of this work, we will render local transaction results pages based on devolved administration availability data from the content item. 

Currently the functionality to display different results pages based on devolved administration availability exists, but the availability data is read from the `local_transactions_config.yml` file. The existing functionality has been removed as part of this PR.

As well as now reading the devolved administration availability data from the content item, the views and content have been updated. See the [Trello](https://trello.com/c/ahXDKAYp/402-update-frontend-for-local-transactions-with-devolved-administration-data) card to understand what the design/content should be.

# Why?
At the moment, adding service data to the yml file involves developer intervention. Adding this as a feature to the publishing application will make things easier for content designers and developers. Although there are only 2 services specified in the yml file, we have identified that there are many other services that are unavailable to devolved administrations. Publishers will also be able to update these services via the publishing tool, which should make for better results to users.

#### Devolved administration availability with alternative URL (`devolved_administration_service.html.erb`)

devolved-administration-service

#### Service unavailable to devolved administration (`unavailable_service.html.erb`)

unavailable-service

#### DA default behaviour - no alternative url provided (`results.html.erb`)

da-no-alternative-url

## Non devolved administration availability (`results.html.erb`)

england-results


app/controllers/local_transaction_controller.rb
```diff
@@ -34,19 +34,27 @@ def results
     @interaction_details = interaction_details
     @local_authority = LocalAuthorityPresenter.new(@interaction_details["local_authority"])
     @country_name = @local_authority.country_name
+    @devolved_administration_availability = devolved_administration_availability
     @service = LocalTransactionService.new(
-      @publication.title, lgsl, @country_name, @local_authority.url
+      @publication.title, lgsl, @country_name, @local_authority.url, @devolved_administration_availability
```
kevindew, line 39: Your reasoning about not wanting lots of arguments sounds good. I think rather than having a dedicated class for devolved_administration_availability it'd probably make sense to use the [LocalTransactionPresenter](https://github.com/alphagov/frontend/blob/master/app/presenters/local_transaction_presenter.rb) as a place to ask our devolved_administration questions. This is already set in this class and is @publication. We could then adapt the LocalTransactionService to accept the presenter as an argument rather than just the title - however I do question the need for LocalTransactionService in a later comment.

app/models/local_transaction_service.rb
```diff
@@ -1,41 +1,43 @@
...
-    @unavailable_content = LocalTransactionService.config.dig(
-      lgsl, "unavailable_in", @country_name
-    ) || {}
+    @devolved_administration_availability = devolved_administration_availability
   end
 
-  def body
-    @unavailable_content["body"]
+  def unavailable?
+    find_devolved_administration_availability[country_name] == "unavailable"
```
kevindew, line 13: If we had the presenter class available here I wonder if we could pass some of the work of this class that way.

For example this code could be:
```
def unavailable?
  @presenter.unavailable?(country_name)
end
```

and then in the presenter we have something like
```
def unavailabe?(country_name)
  key = country_name.downcase.gsub(" ", "_")
  @content_item.dig("#{key}_availability", "type") == "unavailable"
end
```
or similar - since we probably don't want this class to know too much about the content item details.

app/models/local_transaction_service.rb
```diff
@@ -1,41 +1,43 @@
...
   end
 
-  def button_text
-    @unavailable_content["button_text"] || "Find other services"
+  def devolved_administration?
+    find_devolved_administration_availability[country_name] == "devolved_administration_service"
```
kevindew, line 17: Similar point here that we could moved some of the logic here into the presenter.

app/models/local_transaction_service.rb
kevindew: As I think about this more I wonder whether we're going to need this LocalTransactionService class by moving the devolved_administration stuff into the Presenter. I think we're going to leave this class in a strange place by adapting it as most of it's concerns seem to be based around unavailable handling - with much of it becoming redundant by this work.

I'm imagining we can probably adapt the controller so it's like:

```
    if @publication.unavailable?(@country_name)
      render :unavailable_service
    elsif @publication.devolved_administration?(@country_name)
      render :devolved_administration_service
    else
      render :results
    end
```

And enable us to delete this class

app/models/local_transaction_service.rb
```diff
@@ -1,41 +1,43 @@
...
   end
 
   def button_link
-    @unavailable_content["button_link"] || @local_authority_homepage
+    if devolved_administration?
+      alternative_url[country_name]
+    end
   end
```
kevindew, line 24: I feel like this button_link method doesn't really belong in this class (it seems an odd choice historically) as it's very much a presentation concern. I'd suggest we get rid of it and have a `devolved_administration_alternative_url(country_name)` method on the presenter and call that directly from the devolved_administration view

app/views/local_transaction/unavailable_service.html.erb
```diff
@@ -8,27 +8,24 @@
...
 
-  <% if @publication.more_information.present? %>
-    <section class="more">
-      <div class="more">
-        <%= render "govuk_publishing_components/components/govspeak", {} do %>
-          <%= raw @publication.more_information %>
-        <% end %>
-      </div>
-    </section>
-  <% end %>
```
kevindew, line 33: This seems good to remove - it's seems optimistic that this will always complement the unavailable copy, and it will have just been on the previous page.

config/environments/test.rb
```diff
@@ -35,8 +35,4 @@
 
   # Raises error for missing translations.
   config.action_view.raise_on_missing_translations = true
-
-  config.after_initialize do
-    Frontend.local_transactions_config = "test/fixtures/unavailable_services.yml"
-  end
```
kevindew, line 41: Great stuff 👍

kevindew (commented):
Thanks Rosa - great to see this working.

I've focused most of my comments on the initial modelling where we've got 3 classes in play LocalTransactionPresenter, DevolvedAdministrationAvailability and LocalTransactionService. I'm wondering whether we can utilise the presenter class more and then not need either of the other classes. Hope that makes sense.

On the analytics front I wonder if we're going to want any event logging when the user views the page - rather than interacting with the button (so that we log they've got an unavailable.

Rosa-Fox:
@kevindew thanks for your review, it is ready for re-review. 

I have re-written it to make use of the LocalTransactionPresenter and it a way better approach.

I think it could also be good to get a review from a content writer/designer. I will take some screen grabs on Monday using integration and ask for them to look them over. I am not 100% confident due to the content example (word doc) not including as much text as the web pages we used for design reference. I am not sure if the lines not included in the content don't need changing, or if they shouldn't be included.

test/functional/local_transaction_controller_test.rb
```diff
@@ -137,6 +140,52 @@ class LocalTransactionControllerTest < ActionController::TestCase
...
+                "govuk_slug" => "edinburgh",
+                "country_name" => "Scotland",
+              },
+            ],
+          )
+
+          post :search, params: { slug: "send-a-bear-to-your-local-council", postcode: "EH8 8DX" }
+        end
+
+        should "sanitize postcodes and redirect to the slug for the appropriate authority tier" do
```
kevindew, line 161: I don't think this should is quite right for the scenario - particularly as the postcode isn't malformed. We should rephrase this more focused on the scenario we're testing.

test/integration/local_transactions_test.rb
```diff
@@ -51,6 +51,8 @@ class LocalTransactionsTest < ActionDispatch::IntegrationTest
         lgil_override: 8,
```
kevindew, line 51: Hmm there seems to be quite a high degree of overlap between the purpose of this integration test and the controller tests. It seems unfortunate but I'm not sure there's a quick way out of this problem - I guess this is a sign that there's either too many integration tests or that the controller tests are testing too much.

If we're not testing any scenarios in the controller tests that aren't covered by the integration tests we should just not add them.

test/integration/local_transactions_test.rb
```diff
@@ -365,7 +367,7 @@ class LocalTransactionsTest < ActionDispatch::IntegrationTest
     end
 
     should "show advisory message that we have no url" do
-      assert page.has_content?("We don't have a link for their website.")
+      assert page.has_content?("We don't have a link for their website")
```
kevindew, line 370: Do we need to drop this period? I wasn't sure if it was actually a mistake.

test/integration/local_transactions_test.rb
```diff
@@ -403,91 +405,180 @@ class LocalTransactionsTest < ActionDispatch::IntegrationTest
...
+      assert page.has_no_link?("Go to their website")
+    end
+
+    should "show advisory message that we have no url" do
+      assert page.has_content?("We don't have a link for their website.")
+      assert page.has_link?("local council search", href: "/find-local-council")
+    end
+  end
+
+  context "when a service is available for a devolved administration" do
```
kevindew, line 501:
```suggestion
  context "when a service is handled differently for a devolved administration" do
```

I think we have to be careful we don't make this sound like the normal situation but for a devolved administration postcode.

test/integration/local_transactions_test.rb
```diff
@@ -403,91 +405,180 @@ class LocalTransactionsTest < ActionDispatch::IntegrationTest
...
+        country_name: "Scotland",
+        status: "ok",
+      )
+
+      visit "/pay-bear-tax"
+      fill_in "postcode", with: "EH8 8DX"
+      click_on "Find"
+    end
+
+    should "render devolved_administration_service view" do
```
kevindew, line 530: We should avoid referencing the view name here since it is a user facing concern and instead focus this on the visible behaviour.

test/integration/local_transactions_test.rb
```diff
@@ -403,91 +405,180 @@ class LocalTransactionsTest < ActionDispatch::IntegrationTest
...
+    should "show a button that links to an alternate service provider" do
+      assert_has_button_as_link(
+        "Find information for Scotland",
+        href: "https://scot.gov/service",
+        rel: "external",
+      )
+    end
+  end
+
+  context "when availability data is not provided for a devolved administration" do
```
kevindew, line 543: I don't think we need this test as it doesn't exercise any different code paths. I think we should just make sure one of the earlier integration tests of the feature has a non England country name so we can implicitly cover it.

test/unit/presenters/local_transaction_presenter_test.rb
```diff
@@ -16,4 +24,37 @@ def subject(content_item)
   test "#need_to_know" do
     assert_equal "foo", subject(details: { need_to_know: "foo" }).need_to_know
   end
+
+  test "#scotland_availability" do
+    assert_equal "devolved_administration_service", subject(@content_item_with_devolved_administration_availability, "Scotland").scotland_availability["type"]
+    assert_equal "https://www.gov.scot/stray-dog", subject(@content_item_with_devolved_administration_availability, "Scotland").scotland_availability["alternative_url"]
```
kevindew, line 30: You can probably make this test self-contained like the ones that precede it - which makes it easier to understand.

```suggestion
    devolved_administration = { "type" => "devolved_administration_service", "alternative_url" => "https://gov.scot" }
    assert_equal devolved_administration, subject(details: { scotland_availability: devolved_administration })
```

test/unit/presenters/local_transaction_presenter_test.rb
```diff
@@ -16,4 +24,37 @@ def subject(content_item)
...
+    assert_equal "unavailable", subject(@content_item_with_devolved_administration_availability, "Wales").wales_availability["type"]
+    assert_equal nil, subject(@content_item_with_devolved_administration_availability, "Wales").wales_availability["alternative_url"]
+  end
+
+  test "#northern_ireland_availability" do
+    assert_equal "devolved_administration_service", subject(@content_item_with_devolved_administration_availability, "Northern Ireland").northern_ireland_availability["type"]
+    assert_equal "https://www.nidirect.gov.uk/stray-dog", subject(@content_item_with_devolved_administration_availability, "Northern Ireland").northern_ireland_availability["alternative_url"]
+  end
+
+  test "#unavailable?" do
```
kevindew, line 43: I think this should be broken up into tests for the different scenarios:

```
test "#unvailable? returns true for a devolved administration marked as unavailable"
test "#unvailable? returns false for a devolved administration marked as devolved_administration_service"
test "#unvailable? returns false for a non devolved administration"
```

I think these can also be self contained which should make them a little easier to understand.

test/unit/presenters/local_transaction_presenter_test.rb
```diff
@@ -16,4 +24,37 @@ def subject(content_item)
...
+    assert_equal true, subject(@content_item_with_devolved_administration_availability, "Wales").unavailable?
+    assert_equal false, subject(@content_item_with_devolved_administration_availability, "Scotland").unavailable?
+    assert_equal false, subject(@content_item_with_devolved_administration_availability, "Northern Ireland").unavailable?
+  end
+
+  test "#devolved_administration?" do
```
kevindew, line 49: Similar point about `#unavailable?` for being broken up.

test/unit/presenters/local_transaction_presenter_test.rb
```diff
@@ -16,4 +24,37 @@ def subject(content_item)
...
+    assert_equal true, subject(@content_item_with_devolved_administration_availability, "Scotland").devolved_administration?
+    assert_equal true, subject(@content_item_with_devolved_administration_availability, "Northern Ireland").devolved_administration?
+    assert_equal false, subject(@content_item_with_devolved_administration_availability, "Wales").devolved_administration?
+  end
+
+  test "#devolved_administration_alternative_url" do
```
kevindew, line 55: Similar point about `#unavailable?` for being broken up.

kevindew (commented):
Thanks @Rosa-Fox this looks a lot cleaner without those extra classes. I've added a bunch of comments to refine this.

Should we look at handing this work over to another team member? I'm assuming it's unlikely you'll have a chance to look at this today.

app/presenters/local_transaction_presenter.rb
```diff
@@ -1,13 +1,53 @@
 class LocalTransactionPresenter < ContentItemPresenter
+  attr_reader :scotland_availability, :wales_availability, :northern_ireland_availability, :country_name
```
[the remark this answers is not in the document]
Rosa-Fox: 👍

app/presenters/local_transaction_presenter.rb
```diff
@@ -1,13 +1,53 @@
...
+
+  def initialize(content_item, country_name = "England")
+    super(content_item)
+    @country_name = country_name
+  end
```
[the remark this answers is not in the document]
Rosa-Fox: 👍

app/presenters/local_transaction_presenter.rb
```diff
@@ -1,13 +1,53 @@
...
   PASS_THROUGH_DETAILS_KEYS.each do |key|
     define_method key do
       details[key.to_s] if details
     end
   end
+
+  def unavailable?
+    return false if devolved_administration_availability_is_not_present
+
+    send("#{key}_availability")["type"] == "unavailable"
```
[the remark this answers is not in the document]
Rosa-Fox: 👍

app/presenters/local_transaction_presenter.rb
```diff
@@ -1,13 +1,53 @@
...
+  end
+
+  def devolved_administration?
```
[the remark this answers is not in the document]
Rosa-Fox: 👍

app/presenters/local_transaction_presenter.rb
```diff
@@ -1,13 +1,53 @@
...
+    send("#{key}_availability")["type"] == "devolved_administration_service"
+  end
+
+  def devolved_administration_alternative_url
+    return if devolved_administration_availability_is_not_present
+
+    if devolved_administration?
+      send("#{key}_availability")["alternative_url"]
+    end
+  end
```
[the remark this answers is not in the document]
Rosa-Fox: 👍

app/presenters/local_transaction_presenter.rb
```diff
@@ -1,13 +1,53 @@
...
+
+private
+
+  def key
```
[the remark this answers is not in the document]
Rosa-Fox: 👍

test/unit/presenters/local_transaction_presenter_test.rb
Rosa-Fox: Ooh I did that as I wanted to make use of the content-schema example... but I suppose defining the structure in these tests saves someone having to go and find the schema. Have updated!

test/unit/presenters/local_transaction_presenter_test.rb
Rosa-Fox: 👍

Rosa-Fox: 👍

test/unit/presenters/local_transaction_presenter_test.rb
Rosa-Fox: 👍

app/views/local_transaction/devolved_administration_service.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  edition: @edition,
+} do %>
+  <div class="interaction">
+    <p class="govuk-body">
+      <%= "Get information on this service, or a similar service, on the #{@country_name} Government website."%>
+    </p>
+    <div class="local-authority-result"
+         data-module="auto-track-event"
+         data-track-category="userAlerts:local_transaction"
+         data-track-action="ResultsShownAvailableDa">
```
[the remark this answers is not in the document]
Rosa-Fox: I changed the value to start with a lower case character. I did suggest some values to Harriet but she chose the ones I added so I think we should keep them? Otherwise I can try asking her what she thinks about those options?

app/views/local_transaction/unavailable_service.html.erb
```diff
@@ -8,27 +8,34 @@
       We've matched the postcode to <span class="local-authority"><%= @local_authority.name %></span>.
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks. I added to the locale file along with 'to'.

app/views/local_transaction/unavailable_service.html.erb
Rosa-Fox: I have added this to the unavailable page. Do you think we would need it for the devolved administration page? I thought not at first as I am worried that England related content will display there. However there is additional info in the current YML file. It feels safer to leave it out...

app/views/local_transaction/unavailable_service.html.erb
```diff
@@ -8,27 +8,34 @@
...
-  <% end %>
+    <% else %>
+      <div class="local-authority-result"
+        data-module="auto-track-event"
+        data-track-category="userAlerts:local_transaction"
+        data-track-action="unavailableDaNoUrl">
+        <p class="govuk-body">We don't have a link for their website. Try the <a href="/find-local-council">local council search</a> instead.</p>
+      </div>
+    <% end %>
+  </div>
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks, I moved it to a translation file.

test/functional/local_transaction_controller_test.rb
Rosa-Fox: I changed it to `should "redirect to the local authority slug"`

test/integration/local_transactions_test.rb
Rosa-Fox: Yeah I wasn't 100% sure. I did find it quite useful to have all of these integration tests, but I understand they are quite cumbersome. Is it ok to keep them?

test/integration/local_transactions_test.rb
Rosa-Fox: I am not 100% sure but the [current results page](https://www.gov.uk/report-stray-dog/milton-keynes) doesn't have one, so I took it out.

test/integration/local_transactions_test.rb
Rosa-Fox: 👍

test/integration/local_transactions_test.rb
Rosa-Fox: 👍

test/integration/local_transactions_test.rb
Rosa-Fox: I removed it. I guess having the other tests for devolved administrations / the Westminster postcode for England for the tests at the top of the file cover this.

test/integration/local_transactions_test.rb
kevindew: Yeah they can be kept. There's probably no need for the country based functional tests as they're not really testing that isn't tested.

test/integration/local_transactions_test.rb
```diff
@@ -365,8 +367,8 @@ class LocalTransactionsTest < ActionDispatch::IntegrationTest
     end
 
     should "show advisory message that we have no url" do
-      assert page.has_content?("We don't have a link for their website.")
-      assert page.has_link?("local council search", href: "/find-local-council")
+      content = "We don't have a link for their website. Try the &lt;a href=\"/find-local-council\"&gt;local council search&lt;/a&gt; instead."
```
kevindew, line 370: Oh you're actually asserting that plain HTML is output here rather than actual HTML. It should be asserting against actual HTML - or perhaps better to do an I18n lookup.


(if you get two very similar comments on this it's because I wrote one and submitted it and GitHub no longer seems to have it)

kevindew (commented):
Looking great @Rosa-Fox just a few things I flagged left.

app/views/local_transaction/unavailable_service.html.erb
kevindew: Hmm yeah it's tricky to work out what's best. I don't think it should be on unavailable but not on devolved_administration_service - as I imagine any arguments that it should be on unavailable also apply to devolved_administration_service. 

I'd say either have it on both unavailable and devolved_administration_service or neither, I don't mind too much which both have pro's and con's so go with which one you're most comfortable with.

app/views/local_transaction/devolved_administration_service.html.erb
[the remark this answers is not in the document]
kevindew: Yeah I think we should check these suggestions first before or at least give Harriet an opportunity to refine a bit. I'll drop a quick email on the thread to see how these go.

app/views/electoral/results.html.erb
```diff
@@ -8,7 +8,7 @@
   <%= render "govuk_publishing_components/components/title", title: @content_item["title"] %>
   <% if @presenter.electoral_services.present? %>
     <p class="govuk-body">
-      <%= t('formats.local_transaction.matched_postcode') %> <strong><%= @postcode.sanitized_postcode %></strong> to <strong><%= @presenter.electoral_service_name %></strong>.
+      <%= t('formats.local_transaction.matched_postcode') %> <strong><%= @postcode.sanitized_postcode %><%= @presenter.electoral_service_name %></strong>.
```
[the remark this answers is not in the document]
Rosa-Fox: I think this one might do due to it being electoral service related?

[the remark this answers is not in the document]
Rosa-Fox: I updated it because it used the same translation key but it's sort of unrelated. I could split into another commit potentially?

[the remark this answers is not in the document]
kevindew: Oh yes, sorry I hadn't appreciated that this was a different controller. Yes this should have its own entry in the locale file with the other electoral keys and include the postcode.

test/integration/local_transactions_test.rb
kevindew: I can't see that text on the example you linked (which means I'm pleased Milton Keynes does have a service for reporting stray dogs)

I can see the period in the source code though: https://github.com/alphagov/frontend/blob/a61c3bd90c60e11d50228b6b6983ed8b82746a00/app/views/local_transaction/results.html.erb#L49 unless we're somehow talking about different things?

test/integration/local_transactions_test.rb
Rosa-Fox: I removed the functional country based tests (I left a NI test in there,  but it was already there before I added anything so I decided to keep it there).

app/views/local_transaction/unavailable_service.html.erb
Rosa-Fox: I removed them both for now. If design/content people decide that they want it, then it will be easy enough to add it in again.

config/locales/en.yml
```diff
@@ -160,8 +160,9 @@ en:
       invalid_postcode_sub: Check it and enter it again.
       invalid_uprn: This isn't a valid address.
       invalid_uprn_sub: Enter the postcode again.
-      matched_postcode: We've matched the postcode
+      matched_postcode: We've matched the postcode to
       no_local_authority: We couldn't find a council for this postcode.
+      no_local_authority_url: We don't have a link for their website. Try the <a href="/find-local-council">local council search</a> instead.
```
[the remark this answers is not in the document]
Rosa-Fox: ah yeah its so obvious now I see it 😅 . I have updated the views/tests to use the locales.

app/views/electoral/results.html.erb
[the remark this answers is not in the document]
Rosa-Fox: Thanks. I made a separate key in the `electoral` section.

test/integration/local_transactions_test.rb
Rosa-Fox: 👍

Rosa-Fox:
Thanks @kevindew. It is ready for re-re-review. I updated the screen grabs in the PR description.

config/locales/en.yml
```diff
@@ -106,7 +106,9 @@ en:
       description: Need help? Get in touch with your local electoral registration team. They can tell you if you’re on the electoral register, or if you’ve registered for a postal or proxy vote.
       title: Get help with electoral registration
     service:
-      description: For questions about your poll card, polling place, or about returning your postal voting ballot, contact your council."
+      matched_postcode_html: We've matched the postcode to <strong>%{postcode} %{electoral_service_name}</strong>.
```
kevindew, line 109: Looking at the [existing code](https://github.com/alphagov/frontend/blob/48268f18f18c12ba3c33f77be551d1e59829bc9f/app/views/electoral/results.html.erb#L11) I think this should be:

```suggestion
      matched_postcode_html: We've matched the postcode <strong>%{postcode}</strong> to <strong>%{electoral_service_name}</strong>.
```

app/views/local_transaction/unavailable_service.html.erb
kevindew: Great stuff 👍 I think the last thing to do is to document that decision with the commit message.

test/integration/local_transactions_test.rb
```diff
@@ -400,94 +403,141 @@ class LocalTransactionsTest < ActionDispatch::IntegrationTest
...
-      should "show a button that links to an appropriate alternate service provider" do
-        assert_has_button_as_link(
-          "Custom button text",
-          href: "https://gov.scot",
-          rel: "external",
-          start: true,
-        )
-      end
+    should "render matching the postcode to local authority" do
+      content = "We've matched the postcode to <span class=\"local-authority\">Cardiff</span>."
```
kevindew, line 444: Shall we use the locale key for this?

test/integration/local_transactions_test.rb
```diff
@@ -400,94 +403,141 @@ class LocalTransactionsTest < ActionDispatch::IntegrationTest
...
+    end
+
+    should "render matching the postcode to local authority" do
+      content = "We've matched the postcode to <span class=\"local-authority\">Cardiff</span>."
+      assert page.body.include?(content)
+    end
+
+    should "not link to the authority" do
+      assert page.has_no_link?("Go to their website")
+    end
```
kevindew, line 495: Similar point about I18n to above.

kevindew (commented):
Looks super close now Rosa, just a few more snags. I've still got a chat going with Harriet to resolve the analytics naming. Hopefully will sort that by your return.

config/locales/en.yml
Rosa-Fox: 😱 good spot! I have changed it.

test/integration/local_transactions_test.rb
Rosa-Fox: 👍 done!

test/integration/local_transactions_test.rb
Rosa-Fox: 👍 done!

Rosa-Fox:
Thanks @kevindew it is ready for review again. I updated the commit message to explain why the `more_information` section isn't rendered. I also updated the GA values and added the rationale from your email into the commit.

app/views/local_transaction/unavailable_service.html.erb
Rosa-Fox: 👍

app/views/local_transaction/devolved_administration_service.html.erb
```diff
@@ -0,0 +1,25 @@
...
+} do %>
+  <div class="interaction">
+    <p class="govuk-body">
+        <%= t('formats.local_transaction.info_on_country_website', country_name: @country_name) %>
+    </p>
+    <div class="local-authority-result"
+         data-module="auto-track-event"
+         data-track-category="postcodeSearch:local_transaction"
+         data-track-action="postcodeResultShown:devolvedAdministrationService"
+         data-track-label=<%= @country_name %>>
```
kevindew, line 14:
```suggestion
         data-track-label="<%= @country_name %>">
```

I assume you just missed this, it wouldn't matter if it wasn't for "Northern Ireland"

app/views/local_transaction/results.html.erb
```diff
@@ -8,16 +8,16 @@
        data-module="auto-track-event"
        data-track-category="postcodeSearch:local_transaction"
        data-track-action="postcodeResultShown"
-       data-track-label="<%= @local_authority.name %>">
-      <%= t('formats.local_transaction.matched_postcode') %> <span class="local-authority"><%= @local_authority.name %></span>.
+       data-track-label=<%= @local_authority.name %>>
```
kevindew, line 11: Quotes on this too.

app/views/local_transaction/unavailable_service.html.erb
```diff
@@ -1,34 +1,42 @@
...
-        <% end %>
+    <% if @local_authority.url.present? %>
+      <p class="govuk-body">
+        <%= t('formats.local_transaction.find_other_services') %>
+      </p>
+      <div class="local-authority-result"
+           data-module="auto-track-event"
+           data-track-category="postcodeSearch:local_transaction"
+           data-track-action="postcodeResultShown:devolvedAdministrationUnavailable"
+           data-track-label=<%= @country_name %>>
```
kevindew, line 23: This event should fire regardless of whether there is a local_authority url present or not, so should be set before the conditional. (label needs quotes too)

app/views/local_transaction/unavailable_service.html.erb
```diff
@@ -1,34 +1,42 @@
...
       </div>
-    </section>
-  <% end %>
+    <% else %>
+      <div class="local-authority-result"
+      data-module="auto-track-event"
+      data-track-category="userAlerts:local_transaction"
+      data-track-action="postcodeResultShown:laMatchNoLinkNoAuthorityUrl">
+        <p class="govuk-body"><%= t('formats.local_transaction.no_local_authority_url_html') %></p>
+      </div>
```
kevindew, line 39: Shall we put this in a partial so that it doesn't need to be duplicated between this and results? That seems to offer a slightly better chance that this stays up to date. Could probably call it `_no_local_authority_url.html.erb` to match the translation key.

config/locales/en.yml
kevindew: Thanks - by the way the change hasn't come through, maybe you've not pushed it or one of us has a wire crossed?

kevindew (commented):
Super nearly there now! Just a few small issues to fix.

app/views/local_transaction/devolved_administration_service.html.erb
Rosa-Fox: thanks! I thought the syntax seemed strange with the ""! Have added back in :)

app/views/local_transaction/results.html.erb
Rosa-Fox: 👍

app/views/local_transaction/unavailable_service.html.erb
Rosa-Fox: 👍

app/views/local_transaction/unavailable_service.html.erb
Rosa-Fox: Good idea! Have put in a partial.

Rosa-Fox:
@kevindew thanks! Have made the updates. Sorry about the `matched_postcode_html` I must have messed up rewriting the commit somehow.

kevindew (approved):
All looks good to me 👍  I've given it a spin on integration and it worked well too 🎉

Rosa-Fox:
Thanks @kevindew :)

PR #2824, opened 2021-06-16
Add redirect_path validation to the SessionsController

barrucadu:
This validation matches that in account-api, so it will also reject
these paths with a 422 error.  However, we had an incident with
account-api rejecting paths too strictly, and so have decided to start
treating 422 errors from account-api as important for our SLO.

Since we control all the account-api clients (other GOV.UK apps), this
is pretty reasonable, as other GOV.UK apps shouldn't be sending us
invalid data.

However, there are a few places where user input is passed directly to
the account-api, and we need to account for a user tampering with
these:

- redirect_path (or referer) in the sign in controller: we can
duplicate the validation from account-api.

- code & state params in the auth callback controller: if invalid,
these return a 401, not a 422 - so that's fine.

- page_path in the saved pages controller: this isn't live yet, so
let's leave it for now and monitor (as we don't really want to add a
second content-store request in there if it can be avoided).

---

[Trello card](https://trello.com/c/siLdMmuu/848-review-api-tests-for-missing-coverage)

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -5,12 +5,13 @@ class SessionsController < ApplicationController
 
   def create
     level_of_authentication = params[:level_of_authentication]
-    unless %w[level0 level1].include? level_of_authentication
-      level_of_authentication = nil
-    end
+    level_of_authentication = nil unless is_valid_level_of_authentication? level_of_authentication
+
+    redirect_path = params[:redirect_path] || fetch_http_referrer
+    redirect_path = nil unless is_valid_redirect_path? redirect_path
 
     redirect_with_ga GdsApi.account_api.get_sign_in_url(
-      redirect_path: params[:redirect_path] || fetch_http_referrer,
+      redirect_path: redirect_path,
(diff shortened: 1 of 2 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -54,4 +55,17 @@ def fetch_http_referrer
 
     http_referrer.delete_prefix Plek.new.website_root
   end
+
+  def is_valid_level_of_authentication?(value)
+    %w[level0 level1].include? value
+  end
+
+  def is_valid_redirect_path?(value)
```
huwd, line 63: Do we think this will ever end up anywhere else outsdie of `frontend` and `account-api`?
I'm wondering if a Validations module in the personalisations gem might be a better home?

But if it's only ever really going to be these two then that might be premature

huwd (approved):

app/controllers/sessions_controller.rb
barrucadu: :+1: For now I can't think where else we'd want to use this validation, but if a third use does crop I agree we should put it in the gem.

PR #2839, opened 2021-06-24
Bump pact_broker-client to fix $HOME access issues and add it to Dependabot

kevindew:

This fixes an issue where running `./startup.sh --live` can fail because `pact_broker-client` seems to (rather surprisingly) try to read files from $HOME when a developer is using rbenv - see https://github.com/pact-foundation/pact_broker-client/issues/88 for more information.

As we use pact as a testing framework, I've added the gems for that into the dependabot file to monitor for updates to the gems.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,6 +28,10 @@ updates:
         dependency-type: direct
       - dependency-name: sassc-rails
         dependency-type: direct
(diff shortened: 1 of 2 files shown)


richardTowers (approved):
Thank you for getting to the bottom of this bizarre mystery.

PR #2840, opened 2021-06-25
Remove redirect_path and level_of_authentication from sign in

barrucadu:
Originally we planned for this to be the only place which construct an
OAuth auth URI, and other apps which need to sign the user in would
redirect via https://www.gov.uk/sign-in?redirect_path=...

We've since changed that approach, with other apps calling account-api
to get a sign in URI directly.  So we can remove these parameters, as
they're unused.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -4,16 +4,10 @@ class SessionsController < ApplicationController
   before_action :set_no_cache_headers
 
   def create
-    level_of_authentication = params[:level_of_authentication]
-    level_of_authentication = nil unless is_valid_level_of_authentication? level_of_authentication
-
-    redirect_path = params[:redirect_path] || fetch_http_referrer
+    redirect_path = http_referer_path
     redirect_path = nil unless is_valid_redirect_path? redirect_path
 
-    redirect_with_ga GdsApi.account_api.get_sign_in_url(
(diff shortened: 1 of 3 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -48,16 +42,15 @@ def account_manager_url
...
-  def fetch_http_referrer
-    http_referrer = request.headers["HTTP_REFERER"]
-
-    return nil unless http_referrer&.start_with?(Plek.new.website_root)
-
-    http_referrer.delete_prefix Plek.new.website_root
-  end
+  def http_referer_path
+    @http_referer_path ||=
+      begin
```
huwd, line 47: Interesting there's no rescue here - and you are using safe navigators etc.
So I can't see a case where an error could bubble up?

In this scenario what does the `begin` do for us?

barrucadu: It lets us use a multi-line block of code, `@http_referer_path` ends up with the value of:

```ruby
http_referer = request.headers["HTTP_REFERER"]
if http_referer&.start_with?(Plek.new.website_root)
  http_referer.delete_prefix Plek.new.website_root
end
```

So, `/foo` if it's of the form `https://www.gov.uk/foo` and `nil` otherwise

huwd (approved):

PR #2841, opened 2021-06-25
Add account homepage 

danacotoran:
Move the "Your account" page from the account manager to `gov.uk/account/home`.

This is desirable because the account manager prototype will go away when we migrate to Digital Identity, so we'll need this at some point. It is also potentially necessary for upcoming work on saved pages and/or single page notifications – because we're storing the saved pages in the `account-api`, not the attribute service prototype. 

Visually, there should be no difference between this "new" account home page and the "old" one on `govuk-account-manager-prototype`.

-----

https://trello.com/c/lIDQNATi/813-implement-the-govuk-account-home-page

Co-authored-by @barrucadu and @huwd 

----

### DIFF
--- a/app/assets/stylesheets/views/_accounts.scss
+++ b/app/assets/stylesheets/views/_accounts.scss
@@ -1,0 +1,73 @@
+.account .gem-c-layout-for-public__global-banner-wrapper {
+  // TODO: this is a localised fix for the issue described here https://github.com/alphagov/govuk_publishing_components/issues/2159
+  // this rule should be removed if/when the bug is fixed
+  margin-top: 0;
+}
+
+.accounts-menu {
+  margin: 0 0 govuk-spacing(6) 0;
+  padding: 0;
+
+  @include govuk-media-query($from: desktop) {
+    margin-right: govuk-spacing(4);
+  }
+}
+
+// Used to determine width of blue location indicator in navigation mention
+$_current-indicator-width: 4px;
+
+.accounts-menu__item {
+  list-style-type: none;
+  padding: 12px 0;
+
+  @include govuk-media-query($from: tablet) {
+    margin: 0 0 govuk-spacing(4);
+    padding: govuk-spacing(1) 0;
+  }
+}
+
+.accounts-menu__item--current {
+  margin-left: -(govuk-spacing(2) + $_current-indicator-width);
+  padding-left: govuk-spacing(2);
+  border-left: $_current-indicator-width solid govuk-colour("blue");
+
+  .accounts-menu__link {
+    text-decoration: none;
+  }
+}
+
+.accounts-menu__link {
+  @include govuk-font(19, $weight: "bold");
+
+  &:not(:focus):hover {
+    color: $govuk-link-colour;
+  }
+}
+
+.accounts-panel {
+  padding: govuk-spacing(6);
+  border: solid 1px govuk-colour("mid-grey");
+  border-top: solid 3px $govuk-brand-colour;
+}
+
+.feedback-footer {
+  margin-bottom: govuk-spacing(4);
+  padding: govuk-spacing(3);
+  background: govuk-colour("light-grey");
+
+  @include govuk-media-query($from: tablet) {
+    padding: govuk-spacing(6);
+    margin-top: govuk-spacing(2);
+    margin-bottom: govuk-spacing(8);
+  }
+}
+
+.accounts-your-account__email {
+  margin-bottom: govuk-spacing(7);
+}
+
+.user-reminder {
+  padding: govuk-spacing(3);
+  margin-bottom: govuk-spacing(7);
+  background: govuk-colour("light-grey");
+}
--- a/app/controllers/account_home_controller.rb
+++ b/app/controllers/account_home_controller.rb
@@ -1,0 +1,42 @@
+class AccountHomeController < ApplicationController
+  include GovukPersonalisation::AccountConcern
+  before_action -> { set_slimmer_headers(template: "gem_layout_account", remove_search: true, show_accounts: "signed-in") }
+
+  def show
+    @is_account = true
+    @user = GdsApi.account_api.get_user(govuk_account_session: account_session_header).to_h
+  rescue GdsApi::HTTPUnauthorized
+    logout!
+    redirect_with_ga GdsApi.account_api.get_sign_in_url(redirect_path: account_home_path).to_h["auth_uri"]
+  end
+
+private
+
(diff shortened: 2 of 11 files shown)


app/controllers/account_home_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    @is_account = true
+    @user = GdsApi.account_api.get_user(govuk_account_session: account_session_header).to_h
+  rescue GdsApi::HTTPUnauthorized
+    logout!
+    redirect_with_ga GdsApi.account_api.get_sign_in_url(redirect_path: account_home_path).to_h["auth_uri"]
+  end
+
+  helper_method :show_confirmation_reminder?
+
+  def show_confirmation_reminder?
```
huwd, line 15: can we make move this to private method?
The tests still pass if we do

That way it keeps the controller public method section to RESTful method names.
It is slightly odd having a private method that is then called as a helper method from the view, but I think that makes more sense?

Or alternatively, could we make these take the `@user` in as a parameter and move them out of the controller into a Helper module alltogether?

(I think they will still have access to @user if they are included in as a helper... but I always find helpers as external modules calling instance methods even stranger... who knows where you might include it into!)

app/controllers/account_home_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    return false unless @user
+
+    !@user["email_verified"] || @user["has_unconfirmed_email"]
+  end
+
+  helper_method :confirmation_banner_prompt_type
+
+  def confirmation_banner_prompt_type
```
huwd, line 23: same thought as [here](https://github.com/alphagov/frontend/pull/2841/files#r659941055)

huwd, line 23: same thought as L15 comment

app/controllers/account_home_controller.rb
huwd, line 15: can we make move this to private method?
The tests still pass if we do

That way it keeps the controller public method section to RESTful method names.
It is slightly odd having a private method that is then called as a helper method from the view, but I think that makes more sense?

Or alternatively, could we make these take the `@user` in as a parameter and move them out of the controller into a Helper module alltogether?

(I think they will still have access to @user if they are included in as a helper... but I always find helpers as external modules calling instance methods even stranger... who knows where you might include it into!)

huwd (changes requested):

huwd (approved):
Approving on the grounds of that controller code having been moved around.
I think the frontend needs a separate check.

So lets wait for ✅✅

app/views/account_home/show.html.erb
```diff
@@ -0,0 +1,82 @@
...
+        font_size: "l",
+        margin_bottom: 4,
+      } %>
+      <% if @user["email"] %>
+        <p class="govuk-body accounts-your-account__email">
+          <%= @user["email"] %>
+        </p>
+      <% end %>
+      <% if @user.dig("services", "transition_checker") == "yes" %>
+        <p class="govuk-body">You've used the transition checker, so we should put the non-empty service card here.</p>
```
barrucadu, line 28: This line was just a placeholder, it should be removed

app/views/account_home/show.html.erb
```diff
@@ -0,0 +1,82 @@
...
+            text: t("account.your_account.account_not_used.heading"),
+            heading_level: 2,
+            font_size: "m",
+            margin_bottom: 4,
+          } %>
+
+          <p class="govuk-body"><%= t("account.your_account.account_not_used.description") %></p>
+
+          <p class="govuk-body"><%= sanitize(t("account.your_account.account_not_used.action", link: link_to(t("account.your_account.account_not_used.link_text"), "#{transition_checker_path.to_s}/edit-saved-results", html_options = { class: "govuk-link"} ))) %></p>
+        </div>
```
barrucadu, line 65: The brexit checker has been used in this case, but the user will need to do MFA again to access it (as they signed in at a lower level of authentication).

Since we don't have any designs for that yet, and we don't have the ability to sign in at a lower level switched on in production, this can be combined with the case above, eg

```
<% if has_used? "transition_checker" %>
  "used the checker" case goes here
<% else %>
  "account not used" case goes here
<% end %>
```

app/views/account_home/show.html.erb
danacotoran: Thanks, I definitely removed that at one point, it probably re-added itself when I did a rebase/merge 🤦🏻‍♀️

app/assets/stylesheets/views/_accounts.scss
```diff
@@ -0,0 +1,72 @@
...
+  padding: govuk-spacing(6);
+  border: solid 1px govuk-colour("mid-grey");
+  border-top: solid 3px $govuk-brand-colour;
+}
+
+.feedback-footer {
+  margin-bottom: govuk-spacing(4);
+  padding: govuk-spacing(3);
+  background: govuk-colour("light-grey");
+  @include govuk-media-query($from: tablet) {
```
andysellick, line 57: Nitpick: we normally put a newline before nested things like media queries.

app/views/application/_account-navigation.html.erb
```diff
@@ -0,0 +1,30 @@
+<% page_is ||= nil %>
+
+<nav aria-label="Account management">
+  <ul class="accounts-menu">
+    <li class="accounts-menu__item <%= "accounts-menu__item--current" if page_is == "your-account" %>">
```
andysellick, line 5: Not for this PR, but it should be possible to output these menu items using a loop and some kind of data object. Probably only worth implementing if there'll be more than three of them in future though, I guess.

app/views/layouts/application.html.erb
```diff
@@ -34,8 +34,17 @@
...
         <%= render 'govuk_publishing_components/components/phase_banner', phase: "beta" %>
+      <% elsif @is_account %>
+        <% message = capture do %>
+          <%= t("account.feedback.banners.phase_intro") %>
+          <a class="govuk-link" href=<%= "#{Plek.find('account-manager')}/feedback" %>><%= t("account.feedback.banners.phase_link") %></a>
+          <%= t("account.feedback.banners.phase_outro") %>
+        <% end %>
+        <%= render "govuk_publishing_components/components/phase_banner", {
+        phase: "alpha",
+        message: message
```
andysellick, line 45: Nitpick: lines 44 to 46 need indenting.

andysellick (commented):
Frontend looks okay 👍

barrucadu (approved):

andysellick (approved):

PR #2862, opened 2021-07-06
Set cookies_policy after sign in

barrucadu:
If the user has saved cookie preferences, apply them after logging in.

Won't have any effect until https://github.com/alphagov/govuk-puppet/pull/11189 is deployed, as our varnish config removes cookies currently.


test/unit/presenters/local_transaction_presenter_test.rb
```diff
@@ -28,7 +28,7 @@ def subject(content_item)
     devolved_administration = { "details" => { "wales_availability" => { "type" => "unavailable" } } }
 
     assert_equal "unavailable", subject(devolved_administration).wales_availability["type"]
-    assert_equal nil, subject(devolved_administration).wales_availability["alternative_url"]
+    assert_nil subject(devolved_administration).wales_availability["alternative_url"]
```
huwd, line 31: nice!

huwd (approved):

PR #2864, opened 2021-07-07
Popular links CSS states adjustments

danacotoran:
The "Popular links" on the homepage are over a blue background on mobile, and a black background otherwise. 
The text has a grey colour on the active state, which [does not meet minimum contrast requirements](https://webaim.org/resources/contrastchecker/?fcolor=B1B4B6&bcolor=1D70B8) against the blue background. 
This eliminates the special styling for `:active` on the mobile variant, which ensures these links are more in line with the way we style white links over a blue background in other areas on GOVUK (for example [here](https://www.gov.uk/coronavirus))

Active on blue background before | Active on blue background after
------------ | -------------
Screenshot 2021-07-07 at 13 45 41 | Screenshot 2021-07-07 at 13 45 19

The minimum contrast requirement is also not achieved in the `:active:focus` state on desktop:  the text is grey over a
yellow background which is illegible (this is only an issue in some browsers – I don't see the issue on Safari on my Mac, but I do on the latest versions of Chrome and Firefox). 

### `:active:focus` before
Screenshot 2021-07-07 at 13 42 36

### `:active:focus` after
Screenshot 2021-07-07 at 13 43 03

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -50,9 +50,13 @@
   @include govuk-font($size: 16, $weight: bold);
   color: govuk-colour("white");
 
-  &:active,
-  &:not(:focus):hover {
-    color: $govuk-hover-colour;
+  @include govuk-media-query($from: tablet) {
+    &:active,
+    &:hover {
+      &:not(:focus) {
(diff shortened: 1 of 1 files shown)


injms (approved):
Super nice! Thanks for sorting this out 👍

PR #2874, opened 2021-07-14
Update readme and startup.sh script

edwardkerry:
Updates README in favour of GOVUK Docker and limits startup script to only operate in live mode.

[trello](https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
-<h1>GOV.UK Mainstream, Home & Search Frontend</h1>
+# Frontend
 
-<p>Application to serve some mainstream formats and the homepage for GOV.UK.</p>
+Frontend renders the citizen-facing part of formats stored in the Content Store, and
+some hard-coded routes including the GOV.UK homepage.
 
 ## Live examples
 
@@ -26,8 +27,6 @@ Transaction start pages:
 * https://www.gov.uk/roadmap (GOV.UK public facing roadmap)
 
 ### Licence finders
-
-Some examples:
 
 * https://www.gov.uk/premises-licence
 * https://www.gov.uk/temporary-events-notice
@@ -81,32 +80,7 @@ Assisted digital satisfaction surveys:
 
 ## Technical documentation
 
-Frontend is a Ruby on Rails application that renders the citizen-facing part of formats stored in the Content Store. It looks up the passed-in slug in the Content Store.
-
-It also serves the homepage as a hard-coded route.
-
-Calendar JSON data files are stored in `lib/data/<scope>.json`, with a `divisions` hash for separate data per region (`united-kingdom`, `england-and-wales`, `scotland` or `northern-ireland`).
-
-Each scope's data file contains a list of divisions, containing a list of years, each with a list of events:
-
-```json
-{
-  "title": "UK bank holidays",
-  "description": "UK bank holidays calendar - see UK bank holidays and public holidays for 2012 and 2013",
-  "divisions": {
-    "england-and-wales": {
-      "title": "England and Wales",
-      "2011": [{
-        "title": "New Year's Day",
-        "date": "02/01/2011",
-        "notes": "Substitute day"
-      }]
-    }
-  }
-}
-```
-
-The division `title` attribute is optional.  If this is not present the slug will be humanized and used instead.
+Frontend is a Ruby on Rails application and should follow [our Rails app conventions](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html).
 
 ### Dependencies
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -81,7 +81,7 @@ Assisted digital satisfaction surveys:
 
 ## Technical documentation
 
-Frontend is a Ruby on Rails application that renders the citizen-facing part of formats stored in the Content Store. It looks up the passed-in slug in the Content Store.
+Frontend is a Ruby on Rails application and should follow [our Rails app conventions](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html). It renders the citizen-facing part of formats stored in the Content Store. It looks up the passed-in slug in the Content Store.
```
ChrisBAshton, line 84: I think the "It renders the..." part onwards can be moved to the top of the README. We should also remove the HTML at the top of the README (part of this work is standardising the README structure, and that HTML is not standard!)

README.md
```diff
@@ -81,7 +81,7 @@ Assisted digital satisfaction surveys:
...
 
 It also serves the homepage as a hard-coded route.
```
ChrisBAshton, line 86: This part can also be moved to the intro.

The "Calendar JSON" section below feels like something that would be better moved into its own document in the `docs/` folder, which we can link to from a "Further documentation" subsection that we've been adding to READMEs.

Finally, the `Additional information for calendars` sections onwards could also have the same treatment applied. (Sorry I can't comment on the actual lines - GitHub seems to have removed a feature!)

ChrisBAshton (commented):
Looking good, Ed! I just have a few comments below:

README.md
edwardkerry: 👍

README.md
edwardkerry: 👍

docs/calendars.md
```diff
@@ -0,0 +1,53 @@
+
```
ChrisBAshton, line 1:
```suggestion
```

README.md
```diff
@@ -1,7 +1,7 @@
-<h1>GOV.UK Mainstream, Home & Search Frontend</h1>
-
-<p>Application to serve some mainstream formats and the homepage for GOV.UK.</p>
+# Frontend
 
+Frontend renders the citizen-facing part of formats stored in the Content Store, and
+some hard-coded routes including the GOV.UK homepage.
```
ChrisBAshton, line 4:
```suggestion
some hard-coded routes including the GOV.UK homepage.

```

README.md
```diff
@@ -118,9 +91,11 @@ The division `title` attribute is optional.  If this is not present the slug wil
 
 ### Running the application
 
-If you are [using docker](https://github.com/alphagov/govuk-docker) to run the application (which is advised) it will be available on the host at http://frontend.dev.gov.uk/
+You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) or the local `startup.sh` script to run the app. Read the [guidance on local frontend development](https://docs.publishing.service.gov.uk/manual/local-frontend-development.html) to find out more about each approach, before you get started.
+
+If you are using GOV.UK Docker, remember to combine it with the commands that follow. See the [GOV.UK Docker usage instructions](https://github.com/alphagov/govuk-docker#usage) for examples.
 
-To run the application standalone, run [static](https://github.com/alphagov/static) and execute the following command:
+If you are using the `startup.sh` script, first run [static](https://github.com/alphagov/static) and execute the following command:
```
ChrisBAshton, line 98: This section below needs updating, as Static no longer lives on localhost:3013. It is [now started with govuk-docker](https://github.com/alphagov/static#technical-documentation), so we'd have to set `PLEK_SERVICE_STATIC_URI` to http://static.dev.gov.uk.

startup.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-bundle install
+bundle check || bundle install
```
ChrisBAshton, line 3: Nice! Though is this change necessary? I don't think I've seen this pattern used elsewhere, and from memory, running `bundle install` when I have all the local dependencies already doesn't re-download/re-install the gems, it just exists quite quickly 🤔

ChrisBAshton (approved):
Looking good! A couple of syntax tweaks below, but ready to go after that. I've approved the PR ahead of time. 👍

startup.sh
edwardkerry: I borrowed it from here
https://github.com/alphagov/info-frontend/pull/891/commits/fe012d5e8120dab6145459fea888af2f6ee5325b
I believe that `check` is quicker than `install`, but it is probably a marginal gain for apps without a huge amount of dependencies. Happy for to remain or be removed!

README.md
```diff
@@ -118,49 +92,28 @@ The division `title` attribute is optional.  If this is not present the slug wil
 
 ### Running the application
 
-If you are [using docker](https://github.com/alphagov/govuk-docker) to run the application (which is advised) it will be available on the host at http://frontend.dev.gov.uk/
+You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) or the local `startup.sh` script to run the app. Read the [guidance on local frontend development](https://docs.publishing.service.gov.uk/manual/local-frontend-development.html) to find out more about each approach, before you get started.
 
-To run the application standalone, run [static](https://github.com/alphagov/static) and execute the following command:
+If you are using GOV.UK Docker, remember to combine it with the commands that follow. See the [GOV.UK Docker usage instructions](https://github.com/alphagov/govuk-docker#usage) for examples.
+
+If you are using the `startup.sh` script, first run [static]()https://github.com/alphagov/static) and execute the following command:
```
benthorner, line 99: Just a fly-by comment in case you want to fix: this is incorrect, since `./startup.sh --live` will point at the live instance of Static, so you don't need it running locally (or to set the corresponding environment variable).

PR #2882, opened 2021-07-20
Update local transaction country references

leenagupte:
Trello: https://trello.com/c/KiHIycIv/505-update-references-to-northern-ireland-in-local-lookups

# What's changed and why?

We have standard content for devolved administration results in local lookups, but we need to change the content for how we refer to the Northern Ireland Government is not technically a government so it is factually incorrect to call it that. Executive is right.

This means we will need to make unique content for the name of each nation.

# Expected changes
## Before

|Scotand|Wales|Northern Ireland|
|--------|------|----------------|
|Screenshot 2021-07-20 at 10 34 36|Screenshot 2021-07-20 at 10 34 20|Screenshot 2021-07-20 at 10 33 09|

## After
|Scotand|Wales|Northern Ireland|
|--------|------|----------------|
|Screenshot 2021-07-20 at 10 48 35|Screenshot 2021-07-20 at 10 48 46|Screenshot 2021-07-20 at 10 56 50|

Examples taken from: 

- https://www.gov.uk/test-and-trace-support-payment/glasgow
- https://www.gov.uk/test-and-trace-support-payment/cardiff
- https://www.gov.uk/test-and-trace-support-payment/belfast

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -178,7 +178,10 @@ en:
       no_local_authority_url_html: We don't have a link for their website. Try the <a href="/find-local-council">local council search</a> instead.
       local_authority_website: Go to %{local_authority_name} website
       no_website: We don't have a link for their website.
-      info_on_country_website: Get information on this service, or a similar service, on the %{country_name} Government website.
+      info_on_country_website:
+        northern_ireland: Get information on this service, or a similar service, on the Northern Ireland Executive website.
+        scotland: Get information on this service, or a similar service, on the Scottish Government website.
+        wales: Get information on this service, or a similar service, on the Welsh Government website.
       find_other_services: Find other services on your council website.
       find_info_for: Find information for %{country_name}
       service_not_available: This service is not available in %{country_name}
--- a/app/views/local_transaction/devolved_administration_service.html.erb
+++ b/app/views/local_transaction/devolved_administration_service.html.erb
@@ -1,4 +1,4 @@
-<%= render layout: 'shared/base_page', locals: {
+<%= render layout: "shared/base_page", locals: {
   title: @publication.title,
   publication: @publication,
   edition: @edition,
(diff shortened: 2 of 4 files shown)


config/locales/en.yml
```diff
@@ -178,7 +178,10 @@ en:
       no_local_authority_url_html: We don't have a link for their website. Try the <a href="/find-local-council">local council search</a> instead.
       local_authority_website: Go to %{local_authority_name} website
       no_website: We don't have a link for their website.
-      info_on_country_website: Get information on this service, or a similar service, on the %{country_name} Government website.
+      info_on_country_website:
+        northern_ireland: Get information on this service, or a similar service, on the Northern Ireland Executive website.
+        scotland: Get information on this service, or a similar service, on the Scottish Government website.
+        wales: Get information on this service, or a similar service, on the Welsh Government website.
```
gclssvglx, line 184: The duplication here seems a bit unnecessary, I wonder if it can be removed 🤔 

Maybe something like...

```
government_name:
  northern_ireland: Northern Ireland Executive
  scotland: Scottish Government
  wales: Welsh Government
info_on_country_website: Get information on this service, or a similar service, on the %{government_name} website.
```

Then call it...

``` 
<%= t("formats.local_transaction.info_on_country_website", government_name: t("formats.local_transaction.government_name.#{@country_name.parameterize(separator: "_")}")) %>
```

gclssvglx (approved):
This looks great - thanks @leenagupte.  

No biggie, and don't feel obliged - but I do have a suggestion on how we could remove the duplication.

Nice one 👍

config/locales/en.yml
leenagupte: As there are only 3 countries, I think it's fine to update content 3 times in the locale file. If there were over 10, then yes, we'd have to think of a nicer way to generate the content. I think that accessing two locale entries in the view like that will make it harder for future developers to understand what is going on as well.

PR #2900, opened 2021-08-10
Exclude the Accounts pages from menu A/B test

injms:
## What

Set the Account pages template to always be `gem_layout_account`, regardless of
which A/B variant the user is.

## Why

The explore menu doesn't have the sign in / sign out / account links - these are needed by the people who want to use their GOV.UK account.

---

### DIFF
--- a/app/controllers/account_home_controller.rb
+++ b/app/controllers/account_home_controller.rb
@@ -39,4 +39,8 @@ private
 
     @user["email_verified"] && @user["has_unconfirmed_email"]
   end
+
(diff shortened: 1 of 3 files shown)


barrucadu (approved):
I've tried this locally and it does the right thing, thanks

injms:
I'm going to try and force push or something to get the tests to run - otherwise I don't think that this will deploy.

barrucadu (approved):
Looks good to me

PR #2907, opened 2021-08-13
Use new account layout wrapper for `/account/home`

danacotoran:
A new and improved layout template was added, with the intention of being used on pages that should have the account layout, while not being rendered by the `govuk-account-manager-prototype` app.

The improvement in this new template compared to the old one is that all the page furniture for accounts (including the account nav, account feedback footer, account grid layout) is now contained in the template.

Therefore, we can remove these elements, and the content associated with them, from the `frontend` app, and simply request the `gem_layout_account_manager` template, without any visual differences taking place.

-----
Related (might be relevant for context):
- Creating the new layout template: https://github.com/alphagov/static/pull/2571 https://github.com/alphagov/govuk_publishing_components/pull/2255
- Applying the new layout template to select pages on `email-alert-frontend`: https://github.com/alphagov/email-alert-frontend/pull/1106

----

The next step in the process (after the changes in this PR are live) will be to remove the old `gem_layout_account` template from `static`.


barrucadu (approved):
Tried it locally, looks good to me.

PR #2909, opened 2021-08-18
Add veterans promo to homepage

barrucadu:
This is a placeholder image, we're waiting for the final one.


barrucadu:
We're going ahead with this image, and will swap it later.

himal-mandalia (approved):

PR #2910, opened 2021-08-18
Use new veterans promo image

barrucadu:
This has a darker background and a slightly smaller file size.


himal-mandalia (approved):
Reviewed image/file - all good

PR #2649, opened 2021-03-04
Add 2021 Budget call out to homepage

kevindew:
This also updates the documentation (I'm assuming 608 x 405 is out of date information) and cleans up past images.

Before:

Screenshot 2021-03-04 at 15 48 49

After:

Screenshot 2021-03-04 at 17 07 02

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #2911, opened 2021-08-18
Resize & recompress veterans promo image

barrucadu:
Shrunk to 610x407, and recompressed at 70%.


kevindew:
Thank you, you should also do the other sizes listed on https://github.com/alphagov/frontend/blob/main/app/assets/images/homepage/_README.md and apply the srcset markup of https://github.com/alphagov/frontend/pull/2649/files#diff-b0533c1f365ffe916d27cb02d89deee3e7e1d586f289692af1c69b6036d88449R117-R130 so it loads different images based on the screensize

himal-mandalia (approved):

barrucadu:
I don't really understand the `sizes` attribute from the MDN docs, so I've used the value from that diff.

kevindew (approved):
Awesome thank you - looks a massive file size saving ⭐ 

> I don't really understand the sizes attribute from the MDN docs, so I've used the value from that diff.

Yup, that's cool that shouldn't change between images (it's a bit unfortunate that we had 3 PNG's previously so didn't have this in the source code already).

As I understand it means: 

- when the viewport width is <= 640px use the whole viewport width in calculating the most appropriate image to use; so for a 400px viewport use the 480px wide image.
- when the viewport width is > 640px but <= 1020px consider 33% of the viewport width in choosing an image; so for a 810px viewport look for best image that fits 270px which is the 320px image
- when viewport is wider than > 1020px use the best image that fits 300px wide; which is the 320px image

PR #2935, opened 2021-09-15
Remove jQuery from modules/cookie-settings.js

kevindew:
Trello: https://trello.com/c/hKWh8Mjy/590-remove-jquery-from-frontend-modules-cookie-settingsjs-low-js-experience-required

This is part of an ongoing effort to remove jQuery from the GOV.UK
public tech stack. This file was already written largely without a
jQuery dependency so this was a simple conversion.

### DIFF
--- a/app/assets/javascripts/modules/cookie-settings.js
+++ b/app/assets/javascripts/modules/cookie-settings.js
@@ -2,11 +2,11 @@ window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  function CookieSettings () {}
+  function CookieSettings ($module) {
+    this.$module = $module
+  }
 
-  CookieSettings.prototype.start = function ($module) {
-    this.$module = $module[0]
-
+  CookieSettings.prototype.init = function () {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/cookie-settings.js
```diff
@@ -2,11 +2,11 @@ window.GOVUK = window.GOVUK || {}
...
 (function (Modules) {
-  function CookieSettings () {}
-
-  CookieSettings.prototype.start = function ($module) {
-    this.$module = $module[0]
+  function CookieSettings ($module) {
+    this.$module = $module
+  }
 
+  CookieSettings.prototype.init = function ($module) {
```
leenagupte, line 9: Does `$module` need to be passed in here? I thought `this.$module` that's set in the constructor would be used in this function?

leenagupte (commented):
One question, otherwise looks good!

app/assets/javascripts/modules/cookie-settings.js
andysellick: Shouldn't be here, no.

kevindew: Oops! good catch

andysellick (approved):

PR #2936, opened 2021-09-15
Remove jQuery from views/save-locations.js

kevindew:
Trello: https://trello.com/c/Zn56DzOv/591-remove-jquery-from-frontend-modules-save-locationjs-low-js-experience-required

This is part of an ongoing effort to remove jQuery from the GOV.UK public tech stack. 

The JS file itself was a simple conversion. I then went a bit rogue and removed jQuery from the tests. I then renamed the file to match module conventions. See commits for further details.


spec/javascripts/unit/save-location.spec.js
```diff
@@ -1,184 +1,181 @@
...
+    element.setAttribute('data-save-button-text', 'Save')
+    element.setAttribute('data-undo-description', 'You\'ve saved # as your location for bank holidays')
+    element.setAttribute('data-undo-button-aria-label', 'Remove # as your saved location for bank holidays')
+    element.setAttribute('data-undo-button-text', 'Undo')
+    element.innerHTML =
+      '<p class="js-nation-description" aria-live="polite">Do you want to save # as your location for bank holidays?</p>' +
+      '<button class="js-nation-button" type="submit" aria-label="Save # as your location for bank holidays">Save</button>' +
+      '<p class="js-nation-link">' +
+        '<a href="#" class="govuk-link">We\'ll add a cookie to your device</a>' +
+      '</p>'
```
leenagupte: To be honest, I'm not sure this makes it any easier to read. I think a better formatted html block might have been nicer.

spec/javascripts/unit/save-location.spec.js
```diff
@@ -1,184 +1,181 @@
...
-    beforeEach(function () {
-      $locations = $(html)
-      $('body').append($locations)
-    })
-
     it('shows the bank holiday option if cookies are allowed', function () {
-      new GOVUK.Modules.SaveBankHolidayNation($locations[0]).init()
-      expect($locations[0].style.display).toEqual('block')
+      var bank = new GOVUK.Modules.SaveBankHolidayNation(element)
+      bank.init()
```
leenagupte: Bit of a nitpick, but on the other lines these two lines are combined.

leenagupte (commented):
I've added a couple of inline comments, though they're nothing major.

The changes in the first and third commit look fine to me, but I don't have enough jQuery/JS test knowledge to comment on the second.

I haven't been able to test this locally, and the "Do you want to save..." section doesn't display for me in the development environment, though it works well enough in production. And yes, I have accepted cookies! 
Have you been able to run this successfully locally or in a test environment? Could you add some screenshots?

app/assets/javascripts/application.js
```diff
@@ -12,5 +12,3 @@
 //= require transactions
 //= require support
 //= require_tree ./modules
-
-//= require ./views/save-location
```
andysellick, line 16: Why has this been removed? I've just checked the deployment and it looks like the script no longer runs. Is it included somewhere else somehow or is this a mistake?

andysellick (changes requested):

app/assets/javascripts/application.js
kevindew: It's now moved to modules (see: https://github.com/alphagov/frontend/pull/2936/commits/de878c157cde92137ca4648f90fad8fb640dda21) so automatically required by line 14.

It might not have worked for you because the include for this is inside an AB test. As far as I can tell it's all working correctly.

spec/javascripts/unit/save-location.spec.js
kevindew: Ah yes thanks, Fixed.

spec/javascripts/unit/save-location.spec.js
kevindew: That's fair. I didn't make the change to improve readability the goal was removing jQuery and, well, jQuery is inherently more readable than the DOM APIs 😆 .

I can see how this ended up rather clunky as there is so many data attributes. I've switched the approach of this to another technique that should hopefully resolve your concern.

app/assets/javascripts/application.js
andysellick: Ah of course. Sorry.

andysellick (approved):

spec/javascripts/unit/save-location.spec.js
leenagupte: The `div.innerHTML` is much nicer thanks!

For me the benefit of the html block is that it's easier to see which bits of the html match which element attributes. I

leenagupte:
@kevindew One last thing. I was not able to display the "Do you want to save..." section locally. Have you checked that it still works as expected?

andysellick:
> @kevindew One last thing. I was not able to display the "Do you want to save..." section locally. Have you checked that it still works as expected?

I had the same thing, but I forgot that it's behind an AB Test 🤦

leenagupte:
> > @kevindew One last thing. I was not able to display the "Do you want to save..." section locally. Have you checked that it still works as expected?
> 
> I had the same thing, but I forgot that it's behind an AB Test 🤦

Oh! So how do you test it then?

kevindew:
> > > @kevindew One last thing. I was not able to display the "Do you want to save..." section locally. Have you checked that it still works as expected?
> > 
> > 
> > I had the same thing, but I forgot that it's behind an AB Test 🤦
> 
> Oh! So how do you test it then?

I just edited the conditional to be `<% if true %>` locally to see it as a quick view. I also tried it out on integration with the browser extension.

PR #2939, opened 2021-09-15
Remove jQuery from track-smart-answer.js

leenagupte:
Trello: https://trello.com/c/CBfKxNTs

# What's changed?
Removes jQuery from modules/track-smart-answer.js in Frontend.
Also changes the module to be initialised with an `init` method rather than a `start` method.

# Why?
jQuery is being removed from GOV.UK.

# Testing the changes locally
The changes were tested by visiting `/settle-in-the-uk` locally. 
The "Completed" event still fires when the outcome is reached with the changes:

Screenshot 2021-09-15 at 13 08 44

### DIFF
--- a/app/assets/javascripts/modules/track-smart-answer.js
+++ b/app/assets/javascripts/modules/track-smart-answer.js
@@ -4,33 +4,19 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   'use strict'
 
   function TrackSmartAnswer (element) {
-    this.$module = element
+    this.nodeType = element.getAttribute('data-smart-answer-node-type')
+    this.flowSlug = element.getAttribute('data-smart-answer-slug')
   }
 
   TrackSmartAnswer.prototype.init = function () {
-    var nodeType = this.$module.getAttribute('data-smart-answer-node-type')
-    var flowSlug = this.$module.getAttribute('data-smart-answer-slug')
+    if (!this.flowSlug || !GOVUK.analytics || !GOVUK.analytics.trackEvent) return
 
-    if (!nodeType || !flowSlug) {
-      return
-    }
-
-    var trackingOptions = {
-      label: flowSlug,
-      nonInteraction: true,
-      page: this.currentPath()
-    }
-
-    var trackSmartAnswer = function (category, action) {
-      if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
-        GOVUK.analytics.trackEvent(category, action, trackingOptions)
-      }
-    }
-
-    switch (nodeType) {
-      case 'outcome':
-        trackSmartAnswer('Simple Smart Answer', 'Completed', trackingOptions)
-        break
+    if (this.nodeType === 'outcome') {
+      GOVUK.analytics.trackEvent('Simple Smart Answer', 'Completed', {
+        label: this.flowSlug,
+        nonInteraction: true,
+        page: this.currentPath()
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/track-smart-answer.js
```diff
@@ -3,35 +3,40 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-      var nodeType = element.data('smart-answer-node-type')
-      var flowSlug = element.data('smart-answer-slug')
+  function TrackSmartAnswer (element) {
+    this.$module = element
+  }
 
-      if ((nodeType === undefined) || (flowSlug === undefined)) {
-        return
-      }
+  TrackSmartAnswer.prototype.init = function () {
```
kevindew, line 10: With the change of initialiser we could probably simplify the rest of this module quite significantly by being a bit tactical.

```
TrackSmartAnswer.prototype.init = function () {
  if (!this.flowSlug || !GOVUK.analytics || !GOVUK.analytics.trackEvent) return
  
  if (this.nodeType === 'outcome') {
    GOVUK.analytics.trackEvent('Simple Smart Answer', 'Completed', {
      label: this.flowSlug,
      nonInteraction: true,
      page: window.location.pathname      
    })
  }
}
```

kevindew (commented):
Thanks @leenagupte, this change looks good - you've nicely removed the jQuery.

I've just made some suggestions to make this all more idiomatic as the past implementation seems to have built up a lot of cruft along the way.

app/assets/javascripts/modules/track-smart-answer.js
leenagupte: I think setting `page: window.location.pathname` like this will break the tests as one of them relies on being able to mock the path.

Also, why does the suggested guard clause stop checking if `nodeType` is null? 👇 
`if (!this.flowSlug || !GOVUK.analytics || !GOVUK.analytics.trackEvent) return`

To be honest the brief in the card was to just remove jQuery, I think completely restructuring the code might be a tad out of scope.

kevindew: > I think setting `page: window.location.pathname` like this will break the tests as one of them relies on being able to mock the path.

Ah yes, that makes sense. I did wonder why such a trivial method existed when there was so much other complexity.

> 
> Also, why does the suggested guard clause stop checking if `nodeType` is null? 👇
> `if (!this.flowSlug || !GOVUK.analytics || !GOVUK.analytics.trackEvent) return`

it doesn't need to check if that is null because there's an explicit check that it matches 'outcome' (`if (this.nodeType === 'outcome')` in my code, and the [very unusual 1 outcome switch](https://github.com/alphagov/frontend/blob/546e11b8bd71484bc77cd13f0ab2489ce169f3b1/app/assets/javascripts/modules/track-smart-answer.js#L30-L35) in the existing 

> To be honest the brief in the card was to just remove jQuery, I think completely restructuring the code might be a tad out of scope.

I'd consider basic simplifications in scope, it seems unnecessary to leave the WTFs around for a future person to ponder.

leenagupte:
@kevindew Thanks for your review. I've tied up the code as per your suggestion with the exception of replacing the `currentPath` function.  The tests all still pass (locally) and the "Completed" event still fires on the outcome page.

kevindew (approved):
Super, that looks great 👍

PR #2948, opened 2021-09-23
Apply the big number component to the govuk homepage

owenatgov:
## What
Applies the [big number component](https://components.publishing.service.gov.uk/component-guide/big_number) to the homepage and removes custom code which handled this pattern previously.

## Why
To make use of our new big number component to reduce repeated implementation of this pattern. See https://github.com/alphagov/govuk_publishing_components/issues/2221 for more details.

## Visual changes
| Before | After |
| --- | --- |
| Screenshot 2021-09-23 at 14 46 22 | Screenshot 2021-09-23 at 14 46 29 |

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -4,6 +4,7 @@ $govuk-new-link-styles: true;
 // Components from govuk_publishing_components gem
 @import 'govuk_publishing_components/govuk_frontend_support';
 @import 'govuk_publishing_components/components/back-link';
+@import 'govuk_publishing_components/components/big-number';
 @import 'govuk_publishing_components/components/breadcrumbs';
 @import 'govuk_publishing_components/components/button';
 @import 'govuk_publishing_components/components/character-count';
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -4,6 +4,7 @@ $govuk-new-link-styles: true;
 // Components from govuk_publishing_components gem
 @import 'govuk_publishing_components/govuk_frontend_support';
 @import 'govuk_publishing_components/components/back-link';
+@import 'govuk_publishing_components/components/big-number';
```
andysellick, line 7: The new focus and focus + hover state for this seems unnecessarily busy with the extra underline. Did you consider mimicking the existing appearance for the component?

Before | After
------ | ------
Screenshot 2021-09-24 at 07 54 35 | Screenshot 2021-09-24 at 07 54 27

(this isn't the place for a longer conversation on this, so I'm going to approve regardless because I think using a component for this is a good idea)

andysellick (approved):
Good stuff! Just one small observation, non-blocking.

app/assets/stylesheets/application.scss
owenatgov: Good spot. I'll raise a PR on the gem soon as to refine this a bit more. I'll merge this in for now.

PR #2940, opened 2021-09-15
Add new Promo image for ministerial appointments

huwd:
Changes homepage promo image for a news event

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #2960, opened 2021-10-01
Update homepage promo for HGV drive shortage topical event

brucebolt:
Updates the left homepage promo box to link to the HGV driver shortage topical event.

Also renames the locale file keys for these boxes, to reduce the amount of work needed each time these boxes are updated.  Also normalised the locale files after making these changes.


danacotoran:
Could we optimise the new image? it's quite large for what it is

brucebolt:
> Could we optimise the new image? it's quite large for what it is

Good idea.  I've optimised it down to 25KB.

config/locales/en.yml
```diff
@@ -305,19 +304,26 @@ en:
...
-      uk_bank_holidays: &uk_bank_holidays UK bank holidays
+      travel_abroad:
+      travel_abroad_check:
+      uk_bank_holidays: UK bank holidays
       uk_bank_holidays_dates_html: Check the dates for <a href="/bank-holidays" class="govuk-link" data-track-category="homepageClicked" data-track-action="promoTextLink" data-track-label="/bank-holidays" data-track-value="1" data-track-dimension="%{dimension}" data-track-dimension-index="29">bank holidays</a> in England, Wales, Scotland and Northern Ireland.
       universal_credit: Sign in to your Universal Credit account
-      vaccination: COVID-19 vaccinations
-      vaccination_description: Get information about vaccinations on NHS.UK.
+      vaccination:
+      vaccination_description:
```
danacotoran, line 326: are we still referencing `vaccination` and `vaccination_description` somewhere (same question applies for `travel_abroad` and `travel_abroad_check`)?

brucebolt: Thanks, looks like these got brought across from the Welsh translation file.  Have now removed.

danacotoran (approved):

kevindew:
The image looks smaller in dimensions than we normally have for JPEG's. Typically for JPEG's there are multiple image sizes and srcset used (see: https://github.com/alphagov/frontend/pull/2940/files#diff-b0533c1f365ffe916d27cb02d89deee3e7e1d586f289692af1c69b6036d88449L217-L224) 

There is documentation in https://github.com/alphagov/frontend/blob/0e0efd06ad1f10fbb78a7447f3615520fb36f3e9/app/assets/images/homepage/_README.md but this seems to get missed every time this change occurs so I guess it needs to be moved somewhere else, did you follow any documentation for this or just worked off precedent? I'm assuming the latter so I wonder if a note needs to be in the PR template or in a code comment.

brucebolt:
Thanks for pointing out that documentation @kevindew, I had no idea it existed.  I followed the instructions in the [dev docs](https://docs.publishing.service.gov.uk/manual/update-promo-boxes.html) which appear to not include this information.

I've raised [a PR](https://github.com/alphagov/frontend/pull/2961) to resize the images.

kevindew:
Ah ha I didn't know about that doc. I'll have a look at getting that updated. I've opened a [thread on Slack with the govuk-frontenders](https://gds.slack.com/archives/CARGH33JS/p1633100799046200) to try work out how we can improve the knowledge sharing here

PR #2973, opened 2021-10-14
Add Dockerfile specifically for Kubernetes

sophiebrown01:
This PR adds a new Dockerfile specifically targeted at Kubernetes, which switches from the use of Foreman for running Rails to instead running Rails directly. Foreman isn't necessary when running on Kubernetes, as the controller managers within Kubernetes will take care of ensuring that our app is running without the need of a separate process manager.

The reason behind creating a separate Dockerfile in this PR is because the existing Dockerfile is used by the publishing end-to-end tests; once these tests are deprecated (as is the long-term plan) then `Dockerfile.k8s` can become `Dockerfile`.

Trello: https://trello.com/c/irrbEFMV

### DIFF
--- a/Dockerfile.k8s
+++ b/Dockerfile.k8s
@@ -1,0 +1,23 @@
+ARG base_image=ruby:2.7.3
+FROM ${base_image}
+RUN apt-get update -qq && apt-get upgrade -y
+RUN apt-get install -y build-essential nodejs && apt-get clean
+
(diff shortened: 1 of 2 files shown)


karlbaker02:
Superseded and obsolete (for now) as a result of https://github.com/alphagov/govuk-helm-charts/pull/12.

PR #2990, opened 2021-10-25
Update intervention usage

maxf:
The previous version of the intervention component would render a banner even if no parameter was passed.
This behaviour has now changed as of govuk_publishing_components v28.1: if no parameter is passed then
then no banner is rendered. Hence previous tests failed.

We've now added the `suggestion_parameter` test to the component usage in the tests so that the component 
still gets rendered as needed

### DIFF
--- a/app/views/local_transaction/search.html.erb
+++ b/app/views/local_transaction/search.html.erb
@@ -11,7 +11,7 @@
   edition: @edition,
 } do %>
 
-  <%= render 'govuk_publishing_components/components/intervention' if should_show_sab_intervention? %>
+  <%= render 'govuk_publishing_components/components/intervention', { suggestion_text: "Something" } if should_show_sab_intervention? %>
 
   <% if @publication.introduction.present? %>
     <section class="intro">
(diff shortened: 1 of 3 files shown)


owenatgov (approved):
LGTM

app/views/local_transaction/search.html.erb
```diff
@@ -11,7 +11,7 @@
   edition: @edition,
 } do %>
 
-  <%= render 'govuk_publishing_components/components/intervention' if should_show_sab_intervention? %>
+  <%= render 'govuk_publishing_components/components/intervention', { suggestion_text: "Something" } if should_show_sab_intervention? %>
```
chao-xian, line 14: Is the `Something` shown to users? Where did this content come from?

maxf: Nothing is shown because `should_show_sab_intervention` is never set. This change is a quick fix to get the tests to pass, but the plan is to adapt most of the `sab` code for the next experiment. Then, `Something` will be replaced by something the user will see.

PR #3018, opened 2021-11-09
Update account home page for email subscriptions

alex9smith:
So it shows users' information about their email subscriptions, if they
have any, as well as their Brexit checker results.

Previously, the only service users could use their GOV.UK account for
was the Brexit checker. If they hadn't saved their checker results to
their account, the home page showed a prompt to use the Brexit checker.

Now users will also be able to sign up for email notifications with
their GOV.UK account so it's not correct to show the "You have not used
your account to access any services yet" message if they've not used the
Brexit checker.

We are treating email subscriptions in a slightly different way to other
services that might use the GOV.UK account, which is why I've added a new
helper method to check if a user has subscriptions rather than using the
existing `has_used?`

[Trello](https://trello.com/c/PErJyjJD/1131-update-account-home-page-for-email-subscriptions)

### User with no email subscriptions and no brexit checker results:
frontend dev gov uk_account_home(iPhone X)

### User with email subscriptions and no brexit checker results:
frontend dev gov uk_account_home(iPhone X) (3)

### User with no email subscriptions and brexit checker results
frontend dev gov uk_account_home(iPhone X) (2)

### User with email subscriptions and brexit checker results:
frontend dev gov uk_account_home(iPhone X) (1)

### DIFF
--- a/app/controllers/account_home_controller.rb
+++ b/app/controllers/account_home_controller.rb
@@ -5,6 +5,7 @@ class AccountHomeController < ApplicationController
   def show
     @is_account = true
     @user = GdsApi.account_api.get_user(govuk_account_session: account_session_header).to_h
+    @has_email_subscriptions = has_email_subscriptions?
   rescue GdsApi::HTTPUnauthorized
     logout!
     redirect_with_analytics GdsApi.account_api.get_sign_in_url(redirect_path: account_home_path)["auth_uri"]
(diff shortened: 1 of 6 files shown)


app/controllers/account_home_controller.rb
```diff
@@ -39,4 +40,15 @@ def confirmed_user_changed_email?
...
     @user["email_verified"] && @user["has_unconfirmed_email"]
   end
+
+  def has_email_subscriptions?
+    subscriber = GdsApi.email_alert_api.authenticate_subscriber_by_govuk_account(
+      govuk_account_session: @account_session_header,
+    ).to_h.fetch("subscriber")
+
+    subscriptions = GdsApi.email_alert_api.get_subscriptions(id: subscriber.fetch("id")).to_h.fetch("subscriptions", [])
+    !subscriptions.empty?
```
barrucadu, line 50: Design question: do we want to show the "you have subscriptions"-state service card to *all* users with subscriptions, or only those who have also linked their notifications to their account?  You can check that by seeing if `subscriber["govuk_account_id"]` is present.

alex9smith: Checked with the team and we want to show this panel to anyone with a subscription

barrucadu (commented):
Looks good, I'll deploy this to staging before our QA session later, and hold off the final approve / merge until we're good to go.

barrucadu:
Whoops, we ned to give frontend a bearer token for email-alert-api

barrucadu (approved):

PR #3022, opened 2021-11-12
Update 'Cookies on GOV.UK'

Tetrino:
This implements changes as requested by the Content team for theCookies setting page.

https://trello.com/c/6wk5iHFW/2776-update-cookies-on-govuk-3

### DIFF
--- a/app/views/help/cookie_settings.html.erb
+++ b/app/views/help/cookie_settings.html.erb
@@ -16,6 +16,7 @@
   } do %>
     <p><%= t('cookies.explanation') %></p>
     <p><%= t('cookies.how_we_use') %></p>
+    <p><%= t('cookies.explanation_html') %></p>
   <% end %>
 
   <%= render "govuk_publishing_components/components/heading", {
@@ -44,13 +45,15 @@
 
       <% cookies_usage_hint = capture do %>
         <p class="govuk-body govuk-hint"><%= t('cookies.usage_info') %></p>
-        <p class="govuk-body govuk-hint"><%= t('cookies.google_info') %>:</p>
-        <ul class="govuk-list govuk-list--bullet govuk-hint">
+        <p class="govuk-body govuk-hint"><%= t('cookies.google_info') %></p>
+        <ul class="govuk-list--bullet govuk-hint">
           <li><%= t('cookies.how_you_got') %></li>
           <li><%= t('cookies.pages_visited') %></li>
           <li><%= t('cookies.clicks') %></li>
         </ul>
+        <p class="govuk-body govuk-hint"><%= t('cookies.lux_explain') %></p>
         <p class="govuk-body govuk-hint"><%= t('cookies.lux_info') %></p>
+        <p class="govuk-body govuk-hint"><%= t('cookies.google_share') %></p>
       <% end %>
 
       <%= render "govuk_publishing_components/components/radio", {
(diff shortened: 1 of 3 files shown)


config/locales/cy.yml
```diff
@@ -99,24 +99,28 @@ cy:
     upcoming_bank_holidays: Gwyliau banc i ddod
   continue:
   cookies:
-    additional:
     always_on:
     clicks:
     confirmation_message:
     confirmation_previous_referrer_link:
     confirmation_title:
+    explaination:
```
ChrisBAshton, line 107:
```suggestion
    explanation:
```

The view references to this key will need updating too.

config/locales/en.yml
```diff
@@ -99,42 +99,48 @@ en:
...
-    additional: Most services we link to are run by different government departments, for example DWP’s Universal Credit online, DVLA’s vehicle tax service, or HMRC’s webchat. These services may set additional cookies and, if so, will have their own cookie policy and banner linking to it.
     always_on: They always need to be on.
-    clicks: what you click on while you're visiting the site
+    clicks: what you click on while you're visiting these sites
     confirmation_message: Government services may set additional cookies and, if so, will have their own cookie policy and banner.
     confirmation_previous_referrer_link: Go back to the page you were looking at
     confirmation_title: Your cookie settings were saved
+    explaination:
+      body: This page has a brief explanation of each type of cookie we use. If you want more details,
+      link: read our detailed cookie information
```
ChrisBAshton, line 109: Rather than split these into two keys, I'd suggest having one HTML field instead, like in https://github.com/alphagov/frontend/blob/d72c1dff5eb83bb3d1afaa1149a49fab3644200a/config/locales/en.yml#L8. That makes our approach consistent with what's in Frontend already, and should make it easier to accommodate things like RTL languages in future, as they wouldn't be locked into the markup we've hardcoded into the view.

ChrisBAshton (commented):
Looking good 👍 just one typo, and a suggestion for supporting i18n.

config/locales/en.yml
Tetrino: We seem to have inconsistency on this based on which repo we're in. Other repos have the links separated rather than a large HTML string in the translation, which is much cleaner. I have done as you've stated in this case, but I do not feel they're particularly sensible as a solution.

config/locales/cy.yml
Tetrino: Sorry about this one.

ChrisBAshton (approved):
LGTM 👍 thanks.

app/views/help/cookie_settings.html.erb
```diff
@@ -44,13 +45,15 @@
 
       <% cookies_usage_hint = capture do %>
         <p class="govuk-body govuk-hint"><%= t('cookies.usage_info') %></p>
-        <p class="govuk-body govuk-hint"><%= t('cookies.google_info') %>:</p>
-        <ul class="govuk-list govuk-list--bullet govuk-hint">
+        <p class="govuk-body govuk-hint"><%= t('cookies.google_info') %></p>
+        <ul class="govuk-list-body govuk-list--bullet govuk-hint">
```
injms, line 49: Just wondering - what's the `govuk-list-body` class doing?

Tetrino: @injms It's ensuring the same text style as `govuk-body` while allowing the list styling. Previously the text was much harsher in colour than the rest of the page.

injms: Thanks for the explanation. I asked as there aren't any styles associated with the `govuk-list-body` class - so you'll get the same effect by removing the class:

```suggestion
        <ul class="govuk-list--bullet govuk-hint">
```

Tetrino: @injms Sorry for being late to this - you're correct, and will apply the change shortly.

ChrisBAshton:
Superseded by #3097.

PR #3025, opened 2021-11-18
Add new inverse header to the homepage

injms:
, [Jira issue NAV-2807](https://gov-uk.atlassian.net/browse/NAV-2807)## What

This adds the new design for the inverse header to the homepage.

This should be merged into the `new-homepage` feature branch - so purposefully key features of the homepage are missing. This pull request is only about the blue banner at the top of the page.

Ticket: https://trello.com/c/g16CgpOP

## Why

## Visual changes

| Screen size | Screenshot |
| - | - |
| 320px |  |
| 375px |  |
| 425px |  |
| 768px |  |
| 1024px |  |
| 1440px |  |

---

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -1,63 +1,44 @@
-.home-top {
-  overflow: hidden;
-  background: $govuk-brand-colour;
+.homepage-inverse-header {
+  background-color: $govuk-brand-colour;
+  padding-bottom: govuk-spacing(7);
+  padding-top: govuk-spacing(8);
+
+  @include govuk-media-query($from: desktop) {
+    padding-bottom: govuk-spacing(8);
+    padding-top: govuk-spacing(9);
+  }
+}
+
+.homepage-inverse-header__title {
   color: govuk-colour("white");
-}
-
-.home-top__inner {
-  position: relative;
-}
-
-.home-top__intro {
-  @include govuk-font(19);
-  margin: 0;
-}
-
-.home-top__intro--simpler {
+  font-size: 32px;
+  font-size: govuk-px-to-rem(32);
   font-weight: bold;
-  margin-bottom: govuk-spacing(4);
-}
+  line-height: 1;
+  margin: 0;
+  padding-bottom: govuk-spacing(6);
 
-.home-top__links {
-  padding: 0 0 govuk-spacing(4) 0;
-
-  @include govuk-media-query($from: tablet) {
-    position: relative;
-    z-index: 1;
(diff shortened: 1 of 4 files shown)


mia-allers-gds:
@injms this looks to me!

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -1,65 +1,46 @@
...
+  @include govuk-media-query($from: desktop) {
+    padding-bottom: govuk-spacing(8);
+    padding-top: govuk-spacing(9);
+  }
 }
 
-.home-top__intro--simpler {
+.homepage-inverse-header__title {
+  color: govuk-colour("white");
+  font-size: 32px;
```
chao-xian, line 14: I get that this is a fallback, but is it necessary if it's IE9 and 10 that we're worried about as they do have partial support for rems? [It's only if we used the `font` shorthand or on pseudo elements where it doesn't work from what I gather](https://caniuse.com/?search=rem).

app/views/homepage/index.html.erb
```diff
@@ -1,108 +1,12 @@
 <% content_for :extra_headers do %>
-  <link rel="canonical" href="<%= Frontend.govuk_website_root + root_path %>" />
-  <meta name="description" content="GOV.UK - The place to find government services and information - Simpler, clearer, faster" />
+  <link rel="canonical" href="<%=  Frontend.govuk_website_root + root_path %>">
+  <meta name="description" content="<%= t("homepage.index.meta_description") %>">
```
chao-xian, line 3: 👍

chao-xian (approved):
Just a minor query but non-blocking from me.

app/assets/stylesheets/views/_homepage.scss
injms: I think the fallback should be kept in because:

 - this is what the [guidance in the GDS way advises to do][1]
 - the support is very good, but not complete - about 2.6% of visitors to GOV.UK in the last month used a browser that didn't support `rem` units

As [Frontend doesn't  `govuk_template` any more][2] we don't need to wrap this in a conditional statement either, which is nice.

[1]: https://gds-way.cloudapps.digital/manuals/programming-languages/css.html#include-fallbacks-for-rem-spacing
[2]: https://github.com/alphagov/frontend/blob/fd8de1514f2973115fa5a89c188a17e5380a8011/app/assets/stylesheets/application.scss#L1

chao-xian: Cool thanks! Didn't know it's in GDS Way - I'll double check it.

PR #3027, opened 2021-11-18
Update homepage layout

injms:
## What
Updates the design of the govuk homepage. Specifically makes the following technical changes:

- Moves all individual sections of the homepage into partials
- Moves all homepage content into the `en` yml file, using it as a content source for the templates to render from

## Why
Part of the Explore team's aim to make the homepage easier to use on mobile.

There are a number of decisions we made at this stage in order to deliver at pace and plan to update these post launch. These future updates are recorded [here](https://trello.com/c/RGlBCzuC/186-record-of-the-tech-debt-created-from-homepage-sprints-that-needs-clearing-up-post-deploy).

### Before
www gov uk_

### After
govuk-fronte-new-homepa-uyhk3t herokuapp com_ (1)


chao-xian (approved):
🚀

PR #3028, opened 2021-11-18
Update homepage search and most viewed layout

injms:
, [Jira issue NAV-3048](https://gov-uk.atlassian.net/browse/NAV-3048)## What

This adds the new design for the search and the most viewed links list.

(This is being merged into the `new-homepage` feature branch - so purposefully key features of the homepage are not yet updated.)

Ticket: https://trello.com/c/jmpwnVpI

Relies on changes in https://github.com/alphagov/govuk_publishing_components/pull/2462 being released.

## Why

## Visual changes

| 320px | 375px | 550px | 768px | 1024px | 1440px | 
| - | - | - | - | - | - |
|  |  |  |  |  |  |
---

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -279,3 +279,37 @@
     margin-bottom: govuk-spacing(6);
   }
 }
+
+.homepage-links-and-search {
+  background: govuk-colour("light-grey");
+  padding-bottom: govuk-spacing(8);
+  padding-top: govuk-spacing(6);
+}
+
+.homepage-search {
+  @include govuk-media-query($until: "tablet") {
+    max-width: 420px;
+  }
+}
+
+.homepage-most-viewed-list {
+  list-style: none;
+  margin: 0;
+  padding: 0 0 govuk-spacing(7) 0;
+
+  @include govuk-media-query($from: "tablet") {
+    padding-bottom: 0;
+  }
+
+  & > li {
+    margin-bottom: govuk-spacing(4);
+
+    @include govuk-media-query($from: "tablet") {
+      margin-bottom: govuk-spacing(2);
+    }
+  }
+}
+
+.homepage-most-viewed-list__item {
+  @include govuk-font($size: 19, $weight: "bold");
+}
--- a/app/views/homepage/_links_and_search.html.erb
+++ b/app/views/homepage/_links_and_search.html.erb
@@ -1,0 +1,41 @@
+<section class="homepage-links-and-search">
+  <div class="govuk-width-container">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-one-half govuk-grid-column-two-thirds-from-desktop">
+        <%= render "govuk_publishing_components/components/heading", {
+          font_size: "m",
+          margin_bottom: 4,
+          text: t("homepage.index.most_viewed"),
+        } %>
+        <ul class="homepage-most-viewed-list">
+          <% t("homepage.index.most_viewed_list").each do | item | %>
+            <li>
+              <%= link_to(item[:text], item[:href], {
+                class: "govuk-link govuk-link--no-visited-state homepage-most-viewed-list__item",
+                data: {
+                  track_category: "homepageClicked",
+                  track_action: "popularLink",
+                  track_label: item[:href],
+                  track_value: 1,
+                  track_dimension_index: 29,
+                  track_dimension: item[:text],
+                }
+              }) %>
+            </li>
+          <% end %>
+        </ul>
+      </div>
+      <div class="govuk-grid-column-one-half govuk-grid-column-one-third-from-desktop homepage-search">
+        <%= render "govuk_publishing_components/components/search", {
+          button_text: t("homepage.index.search_button"),
+          inline_label: false,
+          label_margin_bottom: 4,
+          label_size: "m",
+          label_text: t("homepage.index.search_label"),
+          margin_bottom: 0,
+          wrap_label_in_a_heading: true,
+        } %>
+      </div>
+    </div>
+  </div>
+</section>
--- a/app/views/homepage/index.html.erb
+++ b/app/views/homepage/index.html.erb
@@ -7,7 +7,7 @@
 
 <main id="content" role="main">
   <%= render partial: "homepage/inverse_header" %>
-
+  <%= render partial: "homepage/links_and_search" %>
   <%= render partial: "homepage/services_and_information" %>
 
   <div id="homepage" class="govuk-width-container">
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -321,7 +321,7 @@ en:
       - title: Policy papers and consultations
         description: Consultations and strategy
         link: /search/policy-papers-and-consultations?content_store_document_type[]=open_consultations&content_store_document_type[]=closed_consultations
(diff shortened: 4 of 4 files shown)


app/assets/stylesheets/views/_homepage.scss
```diff
@@ -159,3 +159,48 @@
...
   color: govuk-colour("white");
 }
+
+.homepage-links-and-search {
+  background: govuk-colour("light-grey");
+  padding-bottom: govuk-spacing(8);
+  padding-top: govuk-spacing(6);
+
+  @include govuk-media-query($from: "desktop") {
+    padding-bottom: govuk-spacing(8);
```
owenatgov, line 169: Redundant spacing as it's already specified in the mobile styling on line 165.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -159,3 +159,48 @@
...
+  }
+}
+
+.homepage-search {
+  @include govuk-media-query($until: "tablet") {
+    max-width: 420px;
```
owenatgov, line 175: Not a suggestion but a musing that I wonder if this should be specified at the gem level. I can't think of any scenarios that are made better with great big long search bars.

app/views/homepage/_links_and_search.html.erb
```diff
@@ -0,0 +1,25 @@
+<section class="homepage-links-and-search">
+  <div class="govuk-width-container">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-one-half govuk-grid-column-two-thirds-from-desktop homepage-links">
```
owenatgov, line 4: `homepage-links` doesn't look like it's doing anything here so it can be removed.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -159,3 +159,48 @@
...
+  list-style: none;
+  margin: 0;
+  padding: 0 0 govuk-spacing(7) 0;
+
+  @include govuk-media-query($from: "tablet") {
+    padding-bottom: 0;
+  }
+
+  & > li {
+    padding-bottom: govuk-spacing(4);
```
owenatgov, line 189: This doesn't feel quite right using padding to dictate spacing between the list items. Could it be margin instead?

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -159,3 +159,48 @@
...
+  }
+}
+
+.homepage-search {
+  @include govuk-media-query($until: "tablet") {
+    max-width: 420px;
+  }
+}
+
+.homepage-most-viewed-list {
```
owenatgov, line 179: Whilst it doesn't add a huge amount here, using `govuk-list` alongside this class to dictate spacing feels apt as the element is a list in this case, not something we want to just use list markup for. By my calculation it would mean we're writing less CSS here and using a class that get's loaded anyway.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -159,3 +159,48 @@
...
+  }
+}
+
+.homepage-most-viewed-list__item {
+  font-size: 16px;
+  font-size: govuk-px-to-em(16);
+  font-weight: bold;
+
+  @include govuk-media-query($from: desktop) {
+    font-size: 19px;
```
owenatgov, line 203: We've been avoiding using `govuk-font` in a few places recently, but this feels like an appropriate use case for `govuk-font`.

owenatgov (commented):
Couple a' comments. Looking good.

app/assets/stylesheets/views/_homepage.scss
injms: Using `govuk-list` and then overriding the spacing was the reason for the [recent layout bug in the navigation header][1] - each page having two  stylesheets means that the order of the selectors can't always be controlled.

With that in mind, I'd rather avoid using `govuk-list` when anything needs to be overridden.

[1]: https://github.com/alphagov/govuk_publishing_components/pull/2460

app/views/homepage/_links_and_search.html.erb
injms: Good spot - removed in https://github.com/alphagov/frontend/pull/3028/commits/34273151fa0bc5d55d00653845d73cb5aa836937.

app/assets/stylesheets/views/_homepage.scss
injms: Sure - swapped in https://github.com/alphagov/frontend/pull/3028/commits/05f881c70c9ff75a0a12080d09ffdd1ee4aa8ca2.

app/assets/stylesheets/views/_homepage.scss
injms: Good spot - it's a nice mixin that fits this use perfectly. Updated in https://github.com/alphagov/frontend/pull/3028/commits/b4b96872f55821b03d9cd910b429c43fda4fa11a.

app/assets/stylesheets/views/_homepage.scss
injms: So the Frontend has the ability to [set a character width on the text input][1] (seems like a nice use of `max-width`)  which appears to be available on the [gem's input component][2].

Whilst it shouldn't be too difficult to get the search component to use this, I'd suggest that this improvement should be a separate thing done in the component gem - otherwise it'd block this work. Happy to raise this as an issue.

[1]: https://design-system.service.gov.uk/components/text-input/#fixed-width-inputs
[2]: https://components.publishing.service.gov.uk/component-guide/input/with_custom_width

app/assets/stylesheets/views/_homepage.scss
injms: Removed in https://github.com/alphagov/frontend/pull/3028/commits/53a013dac4444d551a5682b4e30278f437f79f44.

app/assets/stylesheets/views/_homepage.scss
owenatgov: Completely agree, a tomorrow problem. I should've been clearer in my verbosity that this is something to think about later. Resolving for now but will make a note to jot up an issue and reference this comment.

app/assets/stylesheets/views/_homepage.scss
owenatgov: Gotcha, I'll back down on this one. Resolving.

owenatgov (approved):
Looking _tidy_. I'm on the fence about the commit history on this and how much it needs a tidy up. This is me trying to prompt you for thoughts Ian without being annoying 😁

injms:
@owenatgov agreed this needs a squash - I think a nice thing about working on branches is that we can add this tiny small commits for clarity, and then squash once the conversation has happened. The commits should remain for posterity, albeit detached from this branch.

mia-allers-gds:
@injms I've spotted an amendment I'd like to make to the spacing after all H2s on the page (mobile). Could we make the space after the H2s 30px instead of 20px on mobile? It's looking a bit squashed at the moment.

injms:
> @injms I've spotted an amendment I'd like to make to the spacing after all H2s on the page (mobile). Could we make the space after the H2s 30px instead of 20px on mobile? It's looking a bit squashed at the moment.

@mia-allers-gds as discussed, I'll merge this in so these changes can be seen as part of the whole page.

PR #3033, opened 2021-11-23
Add services and information / topics to homepage

injms:
, [Jira issue NAV-3280](https://gov-uk.atlassian.net/browse/NAV-3280)## What

This adds the redesigned Topics / Services and information section to the new homepage design.

This is being merged into the `new-homepage` feature branch - so this pull request only focuses on the slice being added.

Ticket: https://trello.com/c/IIgmi2TA

## Why

## Visual changes

| 320px | 375px | 550px | 768px | 1024px | 1440px | 
| - | - | - | - | - | - |
|  |  |  |  |  |  |

---

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -1,3 +1,6 @@
+@import "govuk_publishing_components/components/mixins/prefixed-transform";
+@import "../mixins/grid-helper";
+
 .homepage-inverse-header {
   background-color: $govuk-brand-colour;
   padding-bottom: govuk-spacing(7);
@@ -158,4 +161,84 @@
 .home-more-promo__link-content {
   @include govuk-font(36, $weight: bold);
   color: govuk-colour("white");
+}
+
+.homepage-services-and-info {
+  border-bottom: $govuk-focus-width solid govuk-colour("black");
+  padding-bottom: govuk-spacing(7);
+  padding-top: govuk-spacing(6);
+
+  @include govuk-media-query($from: desktop) {
+    padding-bottom: govuk-spacing(6);
+  }
+}
+
+.homepage-services-and-info__list {
+  @include columns($items: 16, $columns: 1, $selector: "li");
+  list-style: none;
+  margin: 0 govuk-spacing(-3);
+  padding: 0;
+
+  @include govuk-media-query($from: "tablet") {
+    @include columns($items: 16, $columns: 2, $selector: "li");
+  }
+
+  @include govuk-media-query($from: "desktop") {
+    @include columns($items: 16, $columns: 3, $selector: "li");
+  }
+}
+
+.chevron-card {
+  border-top: 1px solid $govuk-border-colour;
+  margin: 0 govuk-spacing(3);
+  padding: govuk-spacing(2) 0 govuk-spacing(4) 0;
+}
+
+.chevron-card__wrapper {
+  padding: govuk-spacing(2) govuk-spacing(6) govuk-spacing(2) 0;
+  position: relative;
+}
+
+.chevron-card__description {
+  margin: 0;
+}
+
+.chevron-card__link {
+  &:after {
+    bottom: 0;
+    content: "";
+    display: block;
+    left: 0;
+    position: absolute;
+    right: 0;
+    top: 0;
+  }
+
+  &:before {
+    $dimension: 7px;
+    $width: 3px;
+
+    border-right: $width solid $govuk-brand-colour;
+    border-top: $width solid $govuk-brand-colour;
+    content: "";
+    display: block;
+    height: $dimension;
+    position: absolute;
+    right: govuk-spacing(1);
+    top: govuk-spacing(3);
+    @include prefixed-transform($rotate: 45deg);
+    width: $dimension;
+  }
+
+  &:hover {
+    &:before {
+      border-color: $govuk-link-hover-colour;
+    }
+  }
+
+  &:focus {
+    &:before {
+      border-color: $govuk-focus-text-colour;
+    }
+  }
 }
--- a/app/views/homepage/_chevron_card.html.erb
+++ b/app/views/homepage/_chevron_card.html.erb
@@ -1,0 +1,31 @@
+<%
+  element ||= :li
+
+  tracking_attributes = {
+    track_action: track_action,
+    track_category: "homepageClicked",
+    track_dimension_index: 29,
+    track_dimension: raw(title),
+    track_label: link,
+    track_value: 1,
+  }
+
+  link = capture do
+    link_to(raw(title), link, {
+      class: "govuk-link govuk-link--no-visited-state chevron-card__link",
+      data: tracking_attributes,
+    })
+  end
+%>
+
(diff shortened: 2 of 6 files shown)


app/assets/stylesheets/views/_homepage.scss
```diff
@@ -159,3 +159,87 @@
...
+    grid-template-rows: 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr;
+  }
+
+  @include govuk-media-query($from: "desktop") {
+    grid-template-columns: 1fr 1fr 1fr;
+    grid-template-rows: 1fr 1fr 1fr 1fr 1fr 1fr;
+  }
+}
+
+.chevron-card {
```
owenatgov, line 192: Should `chevron-card` have its own sass file?

app/views/homepage/_chevron_card.html.erb
```diff
@@ -0,0 +1,31 @@
+<%
+  element ||= :li
```
owenatgov, line 2: Nifty pre-empt there

mia-allers-gds:
Looks good to me!

app/views/homepage/_services_and_information.html.erb
```diff
@@ -0,0 +1,18 @@
+<section class="homepage-services-and-info govuk-width-container" aria-labelledby="services-and-information-label" id="services-and-information">
```
owenatgov, line 1: What's the value of having that `services-and-information` label here and then a separate `id/labelledby` relationship with another id with a `-label` suffix? What's more, do we need to have this `aria-labelledby` relationship if the heading is the very next element? From a very quick test with VO, it just skips over this `section` element so this might be redundant code.

app/views/homepage/_chevron_card.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+  link = capture do
+    link_to(raw(item[:title]), item[:href], {
+      class: "govuk-link govuk-link--no-visited-state chevron-card__link",
+      data: tracking_attributes,
+    })
+  end
+%>
+
+<%= content_tag(element, class: "chevron-card") do %>
```
owenatgov, line 21: Should this be `app-c-chevron-card`?

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -159,3 +159,87 @@
...
+
+  @include govuk-media-query($from: desktop) {
+    padding-bottom: govuk-spacing(6);
+  }
+}
+
+.homepage-services-and-info__list {
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-template-rows: 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr;
```
owenatgov, line 176: Could we add a comment explaining why we're doing this hard-coded row thing, maybe even a comment like we have in the super nav translation yml in the equivalent in frontend encouraging folks to update the settings here on the rare chance a new topic needs ot be added to the homepage?

owenatgov (commented):
A few comments, mostly convention/style choices. Looking sharp, again ⚔️

owenatgov:
One other thing I forgot to comment on was if we reckon the chevron card needs tests. We'll need them anyway when we transfer them to the gem but it's a question on if we do this now or later.

app/assets/stylesheets/views/_homepage.scss
injms: Maybe when it's used by pages that aren't the homepage?

app/views/homepage/_chevron_card.html.erb
injms: I'd say not yet as it's not a proper component - I've tried to make it as component-like as possible, but at the moment I don't think it needs the prefix.

app/assets/stylesheets/views/_homepage.scss
injms: A note is a good idea. I need to replace this with a mixin for backwards compatibility, so that should give opportunity to document this better.

It's a bit ugly, but it does the job. Unfortunately `repeat()` hasn't got quite the support needed yet.

app/views/homepage/_services_and_information.html.erb
injms: I think that this is probably a hangover from when this was visually hidden - so, as you correctly point out, it's redundant. Removed in https://github.com/alphagov/frontend/pull/3033/commits/f1a0ce477b0a7b49f8ac5a318f0eee8a3f7e6f2c.

app/assets/stylesheets/views/_homepage.scss
owenatgov: Seems suitable. We can clear it up once we move this to the gem. Happy to call this resolved.

app/views/homepage/_chevron_card.html.erb
owenatgov: I can live with that. Resolving.

app/assets/stylesheets/views/_homepage.scss
injms: I've updated the `columns()` mixin from the gem in https://github.com/alphagov/frontend/pull/3033/commits/6e86a62ccca9f658b2b7067e7f9006fddbf5c0a1 - in the interests of speed I've added the mixin here for review, but I'll add the updated version to the gem in the future and then use that version here.

config/locales/en.yml
```diff
@@ -267,38 +267,54 @@ en:
...
-      housing_local_services_description: Owning or renting and council services
-      justice: Crime, justice and the law
-      justice_description: Legal processes, courts and the police
-      tax: Money and tax
-      tax_description: Includes debt and Self Assessment
-      visas_immigration: Visas and immigration
-      visas_immigration_description: Apply to visit, work, study, settle or seek asylum in the UK
-      working: Working, jobs and pensions
-      working_description: Includes holidays, finding a job and redundancy
+      - title: Benefits
```
owenatgov, line 270: Could we add a comment here reminding folks to update columns like we have for the super nav?

owenatgov (commented):
One final comment and then this can go in.

config/locales/en.yml
injms: Good idea - done in https://github.com/alphagov/frontend/pull/3033/commits/3de50eb9bf9dbe3add9f16ed084eebec5f87d7fa.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -1,3 +1,6 @@
+@import "govuk_publishing_components/components/mixins/prefixed-transform";
```
owenatgov, line 1: Eyyyyy nice 😎  During cleanup we can separate out your fancy grid work in the gem and expose it in the same way.

owenatgov (approved):
💳

app/assets/stylesheets/views/_homepage.scss
injms: Yeah, I just didn't want putting the mixin in the gem to delay this work. Since it's a mixin it won't add any extra to the compiled CSS, so I didn't think it too bad a compromise.

PR #3035, opened 2021-11-24
Activate use of Elections API

sihugh:
Tweaks monitoring so that we can track all status codes received from the API

Activates use of the Electoral Commission's API in production by using the real route for the `/contact-electoral-registration-office` page. 

If this route is commented out, then we will fall back to using the standard `LocalTransactionController` for this service again (and therefore uses links from local-links-manager again).

### DIFF
--- a/app/services/electoral_service.rb
+++ b/app/services/electoral_service.rb
@@ -11,10 +11,12 @@ class ElectoralService
   def make_request
     @body = begin
       with_caching do
-        json = RestClient.get(request_url, headers)
-        JSON.parse(json)
+        response = RestClient.get(request_url, headers)
+        report_status_code(response.code)
+        JSON.parse(response)
       end
     rescue RestClient::Exception => e
+      report_status_code(e.http_code)
       @error = assemble_error(e)
       {}
     end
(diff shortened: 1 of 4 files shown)


app/services/electoral_service.rb
```diff
@@ -46,17 +46,17 @@ def assemble_error(exception)
     end
   end
 
-  def report_error(code)
-    GovukStatsd.increment("elections.errors.#{code}")
+  def report_status_code(code)
+    GovukStatsd.increment("elections_api.#{monitoring_path}.#{code}")
```
hannako, line 50: Is it fine that before you changes we were logging `elections` and `elections_api` but now we're just logging `elections_api` ?

docs/elections-api.md
```diff
@@ -0,0 +1,86 @@
...
+
+Click "Find"
+
+See results on https://www.gov.uk/contact-electoral-registration-office?postcode=w1a+1aa
+
+### Ambiguous postcode
+
+Land on https://www.gov.uk/contact-electoral-registration-office
+
+Enter unambiguous postcode (eg WV14 8TU)
```
hannako, line 52: Enter ambiguous postcode? 🙃

hannako (approved):

app/services/electoral_service.rb
sihugh: Yep, I tweaked them so that they line up. It's arbitrary what we call them though.

docs/elections-api.md
sihugh: Ta

PR #3043, opened 2021-11-29
Remove Brexit checker card from account home

alex9smith:
The Brexit checker is being retired at the end of the month. Remove the
card about it from the account home page so we're not directing users to
do something that's no longer possible.

Getting this PR ready ahead of time so it can be reviewed and ready to merge when we've decided on the right time.

[Trello](https://trello.com/c/jUukpLeb/1170-remove-brexit-checker-service-card)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -28,9 +28,4 @@ module ApplicationHelper
   def current_path_without_query_string
     request.original_fullpath.split("?", 2).first
   end
-
-  def transition_checker_path
-    base_url = Rails.env.development? ? Plek.find("finder-frontend") : Plek.new.website_root
-    "#{base_url}/transition-check"
-  end
 end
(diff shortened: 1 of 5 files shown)


huwd (changes requested):
I think we would be able to remove this method too, it only appears to be used by the view that goes here.
https://github.com/alphagov/frontend/blob/465a00aae52b6bd762debc14a0f1aba3b61d8dd4/app/helpers/application_helper.rb#L32-L35

huwd:
Ooh the locales too:
https://github.com/alphagov/frontend/blob/aab60326018848b05e71a161858bc8b44c48a76a/config/locales/en.yml#L31-L37

huwd:
Humm not sure about these... 🤔 might be a case of changing the example if that URL is now gone. What do you think?
https://github.com/alphagov/frontend/blob/cb0e28bdfc8695a7d589bf4877553b824280cebf/test/integration/sessions_test.rb#L8-L15

alex9smith:
Ooh good find. I thought we had a test for unused locale keys - not sure why that didn't pick up these

alex9smith:
For the integration tests I think the Brexit team who are actually removing the checker should handle those - we only care about the account home page in this PR and the page does still exist at the moment

huwd (approved):
Nice one! :)

PR #3044, opened 2021-11-29
Fix leaky locale assignment

barrucadu:
The Rails locale is not a request-local property, it's a thread-local
property.  So if two requests are served by the same thread, where the
first request sets the locale to something non-default and the second
does not set it at all, the second request will use the same
non-default locale as the first one.

This is described in the i18n docs:
https://guides.rubyonrails.org/i18n.html

The homepage controller does not set the locale, and so if a user
views a Welsh content item and then the same thread is used to serve
the homepage, the homepage will use the Welsh locale.  This isn't a
new problem, but it didn't cause any issues, because Rails falls back
to the default locale (English in our case) if a translation is
missing, and we didn't have translated content for the homepage.

fe2b0b7 added those missing Welsh translations.

This commit always sets the locale to the default locale at the start
of a request.  A better option would be to use `I18n.with_locale` and
an `around_action` for those controller methods which use a different
locale, but that would be a larger refactor.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,10 @@ class ApplicationController < ActionController::Base
   include Slimmer::Headers
   include Slimmer::Template
 
+  before_action do
+    I18n.locale = I18n.default_locale
+  end
+
   rescue_from GdsApi::TimedOutException, with: :error_503
   rescue_from GdsApi::EndpointNotFound, with: :error_503
   rescue_from GdsApi::HTTPErrorResponse, with: :error_503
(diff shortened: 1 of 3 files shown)


hannako (approved):

app/controllers/application_controller.rb
```diff
@@ -2,6 +2,10 @@ class ApplicationController < ActionController::Base
   include Slimmer::Headers
   include Slimmer::Template
 
+  before_action do
+    I18n.locale = I18n.default_locale
```
kevindew, line 6: Not to block this going out but as a follow-up we should avoid setting I18n.locale directly and globally and instead switch to it being scoped for an action: https://github.com/alphagov/collections/pull/2132/files#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07R41-R44

PR #3045, opened 2021-11-29
Update promo slots on GOVUK homepage

hannako:
Replace COP26 with Coronavirus, to match this archived page:
https://webarchive.nationalarchives.gov.uk/ukgwa/20210805112744/https://www.gov.uk/

### BEFORE
Screenshot 2021-11-29 at 12 24 41

### AFTER
Screenshot 2021-11-29 at 12 24 06

[Release app](https://govuk-fronte-update-pro-dxiidu.herokuapp.com/)


app/views/homepage/_promotion-slots.html.erb
```diff
@@ -5,37 +5,30 @@
...
           loading: "lazy",
           width: 610,
           sizes: "(max-width: 640px) 100vw, (max-width: 1020px) 33vw, 300px",
-          srcset: {
-            "homepage/cop26.jpg": "610w",
-            "homepage/cop26-480.jpg": "480w",
-            "homepage/cop26-320.jpg": "320w",
-            "homepage/cop26-240.jpg": "240w",
-            "homepage/cop26-170.jpg": "170w",
-          },
```
barrucadu, line 30: Could you also add the other image sizes here?  This is good for devices with smaller screens, which can download a smaller image.

hannako: The [docs](https://docs.publishing.service.gov.uk/apps/frontend/update-homepage-promotion-slots.html#3-update-the-markup-and-phrases) suggest we only need the smaller images for photographs (or complex graphics) rather than simple graphics like this one. We also don't provide the `srcset` attributes for the other promo images on the homepage. Do you think this is wrong? I.e. should I update the docs and add smaller images for all three slots?

barrucadu: Ok, if that's only for photos then we don't need to do it here.

barrucadu (approved):

PR #3049, opened 2021-11-30
Update chevron card spacing

injms:
, [Jira issue NAV-3238](https://gov-uk.atlassian.net/browse/NAV-3238)## What

 - Reduce the top spacing on the card.
 - Make the description take up the full width of the card.

Part of https://trello.com/c/fPth3krN

## Why

To make the spacing feel more balanced; and to prevent the text from wrapping too much.

## Visual changes

Preview: https://govuk-fronte-update-che-twkftw.herokuapp.com

### Before:

image

### After:

image

---

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -205,7 +205,7 @@
 .chevron-card {
   border-top: 1px solid $govuk-border-colour;
   margin: 0 govuk-spacing(3);
-  padding: govuk-spacing(2) 0 govuk-spacing(4) 0;
+  padding: govuk-spacing(1) 0 govuk-spacing(4) 0;
 }
 
 .chevron-card__wrapper {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/views/_homepage.scss
```diff
@@ -214,7 +214,7 @@
 }
 
 .chevron-card__description {
-  margin: 0;
+  margin: 0 govuk-spacing(-6) 0 0;
```
injms, line 217: This might seems a strange way of doing it - after all, why not remove the padding-right on `.chevron-card`?

Reasoning is:

- we can't change the right spacing on the heading, as that's a component
- adding padding right to the link (`.chevron-card__link`) affects the shape of the focus state
- adding margin right to the link doesn't do anything, as it's an inline element
- adding an extra wrapper around the heading and link seems overkill

So we're left with this. 

I checked that it works in the browser that are supported, and well as IE10 and IE9.

owenatgov (approved):
I can live with this. There are other ways to do this but they would be adding complexity to the markup elsewhere for the sake of avoiding a negative margin. This is a unique use case and feels like a worthwhile tradeoff.

PR #3055, opened 2021-11-30
Set RAILS_LOG_TO_STDOUT in Dockerfile.

sengi:
This always needs to be set when a Rails app is running in a container, so it makes sense to set it in the Dockerfile rather than making the user remember to set it.


sengi:
Hmm. Having just raised this, I don't think this works the way I'd assumed. Looks like we're doing some fancy footwork in [alphagov/govuk_app_config](https://github.com/alphagov/govuk_app_config/blob/7cef295/lib/govuk_app_config/govuk_logging.rb). So we're already sending logs to stdout and stderr, which is great. It's actually just the buffering that needs fixing, which is presumably something in govuk_app_config and/or Puma.

PR #3058, opened 2021-12-01
Add size: "large" to homepage search

owenatgov:
## What
Add's `size: "large"` to the search input on the new homepage so that it triggers the [search component's large mode](https://components.publishing.service.gov.uk/component-guide/search/large_version).

## Why
So that we are consistent with how we display the search input across the homepage and the super nav header.

[Card](https://trello.com/c/5polNMpk/686-7homepage-increase-searchbox-height-mobile), [Jira issue NAV-3283](https://gov-uk.atlassian.net/browse/NAV-3283)

### Before
Screenshot 2021-12-01 at 17 11 00

### After
Screenshot 2021-12-01 at 17 11 36

### DIFF
--- a/app/views/homepage/_links_and_search.html.erb
+++ b/app/views/homepage/_links_and_search.html.erb
@@ -34,6 +34,7 @@
           label_text: t("homepage.index.search_label"),
           margin_bottom: 0,
           wrap_label_in_a_heading: true,
+          size: "large",
         } %>
       </div>
     </div>


mia-allers-gds:
@owenatgov looks good, any idea of whether we can meet our stretch of keeping it smaller on desktop?

chao-xian (approved):

owenatgov:
@mia-allers-gds We're making good time so maybe. It'll involve a change to the gem so we could quite easily do it a bit later. I've got some thoughts about how to do this and will investigate as I get to the end of our snag list.

PR #3063, opened 2021-12-03
Update spacing for the bank holiday box on the new homepage

owenatgov:
## What
Amends the spacing on the bank holidays blue box on the new homepage design

## Why
So that it lines up with page margins on desktop and looks neat on smaller screens.

[Card](https://trello.com/c/mwUNkxRc/680-11homepage-fix-right-alignment-of-the-uk-bank-holidays-box), [Jira issue NAV-2826](https://gov-uk.atlassian.net/browse/NAV-2826)

## Mobile (375px)
Before:
Screenshot 2021-12-03 at 16 16 24

After:
Screenshot 2021-12-03 at 16 00 32

## Desktop
Before:
Screenshot 2021-12-03 at 16 17 00

After:
Screenshot 2021-12-03 at 16 01 21

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -136,13 +136,20 @@
 
 .homepage-promo {
   display: block;
+  box-sizing: border-box;
   margin: 0;
-  padding: (govuk-spacing(9) + govuk-spacing(4)) govuk-spacing(3); // results in a top and bottom padding of 80px for small screens
+  padding: govuk-spacing(8);
   background: $govuk-brand-colour;
   text-align: center;
+  width: 100%;
   max-width: 240px;
 
   @include govuk-media-query($from: tablet) {
+    max-width: none;
+    padding: (govuk-spacing(9) + govuk-spacing(2)) govuk-spacing(3); // govuk spacing 9 (60px) + 2 (10px) to create a top and bottom of 70px each
+  }
+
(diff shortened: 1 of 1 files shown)


mia-allers-gds:
Looks good, thank you @owenatgov

chao-xian:
It's squished on tablet, is that going to be ok?
govuk-fronte-new-homepa-aqcdsv herokuapp com_(iPad)

owenatgov:
@chao-xian So this happens between approx 705px and 770px. @mia-allers-gds Is Kelv's comment above gonna be a problem? If so I can try and mess with the spacing.

mia-allers-gds:
@owenatgov @chao-xian I think we decided we weren't going to play around too much with it as we aren't sure it will persist in the long term? It looks not great at that size, the height could be increased to make it look better but as I recall Owen said that would be quite a lot of work?

chao-xian:
Ah ok - let's mark it as an iteration thing then and "fix it" in the future. I'll add a card to the planning board.

chao-xian (dismissed):

owenatgov:
@mia-allers-gds @chao-xian So, it's a bit fiddly and becomes awkward at widths where the text goes across multiple lines. I tried googling about your idea Mia of having consistent dimensions like an image and was a little surprised to find that there isn't an easy way to do that. Lemme try to make it a little taller on tablet and see where I get to.

chao-xian (approved):

PR #3068, opened 2021-12-07
Remove uses of `has_unconfirmed_email`

alex9smith:
This attribute is no longer relevant since we switched to using DI's SSO
service - they require users to confirm their email address as part of
the account creation journey.

These helper methods aren't being used in the account home view, so
this commit has no affect on the behaviour or appearance of this app.

[Trello](https://trello.com/c/SdkcF8EO/13-remove-hasunconfirmedemail-attribute)

### DIFF
--- a/app/controllers/account_home_controller.rb
+++ b/app/controllers/account_home_controller.rb
@@ -13,32 +13,10 @@ class AccountHomeController < ApplicationController
 
 private
 
-  helper_method :show_confirmation_reminder?
-
-  def show_confirmation_reminder?
-    return false unless @user
-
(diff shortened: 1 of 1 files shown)


app/controllers/account_home_controller.rb
```diff
@@ -13,14 +13,6 @@ def show
...
-
-  def show_confirmation_reminder?
-    return false unless @user
-
-    !@user["email_verified"] || @user["has_unconfirmed_email"]
-  end
-
   helper_method :confirmation_banner_prompt_type
 
   def confirmation_banner_prompt_type
```
barrucadu, line 18: Remove this method too

barrucadu (approved):

PR #3081, opened 2021-12-13
Remove govuk-link--no-visited-state from links on the homepage

owenatgov:
## What
Remove the class `govuk-link--no-visited-state` from links on the homepage.

## Why
So we are using browser functionality to clearly convey to users which links they have already visited.


jon-kirwan (approved):
👍

PR #3084, opened 2021-12-13
Set the correct number of items for the government activity section on IE11

owenatgov:
## What
Change the number of grid items under the government activity section of the homepage from 5 to 6.

## Why
On IE11, you need to explicitly specify the number of items in a grid in order for the grid to render properly. This had been missed in a recent addition and was causing IE11 and below to look weird.

### Before
Screenshot 2021-12-13 at 17 00 15

### After
Screenshot 2021-12-13 at 17 00 37


andysellick (approved):
🚀

PR #3085, opened 2021-12-13
Apply correct grid markup to cookie page

owenatgov:
## What
Reworks how the grid is used on the cookie page to be semantically correct.

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/2202

### Before
www gov uk_help_cookies

### After
frontend dev gov uk_help_cookies

### DIFF
--- a/app/views/help/cookie_settings.html.erb
+++ b/app/views/help/cookie_settings.html.erb
@@ -1,137 +1,141 @@
 <% content_for :title, "Cookies on GOV.UK" %>
-<main id="content" role="main" class="group help-page cookie-settings govuk-grid-column-two-thirds">
-    <div class="cookie-settings__confirmation" data-cookie-confirmation="true">
-      <%= render "govuk_publishing_components/components/success_alert", {
-        message: t('cookies.confirmation_title'),
-        description: raw("<p class='govuk-body'>#{t('cookies.confirmation_message')}</p>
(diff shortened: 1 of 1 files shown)


app/views/help/cookie_settings.html.erb
```diff
@@ -1,13 +1,16 @@
...
       <%= render "govuk_publishing_components/components/success_alert", {
         message: t('cookies.confirmation_title'),
         description: raw("<p class='govuk-body'>#{t('cookies.confirmation_message')}</p>
-        <a class='govuk-link govuk-notification-banner__link cookie-settings__prev-page' href='#' data-module='gem-track-click' data-track-category='cookieSettings' data-track-action='Back to previous page'>#{t('cookies.confirmation_previous_referrer_link')}</a>").html_safe
+        <a class='govuk-link govuk-notification-banner__link cookie-settings__prev-page' href='#' data-module='gem-track-click' data-track-category='cookieSettings' data-track-action='Back to previous page'>#{t('cookies.confirmation_previous_referrer_link')}</a>").html_safe,
+        margin_bottom: 0,
       } %>
   </div>
 
+  <div class="govuk-grid-column-two-thirds">
```
andysellick, line 12: Why did you add another grid column here? It's the same as the one above - would all of the content not fit into that one?

owenatgov:
@andysellick I've addressed your review and also moved a few classes around as the way we were using `main` here was inconsistent with how we use it elsewhere (no classes, only the content id and `role="main"`) and removed a few redundant classes which I suspect are going to break a few smokey tests but I can fix them upstream. Take a look when you have a sec.

owenatgov:
We've already got a failing test! Investigating.

andysellick (approved):
Nice 🎄

PR #3111, opened 2022-01-06
Fix not passing redirect path to first_time form

barrucadu:
This bug was introduced in 59f09bf.  If you look in that commit, you
can see that it was `callback["redirect_path"]` previously.  I've
added a test to check this case.


danacotoran (approved):
Seems like it should do the right thing based on my limited understanding of how this works

PR #3112, opened 2022-01-07
[WIP] Add Percy for visual regression testing

chao-xian:

### DIFF
--- a/.github/workflows/visual-regression-tests.yml
+++ b/.github/workflows/visual-regression-tests.yml
@@ -4,7 +4,7 @@ on:
       - main
 
 jobs:
-  visual-regression-test-components:
+  visual-regression-test:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/visual-regression-tests.yml
+++ b/.github/workflows/visual-regression-tests.yml
@@ -11,18 +11,10 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true # Also runs `bundle install`
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2.5.1
         with:
           node-version: "12"
-- name: Get Yarn cache directory path
-  id: yarn-cache-directory-path
-  run: echo "::set-output name=dir::$(yarn cache dir)"
-- name: Check for cached Yarn dependencies
-  uses: actions/cache@v2
-  with:
-    path: ${{ steps.yarn-cache-directory-path.outputs.dir }}
-    key: yarn-${{ hashFiles('**/yarn.lock') }}
-    restore-keys: yarn-
+          cache: yarn
 - run: yarn install --frozen-lockfile
 - uses: percy/exec-action@v0.3.1
   with:
--- a/.github/workflows/visual-regression-tests.yml
+++ b/.github/workflows/visual-regression-tests.yml
@@ -18,6 +18,6 @@ jobs:
 - run: yarn install --frozen-lockfile
 - uses: percy/exec-action@v0.3.1
   with:
-    command: bundle exec rspec --tag visual_regression
+    command: bundle exec rake test_unit
   env:
     PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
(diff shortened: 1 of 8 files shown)


.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true # Also runs `bundle install`
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "12"
+This conversation was marked as resolved by injms
```
injms, line 17: ?
```suggestion

```

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: Get Yarn cache directory path
+  id: yarn-cache-directory-path
+  run: echo "::set-output name=dir::$(yarn cache dir)"
+- name: Check for cached Yarn dependencies
+  uses: actions/cache@v2
+  with:
+    path: ${{ steps.yarn-cache-directory-path.outputs.dir }}
+    key: yarn-${{ hashFiles('**/yarn.lock') }}
+    restore-keys: yarn-
```
injms, line 26: If we update the `setup-node` image to version 2, we can replace this slightly odd caching stuff with the built in caching available:

```suggestion 
- uses: actions/setup-node@v2.5.1
   with:
     node-version: "12"
     cache: yarn
```

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+- run: yarn install --frozen-lockfile
+- uses: percy/exec-action@v0.3.1
+  with:
+    command: bundle exec rspec --tag visual_regression
+  env:
+    PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
```
injms, line 31: Could this token be added at an organisation level, since updating it could be a faff once it's being used in 2+ repositories?

chao-xian: [Seems we can ](https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-secrets-and-runners-with-your-organization#sharing-secrets-within-an-organization) but I'll need someone with The Power like from @alex-ju 🙏

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,31 @@
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  visual-regression-test-components:
```
injms, line 7: Just a rename, since we're not looking at components here:

```suggestion
  visual-regression-test:
```

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+        with:
+          bundler-cache: true # Also runs `bundle install`
+      - uses: actions/setup-node@v2.5.1
+        with:
+          node-version: "12"
+          cache: yarn
+- run: yarn install --frozen-lockfile
+- uses: percy/exec-action@v0.3.1
+  with:
+    command: bundle exec rspec --tag visual_regression
```
injms, line 21: I think that this repo doesn't use RSpec for tests, it uses Minitest (at least, that's what it looks like in the `Rakefile`) - which could be why this command isn't doing anything.

Or because none of the tests are tagged with `visual_regression` nothing is being found to run.

I think this command will need to be changed to
```suggestion
    command: bundle exec rake test_unit
```

but that will run _all_ the tests, so we could either 

 - quick and dirty - only run the homepage test for the moment (which still runs all the other non-visual regression tests at the same time)
 - less tech debty but more work - create a folder specifically for adding only visual regression tests, which could be ignored for local runs and only run in this action.

chao-xian: Gonna quick and dirty it first to make sure it works, then another PR/issue to improve it?

.github/workflows/visual-regression-tests.yml
alex-ju: Sure, I'll catch-up with Kelvin on setting this up at the org level

PR #3117, opened 2022-01-13
Apply margin bottom to big number component on homepage

owenatgov:
## What
Apply a `margin_bottom` attribute to the instances of the [big number component](https://components.publishing.service.gov.uk/component-guide/big_number) on the gov.uk homepage and remove the wrapper that was there previously and it's associated styling.

## Why
Now that big numbers have the option to apply a custom margin bottom, it makes sense to use it instead of applying wrappers.

[Card](https://trello.com/c/u0DGQWQY/724-add-margin-bottom-to-big-number-component-and-apply-to-the-homepage), [Jira issue NAV-2886](https://gov-uk.atlassian.net/browse/NAV-2886)

## Visual changes (mobile only)
This change suffers from the same issue a number of elements on the homepage are suffering from in that they are forced to use the design system spacing matrix and we can't control spacing between screen sizes. More details on this issue https://github.com/alphagov/govuk-frontend/issues/2457. In that respect, this change potentially introduces a visual regression.

### Before
Screenshot 2022-01-13 at 16 45 58

### After
Screenshot 2022-01-13 at 16 46 15

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -229,12 +229,6 @@
   padding: 0;
 }
 
-.homepage-big-number__wrapper {
-  &:nth-child(1) {
-    margin-bottom: govuk-spacing(6);
-  }
-}
-
 .homepage-search {
   @include govuk-media-query($until: "tablet") {
     max-width: 420px;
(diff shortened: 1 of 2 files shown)


mia-allers-gds:
@owenatgov after definitely looks better to my eye

PR #3140, opened 2022-02-02
Flesh out signing in integration tests 

barrucadu:
This also addresses two minor points with the sessions controller:

- If the _ga parameter is present but blank, strip it from the
destination URL.  There's no point passing along an empty GA session
id.

- Gracefully handle the lack of a redirect path for a first-time user.
This is strictly speaking a bug, but it's unlikely that a user would
stumble upon this, as they would have to directly enter the URL for
www.gov.uk/account (or /sign-in/redirect), without clicking a link to
it, despite not already having an account.

These tests don't cover the special request / response header logic.
They just stub it out.  I don't think we can properly handle it
without being able to, eg, run code in between capybara receiving a
redirect response and following it (to update the request headers
based on the response).  But the headers are covered by the functional
tests for the controller.

Also, I found the need to stub out the email-alert-api calls a little
awkward, and feel that we should probably move that communication to
account-api as part of the GET /api/user behaviour, rather than doing
it in frontend.

---

[Trello card](https://trello.com/c/8MltdY6B/1211-review-improve-tests-for-authentication)

### DIFF
--- a/test/integration/sessions_test.rb
+++ b/test/integration/sessions_test.rb
@@ -5,39 +5,6 @@ class SessionsTest < ActionDispatch::IntegrationTest
   include GdsApi::TestHelpers::AccountApi
   include GovukPersonalisation::TestHelpers::Features
 
-  context "HTTP Referer" do
-    should "add a redirect_path param to /account from the HTTP Referer header" do
-      stub = stub_account_api_get_sign_in_url(redirect_path: "/transition-check/results?c[]=import-wombats&c[]=practice-wizardry")
-
-      get "/account", headers: { "Referer" => "#{Plek.new.website_root}/transition-check/results?c[]=import-wombats&c[]=practice-wizardry" }
-
-      assert_response :redirect
-      assert_requested stub
-    end
-
-    should "not allow protocol-relative redirects" do
-      stub = stub_account_api_get_sign_in_url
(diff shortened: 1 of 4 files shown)


test/integration/sessions_test.rb
```diff
@@ -1,54 +1,171 @@
...
+      # form, to pluck out the response header and fill in the request
+      # header)
       mock_logged_in_session "!#{@govuk_account_session}"
+    end
 
-      visit new_govuk_session_callback_path(code: "code", state: "state")
-
-      assert page.has_content? "use cookies to collect anonymised analytics"
+    @stub_user_info = stub_account_api_user_info
+    stub_email_alert_api_authenticate_subscriber_by_govuk_account("!#{@govuk_account_session}", "subscriber-id", "email@example.com")
```
huwd, line 118: Right so the proposal is we move the need for three three bits, integrating the data we've stubbed here into GET /api/user.

Makes sense to me.

huwd (approved):

test/integration/sessions_test.rb
barrucadu: Yeah, if we return `services: { notifications: "yes" }` or `services: { notifications: "no" }` from `GET /api/user` then the controller here can just read that.

PR #3138, opened 2022-02-02
Upgrade Ruby to 2.7.5

ChrisBAshton:
Upgrades Ruby to 2.7.5. See commits for more details.
            This PR should be tested before merge, as per https://docs.publishing.service.gov.uk/manual/ruby.html#update-ruby-version-in-the-relevant-repos.

PR #3152, opened 2022-02-23
Fix Dockerfile

fredericfran-gds:
In #3138, they upgraded to Ruby 2.7.5 from 2.7.3 but they use the
"normal" Ruby 2.7.5 image in the Dockerfile rather than the
slim-buster on. This PR fixes this issue.


sengi (approved):
Thanks!

PR #3153, opened 2022-03-01
Replace Levelling up promo with ukraine promo

KludgeKML:
https://govuk.zendesk.com/agent/tickets/4899159


brucebolt (commented):
Looks like there's some files `.DS_Store` that probably shouldn't have been committed (I think they're something Mac related)

maxgds (changes requested):
You probably want to remove the image(s) from the old promo here, unless you think it's likely to be put back at some point. Also, you've supplied different image sizes but I don't see a srcset value list in the yml

Actually, the guidance says to remove the old images regardless as they can be retrieved.

maxgds:
> You probably want to remove the image(s) from the old promo here, unless you think it's likely to be put back at some point. Also, you've supplied different image sizes but I don't see a srcset value list in the yml
> 
> Actually, the guidance says to remove the old images regardless as they can be retrieved.

You did in fact remove the old image, my apologies!

maxgds (approved):

PR #3161, opened 2022-03-04
Change popular links to use content from publishing_components

huwd:
## Dependencies

🚨 Do not merge until [this PR ](https://github.com/alphagov/govuk_publishing_components/pull/2660) has been merged, released and the latest version updated in `frontend` 🚨

Otherwise the update to add `Invasion of Ukraine` popular link will be revered to older content

## What & Why?

Since the header redesign popular links now appear in two places, once
on the home page and one on the header.

If these two list diverge it could be confusing for users, and updating
now has to occur in two places.

However, as the publishin components are a dependency of frontend,
content stored in the locale file here is already loaded into the app.

This means we can remove this content from the application, and load it
directly from the compoent content.

As a result there is a single place we can upload changes.

## But does it work?
Yes!

image

### DIFF
--- a/app/views/homepage/_links_and_search.html.erb
+++ b/app/views/homepage/_links_and_search.html.erb
@@ -5,12 +5,12 @@
         <%= render "govuk_publishing_components/components/heading", {
           font_size: "m",
           margin_bottom: 4,
-          text: t("homepage.index.popular_links_heading"),
+          text: t("components.layout_super_navigation_header.popular_links_heading"),
         } %>
         <ul class="homepage-most-viewed-list" data-module="gem-track-click">
-          <% t("homepage.index.popular_links").each do | item | %>
+          <% t("components.layout_super_navigation_header.popular_links").each do | item | %>
             <li>
-              <%= link_to(item[:text], item[:href], {
+              <%= link_to(item[:label], item[:href], {
                 class: "govuk-link homepage-most-viewed-list__item",
                 data: {
                   track_category: "homepageClicked",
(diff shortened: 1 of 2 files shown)


app/views/homepage/_links_and_search.html.erb
```diff
@@ -5,12 +5,12 @@
         <%= render "govuk_publishing_components/components/heading", {
           font_size: "m",
           margin_bottom: 4,
-          text: t("homepage.index.popular_links_heading"),
+          text: t("components.layout_super_navigation_header.popular_links_heading"),
         } %>
         <ul class="homepage-most-viewed-list" data-module="gem-track-click">
-          <% t("homepage.index.popular_links").each do | item | %>
+          <% t("components.layout_super_navigation_header.popular_links").each do | item | %>
```
kevindew, line 11: This feels a bit fishy to me by the way it uses the `components.layout_super_navigation_header` locale key. As it feels a bit like you're pulling a private API. If you were to make changes to the `layout_super_navigation_header` component you'd not expect it to then reverberate elsewhere in the stack.

I think it'd be ok if the key wasn't specific to a particular component or used a public API of a helper class - it'd be worth checking that tests definitely fail if, whatever is used, is missing in components gem.

FWIW I think we've got things somewhat the wrong way round that components gem knows about this data and other apps pull it in, rather than apps inject it into component gem. But can see that this is harder to solve when the current situation is that apps wouldn't need to duplicate this information.

huwd: Great, thanks for responding Kevin.

I think I'll close this down now rather than leaving it open - hopefully can be a useful reference for whatever we do down the line to sort out the wider FWIW bit :)

PR #3186, opened 2022-03-30
Remove unused partial and unused translations

ChrisBAshton:
I wondered why our LocaleChecker configuration had been set to ignore homepage-related locales. Did a bit of digging and concluded we could re-enable the checker in full by removing some unused code/configs.

See commits for details.


richardTowers (approved):
Thanks for this Chris! The commits all make sense to me. Looks like a nice cleanup, and should prevent us from wasting time maintaining translations that we don't use.

PR #3211, opened 2022-04-28
Add GA4 analytics tracking attributes to tabs

andysellick:

## What
Preliminary attempt to try to get GTM code working on integration. Note that this should only work on integration, where we have enabled the `dataLayer` using the [GTM component](https://components.publishing.service.gov.uk/component-guide/google_tag_manager_script). Applies to the tabs on the `/renew-driving-licence` and `/bank-holidays` pages.

## Why
We want to experiment to see if we can get some data events to work and show up in GA.

## Visual changes
None.

Trello card: https://trello.com/c/V0RFrSMK/43-test-technical-approach-to-tracking-on-tabs-and-accordions

### DIFF
--- a/app/presenters/transaction_presenter.rb
+++ b/app/presenters/transaction_presenter.rb
@@ -30,8 +30,12 @@ class TransactionPresenter < ContentItemPresenter
     content_item["title"]
   end
 
+  def tab_count
+    [more_information, what_you_need_to_know, other_ways_to_apply].count(&:present?)
+  end
+
   def multiple_more_information_sections?
-    [more_information, what_you_need_to_know, other_ways_to_apply].count(&:present?) > 1
+    tab_count > 1
   end
 
   def start_button_text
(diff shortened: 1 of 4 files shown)


kevindew:
Ooh just caught a glimpse of this when looking at this repo for something totally different - this comment may be totally not relevant or helpful as I don't know your direction.

One thing caught my eye that you're going to the level of tracking browser events and then passing that to GTM datalayer. I noticed this because a different direction was taken in Content Publisher with this that you might find is lower maintenance and I'm thinking you probably don't need to write any JS as GTM can take care of it. 

In GTM there was set-up a generic click listener for elements with a data attribute of gtm-click-tracking="true" and then uses other attributes of the element to set the data sent as part of the event. 

Again apologies if this is an unhelpful comment, just didn't want you to be going down a big "we need to write JS for everything approach" if it can be avoided in the early days.

andysellick:
That's really interesting, thanks @kevindew. I'll definitely look into that approach. For now though this code is an attempt to just get some data appearing in GA, so yes, really hoping the finished product doesn't look like this!

gclssvglx (commented):
Looking good @andysellick - just have one question...

gclssvglx (approved):
Nice one @andysellick

PR #3221, opened 2022-05-11
Replace jasmine gem with jasmine-browser-runner npm package

edwardkerry:
This PR swaps the deprecated Jasmine gem with jasmine-browser-runner npm package.

Images of script and specs running as expected in CI:
Screenshot 2022-05-11 at 12 19 40
Screenshot 2022-05-11 at 12 20 00


KludgeKML (approved):
Looks good to me!

spec/support/jasmine-browser.json
```diff
@@ -0,0 +1,22 @@
...
+  ],
+  "helpers": [
+    "helpers/*.js",
+    "vendor/*.js"
+  ],
+  "env": {
+    "stopSpecOnExpectationFailure": false,
+    "stopOnSpecFailure": false,
+    "random": false
+  },
```
hannako, line 20: What are these env variables for? Where are they being used. I've not seen this in the other jasmine browser prs

KludgeKML: I suspect that's my lack of due diligence. Ed probably copied them from Smart Answers, where I used them copied from (I can't remember, but one of the publishing tech apps).

It's merged now, but if these are unnecessary or should have other settings, they'll need changing in Smart Answers too, so leave it with me - If they're not necessary I'll make some PRs to have them removed.

hannako: Ah yep no worries. As per slack message: Looking at the docs, false is the default for stopSpecOnExpectationFailure and stopOnSpecFailure any way, and I feel like we should be running the specs in a random order! So wondering if we should lose them?

KludgeKML: Looks like Ed's added this: https://github.com/alphagov/frontend/pull/3223 I'll do the same for SA.

PR #3223, opened 2022-05-11
Remove unneeded jasmine-browser-runner configuration

edwardkerry:
Ideally tests would run in a random order so we should remove this
configuration. Other configuration options are setting the default in
any case so can be removed for concision and clarity.

We are only using a `/vendors` directory for our helpers in this app, so we
can remove the `/helpers` directory from the `helpers` block.

### DIFF
--- a/spec/support/jasmine-browser.json
+++ b/spec/support/jasmine-browser.json
@@ -10,13 +10,7 @@
     "**/*[sS]pec.js"
   ],
   "helpers": [
-    "helpers/*.js",
     "vendor/*.js"
   ],
-  "env": {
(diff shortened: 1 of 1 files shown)


hannako (commented):
Looks like you've been a bit over zealous with your config trimming and removed the helpers by accident 😂

edwardkerry:
> Looks like you've been a bit over zealous with your config trimming and removed the helpers by accident 😂

We don't have a `/helpers` directory here - just a `/vendors`

hannako (approved):
LGTM

PR #3224, opened 2022-05-11
Remove first time sign in page

alex9smith:
When we first built the account home area, we were expecting it to be
part of `www.gov.uk` with only the authentication pages on
`account.gov.uk`. Since then, the team building the account area has
moved to the Digital Identity directorate and we've decided that all of
the central account pages should be on `account.gov.uk` and `www.gov.uk`
should be treated as any other service that's using GOV.UK Sign In as
an OIDC provider.

The first time sign in page in this app currently has part of the
account privacy policy and asks users for their consent to use cookies
and send them emails to ask for feedback. These questions need to be
moved along with the rest of the account home area to a new central
app on `account.gov.uk` and so will not be stored in GOV.UK's account
API.

This app uses the presence of answers to these questions to decide if a
user is signing in for the first time or not. If the questions are no
longer present, then this will no longer be possible. The only other
content on this page is privacy information which we are going to move
to email alert frontend.

Therefore we can remove the first time sign in page and logic in the
callback endpoint to detect if a user is signing in for the first time. 

This is quite a large commit, apologies, but if I split it up more I'd have to 
spend quite a bit of time refactoring tests that were going to be removed 
in later commits which didn't seem worth it to me. 

---

*DO NOT MERGE* this PR until we are ready to deploy the PR on 
email alert frontend to add the privacy information to that journey.

---
[Trello](https://trello.com/c/ZmuMTQy5/1370-remove-cookies-and-feedback)


app/controllers/sessions_controller.rb
```diff
@@ -19,75 +18,11 @@ def callback
...
-      when "reject"
-        update_saved_cookie_consent = callback["cookie_consent"]
-        "reject"
-      else
-        update_saved_cookie_consent = false
-        callback["cookie_consent"] ? "accept" : "reject"
-      end
-
-    if callback.key?("cookie_consent") && callback.key?("feedback_consent") && (callback["cookie_consent"].nil? || callback["feedback_consent"].nil?)
-      # slightly hacky way of passing the session header along in the
```
huwd, line 36: I guess we don't need this any more becuase now we can just "properly log the user in"?

huwd (approved):
Much simplier, nice one.
Can we give this a test in integration / staging, whilst we wait for the other stuff?

app/controllers/sessions_controller.rb
alex9smith: Yes exactly - the user always gets logged in now when they hit this endpoint (assuming they have a valid code etc.) so there's no need to mess with the session header or check the previously stored attributes at all. 

Now all this endpoint does is check their code, fetch `userinfo`, create an account if they're new and redirect them onto the page they came from. So much simpler!

alex9smith:
Here's a short recording of the journey with this and https://github.com/alphagov/email-alert-frontend/pull/1303 on integration:

https://user-images.githubusercontent.com/6362602/168086367-a4811971-24c7-44bc-ac76-dfaaf3844282.mov

Note there's no first time sign in page any more, and the account privacy details are on the subscription confirmation page.

PR #3253, opened 2022-06-01
Fix column alignment in the "More on GOV.UK" homepage component

matthillco:
In tablet view and above, the columns in this component were misaligned with the columns in the Featured component above. This is due to two incompatible layout approaches.

Several approaches to align these elements were investigated, using combinations of float based layouts and CSS grid. However these approaches did not yield any compatible alignments, possibly due to differences in how browsers calculate `%` widths vs `fr` widths.

The fix was to use CSS `columns` with a `column-gap` matching the gutter in the row/columns layout classes, and limiting the width of the list to two columns out of three (66.66%). This aligns the elements correctly while keeping the required top -> bottom then left -> right behaviour of the links.

Note that the `break-inside: avoid-column` and the padding-top and padding-bottom tweaks are used to prevent triggering a possible bug in Safari where focused anchors in a multi-column layout cause links to jump between columns. As this is not a long-term robust solution, we should investigate this further for a possible refactor.

[Heroku preview](https://frontend-pull-3253.herokuapp.com/)

[Trello card](https://trello.com/c/Qgx5RXtL/1043-align-the-more-on-govuk-homepage-section-columns-to-the-grid), [Jira issue NAV-5463](https://gov-uk.atlassian.net/browse/NAV-5463)

## Before

image

## After

frontend-pull-3253 herokuapp com_

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -102,16 +102,44 @@
   padding: 0;
 
   @include govuk-media-query($from: desktop) {
-    @include columns($items: 10, $columns: 2, $selector: "li", $flow: column);
+    width: 66.66%;
+    columns: 2;
+    column-gap: 15px;
   }
+}
 
-  & > li {
-    margin: 0 0 govuk-spacing(4) 0;
-    @include govuk-font($size: 19);
+.homepage-most-active-list__item {
+  margin: 0 0 govuk-spacing(4) 0;
+  @include govuk-font($size: 19);
+  @include govuk-media-query($from: desktop) {
+    break-inside: avoid-column;
+    margin-bottom: 0;
+    padding-top: .1875rem;
+    padding-bottom: calc(#{govuk-spacing(2)} - .1875rem);
 
-    @include govuk-media-query($from: desktop) {
-      margin-bottom: govuk-spacing(2);
-    }
+    // After much experimentation, I've used break-inside: avoid-column
+    // and some tweaks to padding-top and padding-bottom on the list
+    // items. This prevents triggering a possible bug in Safari where
+    // focused anchors in a multi-column layout cause links to jump
+    // between columns.
+
+    // Magic numbers in CSS aren't recommended, but the .1875rem value
+    // used in the padding is to compensate for the value used in our
+    // global hover states for links, eg:
+
+    // text-decoration-thickness: max(3px, .1875rem, .12em);
+
+    // This value could be tweaked, and may not be "correct" but when
+    // added to the top padding, it prevents Safari from rendering a
+    // sliver of the focus state in the opposite column.
+
+    // The bottom margin was removed and the spacing instead added as
+    // padding-bottom -- this gives some space within the element for
+    // the focus/hover states to render without being cropped off,
+    // another possible Safari bug.
+
+    // This likely isn't a long-term robust solution and may require
+    // a full refactor in the future.
   }
 }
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/views/_homepage.scss
```diff
@@ -102,16 +102,17 @@
   padding: 0;
 
   @include govuk-media-query($from: desktop) {
-    @include columns($items: 10, $columns: 2, $selector: "li", $flow: column);
+    max-width: 66.66%;
+    columns: 2;
+    column-gap: 15px;
```
patrickpatrickpatrick, line 107: I checked on [caniuse?](https://caniuse.com/?search=column-gap) and it seems `column-gap` isn't supported by IE11. I was thinking an alternative approach could be to separate the unordered list of 10 items into two unordered lists of 5 items and then use the same govuk-grid-column classes that are being used in the row above?

patrickpatrickpatrick (changes requested):

patrickpatrickpatrick (approved):
LGTM!

injms:
I'm seeing an odd behaviour in Safari[^1] - when tabbing through the list of links, the "Check your State Pension age" link moves to the left column:

Screenshot 2022-06-07 at 16 07 13

Then when tabbing to the next link, it reverts back to the second column:

Screenshot 2022-06-07 at 16 07 18

[^1]: Safari Version 15.5 (16613.2.7.1.9, 16613)

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -102,16 +102,20 @@
...
-      margin-bottom: govuk-spacing(2);
-    }
+.homepage-most-active-list__item {
+  margin: 0 0 govuk-spacing(4) 0;
+  @include govuk-font($size: 19);
+  @include govuk-media-query($from: desktop) {
+    break-inside: avoid-column;
+    margin-bottom: 0;
+    padding-top: 0.1875rem;
+    padding-bottom: calc(#{govuk-spacing(2)} - 0.1875rem);
```
injms, line 118: You've put a really thorough explanation in the git commit message - could the explanation also be put here as a comment to make it easier to find without using `git blame`?

matthillco: I've updated the PR description with the same commit message, is that OK?

matthillco:
After much experimentation, I have used `break-inside: avoid-column` and some tweaks to `padding-top` and `padding-bottom` on the list items: this prevents triggering a possible bug in Safari where focused anchors in a multi-column layout cause links to jump between columns. 

This isn't a long-term robust solution so we should investigate this further for a possible refactor.

app/assets/stylesheets/views/_homepage.scss
matthillco: I've added a more detailed comment as requested.

matthillco, line 118: After much experimentation, I have used `break-inside: avoid-column` and some tweaks to `padding-top` and `padding-bottom` on the list items. This prevents triggering a possible bug in Safari where focused anchors in a multi-column layout cause links to jump between columns.

I wouldn't normally use magic numbers in CSS like this, but the `0.1875rem` value used in the padding is to compensate for the value used for the `text-decoration-thickness` in our global hover states for links (`text-decoration-thickness: max(3px, .1875rem, .12em);`). This value could be tweaked, and may not be "correct" but when added to the top padding, it prevents Safari from rendering one pixel of the focus state in the opposite column.

The bottom margin was removed and the spacing instead added as `padding-bottom` -- this gives some space within the element for the focus/hover states to render without being cropped off, another possible Safari bug.

This isn't a long-term robust solution so we should investigate this further for a possible refactor.

injms (approved):
Looks good - and thanks for fixing that quite esoteric Safari bug 👍

PR #3254, opened 2022-06-07
Add option for anchor to location form results

patrickpatrickpatrick:
## What

On pages with a location form, an anchor link can now be used in the resulting url.

## Why

On pages with [long content before the location form](https://www.gov.uk/female-genital-mutilation-help-advice), when a user submitted the results the scroll position would be brought to the top of the page and it was unclear if the search had taken place or where the results of search were. If an anchor link is specified, on submit the form will redirect to an anchor link to where the results on the page are. By default this is specified but you can easily disable by setting the results_anchor to nil for a given slug.

[Related Trello Card](https://trello.com/c/oJlNbOFj/1069-fgm-help-and-advice-postcode-search), [Jira issue NAV-5197](https://gov-uk.atlassian.net/browse/NAV-5197)

### DIFF
--- a/app/controllers/place_controller.rb
+++ b/app/controllers/place_controller.rb
@@ -27,14 +27,18 @@ private
   end
 
   def locals
+    locals = {
+      results_anchor: "results",
+    }
+
     if params[:slug] == REPORT_CHILD_ABUSE_SLUG
-      {
+      locals.merge!({
         option_partial: "option_report_child_abuse",
         preposition: "for",
-      }
-    else
-      {}
+      })
     end
+
+    locals
   end
 
   def postcode_provided?
--- a/app/views/application/_location_form.html.erb
+++ b/app/views/application/_location_form.html.erb
@@ -1,3 +1,5 @@
+<% action ||= nil %>
+
 <div class="postcode-search-form"
   data-module="track-submit"
   data-track-category="postcodeSearch:<%= publication_format %>"
@@ -22,7 +24,13 @@
     } %>
   <% end %>
 
-  <form method="post" id="local-locator-form" class="location-form govuk-!-margin-bottom-9">
+  <%= form_with(
+    id: "local-locator-form",
+    url: action,
+    class: "location-form govuk-!-margin-bottom-9",
+    method: "post",
+  ) do | form | %>
+
     <fieldset class="govuk-fieldset">
       <legend class=" govuk-fieldset__legend govuk-visually-hidden"><%= t('formats.local_transaction.postcode_lookup') %></legend>
 
@@ -53,5 +61,5 @@
         class: "govuk-body" 
       %>
     </fieldset>
-  </form>
+  <% end %>
 </div>
--- a/app/views/place/show.html.erb
+++ b/app/views/place/show.html.erb
@@ -1,3 +1,5 @@
+<% results_anchor ||= nil %>
+
 <% content_for :extra_headers do %>
   <%= render 'govuk_publishing_components/components/machine_readable_metadata',
     schema: :article,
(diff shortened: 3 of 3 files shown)


app/controllers/place_controller.rb
```diff
@@ -27,11 +28,16 @@ def publication
...
 
   def locals
-    if params[:slug] == REPORT_CHILD_ABUSE_SLUG
+    case params[:slug]
+    when REPORT_CHILD_ABUSE_SLUG
       {
         option_partial: "option_report_child_abuse",
         preposition: "for",
       }
+    when FGM_SLUG
```
leenagupte, line 37: This seems like a neat way to get people straight to the results section. Is there a reason that this change is being limited to one route? Could `results_anchor: "results"` not be appended to locals in all cases?

patrickpatrickpatrick: It could be appended in all cases, it's just that [this](https://www.gov.uk/female-genital-mutilation-help-advice) was the specific page highlighted where this was an issue and I wasn't sure if I should roll it out to all pages.

app/controllers/place_controller.rb
```diff
@@ -27,13 +27,17 @@ def publication
...
     if params[:slug] == REPORT_CHILD_ABUSE_SLUG
-      {
+      locals.merge({
         option_partial: "option_report_child_abuse",
         preposition: "for",
-      }
+      })
     else
-      {}
+      locals
```
leenagupte, line 40: Thanks for making this more generic. I'm sure that there are other services that will benefit from this anchor.

A minor point: you don't need the `else`, you can just return `locals` in the last line of the method. This may mean the `merge` will need to change to a `merge!` though.

app/views/application/_location_form.html.erb
```diff
@@ -22,7 +24,12 @@
     } %>
   <% end %>
 
-  <form method="post" id="local-locator-form" class="location-form govuk-!-margin-bottom-9">
+  <%= content_tag(:form,
```
leenagupte, line 27: Can you add something to the commit message about why this has changed to a `content_tag`? For example, as this is a form, why wasn't a `form_with` used?

app/views/place/show.html.erb
```diff
@@ -22,26 +24,31 @@
...
-        <%= render "govuk_publishing_components/components/heading", {
-          text: "Results #{preposition ||= "near"} <strong>#{postcode}</strong>".html_safe,
-          margin_bottom: 4,
-        } %>
-        <ol id="options" class="govuk-list">
-          <%= render partial: option_partial ||= "option", locals: { places: publication.places } %>
-        </ol>
-      <% end %>
-    </section>
+    <%= content_tag "section", id: results_anchor, class: "places-results",
```
leenagupte, line 35: Again here, why was this html `section` replaced with a `content_tag`? There is another `section` in this view that hasn't been changed, so now the view is inconsistent.

I think it might be clearer if either: 
1. both sections remain as they are with only the `results_anchor` id being added to this one, or 
2. both sections are changed to a `content_tag` in a separate commit that explains why.

leenagupte (commented):
I've tested this locally on both "/report-child-abuse-to-local-council" and "/female-genital-mutilation-help-advice" and it works really well. 🎉 

I've left a few recommendations in the inline comments.

app/views/place/show.html.erb
patrickpatrickpatrick: I wasn't sure if there was another alternative to writing the conditional attributes of the elements in the template other than this

`<form <% if results_anchor %> id="<%= results_anchor %> <% end %> `

which is why I changed it to content_tag as that handles the case if an attribute is nil a bit more elegantly. Unless the snippet above is OK in which case can revert both instances of `content_tag` to that.

app/controllers/place_controller.rb
patrickpatrickpatrick: ah yeah that makes more sense!

app/views/place/show.html.erb
leenagupte: Ok, that makes sense. Can you add that to the commit message?

leenagupte: I think `[form_with](https://api.rubyonrails.org/v7.0.3/classes/ActionView/Helpers/FormHelper.html#method-i-form_with)` allows an option `id` attribute as well.

leenagupte (approved):
Thanks for updating the commit message 🎉 
LGTM 👍

PR #3263, opened 2022-06-20
Add anchor link to location form on error

patrickpatrickpatrick:
## What

When there is an error on location form submit, anchor link to where the error message is.

## Why

On pages with long content before the location form, if submitting the form resulted in an error then it was not obvious an error had occurred as the page scroll would be reset to the top of the page. 

Related to https://github.com/alphagov/frontend/pull/3254


MartinJJones (approved):
Looks good to me, tested locally and the error message is scrolled into view 👍

PR #3267, opened 2022-06-24
GUA-72: Redirect /account/home to Digital Identity

alex9smith:
We are moving all central account features off GOV.UK and into a DI
owned app to better separate out concerns between DI (the ID provider)
and [`www.gov.uk`](http://www.gov.uk/) (a relying party).

The main page that needs to move is this one. We want to show users
which services they've used on their account home page, which isn't
possible if it doesn't sit within DI where it can access that usage
data.

Redirect the route that this app serves the account home on to a DI
domain. This is done at the app level (rather than publishing a
redirect content item with Router) so that it can do environment-aware
redirects using `GovukPersonalisation`.

Note: this implementation keeps the controller with a single `show`
method which does the redirect. I think it would be more 'Railsy' to
do the redirect in `routes.rb`, but when I tried that the application
kept trying to follow the redirect and render the path from the
destination URL (`/manage-your-account`) as a content item when tested.

Keeping the controller seemed a more pragmatic solution compared to
messing with the 'catch-all' routing fallbacks.

---

DO NOT MERGE this PR until we are ready to remove the account home. We need to also update email-alert-frontend and the DI account management app, and release all three changes at (roughly) the same time.

---

[Jira](https://govukverify.atlassian.net/browse/GUA-72)


huwd:
I think this looks good for the moment.
Your PR description is great and got me thinking... Due to the microservice landscape of GOV.UK routes only "own" the paths that have been assigned to them by router...

So I think your correct about a router.rb being the rails-y way...

But perhaps the GOV.UK-y way is to intervene above this app (eventually?).
It made me remember there is this stuff over here:
[Special Route publisher](https://github.com/alphagov/special-route-publisher/blob/74107527d7cc9db7595f100030c01bc3111620d3/data/special_routes.yaml#L23-L26)

Which i guess we'll want to keep in place in order to get the user _to_ this app _to_ be redirected to the IDP 🤔

app/controllers/account_home_controller.rb
```diff
@@ -1,33 +1,5 @@
...
-  def has_email_subscriptions?
-    subscriber = GdsApi.email_alert_api.authenticate_subscriber_by_govuk_account(
-      govuk_account_session: @account_session_header,
-    ).to_h.fetch("subscriber")
-
-    subscriptions = GdsApi.email_alert_api.get_subscriptions(id: subscriber.fetch("id")).to_h.fetch("subscriptions", [])
-    !subscriptions.empty?
-  rescue GdsApi::HTTPNotFound
-    false
+    redirect_to(GovukPersonalisation::Urls.manage, allow_other_host: true)
```
chao-xian, line 3: So this redirect probably shouldn't be at the frontend rendering app level...

alex9smith:
> But perhaps the GOV.UK-y way is to intervene above this app 

In the case where we're redirecting a piece of content, or even something complicated like a smart answer, which is only used by the public I completely agree that having the redirect at the Router is the correct thing to do.

However, I think this case is more complicated and from what I understand about Router and special routes (admittedly not much!) it can't handle what I think we need to do. This confusion is probably my fault as I put the phrase 'environment-aware redirects' in the PR description and then didn't explain what I meant.

GOV.UK's integration and staging environments use Auth integration environment as the IDP, and GOV.UK production uses Auth's production. I think it's important that we implement this redirect in a way that maintains this relationship because the way the account behaves in GOV.UK's pre-production environments in a way that's different to content pages. If a link on a guidance page in integration takes you to a production campaigns site then that's not a big deal, but we have to step through journeys using the account all the time in integration and staging while we're developing features that use it. It'll be really confusing to be working on something in integration, click a link and suddenly end up in production. 

I think we could set up different redirects in integration and production, but I'm pretty sure the Router database in staging is overwritten by the production database every night by the backup jobs. That would make it impossible to maintain the relationship between GOV.UK staging and Auth integration.

Implementing the redirect at the app level allows us to use `GovukPersonalisation` to provide different redirect URLs to each environment and therefore keep the sign-in relationships.

I suppose what I'm advocating for is to introduce a bit of tech debt in order to make the development and testing experience better, and when I phrase it like that it's less convincing. So I'm open to being convinced that we should put this redirect in Router instead.

app/controllers/account_home_controller.rb
huwd: Richard has raised a good point about GA and us not breaking the cookie banners, which is an interesting one. I think we need to modify this to [redirect_with_ga](https://github.com/alphagov/govuk_personalisation/blob/main/lib/govuk_personalisation/controller_concern.rb#L138-L140)

huwd:
So... assuming I've got that correct about lack of GA support, and whatever happens with DB state...

... my vote is that we change the redirect param to the `govuk_personalisation` one above, then document this as tech debt and link to that trello card in the commit message?

huwd (changes requested):
so [redirect_with_analytics](https://github.com/alphagov/govuk_personalisation/blob/main/lib/govuk_personalisation/controller_concern.rb#L138-L140), and perhaps an additional test to check that those params get passed (happy state only?)... 🤔 that might be doubling up with govuk_personalisation, up to you on the test?

chao-xian:
@huwd the "database" for redirects is publishing-api, and as they are sync'd _I think_ they are also modified to reflect their environment to cover your use case above. [I think it's this](https://github.com/alphagov/govuk-puppet/blob/main/modules/govuk_env_sync/files/govuk_env_sync.sh#L72)?

huwd:
Change looks good, did you want to fixup the commit?

huwd (approved):

PR #3289, opened 2022-07-12
Add ability to lock version if manual deploy

fredericfran-gds:
Ref:
1. [trello card](https://trello.com/c/5uj2U5rS/900-prevent-manual-deploys-being-overridden-by-cd)

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -17,6 +17,11 @@ on:
         - staging
         - production
         default: 'integration'
+      lockIfManualDeploy:
+        description: 'Lock application to this version if manually deployed'
+        required: true
+        type: boolean
+        default: true
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -17,6 +17,11 @@ on:
         - staging
         - production
         default: 'integration'
+      lockIfManualDeploy:
+        description: 'Lock application to this version if manually deployed'
+        required: true
+        type: boolean
+        default: true
```
theseanything, line 24: Nit: Wondering if it'd be safer to have the default to be false. The risk with this feature is that someone manually deploys and then "forgets" to re-enable continuous deployment. If it were false then the person would to explicitly disable it, and which would act as a reminder/check that they understand that CD will be disabled.

theseanything (approved):

fredericfran-gds:
susperseded

PR #3339, opened 2022-08-25
Remove `n/a` attributes on tab components

gclssvglx:
## What

Remove all attributes that are set to `n/a` for all tab component instances.

## Why

We have a [shared schema](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-schemas.js) where all attributes are initially set to `null`. This means we only need to set the attributes that are needed and have a value when required prior to pushing the event. All other attributes will be sent as `null` by default, so we no longer need to set them on purpose.

### DIFF
--- a/app/views/calendar/bank_holidays.html.erb
+++ b/app/views/calendar/bank_holidays.html.erb
@@ -82,8 +82,6 @@
                   text: t(division.title),
                   index: index + 1,
                   index_total: @calendar.divisions.length,
-                  section: 'n/a',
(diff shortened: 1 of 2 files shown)


AshGDS (approved):
All good - thanks @gclssvglx :+1:

AgaDufrat:
This has been open and approved for a while. Do we still want to merge it?

PR #3342, opened 2022-08-30
Set up individual component CSS loading

injms:
## What

Update `frontend` to load component and view CSS only when required by the page being viewed - so the CSS for the radio buttons only loads on pages with radio buttons, and the stylesheet for the homepage is only loaded on the homepage.

## Why

This reduces the amount of CSS required for each page and increases the ability of a browser to cache the stylesheets - this should mean a faster load time for both first time and repeat visitors.

See [RFC #149][rfc] for more details

## Visual changes

None.

---

[rfc]: https://github.com/alphagov/govuk-rfcs/pull/152

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -3,3 +3,17 @@
 //= link application.js
 //= link test-dependencies.js
 //= link application.css
+
+//= link components/_calendar.css
+//= link components/_metadata.css
+//= link components/_subscribe.css
+
+//= link views/_calendars.css
+//= link views/_completed_transaction.css
+//= link views/_cookie-settings.css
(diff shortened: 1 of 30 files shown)


KludgeKML (approved):
Looks good to me.

jon-kirwan (approved):
Looks great to me too and approved 👍 (I wonder if we should create a separate ticket to add an integration test? We have one in the dummy app for publishing components but might be a good idea to include one here too?)

MartinJJones:
> I wonder if we should create a separate ticket to add an integration test? We have one in the dummy app for publishing components but might be a good idea to include one here too?

Thanks, good point, I've just created a separate ticket for this 👍

PR #3370, opened 2022-09-10
Add bank holiday for funeral of Elizabeth II

intUnderflow:
This is per the declaration made in the Accession Council today


richardTowers:
Thanks for this :) We had to wait for the official announcement of the date of the funeral.

Closing as superceded by #3371 (which also has slightly different wording).

PR #3414, opened 2022-10-26
started process of removing gem and updating test files

kashifatcha:

https://trello.com/c/b9PRFqgU/231-applications-use-govuk-content-schema-test-helpers-which-is-an-archived-gem


test/test_helper.rb
```diff
@@ -14,8 +14,8 @@
 require "webmock/minitest"
 WebMock.disable_net_connect!(allow_localhost: true)
 require "timecop"
-
-require "govuk-content-schema-test-helpers"
+require "govuk_schemas/example"
+require "govuk_schemas/validator"
```
hannako, line 18: I don't think we need these two `require` statements.

- Looks like the gem already includes the `Example` class, see [here](https://github.com/alphagov/govuk_schemas/blob/64edf7e5204343c9976ce00808d6a16960975967/lib/govuk_schemas.rb#L5), so all of those [GovukSchemas::Example.find("foo", example_name: "bar")](https://github.com/alphagov/frontend/pull/3414/files#diff-08b3e147ea674189ef67a2ce266bf184a0d570a30a8484f095598c41173d98ceR13) will still work without it.
- The `Validator` class is needed for the `assert_valid_against...` methods, eg [here](https://github.com/alphagov/govuk_schemas/blob/64edf7e5204343c9976ce00808d6a16960975967/lib/govuk_schemas.rb#L5). Those methods are defined in the gem [here](https://github.com/alphagov/govuk_schemas/blob/64edf7e5204343c9976ce00808d6a16960975967/lib/govuk_schemas/assert_matchers.rb), which you can see is requiring the validator class. Ie as we've got ` require "govuk_schemas/assert_matchers"` on all the test files that are using those matcher methods, we get have the validator code for free :)

hannako (commented):
Hi Kashif, this is looking good! Do you want to just remove those extra lines (see comments) and then I will approve?

hannako (approved):

PR #3434, opened 2022-11-10
Remove text from completed transaction page "organ donor" promotion

jon-kirwan:
## What

https://trello.com/c/HLmurT4f/1536-add-new-designs-onto-organ-donation-pages

Remove text from completed transaction page "organ donor" promotion.

## Why
- You do not have to join the register as you’re automatically opted in now. 
- The second link should be removed because it takes away from the real call to action, which is the register a decision to donate 
- Removing the rhetorical question, because this language is quite odd in this context

## Visual changes
- [Give feedback on Book your driving test - organ donor promotion](https://govuk-frontend-app-pr-3434.herokuapp.com/done/book-driving-test)

### DIFF
--- a/app/views/completed_transaction/show.html.erb
+++ b/app/views/completed_transaction/show.html.erb
@@ -17,15 +17,6 @@
         } %>
 
         <p class="govuk-body">
-          <%= t('formats.transaction.organ_donation_laws') %> - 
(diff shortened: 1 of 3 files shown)


MartinJJones (approved):
Looks good to me 👍

PR #3446, opened 2022-11-23
Add support for choosing from a split postcode.

KludgeKML:

## What

Add support for choosing an address to disambiguate a postcode when looking for a council.

## Why

Some postcodes fall into more than one council, and we then need to choose from a list of addresses to get the exact result. This wasn't previously possible, but the move to locations api away from mapit gives us the required information.

[Support split postcodes in Frontend](https://trello.com/c/HjgjrOCq/1580-support-split-postcodes-in-frontend)

## How

When a postcode is initially checked, we determine whether there is more than one unique local custodian code involved. If not, we continue as usual. If there is, we re-query locations api to get a list of all addresses at that postcode, and map the LCCs of those individual addresses to authority slugs. We can then present a chooser that allows us to select an address and move directly on to the correct authority slug.

(Note - the current way I'm doing this is very poorly optimised, I will clean that up by caching results to reduce the number of calls to local links manager once it's proven to work)

## Screenshots

District and Borough Results (Before)

Screenshot 2022-12-13 at 14 06 14

District and Borough Results (After)

Screenshot 2022-12-13 at 14 06 21

Single Results (Before)

Screenshot 2022-12-13 at 14 19 00

Single Results (After)

Screenshot 2022-12-13 at 14 17 24

Split postcode disambiguation page (After only)

Screenshot 2022-12-13 at 14 17 13

### DIFF
--- a/app/controllers/find_local_council_controller.rb
+++ b/app/controllers/find_local_council_controller.rb
@@ -17,7 +17,10 @@ class FindLocalCouncilController < ContentItemsController
       return
     end
 
-    redirect_to "#{BASE_PATH}/#{authority_slug}"
+    return redirect_to "#{BASE_PATH}/#{authority_slug_from_lcc(locations_api_response.local_custodian_codes.first)}" if locations_api_response.single_authority?
+
+    @addresses = address_list
+    render :multiple_authorities
   end
 
   def result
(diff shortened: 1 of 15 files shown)


KludgeKML:
@jon-kirwan the branch of gds-api-adapters used has made it into the main branch (at release 85.0.0) of gds-api-adaptors, so we can update the gemfile/lock - do you want me to do that?

jon-kirwan:
> @jon-kirwan the branch of gds-api-adapters used has made it into the main branch (at release 85.0.0) of gds-api-adaptors, so we can update the gemfile/lock - do you want me to do that?

Updated here: https://github.com/alphagov/frontend/pull/3446/commits/2bf38d4f1a9b8ebe8258ff4c71091bcdac58cd63

MartinJJones (approved):
Looks good to me 👍

app/controllers/find_local_council_controller.rb
```diff
@@ -54,15 +57,37 @@ def locations_api_response
     @locations_api_response ||= fetch_location(postcode)
   end
 
+  def single_authority_from_location_api?
+    locations_api_response.local_custodian_codes == 1
+  end
+
+  def address_list
+    @address_list ||= build_addresses(postcode)
```
edwardkerry, line 70: Do we need to pass `postcode` through here? it looks like it's only used to `fetch_addresses`, and `build_addresses` is able to access postcode` directly anyway?

edwardkerry (approved):
Looks good to me, just one (non-blocking) query

PR #3465, opened 2022-12-07
Tweak License unavailable view for accessibility

KludgeKML:

## What

Remove the negative contraction  and add a full stop.

## Why

Negative contractions hard for some people to read and often misintepreted

### Before

Before

### After

After


1pretz1 (approved):
Looks good. Feels like we should localise the text as it's pretty gross how many times we need to update it following a content change, not a blocker though.

PR #3482, opened 2022-12-21
Ga4 homepage tracking

JamesCGDS:

## What
Adds GA4 tracking to the homepage links using the `ga4-link-tracker`

## Why
It is next on our list of interactions to add.

[Trello card](https://trello.com/c/WAST3s6v/392-add-tracking-homepage)

## Visual changes
N/A

### DIFF
--- a/app/views/homepage/_chevron_card.html.erb
+++ b/app/views/homepage/_chevron_card.html.erb
@@ -8,6 +8,13 @@
     track_dimension: raw(title),
     track_label: link,
     track_value: 1,
+    ga4_link: {
+      'event_name': 'navigation',
+      'type': 'home page',
+      'index': index,
+      'index_total': index_total,
+      'section': section
+    }
   }
 
   link = capture do
--- a/app/views/homepage/_government_activity.html.erb
+++ b/app/views/homepage/_government_activity.html.erb
@@ -1,6 +1,6 @@
 <section class="homepage-section homepage-section__government-activity">
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-full govuk-grid-column-two-thirds-from-desktop" data-module="gem-track-click">
+    <div class="govuk-grid-column-full govuk-grid-column-two-thirds-from-desktop" data-module="gem-track-click ga4-link-tracker">
       <div class="homepage-section__heading">
         <div class="homepage-section__heading-wrapper">
           <%= render "govuk_publishing_components/components/heading", {
@@ -16,12 +16,15 @@
       </div>
 
       <ul class="homepage-services-and-info__list">
-        <% t("homepage.government_activity").each do | item | %>
+        <% t("homepage.government_activity").each_with_index do | item, index | %>
           <%= render partial: "homepage/chevron_card", locals: {
             description: item[:description],
             link: item[:link],
             title: item[:title],
             track_action: "governmentactivityLink",
+            index: "3.#{index + 1}",
+            index_total: ga4_links_count,
+            section: t("homepage.index.government_activity", locale: :en)
           } %>
         <% end %>
       </ul>
@@ -34,8 +37,15 @@
           text: t("homepage.index.departments_and_organisations"),
         } %>
       </div>
+
+      <%# 
+        The "big number" component links below are hardcoded separately (i.e. not in a loop in this instance) and therefore their 'index' properties are also
+        hardcoded. Additionally, in order to account for these hardcoded components in the 'index_total' property, 2 is added to the ga4_links_counts variable 
+        (which is defined in homepage > index.html.erb). If a new "big number" component is added or removed, this will need to be reflected in the 'index' and
+        'index_total' properties to ensure that they remain accurate for tracking/analytics purposes.
+      %>
       
-      <div class="homepage-section__depts" data-module="gem-track-click">
+      <div class="homepage-section__depts" data-module="gem-track-click ga4-link-tracker">
         <%= render "govuk_publishing_components/components/big_number", {
           number: t("homepage.index.ministerial_departments_count"),
           label: t("homepage.index.ministerial_departments"),
@@ -48,6 +58,14 @@
             "track-dimension": "#{t("homepage.index.ministerial_departments_count")} #{t("homepage.index.ministerial_departments")}",
             "track-dimension-index": 29,
             "track-value": 1,
+            "ga4_link": {
+              "event_name": "navigation",
+              "type": "home page",
+              "index": 4.1,
+              "index_total": ga4_links_count,
+              "section": t("homepage.index.departments_and_organisations", locale: :en),
+              "text": "#{t("homepage.index.ministerial_departments_count")} #{t("homepage.index.ministerial_departments", locale: :en)}"
+            }
           },
         } %>
         <%= render "govuk_publishing_components/components/big_number", {
@@ -62,6 +80,14 @@
             "track-dimension": "#{t("homepage.index.other_agencies_count")} #{t("homepage.index.other_agencies")}",
             "track-dimension-index": 29,
             "track-value": 1,
+            "ga4_link": {
+              "event_name": "navigation",
+              "type": "home page",
+              "index": 4.2,
+              "index_total": ga4_links_count,
+              "section": t("homepage.index.departments_and_organisations", locale: :en),
+              "text": "#{t("homepage.index.other_agencies_count")} #{t("homepage.index.other_agencies", locale: :en)}"
+            }
           },
         } %>
       </div>
--- a/app/views/homepage/_links_and_search.html.erb
+++ b/app/views/homepage/_links_and_search.html.erb
@@ -7,8 +7,8 @@
           margin_bottom: 4,
           text: t("homepage.index.popular_links_heading"),
         } %>
-        <ul class="homepage-most-viewed-list" data-module="gem-track-click">
-          <% t("homepage.index.popular_links").each do | item | %>
+        <ul class="homepage-most-viewed-list" data-module="gem-track-click ga4-link-tracker">
+          <% t("homepage.index.popular_links").each_with_index do | item, index | %>
             <li>
               <%= link_to(item[:text], item[:href], {
                 class: "govuk-link homepage-most-viewed-list__item",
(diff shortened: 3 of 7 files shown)


app/views/homepage/_links_and_search.html.erb
```diff
@@ -19,6 +19,13 @@
                   track_value: 1,
                   track_dimension_index: 29,
                   track_dimension: item[:text],
+                  ga4_link: {
+                    'event_name': 'navigation',
+                    'type': 'home page',
+                    'index': "#{index}",
+                    'index_total': "#{t("homepage.index.popular_links").length}",
```
andysellick, line 26: Could this value be pre-calculated rather than once through every pass of the loop?

app/views/homepage/_chevron_card.html.erb
```diff
@@ -8,6 +8,13 @@
     track_dimension: raw(title),
     track_label: link,
     track_value: 1,
+    ga4_link: {
+      'event_name': 'navigation',
+      'type': 'home page',
+      'index': index ||= nil,
+      'index_total': index_total ||= nil,
+      'section': section ||= nil
```
andysellick, line 16: Just curious, don't really understand how this fits together, but these values default to `nil` if they're not available - is that possible? Is that bad?

app/views/homepage/_services_and_information.html.erb
```diff
@@ -10,13 +10,16 @@
...
-        <% t("homepage.categories").each do | item | %>
+      <ul class="homepage-services-and-info__list" data-module="gem-track-click ga4-link-tracker">
+        <% t("homepage.categories").each_with_index do | item, index | %>
           <%= render partial: "homepage/chevron_card", locals: {
             description: item[:description],
             link: item[:link],
             title: item[:title],
             track_action: "topicsLink",
+            index: index,
+            index_total: t("homepage.categories").length,
```
andysellick, line 21: Could this be pre-calculated?

app/views/homepage/_government_activity.html.erb
```diff
@@ -16,12 +16,15 @@
...
       <ul class="homepage-services-and-info__list">
-        <% t("homepage.government_activity").each do | item | %>
+        <% t("homepage.government_activity").each_with_index do | item, index | %>
           <%= render partial: "homepage/chevron_card", locals: {
             description: item[:description],
             link: item[:link],
             title: item[:title],
             track_action: "governmentactivityLink",
+            index: index,
+            index_total: t("homepage.government_activity").length,
```
andysellick, line 26: Could this be pre-calculated?

app/views/homepage/_government_activity.html.erb
```diff
@@ -51,6 +51,9 @@
             "track-dimension": "#{t("homepage.index.ministerial_departments_count")} #{t("homepage.index.ministerial_departments")}",
             "track-dimension-index": 29,
             "track-value": 1,
+            "index": 0,
+            "index_total": 2,
+            "section": t("homepage.index.departments_and_organisations")
```
andysellick: I'm confused here because looking at the code I don't think this should work, but looking at the deployment it appears to. I'd thought that the GA4 tracking attributes would need to be included in an object and then added here something like:

```
"ga4_link": { "attributes" }
```

Am I looking at the wrong thing? Or is it just Friday afternoon and I'm slipping?

app/views/homepage/_government_activity.html.erb
```diff
@@ -51,9 +51,14 @@
...
             "track-value": 1,
-            "index": 0,
-            "index_total": 2,
-            "section": t("homepage.index.departments_and_organisations")
+            "ga4_link": {
+              "event_name": "navigation",
+              "type": "home page",
+              "index": 0,
+              "index_total": 2,
+              "section": t("homepage.index.departments_and_organisations"),
```
andysellick, line 59: We had a thing on some other tracking where it was initially using the translation text but we realised that we'd get different data for the same page in different translations, which was undesirable in that situation. I think in the end we had to hard code the text we needed in the markup (it was the feedback component, I remember now) but a better approach is to call the translation but always force the English version - assuming that's appropriate in this situation.

app/views/homepage/_promotion-slots.html.erb
```diff
@@ -22,6 +22,13 @@
             track_value: "1",
             track_dimension_index: "29",
             track_dimension: item[:title],
+            ga4_link: {
+              event_name: 'navigation',
+              type: 'home page',
+              index: index,
+              index_total: t("homepage.index.promotion_slots").length,
```
andysellick, line 29: Pre-calculate?

app/views/homepage/_more_on_govuk.html.erb
```diff
@@ -20,6 +20,13 @@
             "track-value": "1",
             "track-dimension": item[:title],
             "track-dimension-index": "29",
+            "ga4_link": {
+              'event_name': 'navigation',
+              'type': 'home page',
+              "index": index,
+              "index_total": t("homepage.index.more_links").length,
```
andysellick, line 27: Pre-calculate?

app/views/homepage/_links_and_search.html.erb
JamesCGDS: Thank you good spot, have updated

app/views/homepage/_services_and_information.html.erb
JamesCGDS: Thank you good spot, have updated

app/views/homepage/_government_activity.html.erb
JamesCGDS: Thank you good spot, have updated

app/views/homepage/_more_on_govuk.html.erb
JamesCGDS: Thank you good spot, have updated

app/views/homepage/_chevron_card.html.erb
JamesCGDS: I don't think it's possible for these values to not be available i.e. they wouldn't ever be `nil`. However if I just have it as `'index:' index` (i.e. without `||= nil`), then I get an error when the component is rendered. I think I wrote this just before I left for Christmas and my brain might've been a bit foggy so I can have a poke around to see if there's a better way to do this

app/views/homepage/_government_activity.html.erb
JamesCGDS: Thank you good spot, have updated

app/views/homepage/_government_activity.html.erb
JamesCGDS: Yeah, this is incorrect but I think I addressed it in a later commit (apologies for the confusion). There is an instance in this template where I didn't address this as the component receiving the values handled building the `"ga4-link"` attribute object separately

JamesCGDS:
Thanks for your feedback @andysellick 👍  I've just noticed that the `index` property needs to be a decimal rather than an integer so I just need to make that change

app/views/homepage/_chevron_card.html.erb
JamesCGDS: I’ve had a look around and I think this is just the way to prevent the `undefined local variable or method` error when rendering the template. It looks like the same approach has been adopted in the components gem (e.g. [back link](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_back_link.html.erb), [big number](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_big_number.html.erb), [cards](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_cards.html.erb)) but let me know if you'd like me to look more into it.

andysellick: I've just tried removing the `||= nil` and couldn't get an error to occur. As far as I can see the chevron card is only used on the homepage and isn't called from anywhere else, so we should have complete control over the context in which it's used. 

Yes, using this approach means that if the index etc. isn't passed, an error doesn't occur, but I'd caution against that  - it means that a careless developer could remove those attributes from the call and the analytics would silently fail to report full information. If we force an error when the thing we need isn't passed, that feels like a better protector of the functionality we're adding.

app/views/homepage/_government_activity.html.erb
```diff
@@ -34,8 +37,15 @@
           text: t("homepage.index.departments_and_organisations"),
         } %>
       </div>
+
+      <%# 
+        The "big number" component links below are hardcoded separately (i.e. not in a loop in this instance) and therefore their 'index' properties are also
+        hardcoded. Additionally, in order to account for these hardcoded components in the 'index_total' property, 2 is added to the ga4_links_counts variable
```
andysellick, line 43: Where is 2 added to the `ga4_links_counts` variable? I'm not sure I understand this comment.

andysellick (commented):
Looking good, just a couple of outstanding questions.

app/views/homepage/_government_activity.html.erb
JamesCGDS: It's added in `index.html.erb` - if you'd like to hop on a quick call for me to explain just give me a shout, it's a bit tricky to explain over text

app/views/homepage/_chevron_card.html.erb
JamesCGDS: Cool, I've removed `|||= nil` and it's working for me too now. Apologies for the delay, I was figuring out what was causing the error previously

app/views/homepage/index.html.erb
```diff
@@ -1,3 +1,20 @@
...
+  (please see the _government_activity.html.erb template for further details)
+%>
+
+<% ga4_links_count = 
+  t("homepage.index.popular_links").length + 
+  t("homepage.categories").length + 
+  t("homepage.government_activity").length + 
+  t("homepage.index.more_links").length + 
+  t("homepage.index.promotion_slots").length +
+  2
```
andysellick, line 15: Oh there's the two! I see it now. Didn't see it because it was on a line by itself. Can I suggest a slight reformat?

```Ruby
<% ga4_links_count = 
  t("homepage.index.popular_links").length + 
  t("homepage.categories").length + 
  t("homepage.government_activity").length + 
  t("homepage.index.more_links").length + 
  t("homepage.index.promotion_slots").length + 2 # added to account for the two hardcoded big number links
%>
```

andysellick (approved):
One teeeeny little thing, but otherwise looks good 👍

app/views/homepage/index.html.erb
JamesCGDS: Good shout, thanks! I'll make that change and squash the commits together

PR #3493, opened 2023-01-04
Revert "Banner research panel"

chao-xian:
Reverts alphagov/frontend#3445 to remove the banner on /bank-holidays

The sign ups are more than enough on the linked to studies.

See also https://github.com/alphagov/collections/pull/3111

[Trello](https://trello.com/c/FFHRYpBe/1542-take-down-panel-banner-m), [Jira issue NAV-5412](https://gov-uk.atlassian.net/browse/NAV-5412)


hannako (changes requested):
We still have some recruitment banners on pages rendered by frontend (https://github.com/alphagov/frontend/pull/3472). I think instead of a revert we'll just need to remove from bank hol page, otherwise we'll lose the component and the styling.

chao-xian:
@hannako  Oh damn didn't spot that - ok I'll rework this. Thanks!

PR #3504, opened 2023-01-12
Split postcode local transactions

KludgeKML:

## What

Adds support for split postcodes to the local transaction controller so that people whose postcodes sit across multiple local authorities can be prompted to enter their address and get the correct authority returned.

## Why

We can now support split postcodes, and since we've successfully added them to the find your council endpoint, we need to add them to other endpoints with the same potential problems.

[Trello card](https://trello.com/c/2OjOnhhR/1743-add-split-postcode-support-for-local-transactions-in-frontend)

## How

Copying the work from here: [alphagov/frontend: Pull Request 3446](https://github.com/alphagov/frontend/pull/3446) we make the same changes in the local transactions controller, so that it can detect whether a postcode is split and add a new disambiguation page.

## Screenshots?

Screenshot 2023-01-12 at 15 56 00


theseanything:
@KludgeKML pact tests are still failing as it's re-running an existing workflow that used the old configuration. Fixed [here](https://github.com/alphagov/frontend/pull/3506). Any future PRs should be fine or if you add new commits this PR, as they'll pick up the new config. For now you can ignore this failing test, as its passed in Jenkins (it shouldn't block you merging)

edwardkerry (approved):
Technically I think this looks fine 👍 
Small nitpick that the last commit correcting a typo could be squashed into previous commit - but not a blocker.

PR #3506, opened 2023-01-13
Fix pact tests clone wrong repo and remove mongodb

theseanything:
This pact workflow should be checking out the frontend repo, not content store. Also doesn't need mongodb to run.


hannako (approved):
LGTM!

PR #3510, opened 2023-01-16
Remove specifying version for mail

chao-xian:
Since the issue requiring that pinning in the first place is fixed https://github.com/mikel/mail/issues/1489


chao-xian:
Better one here https://github.com/alphagov/frontend/pull/3513

PR #3525, opened 2023-01-26
Update Place controller with split postcode support

KludgeKML:

## What

Adds support for split postcodes to the place controller so that people whose postcodes sit across multiple local authorities can be prompted to enter their address and get the correct authority returned if the place service is local_authority.

## Why

We can now support split postcodes, and since we've successfully added them to the find your council and local transactions endpoints, we need to add them to other endpoints with the same potential problems.

[Trello card](https://trello.com/c/C4nLJyVO/1778-add-split-postcode-support-for-places-in-frontend)

## How

Requires an update to how Imminence returns data so that Imminence will return a response with choosable addresses in instead of place results if Imminence detects there are more than one local authorities covered by the postcode and the search is local_authority bounded rather than simply nearest.

This involves breaking changes to Imminence's API and the representation in gds-api-adapters. Because this would otherwise mean that we have to synchronise releases very closely, we've left code here: https://github.com/alphagov/frontend/compare/split-postcode-places?expand=1#diff-57c57e2a173b62de57bc0fdf6a3e755f9b6fc110040a1394ffaf8b6215a5513eR17-R23 to handle the old-style responses from the previous version of imminence. That allows us to deploy frontend first and accept old data until imminence is deployed.

- https://github.com/alphagov/imminence/pull/862
- https://github.com/alphagov/gds-api-adapters/pull/1183

### DIFF
--- a/app/controllers/place_controller.rb
+++ b/app/controllers/place_controller.rb
@@ -10,7 +10,22 @@ class PlaceController < ContentItemsController
   REPORT_CHILD_ABUSE_SLUG = "report-child-abuse-to-local-council".freeze
 
   def show
-    @location_error = location_error if request.post?
+    render :show, locals:
+  end
+
+  def find
+    @location_error = location_error
+
+    if imminence_response.addresses_returned?
+      @options = imminence_response.addresses.each.map do |address|
+        { text: address["address"], value: address["local_authority_slug"] }
+      end
+      @change_path = place_path(slug: params[:slug])
+      @onward_path = place_find_path(slug: params[:slug], anchor: "results")
+      @postcode = postcode
+      return render :multiple_authorities
+    end
+
     render :show, locals:
   end
 
@@ -49,6 +64,10 @@ private
     PostcodeSanitizer.sanitize(params[:postcode])
   end
 
+  def local_authority_slug
+    params[:local_authority_slug]
+  end
+
   def location_error
     return LocationError.new(INVALID_POSTCODE) if imminence_response.invalid_postcode? || imminence_response.blank_postcode?
     return LocationError.new(NO_LOCATION) if imminence_response.places_not_found?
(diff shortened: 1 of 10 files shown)


app/controllers/place_controller.rb
```diff
@@ -61,11 +80,11 @@ def imminence_response
   def places_from_imminence
     if postcode.present?
       begin
-        places = Frontend.imminence_api.places_for_postcode(content_item_hash["details"]["place_type"], postcode, Frontend::IMMINENCE_QUERY_LIMIT)
+        imminence_data = Frontend.imminence_api.places_for_postcode(content_item_hash["details"]["place_type"], postcode, Frontend::IMMINENCE_QUERY_LIMIT, local_authority_slug)
```
1pretz1, line 83: Could we make this multiline?

app/models/imminence_response.rb
```diff
@@ -2,20 +2,42 @@ class ImminenceResponse
...
 
-  def initialize(postcode, places, error)
+  def initialize(postcode, imminence_response, error)
     @postcode = postcode
-    @places = places
     @error = error
+    @places = []
+    @addresses = []
+
+    return if @error
```
1pretz1, line 13: Could we move this logic out of the initializer? Seems like it could live in the [place_controller](https://github.com/alphagov/frontend/pull/3525/files#diff-5d900c996c47b6c84232c18ba346309a47450e5bacdb2d2494b106c244cf9c58R88) and pass the variables `places and `addresses` in

1pretz1 (commented):
Nice work Keith, added a couple of small comments.

app/views/place/multiple_authorities.html.erb
```diff
@@ -0,0 +1,38 @@
+<%= render layout: "shared/base_page", locals: {
+  title: publication.title,
+  publication: publication,
+  edition: @edition,
+} do %>
+  <p class="govuk-body">
+    <strong><%= @postcode %></strong> <%= link_to t("formats.place.change"), @change_path, class: "govuk-link" %>
+  </p>
```
jon-kirwan, line 8: I notice that the postcode doesn't seem to display here (for when I tested https://www.integration.publishing.service.gov.uk/find-your-local-councillors with "WV14 8TU")

Screenshot 2023-02-17 at 11 05 24

jon-kirwan (commented):
Looking really good. I've added one issue below.

Also, with the previous changes on "Find your local council" - we included a title for "Postcode". Maybe we should do the same here:

```
<%= render layout: "shared/base_page", locals: {
  title: publication.title,
  publication: publication,
  edition: @edition,
} do %>
    text: t("formats.local_transaction.postcode"),
    heading_level: 2,
    margin_bottom: 2
  } %>
  
  <p class="govuk-body">
    <strong><%= @postcode %></strong> <%= link_to t("formats.place.change"), @change_path, class: "govuk-link" %>
  </p>
```

app/views/place/multiple_authorities.html.erb
KludgeKML: Okay, I've added this to places. If you go to https://www.integration.publishing.service.gov.uk/register-offices and use that same postcode, you should see the title/postcode displayed.

(Note that find your local councillors is another location-based finder, which I haven't yet added the title to - there's another one that needs to be updated too. I will do that in another card, just so we keep this PR places-based)

jon-kirwan (approved):
Thanks for the changes @KludgeKML - all looks good to me

PR #3542, opened 2023-02-13
Form Error Handling Changes

patrickpatrickpatrick:

## What

Validation for postcode for the location form on /contact-electoral-registration-office will now generate errors for an empty postcode (but only if the user has submitted an empty postcode and not on initial page load).

/contact-electoral-registration-office now uses the shared location form

All pages will now change the title of the to "Error: Title of the Page" in the event of an error on the submitted form

Simple Smart Answer will now display an error summary in addition to the error message in the form itself

## Why

In an accessibility audit, some issues with validation errors on the location form were raised. After some investigation it was found that

- The error messages for /contact-electoral-registration-office were not always rendering
- /contact-electoral-registration-office uses a different form from other uses of the location form even though it resembles the shared location form in most ways
- The title of the page would only change for /find-local-council when a form errored
- No logic for adding error summary to the Simple Smart Answer forms existed

Here we fix the issue with the error validation and also address the tech debt by making /contact-electoral-registration-office use the same shared location form. We also update the shared `base_page` partial so that it will display an error message on all location forms.

## How

The previous check of `postcode.present?` would always be true because postcode is an object that is always created by `ElectionPostcode.new(postcode_params)` so it is therefore always present. By checking to see if the `postcode_params` are nil we can check to see if there has been a postcode entered at all.

[Trello card](https://trello.com/c/oU9gNPFM/1634-error-message-components-do-not-conform-with-govuk-design-system-guidelines-675-l), [Jira issue NAV-8299](https://gov-uk.atlassian.net/browse/NAV-8299)


patrickpatrickpatrick:
Tested on integration. The location forms all work correctly after this merger of concerns.

patrickpatrickpatrick:
"find_local_council" still uses separate routes and controllers to the rest of the location forms despite the functionality being similar. Could investigate how to also merge this but perhaps this is beyond the scope of this PR.

hannalaakso (commented):
Looks like a nice tidy up and I can see that we're now appending 'Error' to the title on all the pages 🎉 

I haven't been able to check whether electoral office results page `/contact-electoral-registration-office?postcode=SW1A+2AA` is still rendering correctly with a valid postcode but I don't why it wouldn't and it sounds like you checked this on integration already @patrickpatrickpatrick .

There are a couple of divergences from the Design System error summary pattern on some of the other pages:
- `/landlord-immigration-check/y` is missing the error summary
- The pages apart from the electoral office one have the error summary squished right up against the subheading below it, I think we need a bit of margin there to separate the two. It might be good to do it as part of this ticket even thought it wasn't flagged on the original card.
Screenshot 2023-02-20 at 15 40 27

patrickpatrickpatrick:
https://github.com/alphagov/frontend/pull/3545 connected issue, where I consolidate another partial that had duplicated functionality

hannalaakso (approved):
Thanks for addressing my comments 👍

PR #3545, opened 2023-02-15
Find Local Council use shared base_page partial

patrickpatrickpatrick:

## What

Remove the base page for Find Local Council. Find Local Council now uses the shared base_page.

## Why

Found this while implementing changes to location forms on frontend. All pages on frontend use a 'base_page' partial but the Find Local Council had a different 'base_page'. It was found that the shared base_page could be used with some small changes to the find_local_council partial.


chao-xian:
Pls attach related PRs and Trello card if so...

PR #3556, opened 2023-03-06
Fix frontend failing with empty postcode field in places

KludgeKML:

## What

Make the imminence_response_from_data method quick-return if no postcode is supplied

## Why

Currently page is 500-ing when the user presses find without entering a postcode. This isn't really causing serious problems for users (because they can just back-button and enter a postcode), but it's not ideal and it's causing sentry noise.

[Trello card](https://trello.com/c/QcVV7Wee/1903-ps-7-fix-empty-imminence-response-bug-in-frontend)

## How

Add postcode.blank? to the guard clause at the top of the method so that it doesn't try to parse missing imminence response data.

## Screenshots?

N/A

### DIFF
--- a/app/controllers/place_controller.rb
+++ b/app/controllers/place_controller.rb
@@ -16,7 +16,7 @@ class PlaceController < ContentItemsController
   def find
     @location_error = location_error
 
-    if imminence_response.addresses_returned?
+    if postcode_provided? && imminence_response.addresses_returned?
       @options = imminence_response.addresses.each.map do |address|
         { text: address["address"], value: address["local_authority_slug"] }
       end
@@ -34,7 +34,7 @@ private
   helper_method :location_error
 
   def publication
-    @publication ||= if request.post? && imminence_response.places_found?
+    @publication ||= if postcode_provided? && imminence_response.places_found?
                        PlacePresenter.new(content_item_hash, imminence_response.places)
                      else
                        PlacePresenter.new(content_item_hash)
@@ -69,7 +69,8 @@ private
   end
 
   def location_error
-    return LocationError.new(INVALID_POSTCODE) if imminence_response.invalid_postcode? || imminence_response.blank_postcode?
+    return LocationError.new(INVALID_POSTCODE) unless postcode_provided?
+    return LocationError.new(INVALID_POSTCODE) if imminence_response.invalid_postcode?
     return LocationError.new(NO_LOCATION) if imminence_response.places_not_found?
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/place_controller.rb
```diff
@@ -94,7 +94,7 @@ def places_from_imminence
   end
 
   def imminence_response_from_data(postcode, imminence_response, error)
-    return ImminenceResponse.new(postcode, [], [], error) if error
+    return ImminenceResponse.new(postcode, [], [], error) if error || postcode.blank?
```
1pretz1, line 97: Could we add the blank postcode logic a little higher in the chain? I think https://github.com/alphagov/frontend/blob/9b362d7879e9e239cb46b46a28ebfe770f978954/app/controllers/place_controller.rb#L72 would be a more suitable place as we're not actually handling an `imminence_response` as the validation occurs before the request.

```
return LocationError.new(INVALID_POSTCODE) if postcode.blank? || imminence_response.invalid_postcode?
```

test/integration/place_test.rb
```diff
@@ -254,6 +254,29 @@ class PlacesTest < ActionDispatch::IntegrationTest
...
+      click_on "Find"
+    end
+
+    should "display error message" do
+      assert page.has_content?("This isn't a valid postcode")
+    end
+
+    should "not show the 'no results' message" do
+      assert page.has_no_content?("We couldn't find any results for this postcode.")
+    end
```
1pretz1, line 269: Do we need this test? Looking at the logic, only one error can be returned and so testing the presence of that one should be suffice, unless I'm misunderstanding something.

1pretz1 (commented):
Nice work! I've left a couple of comments but nothing major. My other though was that It'd be nice to localise these as the keys / text exist in the locale files but as this isn't a pattern in this spec then I think it is out of scope of the PR 👍

test/integration/place_test.rb
KludgeKML: The block was copied from the test below, digging into the blames it looks like in the context of the following test it was added to cover regressions that aren't relevant here (as long as they're covered below), so I've removed it.

app/controllers/place_controller.rb
KludgeKML: @1pretz1 Hmm, I've moved this up, but left it in a separate commit - the way imminence response is used throughout the controller means that a lot of places assume it exists in some form, so I'm not sure the change doesn't do more damage than the original. Have a look, see what you think.

KludgeKML: (It does simplify the code lower down, so it's swings and roundabouts)

1pretz1: Ah yes, good point! I've had a little play based off your work and think the commit below is a reasonable solution as it doesn't really make the code more complicated as we're just moving the conditionals around and addresses the imminence response concern. Also using the `postcode_provided?` method simplifies things further: https://github.com/alphagov/frontend/commit/5b7652f66b820eb5f54c29560891b212661405e7

What do you think?

KludgeKML: I think I like it, works neatly for the two higher up tests. I'll give it one more once-over in the morning, but I think it covers everything. Nice work.

PR #3570, opened 2023-03-24
Update index parameter on homepage links

andysellick:

## What
Updates the data attributes on the homepage for GA4 link tracking, specifically the `index` parameter, which is no longer a string of numbers but several sub attributes that specifically relate to the number of sections, number of links etc.

## Why
Part of the GA4 migration work. Index parameters were inconsistent and hard to parse, so the new index parameter is designed to be much clearer and more specific.

## Visual changes
None.

Trello card: https://trello.com/c/16ngDhBP/493-change-the-index-parameter-part-2-indexsectioncount

### DIFF
--- a/app/views/homepage/_government_activity.html.erb
+++ b/app/views/homepage/_government_activity.html.erb
@@ -22,7 +22,11 @@
             link: item[:link],
             title: item[:title],
             track_action: "governmentactivityLink",
-            index: "3.#{index + 1}",
+            index_values: {
+              index_section: 3,
+              index_link: index + 1,
+              index_section_count: 6,
+            },
             index_total: ga4_links_count,
             section: t("homepage.index.government_activity", locale: :en)
           } %>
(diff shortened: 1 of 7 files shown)


app/views/homepage/_government_activity.html.erb
```diff
@@ -39,12 +43,9 @@
...
 
       <%# 
-        The "big number" component links below are hardcoded separately (i.e. not in a loop in this instance) and therefore their 'index' properties are also
-        hardcoded. Additionally, in order to account for these hardcoded components in the 'index_total' property, 2 is added to the ga4_links_counts variable 
-        (which is defined in homepage > index.html.erb). If a new "big number" component is added or removed, this will need to be reflected in the 'index' and
-        'index_total' properties to ensure that they remain accurate for tracking/analytics purposes.
-      %>
-      
+        The "big number" component links below are hardcoded separately (i.e. not in a loop in this instance) and therefore their 'index' properties are also hardcoded. If a new "big number" component is added or removed, this will need to be reflected in the 
+        'index_link' and 'index_total' properties to ensure that they remain accurate for tracking/analytics purposes.
```
JamesCGDS, line 47: Quick question: seeing as the `index_section` and `index_section_count` properties are also hardcoded, do you think there would there be any mileage in putting a comment in to flag that they'll also need updating if a section is added/removed from the homepage? For example, wherever `index`/`index_values` is being used in the templates, we put something like `The hardcoded properties will need updating whenever sections are added/removed to ensure that tracking/analytics remains accurate.`

Also nitpick: the formatting of the comment is a bit odd

andysellick: Great shout, thanks @JamesCGDS. Have put a comment in the index template, and updated the formatting of this comment as well.

JamesCGDS (approved):
Looks good!

PR #3574, opened 2023-03-30
Add support for District/Unitary pairs

KludgeKML:

## What

Allow the upper tier in a 2 tier system to be a unitary authority in the council finder controller.

## Why

Although in principle 2 tiers will always be in reality county/district,during merge periods, it is sometimes simpler to mark a county that will turn into a unitary authority as a unitary authority before the actual switchover period. This change allows that (previously the code assumed this could never happen, and would break if it did). If we reengineer LLM after April 2023 we can probably remove this, but we'll need to be mindful of similar future events.

[Trello card](https://trello.com/c/HJRF2kzb/210-councils-in-local-links-manager-north-yorks-merge)

## How

Simply allow the upper tier in a county/district pair to be either marked as county or unitary.

### DIFF
--- a/app/controllers/find_local_council_controller.rb
+++ b/app/controllers/find_local_council_controller.rb
@@ -38,10 +38,11 @@ class FindLocalCouncilController < ContentItemsController
 
       render :one_council
     else
-      # NOTE: the data doesn't support the situation where we get > 1 result
-      # and it's anything other than a county and a district, so the obvious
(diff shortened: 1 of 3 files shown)


1pretz1 (approved):
Looks good! 

'the obvious problem with this code *shouldn't* happen. (sorry for when it does)' - it has happened 😄

PR #3588, opened 2023-04-19
Add publication summary to licences.

KludgeKML:

## What

Add the publication.description (ie summary) to the rendered licences for the new licence finder. This makes them render more consistently with other specialist publisher documents (which show the summary in the finder preview and then also on the first page of the document).

## Why

[Trello card](https://trello.com/c/hCfkg3fC/1957-add-summary-to-licences-in-frontend)

## Screenshots?

Before:

Screenshot 2023-04-19 at 13 53 08

After:

Screenshot 2023-04-19 at 14 03 20

### DIFF
--- a/app/views/licence_transaction/start.html.erb
+++ b/app/views/licence_transaction/start.html.erb
@@ -5,6 +5,9 @@
   publication: publication,
   edition: @edition,
 } do %>
+
+  <%= render "govuk_publishing_components/components/lead_paragraph", text: publication.description %>
+
   <article role="article" class="group content-block">
     <div class="inner">
       <% if licence_details.present? %>
--- a/app/views/licence_transaction/continues_on.html.erb
+++ b/app/views/licence_transaction/continues_on.html.erb
@@ -4,6 +4,9 @@
   publication: publication,
   edition: @edition,
 } do %>
+
+  <%= render "govuk_publishing_components/components/lead_paragraph", text: publication.description %>
+
   <article role="article" class="content-block group">
     <div class="inner">
       <div class="intro">


app/views/licence_transaction/start.html.erb
```diff
@@ -5,6 +5,9 @@
   publication: publication,
   edition: @edition,
 } do %>
+
+  <%= render "govuk_publishing_components/components/lead-paragraph", text: publication.summary %>
```
leenagupte, line 9: I think this needs to be `publication.description`

leenagupte: I don't think the "start" template is rendered for every type of licence (local authority / external), so "summary" also needs to be added to the `continues_on` and possibly the `authority` templates.

1pretz1: Nice. Hmm, the authority page also displays the overview (or 'body') so may be worth adding to the authority view too? E.g https://www.integration.publishing.service.gov.uk/find-licences/auction-premises-registration/rotherham. It feels a bit odd that the same information is repeated on the pre and post postcode views but 🤷  

The multiple authority view doesn't return the 'body' from looking at the code so I think we're fine excluding the description from that view (https://github.com/alphagov/frontend/blob/29d7ecbdb40202c59f423d7a21ccc0bc07b38027/app/views/licence_transaction/multiple_authorities.html.erb#L1). I don't have an example of this page though.

Probably worth running it by Alice to double check 👍

KludgeKML: I've asked Alice. The multiple authorities is just a disambiguator for split postcodes, so I think that's fine.

leenagupte: > It feels a bit odd that the same information is repeated on the pre and post postcode views

That is really odd. I wonder how many people actually scroll down to read the information on the postcode lookup page compared to the authority page.

leenagupte (commented):
I think the first two commits should be fixed up, but apart from that, it looks good 👍

KludgeKML:
@leenagupte Okay, commits fixed and comments improved.

1pretz1 (approved):
LGTM

PR #3616, opened 2023-05-17
Make explicit that this test is for multiple licensing authorities

KludgeKML:

## What

Make it clear in this test that the test is for multiple licensing authorities, not multiple local authorities.

## Why

- We no longer support them (see https://github.com/alphagov/frontend/pull/3594/commits/f4c10bb352eba283589f36e8be579628bfe0983a)
- Safe behaviour is to choose the first one if we do happen to get multiple.
- Make it explicit that this is about licencing authorities (rather than local authorities, since there is a concept of multiple authorities being returned for a single postcode in the postcode chooser paths).

[Trello card](https://trello.com/c/3AHvfc4d/2002-investigate-first-authority-selected-flow)

### DIFF
--- a/test/integration/licence_transaction_test.rb
+++ b/test/integration/licence_transaction_test.rb
@@ -337,7 +337,7 @@ class LicenceTransactionTest < ActionDispatch::IntegrationTest
         end
       end
 
-      context "when there are more than one authority" do
+      context "when there are more than one licensing authority" do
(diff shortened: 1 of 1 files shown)


leenagupte (approved):
This is much clearer, thanks!

1pretz1 (approved):

PR #3630, opened 2023-05-31
Add basics for CSV Preview functionality

brucebolt:
, [Jira issue PP-904](https://gov-uk.atlassian.net/browse/PP-904)This adds the basics for providing CSV Preview functionality outside of Whitehall.

Currently this only shows the metadata (e.g. title, department crests and file size) for the CSV file.  The actual preview will be added in later work.

| Whitehall 	| Frontend 	|
|--------	|-------	|
|  Screenshot of CSV preview rendered by Whitehall, including the metadata at the top of the page and a table showing a preview of the CSV file. | Screenshot of a CSV preview rendered by Collections.  Only the metadata is present at the moment.  The appearance is almost identical to Whitehall, except for a few margins being slightly different. |

[Trello card](https://trello.com/c/hn6IpttC)

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -11,6 +11,7 @@
 //= link views/_calendars.css
 //= link views/_completed_transaction.css
 //= link views/_cookie-settings.css
+//= link views/_csv_preview.css
 //= link views/_homepage.css
 //= link views/_travel-advice.css
 //= link views/_report-child-abuse.css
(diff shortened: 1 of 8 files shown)


callumknights (approved):
Code looks good to me, I would have normally questioned the necessity of adding each individual metadata field in its own commit. But given some of them are done rather differently to others I think this way is fine, much easier to understand what has been added that way. 👍

PR #3635, opened 2023-06-06
Fix issue with CSV preview content item retrieval

brucebolt:
, [Jira issue PP-754](https://gov-uk.atlassian.net/browse/PP-754)We previously assumed all `parent_document_url` values from Asset Manager were prefixed `www.gov.uk` regardless of environment.

However we have realised this is an artefact of the data sync. Documents edited in integration (or another non-production environment will have the correct domain).

Therefore updating the controller to only keep the path, regardless of the domain.

[Trello card](https://trello.com/c/QAgy09Z4)


chrisroos (approved):
Looks good to me, @brucebolt.

PR #3639, opened 2023-06-08
Add CSV contents to CSV preview

brucebolt:
, [Jira issue PP-795](https://gov-uk.atlassian.net/browse/PP-795)⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

This adds the ability to see the contents of the CSV file on the CSV preview.

Note: the functionality here around character encoding and truncating has been ported directly from Whitehall to retain the existing behaviour.

## Why

We are removing the rendering of CSV previews out of Whitehall.  This is the next step to adding the functionality into Frontend.

## Screenshots

| Rendered by Whitehall | Rendered by Frontend |
| -- | -- |
| Screenshot of a CSV preview rendered by Whitehall.  Below the metadata header is a table showing the content of the file. | Screenshot of a CSV preview rendered by Frontend.  It is visually identical to the table rendered by Whitehall.  |

[Trello card](https://trello.com/c/LXuEzWzH)

### DIFF
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -1,7 +1,21 @@
+require "csv"
+
 class CsvPreviewController < ApplicationController
+  MAXIMUM_COLUMNS = 50
+  MAXIMUM_ROWS = 1000
+
   def show
-    legacy_url_path = request.path.sub("/preview", "").sub(/^\//, "")
     @asset = GdsApi.asset_manager.whitehall_asset(legacy_url_path).to_hash
+
+    csv_preview = CSV.parse(media, encoding:, headers: true)
+
+    @csv_rows = csv_preview.to_a.map { |row|
+      row.map { |column|
+        {
+          text: column,
+        }
+      }.take(MAXIMUM_COLUMNS)
+    }.take(MAXIMUM_ROWS + 1)
 
     parent_document_path = URI(@asset["parent_document_url"]).request_uri
     @content_item = GdsApi.content_store.content_item(parent_document_path).to_hash
@@ -9,5 +23,33 @@ class CsvPreviewController < ApplicationController
     @attachment_metadata = @content_item.dig("details", "attachments").select do |attachment|
       attachment["url"] =~ /#{legacy_url_path}$/
     end
+  end
+
+private
+
+  def legacy_url_path
+    request.path.sub("/preview", "").sub(/^\//, "")
+  end
+
+  def media
+    GdsApi.asset_manager.whitehall_media(legacy_url_path).body
+  end
+
+  def encoding
+    @encoding ||= if utf_8_encoding?
+                    "UTF-8"
+                  elsif windows_1252_encoding?
+                    "windows-1252"
+                  else
+                    raise FileEncodingError, "File encoding not recognised"
+                  end
+  end
+
+  def utf_8_encoding?
+    media.force_encoding("utf-8").valid_encoding?
+  end
+
+  def windows_1252_encoding?
+    media.force_encoding("windows-1252").valid_encoding?
   end
 end
--- a/app/assets/stylesheets/views/_csv_preview.scss
+++ b/app/assets/stylesheets/views/_csv_preview.scss
@@ -15,3 +15,19 @@
     color: govuk-colour("black");
   }
 }
+
+.csv-preview__outer {
+  border: govuk-spacing(1) solid govuk-colour("light-grey");
+  margin-bottom: govuk-spacing(6);
+}
+
+.csv-preview__inner {
+  border: 1px solid govuk-colour("mid-grey");
+  padding: govuk-spacing(2) govuk-spacing(4) 0;
+  overflow: auto;
+
+  &:focus {
+    border: 3px solid govuk-colour("black");
+    outline: 3px solid govuk-colour("yellow");
+  }
+}
--- a/app/views/csv_preview/show.html.erb
+++ b/app/views/csv_preview/show.html.erb
@@ -48,5 +48,18 @@
         <% end %>
       </div>
     </div>
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full">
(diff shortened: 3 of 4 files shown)


app/controllers/csv_preview_controller.rb
```diff
@@ -1,13 +1,57 @@
+require "csv"
+
 class CsvPreviewController < ApplicationController
   def show
-    legacy_url_path = request.path.sub("/preview", "").sub(/^\//, "")
     @asset = GdsApi.asset_manager.whitehall_asset(legacy_url_path).to_hash
 
+    maximum_columns = 50
```
CristinaRO, line 7: Is there a place where these values could live as constants rather than "magic numbers"? (Maybe there isn't, and this is the only place that will ever need to keep track of them! I have a knee-jerk reaction to numbers in the wild.)

app/controllers/csv_preview_controller.rb
```diff
@@ -1,13 +1,57 @@
...
 
     @attachment_metadata = @content_item.dig("details", "attachments").select do |attachment|
       attachment["url"] =~ /#{legacy_url_path}$/
     end
   end
+
+private
+
+  def legacy_url_path
+    request.path.sub("/preview", "").sub(/^\//, "")
```
CristinaRO, line 30: How stable is this pattern? Where do the path chunks to be substituted come from?

test/fixtures/csv-preview.csv
CristinaRO, line 1: It feels like we should be able to mock a thousand row CSV rather than having a giant fixture, but I trust you that this was the best solution out of a few not-ideal choices. 😅

CristinaRO (approved):
I left a few non-blocking comments. Good work!

app/controllers/csv_preview_controller.rb
brucebolt: Good idea, I'll make them constants.

app/controllers/csv_preview_controller.rb
brucebolt: The route will [always have `/preview` at the end and a leading slash](https://github.com/alphagov/frontend/blob/3f08e1b5b249aab9d0fbd0a282b6675df0f14df3/config/routes.rb#L109), so I think this is safe to substitute.

test/fixtures/csv-preview.csv
brucebolt: Yeah, the file grew and grew.  I've replaced it with a method that will generate a CSV based on a given number of rows and columns.

callumknights (approved):
LGTM 👍

app/controllers/csv_preview_controller.rb
```diff
@@ -1,13 +1,58 @@
...
+
+  def utf_8_encoding?
+    media.force_encoding("utf-8").valid_encoding?
+  end
+
+  def windows_1252_encoding?
+    media.force_encoding("windows-1252")
+    # This regexp checks for the presence of ASCII control characters, which
+    # would indicate we have the wrong encoding.
+    media.valid_encoding? && !media.match(/[\x00-\x08\x0b\x0c\x0e-\x1f]/)
```
BeckaL, line 56: Do we need this regex check if if statement goes through utf8 first (I am assuming these ASCII control characters are utf8 but perhaps I'm wrong)?

app/controllers/csv_preview_controller.rb
```diff
@@ -1,13 +1,58 @@
...
+                  else
+                    raise FileEncodingError, "File encoding not recognised"
+                  end
+  end
+
+  def utf_8_encoding?
+    media.force_encoding("utf-8").valid_encoding?
+  end
+
+  def windows_1252_encoding?
```
BeckaL, line 52: I know this might be a pain, but could we add a test that generates a windows encoded csv?

BeckaL (commented):
Generally looks good to me, some comments / questions around the windows encoding bit

app/controllers/csv_preview_controller.rb
brucebolt: I've added a context where I've converted the file into Windows-1252 encoding, then included the header/table content tests again.

app/controllers/csv_preview_controller.rb
brucebolt: I've no idea TBH.  I just copied it straight from Whitehall as I wasn't sure what problem they were trying to resolve.

Rails has a lot of encoding functionality built into it now, that may not have existed when this code was first written in Whitehall.  It's tempting to remove this regex and hope Rails can deal with correctly now.

BeckaL: Ugh yes I've just gone down a rabbit hole trying to trace back to the [imminence implementation](https://github.com/alphagov/imminence/pull/64/files) this was copied from and then trying to work out why that changed, but failed. Hmm. Is there any automated testing against real data we could do in integration - e.g. loop through some csvs, curl and check we get a 200 response back?

BeckaL: I'm a bit in two minds here - would be nice to remove redundant code if redundant (esp since this line is a bit confusing), but also aware that if we're wrong about it being redundant this might end up with us having to re-investigate bugs we learned the answer to a long time ago 🤷‍♂️

brucebolt: I'm going to remove it.  I also looked back through the commit messages and couldn't understand why the change was made.  We also don't have any examples of what problem the code was trying to fix, so I'm struggling to work out how we would be able to find an affected file for testing.

BeckaL: Yeah I was thinking more take a random subset of csvs and ping them and see that they all worked without the line, but I assume that even if it was useful it'd be for a super niche case so probably wouldn't catch it anyway. Sounds good to me

BeckaL (approved):

PR #3650, opened 2023-06-19
Serve draft CSV previews from the draft asset host

brucebolt:
, [Jira issue PP-873](https://gov-uk.atlassian.net/browse/PP-873)⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

If a user visits a CSV preview but the document is draft, we don't want to serve this document publicly.

As a solution, we need to redirect the user to the draft assets host. Once the user has been redirected, that URL is already behind Signon, so we will authenticate the user.

## Why

We can't serve draft attachment previews to users, as they haven't been published yet.

[Trello card](https://trello.com/c/mbHhuldr)

### DIFF
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -7,6 +7,10 @@ class CsvPreviewController < ApplicationController
   def show
     @asset = GdsApi.asset_manager.whitehall_asset(legacy_url_path).to_hash
 
+    if draft_asset? && !served_from_draft_host?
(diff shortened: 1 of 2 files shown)


chrisroos (approved):
I've not tried this locally but from what I can see in the diff it looks OK to me.

BeckaL (approved):
LGTM

PR #3654, opened 2023-06-21
Resolve issue with `\r\n` line breaks and multi-line fields

brucebolt:
This was originally present in Whitehall, so the fix has been copied straight over from there.

Whitehall fix: https://github.com/alphagov/whitehall/blob/20911ccde160fed879027b1fd03c7cc51a5e8e2d/app/models/csv_preview.rb#L74-L90

### DIFF
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -11,7 +11,19 @@ class CsvPreviewController < ApplicationController
       redirect_to(Plek.find("draft-assets") + request.path, allow_other_host: true) and return
     end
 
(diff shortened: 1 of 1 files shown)


jkempster34 (approved):
This makes sense to me. Are there specs for this that can be copied over? If not, should we write some new ones?

Approving anyway as it's a live issue

brucebolt:
> This makes sense to me. Are there specs for this that can be copied over? If not, should we write some new ones?

We should probably add a test.  Merging this, to fix the immediate issue and we can write a test later.

PR #3621, opened 2023-05-19
Add GA4 tracking to local transactions

AshGDS:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Hi @andysellick, would you be able to approve this if all is OK? Thanks :+1:

## What

Adds the following GA4 events to the `local_transaction` flow (https://www.gov.uk/rubbish-collection-day)
- form_submit when clicking the "Find" button
- form_error when submitting an invalid postcode
- form_complete when you land on a results page 
- information_click if you click the result link

## Why
https://trello.com/c/2qmN8Wjg/546-form-events-formsubmit-formerror-formcomplete-and-informationclick-on-localtransaction-format

- This relies on this change in `govuk_publishing_components` to make the `form_submit` work, so the following PR should be deployed first: https://github.com/alphagov/govuk_publishing_components/pull/3409

## How to test it

1. Run `frontend` locally. If the publishing components PR is not deployed, run `frontend` with a local `static`, and have `static` point to local publishing components. Then make sure your publishing components is on the branch `ga4-local-transaction-changes`
2. Instead of setting up the APIs required for local transactions, we're going to fake the result. To do this open the file `app/controllers/local_transaction_controller.rb` in your editor.
3. Change the `fetch_postcode` function to this:
```Ruby
  def fetch_location(postcode)
    if postcode.present?
      local_custodian_codes = [5990] # hardcoded list of custodian codes
    end
    LocationsApiPostcodeResponse.new(postcode, local_custodian_codes, nil)
  end
```
4. Change the `authority_results` function to this
```Ruby
  def authority_results
    @authority_results = {
        "local_authorities" => [ 
          "name" => 'Example Council',
          "homepage_url" => "http://example.com",
          "country_name" => "England",
          "tier" => "district",
          "slug" => "westminster",
        ]
    }
  end
```
5. Change the `interaction_details` function to this
```Ruby
  def interaction_details
    {
      "local_authority" => {
        "name" => "Hello World",
        "homepage_url" => "http://westminster.example.com",
        "country_name" => "England",
      },
      "local_interaction" => { "url" => "http://westminster.example.com" },
    }
  end
```
6. Now if you go to http://127.0.0.1:3005/rubbish-collection-day/ and submit the form, it should redirect to a results page with our faked data.

PR #3655, opened 2023-06-21
Send location_form action and method data to GA4

Rosa-Fox:
[Trello local_transaction](https://trello.com/c/iQXIg0y8/596-add-action-method-to-form-submit-localtransaction)
[Trello special_route](https://trello.com/c/rRmyCOPB/601-add-action-method-to-form-submit-specialroute)
[Trello place](https://trello.com/c/AgiNiNa0/600-add-action-method-to-form-submit-place)
[Trello license](https://trello.com/c/LK951exF/599-add-action-method-to-form-submit-licence)

Sends "action" and "method" data to GA4.

The location_form partial is used by the following formats:
- place
- local_tranaction
- license_transaction
- special_route

The tests for these formats have been updated to reflect the additional values.

I tested this manually for local transactions using the ['How to test it' instructions in this PR description](https://github.com/alphagov/frontend/pull/3621).

Screenshot 2023-06-21 at 18 11 16

### DIFF
--- a/app/views/application/_location_form.html.erb
+++ b/app/views/application/_location_form.html.erb
@@ -54,6 +54,7 @@
       "module": "ga4-form-tracker",
       "ga4-form": {
         "event_name": "form_submit",
+        "action": "submit",
         "type": ga4_type,
         "text": t("find", locale: :en),
         "section": ga4_section,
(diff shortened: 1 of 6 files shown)


JamesCGDS:
I've had a quick look at this and the `action` is looking good! We might need to check re `method` - I may be wrong here but we've been treating `method` as which mouse button was pressed (i.e. left/right/middle click) when we've implemented GA4 tracking elsewhere so I'm not sure if `method` is referring to the form's submit method (i.e. POST) in this instance.

It can get a bit confusing so I hope I've made sense here - give me a shout if I can explain better!

andysellick:
Yeah thinking about this some more I think a conversation with the analysts is needed on `method` - because for forms it's always going to be `post` so there's a question over how useful that would be - it'd be like recording "click" as a method on links.

Rosa-Fox:
We spoke with a PA offline and we agreed to remove sending through the 'method' for the reason Andy mentioned above.

PR #3656, opened 2023-06-22
Permit dots in the filename for CSV Previews

brucebolt:
Rails does not permit dots in dynamic segments (see https://guides.rubyonrails.org/routing.html#dynamic-segments).

Therefore adding a constraint to override this for CSV Previews, as these can contain dots (and other characters other than slashes) in the filename.


davidgisbey (approved):
lgtm

PR #3657, opened 2023-06-22
Fix homepage grid alignment using the correct number of items

hannalaakso:
The homepage grid hardcodes the number of items in it. Since that number has recently changed in the content item, there is some extra padding appearing at the end of the grid which shouldn't be there. 

The permanent fix will be to update the grid to the gem cards component which doesn't have this issue, we'll pick this up soon.

[Trello](https://trello.com/c/ZSgasiES/1872-bug-spacing-between-topics-and-government-activity-on-mobile-view-is-too-much-s), [Jira issue NAV-8524](https://gov-uk.atlassian.net/browse/NAV-8524)

## Before

image_(1)

## After

govuk-frontend-app-pr-3657 herokuapp com_(iPhone SE)

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -145,17 +145,17 @@
 }
 
 .homepage-services-and-info__list {
-  @include columns($items: 17, $columns: 1, $selector: "li");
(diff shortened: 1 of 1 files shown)


patrickpatrickpatrick (approved):
Looks good to me!

PR #3670, opened 2023-06-27
Single redirect /account to home.account.gov.uk rather than session start if no redirect param passed

KludgeKML:
⚠️ **NOTE: To preserve user journey on /email/manage, this requires DI Home to support the login hints parameter in email-alert-frontend. Do not merge until that is confirmed!** (Now confirmed live)

## What

Redirect requests to /account directly to home.account.gov.uk if there is no redirect param supplied.

## Why

Request from the OneLogin team - redirecting here prevents a problem where rather than being asked to sign in with 2FA to begin with, an uplift scenario starts (where the user logs in without 2FA, then the 2FA is requested later). This is leading to confusing differences in 2FA prompts, and a situation where people resetting their password are asked for their 2FA twice consecutively.

[Trello card](https://trello.com/c/SiAQ3dZU/2086-redirect-govuk-account-to-homeaccountgovuk)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -149,7 +149,7 @@ GEM
       sentry-rails (~> 5.3)
       sentry-ruby (~> 5.3)
       statsd-ruby (~> 1.5)
-    govuk_personalisation (0.13.0)
+    govuk_personalisation (0.14.0)
       plek (>= 1.9.0)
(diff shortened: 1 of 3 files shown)


1pretz1 (approved):
Looks good!

PR #3682, opened 2023-07-05
Remove metadata component

MartinJJones:
## What
Remove the `metadata` app component, replacing it with the `metadata` component from the publishing components gem.

## Why
- Using the `metadata` component from the publishing components gem helps to ensure that the design is consistent across GOV.UK
- Helps keep our codebase DRY, we only need to make changes in one place

Trello: https://trello.com/c/XZyNCfWb/56-identify-which-components-could-be-improved-rationalised-on-govuk

Related issue: https://github.com/alphagov/govuk_publishing_components/issues/3482

### Before

app-c-metadata

### After

gem-c-metadata


andysellick:
@MartinJJones have looked at the code and this looks really good. A few merge conflicts to sort but I'd be happy to approve this if you've got time to finish it off?

PR #3705, opened 2023-07-24
Add GA4 tracking to the Electoral Registration Office results page

AshGDS:

## What
- Adds tracking to the [Electoral Registration Office results page](https://www.gov.uk/contact-electoral-registration-office?postcode=b75+5dd)
- Specifically, the `ga4-auto-tracker` and the `ga4-link-tracker` that we used on other local transactions.

## Why

It was missed when adding tracking to local transactions.

[Trello card](https://trello.com/c/NauEzrMp/625-local-transaction-form-not-firing-formcomplete-and-informationclick-events)

I had an issue testing the `data-ga4-auto` attribute. This is because the JSON contains a single quote in the `text` value. The text comes through as `We've matched the postcode...`. It seems like Capybara doesn't like single quotes in CSS attribute selectors, and we use CSS selectors to test the JSON. So I have just tested that `data-ga4-auto` exists, rather than testing the whole JSON. Feel free to try and fix it, but I tried a range of string combinations and couldn't get it to work.

## How

I couldn't get the page working locally, but if you do `puts page.html` on the test you can see how the HTML is output, and the tests should also hopefully show that it will work.

### DIFF
--- a/app/views/electoral/results.html.erb
+++ b/app/views/electoral/results.html.erb
@@ -22,35 +22,45 @@
       <%= t('electoral.service.matched_postcode_html', postcode: @postcode.sanitized_postcode, electoral_service_name: @presenter.electoral_service_name) %>
     </p>
   <% end %>
-  <% if @presenter.use_electoral_services_contact_details? %>
-    <%= render partial: "contact_details",
-      locals: {
-        presenter: @presenter,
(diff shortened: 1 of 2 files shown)


andysellick (approved):

AshGDS:
Thanks @andysellick :+1:

PR #3716, opened 2023-07-31
Show a message when the csv preview errors

catalinailie:

## What
When publishers have uploaded CSV files that have unsupported characters, the CSVPreviewer throws an error, which Sentry logs and spams our channel, then falls back to trying to show the file/directory directly from the S3 bucket and an XML file is returned stating an error.

Example: 
- CSV with unsupported characters https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1175050/HMRC_Senior_Officials_Travel_October-December_2022.csv/preview
- asset deleted or redirected
https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1172334/desnz-special-advisers-gifts-8-february-march-2023.csv/preview

## Why
The user is presented with a very unfriendly XML message with no useful steps they can take.

Trello card: https://trello.com/c/05Eqs8Z4/1923-stop-showing-users-an-xml-error-for-csv-preview-when-the-csv-file-is-not-valid-m, [Jira issue NAV-8219](https://gov-uk.atlassian.net/browse/NAV-8219)

## Screenshots?

Screenshot 2023-08-04 at 16 56 29

### DIFF
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -13,6 +13,13 @@ class CsvPreviewController < ApplicationController
       redirect_to(Plek.find("draft-assets") + request.path, allow_other_host: true) and return
     end
 
+    parent_document_path = URI(@asset["parent_document_url"]).request_uri
+    @content_item = GdsApi.content_store.content_item(parent_document_path).to_hash
+
+    @attachment_metadata = @content_item.dig("details", "attachments").select do |attachment|
+      attachment["url"] =~ /#{Regexp.escape(legacy_url_path)}$/
+    end
+
     original_error = nil
     row_sep = :auto
     begin
@@ -23,7 +30,7 @@ class CsvPreviewController < ApplicationController
         row_sep = "\r\n"
         retry
       else
-        raise original_error
+        render :malformed_csv and return
       end
     end
 
@@ -34,13 +41,6 @@ class CsvPreviewController < ApplicationController
         }
       }.take(MAXIMUM_COLUMNS)
     }.take(MAXIMUM_ROWS + 1)
-
-    parent_document_path = URI(@asset["parent_document_url"]).request_uri
-    @content_item = GdsApi.content_store.content_item(parent_document_path).to_hash
-
-    @attachment_metadata = @content_item.dig("details", "attachments").select do |attachment|
-      attachment["url"] =~ /#{Regexp.escape(legacy_url_path)}$/
-    end
   end
 
   def access_limited
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -9,6 +9,8 @@ class CsvPreviewController < ApplicationController
   def show
     @asset = GdsApi.asset_manager.whitehall_asset(legacy_url_path).to_hash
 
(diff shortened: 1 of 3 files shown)


kevindew:
Great stuff working on this @catalinailie, just a passing few comments as I'm going to be away for a while and won't have chance to follow up.

I imagine rather than having such a basic error page we can have one that is a bit richer and more helpful to the user. I think this would be good as it should help imply that something isn't broken on GOV.UK and instead is broken with the file the department uploaded. I'm no designer or content designer but I mocked this up as what I thought would be a better experience for the file in your example:

assets publishing service gov uk_government_uploads_system_uploads_attachment_data_file_1175032_HMRC_senior_officials_hospitality_October_to_December_2022 csv_preview

In your Slack comment you also mentioned about not knowing how to access redirects. This can be done from the code that queries asset manager. For example:

```
irb(main):002:0> GdsApi.asset_manager.whitehall_asset("/government/uploads/system/uploads/attachment_data/file/1172334/desnz-special-advisers-gifts-8-february-march-2023.csv").to_ha
sh
=>
{"_response_info"=>{"status"=>"ok"},
 "id"=>"http://asset-manager/assets/64b8e1a32059dc00125d2622",
 "name"=>"desnz-special-advisers-gifts-8-february-march-2023.csv",
 "content_type"=>"text/csv",
 "size"=>530,
 "file_url"=>
  "https://assets.integration.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1172334/desnz-special-advisers-gifts-8-february-march-2023.csv",
 "state"=>"uploaded",
 "draft"=>false,
 "deleted"=>false,
 "redirect_url"=>"https://www.gov.uk/government/publications/desnz-special-advisers-gifts-hospitality-and-meetings-january-to-march-2023",
 "replacement_id"=>"64c292b6331a650014934d1b",
 "parent_document_url"=>"https://draft-origin.publishing.service.gov.uk/government/publications/desnz-special-advisers-gifts-hospitality-and-meetings-january-to-march-2023"}
```

If a response has a `redirect_url` it will mean that the asset will be redirected and that we shouldn't try to render the file ([access the media](https://github.com/alphagov/frontend/blob/19afe2bc1ed8728c325663957408b55362f85d85/app/controllers/csv_preview_controller.rb#L56-L58)).

I imagine the simplest thing to do would be to have code that is along the lines of

```ruby
return error_410 if @asset["deleted"] || @asset["redirect_to"].present?
```

There's probably some cleverer things we could do like `if return some_redirect_code if @asset["redirect_to"].end_with?(".csv")` but I'm not sure that'd be a good idea as there's probably loads of edge cases (check redirect is on GOV.UK, is the file actually a CSV or just pretending etc)

catalinailie:
Thanks @kevindew for taking the time!

> I imagine rather than having such a basic error page we can have one that is a bit richer and more helpful to the user. 

Yes, totally agree. We were talking about having a nicer error page. We asked a content designer in our team for some recommendations. 

>This can be done from the code that queries asset manager

My fault, I was just looking at the `GdsApi.asset_manager.whitehall_media` response. This is very helpful 🙌

Tetrino (approved):
👍

PR #3724, opened 2023-08-14
Remove duplicate licence rendering code

kashifatcha:
## What

We created a new [controller / endpoint / views / presenter etc](https://github.com/alphagov/frontend/pull/3448/files#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5) for the Specialist Publisher licence transaction format as the data structure varied significantly from existing licence endpoint.

Now we’ve deployed the new finder and migrated over the licences, the old rendering code is no longer needed and should be removed. I’ve added a checklist of things to remove (may not be exhaustive).

We don’t want to remove LicenceDetailsPresenter as this is a shared class between both endpoints.

## Why

This code is no longer used and should be deleted.

https://trello.com/c/aiuktMQ6/2138-remove-duplicate-licence-rendering-code-from-frontend


KludgeKML (approved):
LGTM

1pretz1 (approved):
Thanks for splitting the commits and removing the unused instance var + nice work spotting the incorrect paths! Looks great 🍏

PR #3725, opened 2023-08-14
Move Search above Popular for live user testing

matthillco:

## What

Run a live test for 1 week, from 15 August, on moving the search bar on the homepage and have metrics to show the impact of this change

## Why

Get a benchmark to see what happens if we move search up the page.

We realised that it is very difficult to change the look and feel of the header and popular sections without pushing the search bar below the fold on mobile. We know search is the most highly valued and used feature on the homepage, so we don't want to push it below the fold, as it would make it significantly harder to find. One way to avoid pushing search below the fold is to move search above popular. We'll run a live test to move _Search_ above _Popular_ and see what effects it has on user behaviour

[Trello card](https://trello.com/c/SITNaPz9/1962-live-test-to-move-search-bar-m), [Jira issue NAV-8502](https://gov-uk.atlassian.net/browse/NAV-8502)

## BEFORE 

image

## AFTER

image

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -264,10 +264,14 @@
   }
 }
 
+.homepage-popular {
+  padding-top: govuk-spacing(7);
+}
+
 .homepage-most-viewed-list {
   list-style: none;
   margin: 0;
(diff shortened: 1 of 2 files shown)


hannalaakso (approved):
LGTM 👍

PR #3743, opened 2023-09-01
Change layout of homepage

patrickpatrickpatrick:

## What

Change the layout of homepage so that:

- 'Popular on GOV.UK' links now appear in a 3 column grid layout on desktop
- 'Services and information' and "Featured" sections are now next to each other on desktop.
- The 'Featured' section now uses the new "Two Thirds" variation of the image-card component

## Why

As part of several incremental changes being made to the homepage.

[Trello Card](https://trello.com/c/5ZGhnozF/2024-live-test-5-layout-popular-grid-etc-design-dev-work-live-25-sep-m), [Jira issue NAV-3270](https://gov-uk.atlassian.net/browse/NAV-3270)

### Before

homepage-desktop-before

### After

homepage-after

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -145,28 +145,9 @@
 }
 
 .homepage-services-and-info__list {
-  @include columns($items: 16, $columns: 1, $selector: "li");
   list-style: none;
   margin: 0 govuk-spacing(-3);
   padding: 0;
-
-  @include govuk-media-query($from: "tablet") {
-    @include columns($items: 16, $columns: 2, $selector: "li");
-  }
-
-  @include govuk-media-query($from: "desktop") {
-    @include columns($items: 16, $columns: 3, $selector: "li");
-  }
-
-  .homepage-section__government-activity & {
-    @include columns($items: 6, $columns: 1, $selector: "li");
-    margin-bottom: govuk-spacing(4);
-
-    @include govuk-media-query($from: "tablet") {
-      @include columns($items: 6, $columns: 2, $selector: "li");
-      margin-bottom: 0;
-    }
-  }
 }
 
 .chevron-card {
@@ -181,7 +162,7 @@
 }
 
 .chevron-card__description {
-  margin: 0 govuk-spacing(-6) 0 0;
+  margin: 0 govuk-spacing(6) 0 0;
 }
 
 .chevron-card__link {
@@ -206,7 +187,7 @@
     height: $dimension;
     position: absolute;
     right: govuk-spacing(1);
-    top: govuk-spacing(3);
+    top: 50%;
     @include prefixed-transform($rotate: 45deg);
     width: $dimension;
   }
@@ -228,6 +209,10 @@
   padding: govuk-spacing(4) 0 govuk-spacing(6);
 }
 
+.homepage-section--promotion-slots {
+  padding-top: govuk-spacing(6);
+}
+
 .homepage-section--links-and-search {
   background: govuk-colour("light-grey");
   padding: govuk-spacing(6) 0 govuk-spacing(8);
@@ -235,6 +220,10 @@
 
 .homepage-section--services-and-info {
   padding: govuk-spacing(6) 0 0;
+
+  @include govuk-media-query($from: "desktop") {
+    padding-right: govuk-spacing(4);
+  }
 }
 
 .homepage-section__heading {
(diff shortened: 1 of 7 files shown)


app/assets/config/manifest.js
```diff
@@ -12,6 +12,7 @@
 //= link views/_cookie-settings.css
 //= link views/_csv_preview.css
 //= link views/_homepage.css
+//= link views/_homepage_new.css
```
hannako, line 15: I'm finding this second commit a bit hard to follow. I thiiiiink what's happening is you're undoing the css styles you made in the first commit in `homepage.css` , and making those changes in a new file called `homepage_new.css` instead, so that you can make use of the @new_design toggle? But then there are also fresh changes in this commit. So I was wondering if you could rework the first two commits to make it easier to follow. 

BUT I'm also confused as to why we're using the toggle here at all. I was under the impression we'd be using it to incrementally roll out the look and feel changes which is going live mid October - where as this PR is to go live on Monday. Plus you've picked the commits from other cards so it's all rolled into one anyhow. Soooo can we just forget about the toggle in this PR?

Huge apologies if I've totally missed the point of what you're doing here Patrick!

hannako: Ahhh ok so I think you've added this logic (and new homepage css file) because some of the changes form part of the look and feel designs and monica is wanting to include them early, so you're setting the foundations for being able to use the backend toggle later. But I think that will make the roll out of the layout changes for Monday complicated as we'll need to move just these changes out of the the `if` block. 

So if it's ok with you, I think we should remove that logic out of this PR. And we can add it back when/if we need it in subsequent work

hannalaakso (commented):
Some minor comments on the CSS. Still need to review the markup changes.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -277,19 +270,21 @@
     padding-bottom: 0;
   }
 
-  & > li {
+  & > div li {
```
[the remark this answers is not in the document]
MartinJJones: I've updated the styles in a recent commit to use `.homepage-most-viewed-list__item` class instead - https://github.com/alphagov/frontend/pull/3743/commits/8c4fc5b20ce461db1b393b1a8e4fea3500cc7e8c

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -228,13 +209,21 @@
...
 .homepage-section--links-and-search {
   background: govuk-colour("light-grey");
   padding: govuk-spacing(6) 0 govuk-spacing(8);
 }
 
 .homepage-section--services-and-info {
   padding: govuk-spacing(6) 0 0;
+
+  @include govuk-media-query($from: "desktop") {
+    padding-right: 32px;
```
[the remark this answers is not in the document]
MartinJJones: I've updated this to use the `govuk-spacing` mixin

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -245,6 +234,10 @@
   padding: govuk-spacing(3) 0 0;
 }
 
+.homepage-section__container {
+  padding-right: 32px;
```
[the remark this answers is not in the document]
MartinJJones: This class name and styling was removed in a recent commit as it is not required

app/assets/config/manifest.js
MartinJJones: I've removed the logic to set the new styles, so the new homepage layout will be used when the changes are merged

app/views/homepage/_links_and_search.html.erb
```diff
@@ -31,45 +31,52 @@
...
+        text: t("homepage.index.popular_links_heading"),
+        data_attributes: {
+          ga4_scroll_marker: true,
+        },
+      } %>
+      <ul class="homepage-most-viewed-list " data-module="gem-track-click ga4-link-tracker">
+        <% t("homepage.index.popular_links").each_with_index do | item, index | %>
+
+          <% if index % 3 == 0 %>
+            <div class="govuk-grid-row">
```
hannalaakso, line 47: Using a `<div>` here between `<ul>` and `<li>` results in a HTML validation error:
> Element [div](https://html.spec.whatwg.org/multipage/#the-div-element) not allowed as child of element [ul](https://html.spec.whatwg.org/multipage/#the-ul-element) in this context. 

I think we might need to find a different way of building the grid, possibly using CSS grid like we do with the [cards component](https://components.publishing.service.gov.uk/component-guide/cards). Unfortunately that component doesn't currently have a variant that doesn't include the arrows 🤔

hannalaakso (approved):
Great, thanks both @MartinJJones and @patrickpatrickpatrick for getting this into such a good place 🙌  We should probably now update the PR description as it's now longer reflective of the work, there have been quite a few design changes since.

I tested the page on the browsers and devices we usually test in. Didn't spot any issues, apart from how on mobile there seems to be a bit too much grey space below the Popular links. I checked with Monica who agreed that we should reduce the amount of space so that's something to look at tomorrow.  

Screenshot 2023-09-21 at 17 40 55


I'm giving this PR a tick to indicate that I'm happy with everything apart from the above mobile spacing issue so whoever reviews the PR tomorrow can focus on reviewing any small fixes we add tomorrow.

MartinJJones:
@hannalaakso thanks for finding and reporting the spacing issue, I've fixed this issue on a recent commit:

Screenshot 2023-09-22 at 10 19 20

matthillco:
I've also completed a secondary review of changes made today and it looks good to go.

PR #3761, opened 2023-09-13
CSV previews to work with modern urls

yuetylauiris:
**CSV previews to work with modern urls**

**Context**: Currently whitehall uses a semi-hard-coded url called `legacy-url-path` to identify an asset in asset manager.
This url path looks something like `/government/uploads/system/uploads/attachment_data/*`.
While other publishing services use asset-id to identify an asset in asset manager.
We are working on an epic to make whitehall use the asset-id similar to other publishing service.

For csv-preview: Currently with Frontend preview functionality for csv, it uses legacy-url-path to fetch the asset
from asset manager. When legacy url path is removed from Whitehall then frontend would have to use
the asset-id to fetch the asset.

The changes we have made will allow the controller to retrieve asset using asset-id or legacy url path based on the params its called with.

[Trello card](https://trello.com/c/LV4TnjlP/176-story-csv-previews-should-work-with-modern-urls)

### DIFF
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -4,13 +4,12 @@ class CsvPreviewController < ApplicationController
   MAXIMUM_COLUMNS = 50
   MAXIMUM_ROWS = 1000
 
+  before_action :get_asset, only: [:show]
+
   rescue_from GdsApi::HTTPForbidden, with: :access_limited
 
   def show
-    @asset = GdsApi.asset_manager.whitehall_asset(legacy_url_path).to_hash
-
     return error_410 if @asset["deleted"] || @asset["redirect_url"].present?
-
     if draft_asset? && !served_from_draft_host?
       redirect_to(Plek.find("draft-assets") + request.path, allow_other_host: true) and return
     end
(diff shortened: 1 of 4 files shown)


app/controllers/csv_preview_controller.rb
```diff
@@ -51,29 +50,33 @@ def access_limited
 
 private
 
-  def legacy_url_path
+  def path
```
hannako, line 53: Perhaps `asset_path` would be a more informative name here?

app/controllers/csv_preview_controller.rb
```diff
@@ -7,10 +7,9 @@ class CsvPreviewController < ApplicationController
   rescue_from GdsApi::HTTPForbidden, with: :access_limited
 
   def show
-    @asset = GdsApi.asset_manager.whitehall_asset(legacy_url_path).to_hash
+    @asset = params[:legacy] ? GdsApi.asset_manager.whitehall_asset(path).to_hash : GdsApi.asset_manager.asset(params[:id]).to_hash
```
hannako, line 10: What do you think about moving this out to a `before` action? Called something like `fetch_asset`  which runs before the `show` action only, and which sets @asset

hannako (commented):
Thanks for the PR, and for taking the time to update the description which really helped make the intention of the work clearer - much appreciated!!

Overall I'm wondering how long is frontend going to need to support both types of asset preview url? If the answer is "a while" then I think it might be worth refactoring. I've made a couple of small suggestions, but we could pull the asset and media fetching logic out of the controller and put it into a service object or something like that. Which might help remove the duplication in test files.

syed-ali-tw:
> Overall I'm wondering how long is frontend going to need to support both types of asset preview url? If the answer is "a while" then I think it might be worth refactoring. I've made a couple of small suggestions, but we could pull the asset and media fetching logic out of the controller and put it into a service object or something like that. Which might help remove the duplication in test files.

We would need these changes (frontend to handle both routes) for quite some time. Reason being, There are already published contents which require the old Whitehall download media endpoint.
We do have a story in our backlog to decommission this but it would require us to monitor prod to understand how much this endpoint is being used.
We have tried to minimise the test duplication logic and have now moved the test into one file

hannako (approved):
Thanks @syed-ali-tw and @yuetylauiris Great work!

PR #3776, opened 2023-09-25
Implement new header component for homepage

patrickpatrickpatrick:

## What

Move the 'inverse-header' component to the components folder. Make a new 'homepage-header' component. Add logic for displaying the 'inverse-header' when `new_design` parameter is not present. Add logic displaying for displaying the 'homepage-header' when `new_design` parameter is present.

## Why

This is part of changes that will be gradually rolled out to the homepage. As these changes are not going to be made live immediately, I have attempted to split the code in a way that will allow them to be easily toggled by the `new_design` parameter. The additional effect is that this makes the code easier to understand and will allow us to easily make the change to the header permanent. 

## Visual Difference

with `/new_design=true`

Screenshot 2023-10-02 at 11 44 31

[Relevant Trello Card](https://trello.com/c/xgMWdEkA/2058-look-and-feel-homepage-big-banner-m), [Jira issue NAV-8321](https://gov-uk.atlassian.net/browse/NAV-8321)

### DIFF
--- a/app/assets/stylesheets/views/_homepage_header.scss
+++ b/app/assets/stylesheets/views/_homepage_header.scss
@@ -1,0 +1,62 @@
+@import "govuk_publishing_components/individual_component_support";
+
+$pale-blue-colour: #d2e2f1;
+
+.homepage-header {
+  background-color: $govuk-brand-colour;
+  padding-bottom: govuk-spacing(7);
+  padding-top: govuk-spacing(4);
+  @include govuk-typography-common;
+
+  @include govuk-media-query($from: desktop) {
+    padding-bottom: govuk-spacing(8);
+    padding-top: govuk-spacing(6);
+  }
+}
+
+.homepage-header__title {
+  @include govuk-typography-weight-bold;
+  color: govuk-colour("white");
+  font-size: 40px;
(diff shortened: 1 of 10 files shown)


app/assets/stylesheets/views/_homepage_header.scss
```diff
@@ -0,0 +1,62 @@
+@import "govuk_publishing_components/individual_component_support";
+
+$pale-blue-colour: #d2e2f1;
+
+.app-c-homepage-header {
```
MartinJJones, line 5: Now that this has moved back to a view template, I think we should be able to remove the `app-c-` prefix from the class names.

app/views/homepage/_homepage_header.html.erb
```diff
@@ -0,0 +1,49 @@
+<%
+  add_view_stylesheet("homepage_header")
+%>
+
+<header class="app-c-homepage-header">
```
MartinJJones, line 5: We would also need to update this template to remove the `app-c-` prefixes in the class names

MartinJJones (commented):
Changes look good to me overall, just a small nitpick, we should then be able to merge the changes 👍

MartinJJones (approved):

PR #3781, opened 2023-09-29
Fix for csv preview with legacy_url_path

syed-ali-tw:
Reference issue:https://govuk.sentry.io/issues/4265154142/?alert_rule_id=3654813&alert_type=issue&notification_uuid=8ab09d48-0416-4305-9e3c-0de49561a46a&project=202225&referrer=slack

Currently content item when it returns attachments, it returns modern url hence when it tries to find  for legacy url, it fails and @attachment_metadata is empty this cause @attachment_metadata.first to be nil and hence the error.

With the fix we try to not rely on legacy_url_path, instead we look for the filename.

### DIFF
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -16,10 +16,10 @@ class CsvPreviewController < ApplicationController
 
     parent_document_path = URI(@asset["parent_document_url"]).request_uri
     @content_item = GdsApi.content_store.content_item(parent_document_path).to_hash
-
     @attachment_metadata = @content_item.dig("details", "attachments").select do |attachment|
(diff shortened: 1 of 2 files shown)


app/controllers/csv_preview_controller.rb
```diff
@@ -58,6 +58,10 @@ def asset_path
     request.path.sub("/preview", "").sub(/^\//, "")
   end
 
+  def filename
```
richardTowers, line 61: Nitpick, but I'd probably call this method asset_filename to be consistent with asset_path

syed-ali-tw: done

richardTowers (approved):

PR #3786, opened 2023-10-03
Update new_design query string matching to be more specific

hannako:

## What
We have temporary logic in this application to toggle new designs on the presence or absence of the new_design parameter. Eg `new_design=foo`, `new_design=false` etc all result in the new design showing. We should make this more specific.

## How

Only toggle the new design if the query string `new_design=impact` is added

Review app
SHOW DESIGN: https://govuk-frontend-app-pr-3786.herokuapp.com?new_design=impact
HIDE DESIGN: https://govuk-frontend-app-pr-3786.herokuapp.com

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -3,9 +3,14 @@ class HomepageController < ContentItemsController
   slimmer_template "gem_layout_homepage"
 
   def index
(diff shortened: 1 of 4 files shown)


patrickpatrickpatrick (approved):
LGTM!

PR #3796, opened 2023-10-09
Update homepage sections and spacing for new design

MartinJJones:
## What

Create new view templates and stylesheets that are used when `new_design` is true, when requested the layout is updated so that:

- The "More on GOV.UK" is moved next to "Government Activity"
  - Update the styling of the "More on GOV.UK" section so the links appear in a single column
- Section heading font-sizes are increased from `m` to `l`
- Section spacing is updated so that spacing for the top and bottom of each section adds up to 50px on mobile and 60px on desktop
- Removed "Departments and organisations" section
- Updated `index_section_count` and `indexes_section` values are also updated to reflect the new position on the page, this will ensure correct reporting in Google Analytics
- Updated the Chevron card styling so that all content is vertically centred and the top border is removed from the first chevron-card element

## Why

This is part of changes that will be gradually rolled out to the homepage. As these changes are not going to be made live immediately, I have attempted to split the code in a way that will allow them to be easily toggled by the new_design parameter.

[Trello card](https://trello.com/c/5AdPoKbt/2060-look-and-feel-homepage-general-adjustments-m), [Jira issue NAV-8491](https://gov-uk.atlassian.net/browse/NAV-8491)

#### Before
bottom-section-before

#### After
bottom-section-after

#### Before
chevron-card-before

#### After
chevron-card-after

---

### DIFF
--- a/app/views/homepage/index.html.erb
+++ b/app/views/homepage/index.html.erb
@@ -13,7 +13,11 @@
   <%
     # The GA4 tracking on the homepage depends on some hardcoded values for index. If a section is added or removed, these values will need to be updated.
     # Specifically 'index_section_count' (the total number of sections) and 'index_section' (the index of the section e.g. 2 for the 2nd section).
-    index_section_count = 6
+    if new_design?
+      index_section_count = 5
+    else
+      index_section_count = 6
+    end
   %>
 
   <% if new_design? %>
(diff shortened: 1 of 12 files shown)


MartinJJones:
@andysellick would you also be able to help review this PR? thinking specifically about the updates to the sections on the homepage and the related `index_section` and `index_section_count` for Google Analytics when using the new design - https://github.com/alphagov/frontend/pull/3796/files#diff-b0533c1f365ffe916d27cb02d89deee3e7e1d586f289692af1c69b6036d88449

app/views/homepage/index.html.erb
```diff
@@ -13,7 +17,11 @@
   <%
     # The GA4 tracking on the homepage depends on some hardcoded values for index. If a section is added or removed, these values will need to be updated.
     # Specifically 'index_section_count' (the total number of sections) and 'index_section' (the index of the section e.g. 2 for the 2nd section).
-    index_section_count = 6
+    if new_design?
+      index_section_count = 5
+    else
+      index_section_count = 6
+    end
```
andysellick, line 24: I think you could also write this as

```Ruby
index_section_count = 6
index_section_count = 5 if new_design?
```

andysellick (commented):
Index stuff looks good to me 👍

patrickpatrickpatrick:
This is a minor accessibility thing, I'm wondering if [clickable area/touch target of the chevron cards should be adjusted so there's some space between them](https://www.w3.org/TR/mobile-accessibility-mapping/#touch-target-size-and-spacing). They have a pseudo element applied to them which makes the whole element clickable but if you are using zoom it might make distinguishing between the cards difficult.

Screenshot 2023-10-09 at 14 19 42

app/views/homepage/index.html.erb
MartinJJones: Thanks Andy, this looks much cleaner, I've updated the view template to use this approach 👍

MartinJJones:
> This is a minor accessibility thing, I'm wondering if [clickable area/touch target of the chevron cards should be adjusted so there's some space between them](https://www.w3.org/TR/mobile-accessibility-mapping/#touch-target-size-and-spacing). They have a pseudo element applied to them which makes the whole element clickable but if you are using zoom it might make distinguishing between the cards difficult.
> 
> Screenshot 2023-10-09 at 14 19 42

Thanks @patrickpatrickpatrick good spot!, I've updated the padding in a recent commit to ensure that the target size for the chevron-card is the same as before, and updated the padding on `chevron-card__wrapper` to ensure the total padding size is 24px at the top and bottom as needed in the design.

MartinJJones:
> I think the spacing here
> 
> Screenshot 2023-10-09 at 14 10 33
> 
> isn't the same as in the design https://www.figma.com/file/InXsSt89C5y5QRT1DZ3dPM/Homepage-redesign---2023?type=design&node-id=2522-77874&mode=design&t=7uUIoRtetLK4e6ac-0

Thanks @patrickpatrickpatrick I've updated this section so it now closely matches the Figma design

patrickpatrickpatrick:
Looks good to me now, thanks Martin!

patrickpatrickpatrick (approved):

PR #3805, opened 2023-10-13
Update feature flag

hannako:
Update feature flag method so that it only activates the new design on integration


leenagupte (approved):
I've tried it out locally and this method works like a charm.

PR #3817, opened 2023-10-19
Fix summer time date data

me-and:
The data for when the clocks go forward and back is incorrect. It displays correctly at https://www.gov.uk/when-do-the-clocks-change, but the errors are visible in the page metadata and at the generated API endpoint https://www.gov.uk/when-do-the-clocks-change.json.

I've removed the reference to specific years in the metadata – it doesn't seem to be necessary – and I've corrected the year in one of the entries.

It might be better if this data had more validation, and/or were generated programatically, but both of those seem like much more effort than I can spare right now.


hannako:
Thanks for your PR, looks like this has been updated so I'll close.

PR #3825, opened 2023-10-24
Move tagline inside `<h1>` to improve semantics

hannalaakso:
## What
Improve the semantics of the new homepage tagline by moving it inside the `<h1>`. 

## Why
Anika flagged in her accessibility audit that the tagline had been marked up as a paragraph tag but styled to look like a heading. Visually it was only differentiated from the "GOV.UK" text with the use of colour so it looked like a part of the `<h1>` rather than a paragraph tag. This PR makes it part of the `<h1>` so that the semantics reflect the visual design.

## How
Separate the content of the  `<h1>` using span tags for design purposes. Add the visually hidden colon for punctuation to make JAWS and NVDA leave a pause after "GOV.UK" but before the tagline.

An alternative implementation would have been to move the tagline into its own `<h2>`. However, the feedback from Anika (who consulted other accessibility specialists on this) was that having the tagline in the  `<h1>` is much better than having it in a separate `<h2>`. The main reason is because the `<h2>` would be a heading without any real content, or rather the wrong content of the search component.

### DIFF
--- a/app/assets/stylesheets/views/_homepage_header.scss
+++ b/app/assets/stylesheets/views/_homepage_header.scss
@@ -51,6 +51,7 @@ $pale-blue-colour: #d2e2f1;
   margin: 0;
   padding-bottom: govuk-spacing(2);
   line-height: 1.07;
+  display: block;
 
   @include govuk-media-query($from: tablet) {
     font-size: 60px;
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
Changes look good to me 👍

PR #3844, opened 2023-11-07
Update homepage promo slot for the King's Speech

brucebolt:
## What

Replaces a promo slot on the GOV.UK homepage with information about the King's Speech and a link to the campaign site.

## Why

We don't have a user interface for editing these slots, so need to raise a PR to make the change.

[Trello card](https://trello.com/c/0ph5KEaa)

## Screenshots

Screenshot showing the second promo slot has been updated


davidtrussler (approved):
LGTM! 👍 

(I'm not sure why there are keys for both `image_src` and `image_src_new` but this pulls the right image so I guess it's fine.)

PR #3835, opened 2023-10-31
Make design changes to the homepage live

leenagupte:
, [Jira issue NAV-8403](https://gov-uk.atlassian.net/browse/NAV-8403)⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What's changed and why?

This does the minimum to make the homepage changes live. Before the changes were hidden behind two feature flags, one to restrict the changes to the integration environment, and the other to allow reviewers to flip between both versions of the homepage.

Once we're happy that the new changes have been successfully deployed, another PR will be opened to strip out all of the extra code required to support two versions of the homepage.

[Trello card](https://trello.com/c/1xZLQ0cw)

### Before
Screenshot 2023-10-31 at 17 35 09

### After
Screenshot 2023-10-31 at 17 37 29

PR #3846, opened 2023-11-07
Remove new design logic

hannako:
## What

Remove unused templates and css files following https://github.com/alphagov/frontend/pull/3835

## Why

- [Trello card](https://trello.com/c/IhFaIzdI/2220-remove-the-now-redundant-newdesign-code-from-frontend-m), [Jira issue NAV-12392](https://gov-uk.atlassian.net/browse/NAV-12392)

## Follow up work

- In a fresh PR in this application we should rename the files prefixed with `_new` 
   -  app/views/homepage/_more_on_govuk_new.html.erb
   - app/assets/stylesheets/views/_homepage_more_on_govuk_new.scss
   - app/views/homepage/_promotion_slots_new.html.erb
   - app/assets/stylesheets/views/_homepage_new.scss
   - image_src_new

- We also need to remove the env var from helm charts by reverting https://github.com/alphagov/govuk-helm-charts/pull/1316


config/locales/en.yml
```diff
@@ -611,17 +610,14 @@ en:
         - text: Find out payment dates and how to report a payment is missing.
           title: Cost of Living Payment
           href: /guidance/cost-of-living-payment
-          image_src: homepage/cost-of-living-november-2022.jpg
           image_src_new: homepage/cost-of-living.png
```
patrickpatrickpatrick, line 613: Could these be renamed to image_src now since we don't have both image_source and image_source_new?

patrickpatrickpatrick: Oh I see that's in the PR description, never mind

hannako: I made a note in the follow up work section of the PR descriptions as there are quite a few that have a `_new` prefix.

hannako: I can do it now if you think that's better?

patrickpatrickpatrick: Actually I think it's fine splitting this up into multiple PRs because it'll be easier to review

patrickpatrickpatrick (approved):

andysellick (approved):
Analytics looks correct 👍

PR #3857, opened 2023-11-16
Update Dockerfile to force Yarn 3

Tetrino:
Yarn behaviour had changed at some point wherein the version in package.json was no longer the version being installed at a global level, causing build issues.

This forces the version to be the latest version of 3.x Yarn - Version 4 is unsupported.


Tetrino:
Closing as this did not solve the problem.

PR #3858, opened 2023-11-16
Revert "Bump stylelint-config-gds from 1.0.0 to 1.1.0"

Tetrino:
It appears that this dependabot broke the Yarn versioning when implemented. This was not seen until delivery was attempted.


Tetrino:
Closing as this was not the issue.

PR #3859, opened 2023-11-16
Update Dockerfile to force Yarn 1.22.19

Tetrino:
Yarn behaviour had changed at some point in the subsequent releases, so we have to force a global version of 1.22.19 (the last working version) to avoid braking our builds. Further investigation should be made as Yarn continues to upgrade its global version (as of typing, 1.22.21 https://github.com/yarnpkg/yarn/releases )


KludgeKML (approved):
A simple change and easily to revert, so if it works it looks good to me.

sengi:
Good find!

PR #3867, opened 2023-11-23
Adds mechanism to select appropriate licence council

1pretz1:
For some licences, the council being selected after a user submits their postcode is incorrect and the user sees a licence unavailable message. This is because some licences are managed at the county level vs the district level.

This change adds in the ability to query the different councils returned by Local Links Manager and select the first one that has an actionable licence. District and parent authority can be returned by LLM [1].

[1]: https://github.com/alphagov/local-links-manager/blob/f3f95f0a14c7689d30a05e15f06cc6698a9bf62a/app/presenters/local_authority_api_response_presenter.rb#L13

Trello:
https://trello.com/c/DSGOKUXK/2271-investigate-licence-finder-selecting-incorrect-council-for-certain-licences

### DIFF
--- a/app/controllers/licence_transaction_controller.rb
+++ b/app/controllers/licence_transaction_controller.rb
@@ -33,6 +33,10 @@ class LicenceTransactionController < ContentItemsController
     end
 
     if locations_api_response.single_authority?
+      unless local_authority_slug
+        return render :licence_not_found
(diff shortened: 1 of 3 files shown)


KludgeKML (approved):
Looks good to me!

PR #3773, opened 2023-09-21
Add Brand recruitment banner

catalinailie:
The banner will be applied on the following pages:
- [/check-national-insurance-record](https://www.gov.uk/check-national-insurance-record)
- [/check-mot-history](https://www.gov.uk/check-mot-history)

Trello card: https://trello.com/c/EVjvHjqE/2090-brand-team-govuk-user-research-banner-request-to-set-up-m, [Jira issue NAV-8541](https://gov-uk.atlassian.net/browse/NAV-8541)

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## Why

[Trello card?](url)

PR #3876, opened 2023-12-05
Dry up Recruitment Banner logic

georges1996:

## What
To dry up the Recruitment Banner logic to reduce dev toil.

## Why
to reduce dev toil.

[Trello card?](https://trello.com/c/GPsDYWsr/2235-implement-intervention-banner-logic-in-frontend-applications-m-l), [Jira issue NAV-12256](https://gov-uk.atlassian.net/browse/NAV-12256)

## How
To add a yml file and dry up the code to make it easier to update banners

Only caveat is you will need to add:

`<%= render partial: 'shared/intervention_banner' %>` to the view you want the banner to be shown on 
and add `include RecruitmentBannerHelper` to the needed controller

### DIFF
--- a/app/helpers/recruitment_banner_helper.rb
+++ b/app/helpers/recruitment_banner_helper.rb
@@ -1,0 +1,27 @@
+module RecruitmentBannerHelper
+  def recruitment_banner
+    return false if recruitment_banners.nil?
+
+    current_path = request.path
+
+    recruitment_banners.find do |banner|
+      next unless valid?(banner)
+
+      banner["page_paths"]&.include?(current_path)
+    end
+  end
+
+  def recruitment_banners
+    recruitment_banners_urls_file_path = Rails.root.join("lib/data/recruitment_banners.yml")
+    recruitment_banners_data = YAML.load_file(recruitment_banners_urls_file_path)
+    recruitment_banners_data["banners"]
+  end
+
+  def valid?(banner)
+    required_fields.select { |field| banner[field].present? } == required_fields
+  end
+
+  def required_fields
+    %w[survey_url suggestion_text suggestion_link_text page_paths]
+  end
+end
--- a/test/unit/helpers/recruitment_banner_helper_test.rb
+++ b/test/unit/helpers/recruitment_banner_helper_test.rb
@@ -1,0 +1,51 @@
+require "test_helper"
+
+class RecruitmentBannerHelperTest < ActionView::TestCase
+  include RecruitmentBannerHelper
+
+  def setup
+    @recruitment_banners_data = YAML.load_file(Rails.root.join("test/fixtures/recruitment_banners.yml"))
+  end
+
+  def request
+    OpenStruct.new(path: "/")
+  end
+
+  def recruitment_banners
+    @recruitment_banners_data["banners"]
+  end
+
+  test "recruitment_banner returns banners that include the current url" do
+    actual_banners = recruitment_banner
+
+    expected_banners =
+      {
+        "name" => "Banner 1",
+        "suggestion_text" => "Help improve GOV.UK",
+        "suggestion_link_text" => "Take part in user research",
+        "survey_url" => "https://google.com",
+        "page_paths" => ["/"],
+      }
+    assert_equal expected_banners, actual_banners
+  end
+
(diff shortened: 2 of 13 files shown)


banners.yml
```diff
@@ -0,0 +1,19 @@
+banners: 
+  - name: Banner 1
+    suggestion_text: "Help improve GOV.UK"
+    suggestion_link_text: "Take part in user research"
+    survey_url: https://google.com
+    page_urls: 
+      - https://google.com
```
hannako, line 7: We should use a path here, rather than a full url. So that we have a way to review the changes before making them live. ie so the banner would show on the review app (which is on heroku), or on integration.

So you could either replicate the [previous implementation](https://github.com/alphagov/frontend/pull/3773/files) ie include the RecruitmentBanner module into the ContentItem presenter - which means you've access to the base_path from the ContentItem object. Or you could parse the `request.original_url` to extract the path. I feel like the first option is maybe cleaner? But up to you.

app/helpers/recruitment_banner_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+  def there_are_banners?
+    return false if banners.nil?
+
+    true
+  end
+
+  def renderable_banners
+    current_url = request.original_url
+
+    banners.select do |banner|
```
hannako, line 11: We won't ever show two banners on a page. So I think we should select the first banner that matches. Ie the order of the banners in the file is significant.

app/helpers/recruitment_banner_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+      banner["page_urls"]&.include?(current_url)
+    end
+  end
+
+  def banners
```
hannako, line 16: Banner is a bit of a loaded concept on govuk - there are emergency banners and global banners etc. So I think we could be more specific in our naming to avoid confusion... maybe `recruitment_banners` and apply this name to all the related methods you've created?

app/assets/stylesheets/components/_intervention.scss
```diff
@@ -0,0 +1,5 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.gem-c-intervention {
+  margin-top: govuk-spacing(4);
```
hannako, line 4: Is this css styling going to work for all page types that might have a banner? As they all have different templates, did you/could you check and add screenshots to the PR description?

banners.yml
```diff
@@ -0,0 +1,19 @@
+banners:
```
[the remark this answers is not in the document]
hannako: I think you could turn this into a comment. So future us can copy the fields needed.

hannako, line 1: Please could we move banners.yml out of the root directory? Maybe if would be a good fit for lib/data? And rename recruitment_banners.yml as per previous comment

test/unit/helpers/recruitment_banner_helper_test.rb
```diff
@@ -0,0 +1,49 @@
+require "test_helper"
+
+class RecruitmentBannerHelperTest < ActionView::TestCase
+  include RecruitmentBannerHelper
+
+  def setup
+    @banner_data = {
```
hannako, line 7: Could we use a yaml fixture here instead? I think it would be nice to test the parsing code, given the structure of the banners array is really important to this working.

hannako (commented):
Thanks for picking this up George, and getting a PR up so quickly. Hope the comments make sense, happy to chat if not.

test/unit/helpers/recruitment_banner_helper_test.rb
```diff
@@ -0,0 +1,47 @@
...
+        "suggestion_text" => "Help improve GOV.UK",
+        "suggestion_link_text" => "Take part in user research",
+        "survey_url" => "https://google.com",
+        "page_paths" => ["/"],
+      }
+    assert_equal expected_banners, actual_banners
+  end
+
+  test "recruitment_banners yaml structure is valid" do
+    recruitment_banners.each do |banner|
```
hannako, line 33: Sorry maybe I've confused you 😂 I mean check the format of the data in the real file. Not in the fixture.

PR #3877, opened 2023-12-06
Remove image-card overrides

MartinJJones:
## What
Removes the image card override styles as these will be included in the publishing components gem instead.

Depends on https://github.com/alphagov/govuk_publishing_components/pull/3752

## Why
Overrides to the publishing components is not best practice, it should be avoided as it makes the component harder to maintain, behaviour / rendering will be different than expected when following the component guide, it could break if the component is updated in the gem

[Trello card](https://trello.com/c/jZu4Jnsc/2280-homepage-live-test-5-layout-follow-up-card-iterate-the-image-card-component-in-publishing-components-s), [Jira issue NAV-12332](https://gov-uk.atlassian.net/browse/NAV-12332)

---

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -150,36 +150,6 @@
   @include govuk-media-query($from: "desktop") {
     padding: govuk-spacing(9) 0;
   }
-
-  // temp style overrides, will need to update the image-card component in the gem
-  // so these can be removed
-  .gem-c-image-card {
-    @include govuk-media-query($until: "mobile") {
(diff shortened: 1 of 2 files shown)


hannalaakso (approved):
LGTM 👍  I've checked that the removed styles match what was added in https://github.com/alphagov/govuk_publishing_components/pull/3752/files. I've also checked that https://github.com/alphagov/frontend/pull/3889 which pulls in the updated gem release rebased to this branch looks correct and matches what is on live. 

We need to merge in https://github.com/alphagov/frontend/pull/3889 and then merge in this PR.

PR #3897, opened 2023-12-18
Add find-family-hub

KludgeKML:

## What

Add special partial for find-family-hub-local-area place document.

## Why

This new finder has Name + URLs only, but can't be put into LLM because LLM doesn't currently allow arbitrary links that aren't covered by the LGSL/LGIL system. Adding the new partial allows us to publish the finder without maps appearing. (this should really be handled by LLM in future, so adding to a tech debt here: https://trello.com/c/gQIiYsEx/38-there-are-some-very-simple-local-authority-finders-in-imminence-that-might-be-better-off-as-local-links), [Jira issue PNP-5472](https://gov-uk.atlassian.net/browse/PNP-5472)

### DIFF
--- a/app/controllers/place_controller.rb
+++ b/app/controllers/place_controller.rb
@@ -8,6 +8,7 @@ class PlaceController < ContentItemsController
   NO_LOCATION = "validPostcodeNoLocation".freeze
 
   REPORT_CHILD_ABUSE_SLUG = "report-child-abuse-to-local-council".freeze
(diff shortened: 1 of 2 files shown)


hannako (approved):
LGTM (although this controller is getting a bit unwieldy so if another one of these comes up, I think we might need to think of a nicer approach)

KludgeKML:
> LGTM (although this controller is getting a bit unwieldy so if another one of these comes up, I think we might need to think of a nicer approach)

Definitely.

PR #3905, opened 2023-12-29
Ignore vim swap files and tags in .gitignore

unoduetre:

## What

This PR adds vim file types to .gitignore, so it makes it easier and less error prone to develop. Otherwise the developer must always remember to exclude them when commiting.

This PR does not include any functional code, only configuration.

## Why

There is no Trello card, as this is for developer convenience and it does not impact the functional code in any way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ spec/reports/pacts
 .env
 .yarn/cache/
 .yarn/install-state.gz
+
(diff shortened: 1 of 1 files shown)


unoduetre (commented):
Looks good to me.

hannako (approved):

PR #3906, opened 2023-12-29
Add AB test code to HMRC's ready reckoner page

hannako:
## What

Add AB testing logic to one page of interest following a departmental request

Related PR:  https://github.com/alphagov/govuk-fastly/pull/35

## Why

[Trello](https://trello.com/c/GJ5lt8ij/589-set-up-the-next-round-of-a-b-tests), [Jira issue MAIN-2739](https://gov-uk.atlassian.net/browse/MAIN-2739)

## Screenshots

| Variant A and Z | Variant B |
|---|---|
| Screenshot 2023-12-29 at 16 45 55 | Screenshot 2023-12-29 at 16 43 08|

### DIFF
--- a/app/controllers/concerns/hmrc_temporary_ab_testable.rb
+++ b/app/controllers/concerns/hmrc_temporary_ab_testable.rb
@@ -1,0 +1,39 @@
+module HmrcTemporaryAbTestable
+  CUSTOM_DIMENSION = 47 ## Not sure if this is correct
+
+  ALLOWED_VARIANTS = %w[A B Z].freeze
+
+  def self.included(base)
+    base.helper_method(
+      :hmrc_temporary_ab_test_variant,
+      :hmrc_temporary_ab_test_page?,
+      :hmrc_temporary_ab_test_variant_b?,
+    )
+    base.after_action :set_hmrc_temporary_ab_test_response_header
+  end
+
+  def hmrc_temporary_ab_test_variant_b?
+    hmrc_temporary_ab_test_page? && hmrc_temporary_ab_test_variant.variant?("B")
+  end
+
+  def hmrc_temporary_ab_test_page?
+    request.path == "/self-assessment-ready-reckoner"
+  end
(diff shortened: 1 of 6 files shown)


hannako:
@anatron @davidtrussler

I've drafted this PR as second line was quiet today - but I'm without the full context (see below), so I hope it's helpful! Please feel free to close and start fresh if not!

I noted in the content doc that there should be a placeholder string of `{{ab_test_sa_video_ready_reckoner}} ` but as the text is intended to go underneath the button, and mainstream publisher doesn't currently support a text block in that location, it's easier to implement without depending on the presence of the placeholder string. But perhaps that is not the desired behaviour - I wasn't sure.

app/controllers/concerns/hmrc_temporary_ab_testable.rb
```diff
@@ -0,0 +1,39 @@
+module HmrcTemporaryAbTestable
+  CUSTOM_DIMENSION = 47 ## Not sure if this is correct
```
richardTowers, line 2: I'm told it's correct for all of the tests to use 47, so we could remove this comment.

richardTowers (approved):
Nice - I like the use of a Module to include this behaviour. Neato.

Not having the placeholder means that this one will behave slightly differently to the tests in government-frontend.

For those tests, the AB testing code will only run (so users will only have cookies set / sessions tracked) if the placeholder is in the content item. This means content designers can control the start and end of the test period without developer involvement.

Without the placeholder, we don't get that behaviour.

I don't think this is a major issue, so long as we're aware of it. The default Z variant already covers the case where the test is off, and we can enable / disable it using Fastly, as we always have done with AB tests in the past. We'll just have to be aware that it's a bit different.

The only alternative I can think of is putting a placeholder somewhere to use as a flag, and just always replacing it with the empty string. But I think that's super confusing / hacky as the replacement won't appear where the placeholder is.

PR #3931, opened 2024-01-19
Fix video tracking on AB test

andysellick:

## What
Fixes video tracking on https://www.gov.uk/self-assessment-ready-reckoner

Note that this appears on the B variant of the A/B test, use the GOVUK browser extension to switch between them or use a mod header extension.

Frontend components should be implemented in templates or helpers, using the code examples found in https://components.publishing.service.gov.uk/component-guide. Injecting pre-rendered component markup into a page like this causes the following problems:

- component is not kept up to date with any changes upstream
- component JavaScript may not work properly (in this case the `data-image-card-module-started` attribute was included in the markup, which is only added after the JS has finished initialising on the component, so JS was not running on the component)
- styles for the component are not loaded, this is handled by the component template (see https://github.com/alphagov/govuk_publishing_components/blob/main/docs/set-up-individual-component-css-loading.md for more information)

For video tracking and JavaScript in general to work properly on videos they should be wrapped in a [govspeak component](https://components.publishing.service.gov.uk/component-guide/govspeak) and rendered in the page template. Pre-rendered markup for components should not be passed through from the locale files.

## Why
Video tracking for GA4 was not working on this video.

## Visual changes
None.

Trello card: https://trello.com/c/rHlNjPxe/775-video-tracking-not-working-on-a-b-test-pages

### DIFF
--- a/app/views/transaction/show.html.erb
+++ b/app/views/transaction/show.html.erb
@@ -110,9 +110,10 @@
                   margin_bottom: true,
                   data_attributes: data_attributes %>
       <% if hmrc_temporary_ab_test_variant_b? %>
-        <div class="govuk-body govuk-!-padding-bottom-5">
-          <%= "#{t('ab_testing.ready_reckoner_video_test_html')}".html_safe %>
-        </div>
+        <%= render "govuk_publishing_components/components/govspeak", {} do %>
+          <p><%= t('ab_testing.ready_reckoner_video_description') %></p>
+          <p><a href="https://www.youtube.com/watch?v=xHn31myAkio">How do I budget for my Self Assessment tax bill?</a></p>
+        <% end %>
       <% end %>
     </p>
   </section>
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -6,7 +6,7 @@ cy:
     two_versions: Mae 2 fersiwn o'r dudalen hon. Rydych chi'n gweld y
     version: fersiwn
     visit_govuk_html: Ewch i <a href="/" class="govuk-link">hafan GOV.UK</a> i weld gwasanaethau a gwybodaeth y llywodraeth.
-    ready_reckoner_video_test_html:
+    ready_reckoner_video_description:
   account:
     cookies_and_feedback:
       cookie_consent:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -6,7 +6,7 @@ en:
     two_versions: There are 2 versions of this page. You are viewing the
     version: version
     visit_govuk_html: Visit the <a href="/" class="govuk-link">GOV.UK homepage</a> to find government services and information.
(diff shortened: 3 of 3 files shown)


app/views/transaction/show.html.erb
```diff
@@ -110,9 +110,10 @@
                   margin_bottom: true,
                   data_attributes: data_attributes %>
       <% if hmrc_temporary_ab_test_variant_b? %>
-        <div class="govuk-body govuk-!-padding-bottom-5">
-          <%= "#{t('ab_testing.ready_reckoner_video_test_html')}".html_safe %>
-        </div>
+        <%= render "govuk_publishing_components/components/govspeak", {} do %>
+          <p><%= t('ab_testing.ready_reckoner_video_description') %></p>
+          <p><a href="https://www.youtube.com/watch?v=xHn31myAkio">How do I budget for my Self Assessment tax bill?</a></p>
```
hannako, line 115: I was just wondering if we should put the href and link title in the locale file as well?

hannako (approved):

davidtrussler (approved):
Hi @andysellick - this seems to be a very sensible change. My only concern is to wonder why we didn't do it like that originally so I'd be happier if one of the other requested reviewers could approve it as well in case there's a reason for that which I have missed. 🤔 
I would certainly not have expected to implement this in this way more permanently if it was the outcome of the A/B Test.

app/views/transaction/show.html.erb
andysellick: Good shout, will look into.

andysellick:
> Hi @andysellick - this seems to be a very sensible change. My only concern is to wonder why we didn't do it like that originally so I'd be happier if one of the other requested reviewers could approve it as well in case there's a reason for that which I have missed. 🤔 I would certainly not have expected to implement this in this way more permanently if it was the outcome of the A/B Test.

Thanks @davidtrussler. I don't know why this was done like this originally, I think on the surface it seems like a reasonable way to do it if you're not familiar with the way our frontend components work. I think it's been a while since we (the frontend team) did any kind of talk about how frontend stuff works to the wider dev community and we're probably overdue.

I'm glad you mentioned this because I've also been asked with fixing a similar problem with [this change](https://github.com/alphagov/government-frontend/pull/3025) but that one's a bit more complicated to unravel. Can you take a look?

davidtrussler:
> I'm glad you mentioned this because I've also been asked with fixing a similar problem with [this change](https://github.com/alphagov/government-frontend/pull/3025) but that one's a bit more complicated to unravel. Can you take a look?

@andysellick Ah right, so I think on that one there is no other way to do this since the alternate (A/B) content is provided by the translation files by design. I guess that in turn explains why this one was done in a similar way since it followed in the wake of the one you're referring to here (also under some duress over the Xmas period!) We queried this approach with @richardTowers at the time (since embedded videos was a more complex scenario than a simple text change) and it was agreed there was no other way than to approach it like this.

andysellick:
> @andysellick Ah right, so I think on that one there is no other way to do this since the alternate (A/B) content is provided by the translation files by design. I guess that in turn explains why this one was done in a similar way since it followed in the wake of the one you're referring to here (also under some duress over the Xmas period!) We queried this approach with @richardTowers at the time (since embedded videos was a more complex scenario than a simple text change) and it was agreed there was no other way than to approach it like this.

Thanks, that's good to know. If it's just a temporary AB test then I'll take a look at seeing if changing the markup in the locale files could make the video tracking work 👍

PR #3940, opened 2024-01-25
Fix localisation for Continue button in Simple Smart Answers

Tetrino:
We weren't correctly handling situations where the start button text said "Continue". The actual handling of these actions is in publisher so we use string matching to replace with I18n here in frontend.

https://trello.com/c/KbiBLzJ0/2331-fix-welsh-translation-for-continue-action-in-welsh-simple-smart-answers-m, [Jira issue NAV-12242](https://gov-uk.atlassian.net/browse/NAV-12242)

### DIFF
--- a/app/presenters/simple_smart_answer_presenter.rb
+++ b/app/presenters/simple_smart_answer_presenter.rb
@@ -17,6 +17,8 @@ class SimpleSmartAnswerPresenter < ContentItemPresenter
 
     if details["start_button_text"] == "Start now"
       I18n.t("formats.start_now")
+    elsif details["start_button_text"] == "Continue"
+      I18n.t("continue")
     else
       details["start_button_text"]
     end
--- a/test/unit/presenters/simple_smart_answer_presenter_test.rb
+++ b/test/unit/presenters/simple_smart_answer_presenter_test.rb
@@ -8,16 +8,27 @@ class SimpleSmartAnswerPresenterTest < ActiveSupport::TestCase
   context "locale is 'cy'" do
     setup do
       I18n.locale = :cy
-      @item = {
-        details: {
-          start_button_text: "Start now",
-        },
-      }
     end
 
     context "start_button_text is 'Start now'" do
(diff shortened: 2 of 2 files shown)


app/presenters/simple_smart_answer_presenter.rb
```diff
@@ -17,6 +17,8 @@ def start_button_text
 
     if details["start_button_text"] == "Start now"
       I18n.t("formats.start_now")
+    elsif detals["start_button_text"] == "Continue"
```
Tetrino, line 20: As far as solutions go this is miserable, however this appears to be how we handle the labelling here. https://github.com/alphagov/frontend/commit/a6e9db442ada1a983ae47818075776f6c8bcf70a

app/presenters/simple_smart_answer_presenter.rb
```diff
@@ -17,6 +17,8 @@ def start_button_text
 
     if details["start_button_text"] == "Start now"
       I18n.t("formats.start_now")
+    elsif details["start_button_text"] == "Continue"
+      I18n.t("continue")
```
Tetrino, line 21: As far as solutions go this is miserable, however this appears to be how we handle the labelling here. https://github.com/alphagov/frontend/commit/a6e9db442ada1a983ae47818075776f6c8bcf70a

We should be aware that there are potentially more hiccups we don't account for: https://github.com/alphagov/publisher/blob/21541089bd9dc07435efde92e1a6413e1fd659c7/app/views/simple_smart_answers/_fields.html.erb#L30

hannako: Oh dear WHADAMESS. Shall we ship this now as it looks bad. But please can you create a card on the UX tech debt board: https://trello.com/b/Hamq6Ore/govuk-user-experience-tech-debt with a nice explanation of suggested fix.

hannako (approved):
This looks OK for now, but on the basis that we will log and triage the tech debt that this fix has highlighted!

PR #3945, opened 2024-01-29
CSV previews are sometimes broken

unoduetre:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

We have had issues with CSV previews for some time, and work has been done to resolve them. However they seem to be occurring again (see Examples below). The purpose of this card is to investigate what’s going wrong and suggest a fix.

## Examples

- Visit [https://www.gov.uk/government/statistics/postcode-level-gas-statistics-2021-experimental](https://www.gov.uk/government/statistics/postcode-level-gas-statistics-2021-experimental "smartCard-inline") and click view online for the first CSV attachment
- Visit [https://www.gov.uk/government/statistics/legal-aid-statistics-april-to-june-2023-data-files](https://www.gov.uk/government/statistics/legal-aid-statistics-april-to-june-2023-data-files "smartCard-inline") and click view online for the first CSV attachment

## Why

[Trello card](https://trello.com/c/iaw1zR77/2328-csv-previews-are-sometimes-broken-l), [Jira issue NAV-12163](https://gov-uk.atlassian.net/browse/NAV-12163)

## Screenshots?

No user-visible changes.

PR #3952, opened 2024-02-01
CSV previews are sometimes broken - refactoring of CSV previews controller

unoduetre:

## What

This is a refactoring related to the previous work on CSV previews: https://github.com/alphagov/frontend/pull/3945

## Why

[Trello card](https://trello.com/c/iaw1zR77/2328-csv-previews-are-sometimes-broken-l), [Jira issue NAV-12163](https://gov-uk.atlassian.net/browse/NAV-12163)


test/unit/services/csv_preview_service_test.rb
```diff
@@ -1,6 +1,8 @@
 require "test_helper"
 
-class CsvPreviewControllerTest < ActionController::TestCase
+class CsvPreviewServiceTest < ActionController::TestCase
```
hannako, line 3: This is a bit of a confused spec...see docs here for what the [ActionController::TestCase]( https://api.rubyonrails.org/classes/ActionController/TestCase.html) is meant to test. It also seems to be testing all the private methods, which again is not conventional (or particularly informative)

I think in this PR we should at a minimum implement a unit test for the `CSVPreviewService` that only tests the public `csv_rows` method. We can involve the same csv data types to make sure we aren't losing coverage.

Stretch goal:

- a `CsvPreviewServiceControllerTest < ActionController::TestCase ` test, that is actually a controller test. Currently we don't have a controller test at all, just a massive [integration test](https://github.com/alphagov/frontend/blob/main/test/integration/csv_preview_test.rb). Probably the easiest thing to do is to [cherry pick from that integration test](https://github.com/alphagov/frontend/blob/main/test/integration/csv_preview_test.rb) . But this is a nice to have, as we do have test coverage in the integration test. So we could do this later/in a separate piece of work if you don't want to do it now.

hannako (commented):
Thanks for doing this Mat, I think this is a great improvement. I've made a couple of notes, mainly around the spec.

test/unit/services/csv_preview_service_test.rb
unoduetre: Argh, I missed that.

unoduetre: OK, I'll add a more complete unit test for the CSVPreviewService.

app/services/csv_preview_service.rb
```diff
@@ -0,0 +1,84 @@
+class CsvPreviewService
+  MAXIMUM_COLUMNS = 50
+  MAXIMUM_ROWS = 1000
+
+  def initialize(id:, filename:)
+    @id = id
+    @filename = filename
```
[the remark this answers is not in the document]
unoduetre: Yeah, it's a good idea. I will change that.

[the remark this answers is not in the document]
unoduetre: Changed.

test/unit/services/csv_preview_service_test.rb
unoduetre: Added.

app/controllers/csv_preview_controller.rb
```diff
@@ -25,35 +23,20 @@ def show
...
+private
+
   def access_limited
-    render :access_limited, status: :forbidden and return
+    render :access_limited, status: :forbidden
   end
 
-private
+  def malformed_csv
+    render :malformed_csv
```
hannako, line 38: How come we no longer need the `and return` ?

hannako: As there's no controller test, I can't tell if this is an issue or not.

unoduetre: return returns from the function, so it's redundant. What do you think return does in this code?

The controller has originally only had the integration test. I can add tests for that to it, or I can create a functional test and add it there. As this task was a refactoring task and not "write missing tests" task, I've only added tests which relate to the moved code, but I can add the missing controller tests, if that's your preference.

hannako: I was thinking without the return we might get https://api.rubyonrails.org/classes/AbstractController/DoubleRenderError.html 

But if that's not then happy to lose it. Thanks for adding the test.

unoduetre: I've added the test for malformed csv.

unoduetre: https://github.com/alphagov/frontend/blob/8c08ee54269049bbf530846e1e4f8cf6e632073e/test/integration/csv_preview_test.rb#L112

hannako (approved):
Looks great, thanks Mat

PR #3966, opened 2024-02-12
Standardise family-hub specific code

KludgeKML:
- Default to not showing location data if the address or town aren't present, to allow DSA zone finder and family hub to work without bespoke code.
- Report child abuse to your local council service still needs specific styling, so leave that hardcoded for the moment.

## What

As there are likely to be more than one dataset in imminence without physical locations, simplify to remove existing hardcoded slug for family-hub, and detect whether to show the location info based on the presense of an address and/or town (it can't be by postcode, because Imminence at the moment always requires postcodes).

## Why

DSA assessment is due to change soon, to a system where contact details will not include physical addresses. Also mild reduction in tech debt.

[Trello card](https://govuk.zendesk.com/agent/tickets/5557366)

## How

Remove hard-coding and additional partials, add checks to see whether town/address is present and hide location data.

## Screenshots?

No visible changes expected to the affected path (/find-family-hub-local-area)


KludgeKML:
Closing as obsolete.

PR #3977, opened 2024-02-20
Default cookie settings page to reject

andysellick:

## What
Set the options on the `/help/cookies` cookie settings page to be off by default in the HTML. They're currently not set in the HTML.

- this was previously handled by JavaScript, which would set the default reject consent cookie on page load, then set the radio buttons to match, or read the user set consent cookie, and set the radio button options accordingly
- now the radio buttons will appear as 'off' and then if the cookie says otherwise, the JS will update them, so this could be briefly visible (but the buttons are mostly below the fold anyway)
- when JS is disabled these radio buttons don't appear

## Why
Defaulting this to off in the HTML makes forthcoming changes to implement the single cookie consent API easier, because the default consent cookie isn't set anymore, so the JS will instead do nothing if the user has not previously indicated a consent preference. Otherwise we'll have to write something in the JS that detects there's no cookie, and sets all the radio buttons - which seems onerous compared to this.

## Visual changes
None.

Trello card: https://trello.com/c/dmlTAzKB/140-implement-single-consent-api

### DIFF
--- a/app/views/help/cookie_settings.html.erb
+++ b/app/views/help/cookie_settings.html.erb
@@ -73,7 +73,8 @@
               },
               {
                 value: "off",
-                text: t('cookies.off-usage')
+                text: t('cookies.off-usage'),
+                checked: true
               }
             ]
           } %>
@@ -94,7 +95,8 @@
               },
               {
                 value: "off",
-                text: t('cookies.off-campaigns')
+                text: t('cookies.off-campaigns'),
+                checked: true
               }
             ]
           } %>
@@ -111,7 +113,8 @@
               },
               {
                 value: "off",
-                text: t('cookies.off-settings')
+                text: t('cookies.off-settings'),
+                checked: true
               }
             ]
           } %>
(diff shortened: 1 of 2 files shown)


AshGDS (approved):
LGTM 👍 Non-blocking, do you think it's worth writing a test to ensure they stay checked like this by default?

andysellick:
> LGTM 👍 Non-blocking, do you think it's worth writing a test to ensure they stay checked like this by default?

Possibly... can you clarify? It might be slightly tricky as the JS that controls this is now in the gem, but I could write something there perhaps.

AshGDS:
@andysellick 

Within `help_cookes_test.rb`, something like:

```Ruby

    should "have radio buttons set to disable cookies by default" do
      visit "/help/cookies"

      cookies_settings = page.all('input[name=cookies-settings]')
      within "#content" do
        assert_nil cookies_settings[0]['checked']
        assert_equal cookies_settings[1]['checked'], 'checked'
      end
    end
```

AshGDS (approved):
Nice, thanks @andysellick 👍

PR #4013, opened 2024-03-25
Remove tel:// from places views

### LINKED ISSUE #2784: Remove telephone links
Remove all uses of `tel:/` to link to a telephone number as per Design System guidance:

[Do not display telephone numbers as links on devices that cannot make calls](https://design-system.service.gov.uk/patterns/telephone-numbers/)

https://github.com/alphagov/frontend/blob/f42857afbabc764b4adf26cc63311b0e5d19599f/app/views/place/_option_report_child_abuse.html.erb#L19

CodeSonia:
## What

- Removes occurrences of `tel://` from places views and no longer making them appear as links
- Updates places test to reflect the changes 
- Also removed the `tel://` instance from the Text Phone. Note, I was unable to find an example, so I hardcoded a number to check if it was safe to remove the linkable part. 

## Why

[Trello card](https://trello.com/c/Le1dyRqz/2480-frontend-issue-remove-tel-from-templates-s), [Jira issue NAV-12469](https://gov-uk.atlassian.net/browse/NAV-12469)

According to the Design System, it states that we [should not display telephone numbers as links](https://design-system.service.gov.uk/patterns/telephone-numbers/) on devices that cannot make calls, which could be quite misleading/confusing to the user. Therefore, in this PR, we have removed these instances. 

## Visual Changes

| Page Change | Before | After |
| --- | --- | --- |
| Report child abuse | image | image | 
| Report child abuse out of hours | image | image |
| Find a DSA provider | image | image | 

Fixes: [Issue #2784](https://github.com/alphagov/frontend/issues/2784)

### DIFF
--- a/app/views/place/_option.html.erb
+++ b/app/views/place/_option.html.erb
@@ -44,19 +44,19 @@
 
               <% if place["phone"].present? %>
                 <p class="govuk-body">
-                  Phone: <a class="govuk-link tel" href="tel://<%= place["phone"] %>"><%= place["phone"] %></a>
+                  Phone: <%= place["phone"] %>
                 </p>
               <% end %>
 
               <% if place["fax"].present? %>
                 <p class="govuk-body">
-                  Fax: <a class="govuk-link tel" href="tel://<%= place["fax"] %>"><%= place["fax"] %></a>
+                  Fax: <%= place["fax"] %>
                 </p>
               <% end %>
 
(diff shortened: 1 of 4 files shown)


KludgeKML (approved):
LGTM

MartinJJones (approved):
Nice work on fixing the issue Sonia!

PR #4026, opened 2024-04-08
Add truncation notice to CSV previews

unoduetre:

## What

When viewing CSV previews, a notice should indicate to users that the csv has been truncated.

## Why

To ensure our users can view the correct content in a format that works best for them

[Trello card](https://trello.com/c/utAkXLWQ/2473-add-truncation-notice-to-csv-previews-zendesk-m), [Jira issue NAV-12408](https://gov-uk.atlassian.net/browse/NAV-12408)

## Testing

The code has been deployed to integration. It can be tested e.g. [here](https://www.integration.publishing.service.gov.uk/government/publications/register-of-licensed-sponsors-workers).

## Screenshots

screenshot1
screenshot2

### DIFF
--- a/app/services/csv_preview_service.rb
+++ b/app/services/csv_preview_service.rb
@@ -2,8 +2,11 @@ class CsvPreviewService
   MAXIMUM_COLUMNS = 50
   MAXIMUM_ROWS = 1000
 
+  attr_reader :truncated
+
   def initialize(csv)
     @csv = csv
+    @truncated = false
   end
 
   def csv_rows
@@ -26,12 +29,13 @@ class CsvPreviewService
         raise
       end
     end
-    converted_and_restricted_csv(parsed_csv)
+    [converted_and_restricted_csv(parsed_csv), truncated]
   end
 
 private
 
   attr_reader :id, :filename
+  attr_writer :truncated
 
   def newline_or_last_char_index(string, newline_index)
     (0..newline_index).inject(-1) do |current_index|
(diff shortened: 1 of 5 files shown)


MartinJJones (commented):
Frontend changes look good to me 👍, I understand that the notice text may change, happy to review again if anything further changes

unoduetre:
The text has been updated.

app/services/csv_preview_service.rb
```diff
@@ -43,7 +47,9 @@ def newline_or_last_char_index(string, newline_index)
   end
 
   def truncate_to_maximum_number_of_lines(string, maximum_number_of_lines)
-    string[0..newline_or_last_char_index(string, maximum_number_of_lines - 1)]
+    truncation_index = newline_or_last_char_index(string, maximum_number_of_lines - 1)
```
beccapearce, line 50: I wonder if we could add an early return to this method? something like this above this line maybe?
```
return string unless string.length > maximum_number_of_lines
```

beccapearce (approved):
All looks good! I've left a suggestion, but feel free to ignore if it's not worth it! ! :)

app/services/csv_preview_service.rb
```diff
@@ -2,8 +2,11 @@ class CsvPreviewService
   MAXIMUM_COLUMNS = 50
   MAXIMUM_ROWS = 1000
 
+  attr_reader :truncated
```
hannako, line 5: Not sure I follow why we've got to attribute readers?

hannako: Ah got it, public reader, private writer 👍

PR #4041, opened 2024-04-25
Enable Multiarch, Bump to Ruby 3.2.4

dj-maisy:
## What?
This switches Frontend over to use the Multi-arch build process to start giving us ARM-compatible container images.

This also makes sure we're on at least Ruby 3.2.4, to give us ARM supported Ruby Images (and fix a recent CVE).

Paging @hannako for visibility.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG ruby_version=3.2.4
+ARG ruby_version=3.2
 ARG base_image=ghcr.io/alphagov/govuk-ruby-base:$ruby_version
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,9 +1,9 @@
-ARG ruby_version=3.2
+ARG ruby_version=3.2.4
```
sengi, line 1: Best not to specify the patch version here, otherwise you'll prevent it from staying automatically up-to-date in future.
```suggestion
ARG ruby_version=3.2
```

sengi (approved):

PR #4049, opened 2024-05-02
Update homepage featured links

beccapearce:
To be merged on 3rd May

## What

The local elections will end on 2nd May so we want to update the Featured section to remove the Voting ID link and replace it.

## Why

So that the Featured section stays relevant and includes actionable tasks.

[Trello card](https://trello.com/c/P97jqTwh/2570-featured-link-change-to-remove-voting-id-and-add-ni), [Jira issue NAV-12311](https://gov-uk.atlassian.net/browse/NAV-12311)

## How

We have amended the featured links in our en.yml file and removed the old images from our assets

## Screenshots
| Before | After |
|--------|--------|
| image | image |


unoduetre (approved):
Looks good to me.

PR #4074, opened 2024-05-21
Reorder body government activities

CodeSonia:
## What

- Address the inconsistency on the homepage where the order of government activities in the header menu and the page body have drifted out of sync following a redesign. 
- Updated the order of the government activities in page body to match the order in the header menu, which is based on popularity.
- Reorganise the order of the government activities in the `en.yml` locale file to reflect the same order of items in the header.

## Why

[Trello card](https://trello.com/c/2lWoDDKS/2600-synchronise-order-of-headings-in-government-activites), [Jira issue NAV-12452](https://gov-uk.atlassian.net/browse/NAV-12452)

To ensure there is consistency and helps improve user navigation, the order of items in the `government activities` in the page body should match the header menu. 

## Visual Changes
| Before | After |
| --- | --- |
| image|image|

Which now matches the same order of items in the header:
| Government activities list in the header menu|
| -- |
|image|

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -533,18 +533,18 @@ en:
       - title: Departments
         description: Departments, agencies and public bodies
         link: /government/organisations
-      - title: Research and statistics
-        description: Reports, analysis and official statistics
-        link: /search/research-and-statistics
       - title: News
         description: News stories, speeches, letters and notices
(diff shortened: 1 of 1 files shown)


beccapearce (approved):
NICE! ✨

PR #4079, opened 2024-05-23
Fix search bug on foreign travel advice page 

CodeSonia:
## What
- Added a `keydown` event listener to the `CountryFilter` function to prevent the default form submission behaviour when the Enter key is pressed.
- Kept the `keyup` event listener to handle filtering functionality
- The test case in `foreign_travel_advice_spec.js` now checks for `keydown` event instead of `keyup` when Enter key is pressed. 
- Changes on the integration test file `travel_advice_test.rb` where assertions are updated to reflect the visibility changes in the country list items.

## Why

[Trello card](https://trello.com/c/QuR23CRB/2616-bug-search-behaviour-on-foreign-travel-advice-page-s), [Jira issue NAV-12498](https://gov-uk.atlassian.net/browse/NAV-12498)

Previously, the keyup event listener was solely responsible for handling both form submission prevention and filtering. However, a bug was found where if you typed "Spain" in the search and press Enter, the results are cleared and the URL retains the search term. 

## Visual Changes
| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/app/assets/javascripts/views/travel-advice.js
+++ b/app/assets/javascripts/views/travel-advice.js
@@ -13,14 +13,16 @@
 
     this.container = searchInput.closest('.js-travel-container')
 
-    searchInput.addEventListener('keyup', function (event) {
+    searchInput.addEventListener('keydown', function (event) {
       if (event.keyCode === enterKeyCode) {
         event.preventDefault()
-      } else {
-        var filter = this.value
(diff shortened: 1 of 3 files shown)


MartinJJones (approved):
Nice work!, the changes look good to me 👍

PR #4104, opened 2024-06-10
Upgrade to govuk frontend 5.1.0

MartinJJones:
## What

- Move components that rely on govuk-frontend modules to separate `es6-components.js` file
- Remove Sass variables

### Move components that rely on govuk-frontend modules to seperate `es6-components.js` file

In the event that a browser below the target for `govuk-frontend` loads a page with JS on it, attempting to parse the JS from `govuk-frontend` will cause an error. To avoid this from happening, JS that contains `govuk-frontend` JS has been moved to seperate file which will be loaded in a script tag with `type="module"`. This will prevent the JS from being parsed and so prevent the error

### Remove Sass variables

- The `$legacy` attribute in `govuk-colour` has been deprecated and using it will have no effect (other than generating warnings on pre-compilation)
- Removed $govuk-compatibility-govuktemplate as this is no longer available
- Removed $govuk-new-link-styles as this is now set to true by default

The intention is for the PR to include the upgrade to the version of the publishing_components_gem as well, once released.

[Trello](https://trello.com/c/I0dQAkkC/2338-upgrade-frontend-to-run-on-v51-of-govuk-frontend), [Jira issue NAV-12244](https://gov-uk.atlassian.net/browse/NAV-12244)

---

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -162,7 +162,7 @@ GEM
     govuk_personalisation (0.16.0)
       plek (>= 1.9.0)
       rails (>= 6, < 8)
-    govuk_publishing_components (39.2.5)
+    govuk_publishing_components (40.0.0)
       govuk_app_config
       govuk_personalisation (>= 0.7.0)
       kramdown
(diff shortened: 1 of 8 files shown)


hannalaakso (approved):
This looks good to me 👍

PR #4110, opened 2024-06-13
Move the /contact (just the index) from feedback to frontend

georges1996:

## What
Move contact index page from feedback

## Why
As we are condensing the feedback app into frontend as part of our OKR 5

[Trello card](https://trello.com/c/lvIiblLk/182-contact-index-route)

Current implementation: https://www.gov.uk/contact
Local implementation (Run locally): http://frontend.dev.gov.uk/contact

### DIFF
--- a/app/views/contact/index.html.erb
+++ b/app/views/contact/index.html.erb
@@ -1,0 +1,31 @@
+<% content_for :title, "#{t('contact.title')} - GOV.UK" %>
+
+<%= render "govuk_publishing_components/components/breadcrumbs", {
+  collapse_on_mobile: true,
+  breadcrumbs: @breadcrumbs
+} %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= render "govuk_publishing_components/components/title", {
+        title: t('contact.title')
+      } %>
+      
+      <%= render "govuk_publishing_components/components/document_list", {
+        items: @popular_links
+      } %>
+
+      <%= render "govuk_publishing_components/components/details", {
+        title: "More topics"
+      } do %>
+        <%= render "govuk_publishing_components/components/document_list", {
+          items: @long_tail_links
+        } %>
+      <% end %>
+      <hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
+      <p class="govuk-body"><%= t('contact.form_html') %></p>
+      <p class="govuk-body"><%= t('contact.pages_html') %></p>
+    </div>
+  </div>
+</div>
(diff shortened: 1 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -24,4 +24,5 @@
   config.include FeatureHelpers, type: :feature
 
   config.include Helpers::Components, type: :view
+  config.include Capybara::DSL
```
KludgeKML, line 27: Do we still need this?

app/views/contact/index.html.erb
```diff
@@ -0,0 +1,31 @@
+<% content_for :title, "#{t('.title')} - GOV.UK" %>
```
KludgeKML, line 1: I assume this automatically gets this from the contact: part of the locale because it knows about the controller - which is nice, but I think they're explicit in the existing views in frontend, so maybe we should be the same (ie t("contact.title"))

app/views/contact/index.html.erb
```diff
@@ -0,0 +1,31 @@
...
+<%= render "govuk_publishing_components/components/breadcrumbs", {
+  collapse_on_mobile: true,
+  breadcrumbs: @breadcrumbs
+} %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= render "govuk_publishing_components/components/title", {
+        title: t('.title')
```
KludgeKML, line 12: As above?

app/views/contact/index.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      <%= render "govuk_publishing_components/components/details", {
+        title: "More topics"
+      } do %>
+        <%= render "govuk_publishing_components/components/document_list", {
+          items: @long_tail_links
+        } %>
+      <% end %>
+      <hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
+      <p class="govuk-body">Use the <a class="govuk-link" href="/contact/govuk">GOV.UK contact form</a> to send your questions or comments about the website.</p>
+      <p class="govuk-body">Check the <a class="govuk-link" href="/help">GOV.UK help pages</a> to find out about the use of cookies, accessibility of the site, the privacy policy and terms and conditions of use.</p>
```
KludgeKML, line 28: Can we make these lines localisable as well?

bin/rake
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
 require_relative "../config/boot"
 require "rake"
+require "spec_helper"
```
KludgeKML, line 4: Do we need this here?

KludgeKML (approved):
LGTM (you could potentially cut out the unused tests, but we'll be bringing them over).

PR #4113, opened 2024-06-13
Homepage tweaks to support the global banner

matthillco:
## What

Make changes to the homepage to improve it now that the global banner is in place. [Trello](https://trello.com/c/ZCjj63Y0/170-changes-to-the-homepage-to-support-the-global-banner), [Jira issue PNP-7996](https://gov-uk.atlassian.net/browse/PNP-7996)

## Why

* The header bar looks too deep as the blue border on the global banner butts up against it.
* There's not enough space between the bottom of the global banner and the GOV.UK main title text.

## BEFORE/AFTER

image

### DIFF
--- a/app/assets/stylesheets/views/_homepage.scss
+++ b/app/assets/stylesheets/views/_homepage.scss
@@ -251,3 +251,17 @@
     border-top: 4px solid $govuk-brand-colour;
   }
 }
+
+// Temporary overrides for global bar styling on homepage.
+// Doing this here avoids the need for republishing the
+// gem and therefore will be easier to remove later.
+.homepage.global-bar-present .global-bar {
+  border: 0;
+}
+
+.homepage.global-bar-present .homepage-header {
+  @include govuk-responsive-padding(7, "top");
+  @media (min-width: 1281px) { // stylelint-disable-line media-feature-range-notation
+    padding-top: govuk-spacing(8);
+  }
+}


app/assets/stylesheets/views/_homepage.scss
```diff
@@ -251,3 +251,14 @@
...
+
+// Temporary overrides for global bar styling on homepage during election.
+// Doing this here avoids the need for republishing the gem and therefore
+// will be easier to remove later.
+.homepage.global-bar-present .global-bar {
+  border: 0;
+}
+
+.homepage.global-bar-present .homepage-header {
+  padding-top: 55px;
```
andysellick, line 263: This looks quite big on mobile - is that intentional?

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -251,3 +251,14 @@
     border-top: 4px solid $govuk-brand-colour;
   }
 }
+
+// Temporary overrides for global bar styling on homepage during election.
```
andysellick, line 255: Nitpick: this comment feels a bit specific, the changes apply for any time the global bar would appear on the homepage, I suspect we'd leave them in place afterwards in case we need the global bar again at some point.

app/assets/stylesheets/views/_homepage.scss
matthillco: We've added 40px here as the previous padding was 15px, but you're right no mobile design was provided. I'll double check if it needs tweaking for mobile.

app/assets/stylesheets/views/_homepage.scss
matthillco: Good point, I've removed the homepage mention.

app/assets/stylesheets/views/_homepage.scss
matthillco: This has now been updated after pairing with Nikin and tweaking the values for different breakpoints.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -251,3 +251,17 @@
...
+// Temporary overrides for global bar styling on homepage.
+// Doing this here avoids the need for republishing the
+// gem and therefore will be easier to remove later.
+.homepage.global-bar-present .global-bar {
+  border: 0;
+}
+
+.homepage.global-bar-present .homepage-header {
+  @include govuk-responsive-padding(7, "top");
+  @media (min-width: 1281px) { // stylelint-disable-line media-feature-range-notation
```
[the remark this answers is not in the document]
matthillco: Requested by Nikin to match similar tweaks for views beyond 1280px, see previous precedents in `_homepage_header.scss` for example [line 22](https://github.com/alphagov/frontend/blob/af433e2d0b937964023ee8870b0d8b926c6fe071/app/assets/stylesheets/views/_homepage_header.scss#L22).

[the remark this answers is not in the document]
nnagewad: I've reviewed the design. Looks good on multiple viewport sizes 👍🏽

nnagewad:
I've reviewed the design. Looks good on multiple viewport sizes 👍🏽

app/assets/stylesheets/views/_homepage.scss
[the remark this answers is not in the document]
andysellick: I don't think in this case using the hard coded value is appropriate. The changes in `_homepage_header.scss` relate to making things 'extra big' on very large screens, whereas this is to do with how something appears on mobile. I don't want the use of a hard coded number to become a precedent here - we should be using the govuk-frontend breakpoints for consistency wherever possible.

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -251,3 +251,17 @@
...
 }
+
+// Temporary overrides for global bar styling on homepage.
+// Doing this here avoids the need for republishing the
+// gem and therefore will be easier to remove later.
+.homepage.global-bar-present .global-bar {
+  border: 0;
+}
+
+.homepage.global-bar-present .homepage-header {
```
andysellick, line 262: Styles for the `.homepage-header` class are in https://github.com/alphagov/frontend/blob/af433e2d0b937964023ee8870b0d8b926c6fe071/app/assets/stylesheets/views/_homepage_header.scss#L5 - why have they been added here?

app/assets/stylesheets/views/_homepage.scss
```diff
@@ -251,3 +251,17 @@
     border-top: 4px solid $govuk-brand-colour;
   }
 }
+
+// Temporary overrides for global bar styling on homepage.
+// Doing this here avoids the need for republishing the
+// gem and therefore will be easier to remove later.
+.homepage.global-bar-present .global-bar {
```
andysellick, line 258: Is `.homepage.global-bar-present` necessary? Is `.global-bar-present` not specific enough?

app/assets/stylesheets/views/_homepage.scss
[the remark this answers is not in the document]
matthillco: To my knowledge there are only [three breakpoints available](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/settings/_media-queries.scss#L10), hence why this hardcoded 1281 value exists:
```
$govuk-breakpoints: (
  mobile:  320px,
  tablet:  641px,
  desktop: 769px
) !default;
```
Therefore I think it can remain and we look at refactoring this in another piece of work to add more named breakpoints.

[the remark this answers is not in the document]
andysellick: I disagree, for the reasons stated above. To clarify - the padding only needs to be increased on desktop, which we already have a breakpoint for in `govuk-frontend`. Why would this extra padding only need to be added above `1281px`?

app/assets/stylesheets/views/_homepage.scss
andysellick: Great, thanks 👍

app/assets/stylesheets/views/_homepage.scss
matthillco: I reasoned that it made sense to keep the lines in this commit together in the `_homepage.scss` file to identify them as related and to enable quick removal when they are no longer needed (as all this will be refactored). As we need the specific `.homepage.global-bar-present` classes to hook into, I reasoned that these rules can live in the `_homepage.scss` since that's the logical 'parent' SCSS file for this piece of work.

Happy to move them if you think it's necessary, but I'd prefer to keep these together for the reasons given above.

andysellick: Okay, that's reasonable.

app/assets/stylesheets/views/_homepage.scss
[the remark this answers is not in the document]
matthillco: That's not the case, Nikin specifically asked for a little extra space on larger screens because the whole element looks unbalanced without it.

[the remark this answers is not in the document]
andysellick: Ah okay thanks that makes more sense now 👍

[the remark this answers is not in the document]
andysellick: I'm still uncomfortable about using magic numbers... might be something to feed back to the DS team on whether a 'large' breakpoint could be considered.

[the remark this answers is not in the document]
matthillco: I agree, we should avoid them where possible and I think it's probably time to revisit our defined breakpoints and how we use them.

andysellick (approved):

PR #4115, opened 2024-06-17
First proposed solution for the NoMethodError exception (this PR is just for discussion)

unoduetre:

Check: https://govuk-frontend-app-pr-4115.herokuapp.com/renew-driving-licence-at-70/something_funny_to_put_into_the_error_message.js

## Why

[Trello card?](url)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,8 @@ class ApplicationController < ActionController::Base
   include Slimmer::Template
   slimmer_template "gem_layout"
 
+  before_action :force_html_format_before_action
+
   before_action do
     I18n.locale = I18n.default_locale
   end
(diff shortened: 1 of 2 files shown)


app/controllers/transaction_controller.rb
```diff
@@ -2,6 +2,8 @@ class TransactionController < ContentItemsController
   include Cacheable
   include LocaleHelper
 
+  force_html_format
```
unoduetre, line 5: This is how it will look like in controllers. It can also be added to the parent controller.

app/controllers/application_controller.rb
```diff
@@ -27,6 +29,37 @@ class ApplicationController < ActionController::Base
 
   helper_method :content_item, :content_item_hash, :publication
 
+  def self.force_html_format(only: nil, except: nil)
```
unoduetre, line 32: I wonder if it would make sense to extract it into a gem, so we could use it in other apps as well.

unoduetre:
If we agree that's a good approach, I will also add specific format exceptions just in case all formats need to be forced to html besides some.

app/controllers/application_controller.rb
```diff
@@ -23,10 +25,45 @@ class ApplicationController < ActionController::Base
...
 
+  class << self
+  protected
+
+    def force_html_format(only: nil, except: nil)
+      only_copy = only && only.collect(&:to_sym).freeze
+      except_copy = except && except.collect(&:to_sym).freeze
+      define_method(:force_html_format_only) { only_copy }
+      define_method(:force_html_format_except) { except_copy }
+    end
```
hannako, line 36: I can't follow what this code is doing Mat sorry, can you make it more verbose so it's easier to follow?

I'm also wondering more broadly if this level of solution is necessary. I'm wondering if we can do something simpler like having a before action in the application controller to check if the `request.format.js?` and if it is, handling that with a suitable error. 

I'm tagging @KludgeKML here for his opinion as well

unoduetre: Hi. I'll improve the code in the final version. Let me just explain it here now. It created two methods, the first one containing an array of controller actions that should be forced to html (only), the second one containing an array of controller action which should be excepted from being forced to html (except).

With respect to the further issues you mentioned, the bot makes js requests, so if you are ok with maybe returning 415 (Unsupported Media Type) for all non-html requests, should I do that instead of what is suggested in this PR?

unoduetre: @hannako @KludgeKML I've created another PR containing Jess' solution. I think that's better than this one, so I'm closing this PR.

PR #4123, opened 2024-06-19
Turn CSRF protection on by default

hannako:
The rails default CSRF protection was disabled in https://github.com/alphagov/frontend/pull/1317

The POST routes in this application are postcode lookups to an external API. So we don't currently need CSRF protection, but this is being flagged by the scanner as a vulnerability. Let's turn it back on across the application, and just disable it on four specific routes. This means if the application changes in the future we aren't caught out (assuming we're covered by default CSRF protection).

https://trello.com/c/HQaHFTdc/2580-review-and-potentially-fix-code-scanning-alerts-for-our-repos-m, [Jira issue NAV-12235](https://gov-uk.atlassian.net/browse/NAV-12235)

### DIFF
--- a/app/controllers/find_local_council_controller.rb
+++ b/app/controllers/find_local_council_controller.rb
@@ -5,6 +5,7 @@ class FindLocalCouncilController < ContentItemsController
   include SplitPostcodeSupport
 
   skip_before_action :set_locale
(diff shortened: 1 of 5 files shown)


KludgeKML (approved):
LGTM!

PR #4125, opened 2024-06-19
Remove Universal Analytics

AshGDS:
## What / Why
- Removes UA/All references to UA from the whole `frontend` codebase
- I've tried to split it into commits to make it a bit easier to review
- https://trello.com/c/fKGle0gD/187-remove-ua-code-from-applications, [Jira issue PNP-8334](https://gov-uk.atlassian.net/browse/PNP-8334)


andysellick (commented):
Looks good. 

I've found a few instances of `gem-track-click` in [en.yml](https://github.com/alphagov/frontend/blob/main/config/locales/en.yml#L110) that should be removed.

Also [this reference](https://github.com/alphagov/frontend/blob/main/app/views/transaction/show.html.erb#L24) to the old scroll tracker?

app/views/transaction/_additional_information_tabbed.html.erb
```diff
@@ -1,4 +1,4 @@
-<div data-module="track-start-page-tabs">
+<div>
```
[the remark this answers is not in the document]
AshGDS: Thanks, found a page that uses this http://127.0.0.1:3005/renew-driving-licence-at-70 removed the div

AshGDS:
Thanks @andysellick - I had `.yml` files excluded from my VSCode search from a previous trello card, and I've removed that extra meta tag so it should be ready for another review 👍

app/views/transaction/_additional_information_tabbed.html.erb
```diff
@@ -1,4 +1,3 @@
-<div data-module="track-start-page-tabs">
   <%
```
andysellick, line 1: Nitpick: now that the wrapping DIV is gone, should this whole file be un-indented?

andysellick (approved):
Looks good 👏

app/views/transaction/_additional_information_tabbed.html.erb
AshGDS: @andysellick Thanks, done - good spot

PR #4127, opened 2024-06-22
Fix link formating in README.MD

theomiddleton:
## What

Line 84 had an extra close bracket, meaning the url was displayed along with the title, when it should have been hidden. Via [Markdown Guide syntax](https://www.markdownguide.org/basic-syntax/#links)


brucebolt (approved):
Thank you for raising this PR @theomiddleton 🎖️

PR #4136, opened 2024-06-28
 Check ruby test coverage before merging PRs

unoduetre:

## What

Set up automated checking of ruby test coverage by simplecov as a requirement before a PR can be merged. This can be done in two ways. Either a test can be added, which checks simplecov coverage or an additional step can be added to GitHub CI. Investigate which of those two approaches is easier to implement and implement it.

## Why

We need to keep ruby code test coverage at 95% as indicated by simplecov in order to maintain the requirement of dependabot automerger.

[Trello card](https://trello.com/c/c3Cvewr6/2665-check-ruby-test-coverage-before-merging-prs), [Jira issue NAV-12417](https://gov-uk.atlassian.net/browse/NAV-12417)

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,5 +1,12 @@
 require "simplecov"
-SimpleCov.start "rails"
+SimpleCov.start "rails" do
+  enable_coverage :branch
+  minimum_coverage 95
+  add_filter "/lib/api_error_routing_constraint.rb"
+  add_filter "/lib/bank_holiday_generator.rb"
+  add_filter "/lib/format_routing_constraint.rb"
+  add_filter "/lib/frontend.rb"
+end
 
 require "i18n/coverage"
 require "i18n/coverage/printers/file_printer"


test/test_helper.rb
```diff
@@ -1,5 +1,12 @@
 require "simplecov"
-SimpleCov.start "rails"
+SimpleCov.start "rails" do
+  enable_coverage :branch
+  minimum_coverage 95
+  add_filter "/lib/api_error_routing_constraint.rb"
+  add_filter "/lib/bank_holiday_generator.rb"
+  add_filter "/lib/format_routing_constraint.rb"
+  add_filter "/lib/frontend.rb"
```
hannako, line 8: Why are we not checking coverage on these files? 

We've got a test https://github.com/alphagov/frontend/blob/677081b87c10bd49d7d57e9695d7f78553a3c140/test/unit/format_routing_constraint_test.rb#L3 - is not covering everything?

hannako: We should add tests I think rather than skip checking them when we calculate coverage

unoduetre: Yes, you're right, but those files were excluded because simplecov doesn't register they're test-covered (they are). I suppose that's because they're loaded before the testing framework by Rails. After we move to RSpec (I think the other half or the web team is in the progress of doing that), those can be included again.

The routing constraint files are loaded by `config/routes.rb` and tested by `format_routing_constraint_test.rb` and `api_error_routing_constraint_test.rb`.
`frontend.rb` is loaded by some initializers in config e.g. `config/initializers/location_api.rb`. There is not specific test, but the file only adds accessors, so it's covered by the initializers (Rails won't start properly, if those don't work).
 `bank_holiday_generator.rb` is only used in a rake task and tested by `bank_holiday_generator_test.rb`.

coverage

hannako: Ah nice ok. So let's make a card to update the simple cov config. The card will have a dependency on Frontend running RSpec https://github.com/alphagov/frontend/pull/4090

hannako (approved):

test/test_helper.rb
unoduetre: I've added a Trello card for that: https://trello.com/c/WE1eBmlf/1079-reenable-test-coverage-checking-for-excluded-files-after-migrating-to-rspec

PR #1414, opened 2018-02-16
Modernise javascript test execution with some deps bumps

kevindew:
The bulk of this PR swaps the running of the jasmine tests from being a homebrew implementation with an old version of Jasmine to the use of Jasmine 3 via [Jasmine Gem](https://github.com/jasmine/jasmine-gem).

Doing this has unblocked updating to [govuk_app_config](https://github.com/alphagov/govuk_app_config) 1.3 - so the logstasher stuff has been removed and the unicorn config updated.

Finally in a different approach to https://github.com/alphagov/content-tagger/pull/642 I've added foreman to run the app in docker environment via installing it as a gem in a dockerfile which is the approach recommended by foreman - I'm going to be evaluating if this works out easier than adding it to gemfiles or not.

This will need rebasing when https://github.com/alphagov/frontend/pull/1403 is merged - or vice versa if this goes first.

/cc @sihugh

PR #4142, opened 2024-07-02
Implement JavaScript test coverage checking

unoduetre:

## What

Investigate ways of adding a JavaScript test coverage tool to our apps. Potential problems might involve the fact that some of our JavaScript tests run in a browser, so it might be difficult to make the test coverage tool aware of those tests.

If adding a test coverage tool turns out to be possible, discuss and implement the proposed solution.

## Why

In order to automatically check that the requirement of 95% test coverage is met, which is the requirement of https://github.com/alphagov/govuk-dependabot-merger which we use in our repositories to lessen the burden of the developers in relation to merging dependencies upgrades.

[Trello card](https://trello.com/c/6cX3Xipx/2664-investigate-adding-javascript-test-coverage-and-implement-it-if-possible-l), [Jira issue NAV-15267](https://gov-uk.atlassian.net/browse/NAV-15267)

## Dependencies

This PR depends on restructuring JavaScript files, so the code that should be checked for test coverage is separated from the third-party code e.g. components.

The PR where restructuring was done is https://github.com/alphagov/frontend/pull/4164

## Further info

The options of `.nycrc` are documented here: https://github.com/istanbuljs/nyc?tab=readme-ov-file#common-configuration-options

The following options have been set for the following reasons:
* `all=true` - instrument all JavaScript assets in the directory for storing test coverage;
* `exclude=...` - but not the dependencies, e.g. components that should be tested elsewhere;
* `branches/lines=95` - specify required test coverage (`lines` test coverage check is not strictly necessary because we check branches as well, but it doesn't hurt);
* `check-coverage=true` - returns unsuccessful exit code if test coverage requirements are not met (setting this to `false` would allow us to publish an urgent change in the unlikely scenario that we needed to override the code coverage check to do so; alternatively the `/* istanbul ignore next */` syntax could be used to exclude lines of code from the code coverage check) 
* `compact=false` - does not minify the resulting instrumented files;
* `report-dir=...` - specify where the html report should be stored;
* `reporter=...` - show text summary and create the html report;
* `source-map=false` - I had to disable that as it didn't seem to work with our source maps.

### Text summary

Screenshot 2024-07-18 at 14 51 15

### HTML report

Screenshot 2024-07-18 at 14-53-11 Code coverage report
Screenshot 2024-07-18 at 14-53-34 Code coverage report


app/assets/javascripts/views/travel-advice.js
```diff
@@ -147,14 +148,11 @@
     }
   }
 
-  CountryFilter.prototype.pagePath = function () {
-    window.location.pathname.split('/').pop()
```
unoduetre, line 151: This function is not used anywhere and it cannot work, because it doesn't do anything (the "return" is missing). It also doesn't have any tests.

config/application.rb
```diff
@@ -48,12 +48,6 @@ class Application < Rails::Application
     # Enable the asset pipeline
     config.assets.enabled = true
 
-    config.assets.precompile += %w[
```
unoduetre, line 51: Moved to manifest.js

spec/support/reporters/test-coverage.mjs
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  jasmineDone() {
+    const thisFileDirname = path.dirname(new URL(import.meta.url).pathname);
+    const rootDirname = path.join(thisFileDirname, '..', '..', '..');
+    const reportPath = path.join(rootDirname, 'coverage', 'javascript');
+    const nycOutputPath = path.join(rootDirname, '.nyc_output');
+    const assetsPath = path.join(nycOutputPath, 'assets');
+    const testCoveragePath = path.join(nycOutputPath, 'out.json');
+    return Promise.all([
```
unoduetre, line 17: Remove previous report and nyc-instrumented assets as they're not useful anymore and by removing them we make it more error-prone to run it multiple times, as they files related to the previous run are removed.

app/assets/javascripts/all.js
```diff
@@ -1,5 +0,0 @@
-// This is a manifest file used by the runner for JavaScript unit tests.
```
unoduetre, line 1: As confirmed by @MartinJJones, this file is not used anymore, so it can be removed.

config/application.rb
MartinJJones: Just some history on modernizing the JS test approach in frontend app:
- https://github.com/alphagov/frontend/pull/1414 - 2018
added to jasmine.yml file
- https://github.com/alphagov/frontend/pull/3221 - 2022
jasmine.yml file removed, app.js not moved to jamine-browser.json

app/assets/javascripts/all.js
MartinJJones: Just some history on modernizing the JS test approach in frontend app:
- https://github.com/alphagov/frontend/pull/1414 - 2018
added to jasmine.yml file
- https://github.com/alphagov/frontend/pull/3221 - 2022
jasmine.yml file removed, app.js not moved to jamine-browser.json

spec/support/reporters/test-coverage.mjs
```diff
@@ -8,11 +8,14 @@ export default class TestCoverageReporter {
   }
 
   jasmineDone() {
-    const thisFileDirname = path.dirname(new URL(import.meta.url).pathname)
-    const nycOutputPath = path.join(thisFileDirname, '..', '..', '..', '.nyc_output');
+    const thisFileDirname = path.dirname(new URL(import.meta.url).pathname);
+    const rootDirname = path.join(thisFileDirname, '..', '..', '..');
```
hannalaakso, line 12: I'm wondering if there's another way, maybe via Node, to get this root directory name. I'm just thinking that if this file gets moved, this path won't work anymore.

hannalaakso, line 12: I'm wondering if there's another way, maybe via Node, to get this root directory name. I'm just thinking that if this file gets moved, this path won't work anymore.

spec/support/reporters/test-coverage.mjs
hannalaakso: This seems like really useful information, should it be a code comment so that it gets retained for the next developer?

spec/support/reporters/test-coverage.mjs
hannalaakso, line 1: Just a nit pick, but it looks like this commit could be squashed into the previous one as it's amending the new lines added in this PR.

unoduetre: Done

app/assets/javascripts/views/travel-advice.js
unoduetre: Removal of this function has been extracted to a separate PR.

spec/support/reporters/test-coverage.mjs
unoduetre: Hmm, I wonder, if a code comment is the best idea. Could it be documented somewhere else? As this code will be shared between repos in the future, what would be the best place? Where would you search for this kind of info?

config/application.rb
unoduetre: This and related code has been extracted to a separate PR.

app/assets/javascripts/all.js
unoduetre: Removal of all.js has been extracted to a separate PR.

spec/support/reporters/test-coverage.mjs
unoduetre: Or alternatively we could use import.meta.resolve with a toplevel file.

spec/support/reporters/test-coverage.mjs
unoduetre: *sorry, I meant **less** error-prone of course

spec/support/jasmine-browser.json
```diff
@@ -1,16 +1,20 @@
...
     "**/*[sS]pec.js"
   ],
   "helpers": [
+    "helpers/*.mjs",
     "vendor/*.js"
   ],
-  "browser": "headlessChrome"
+  "browser": "headlessChrome",
+  "reporters": [
+    "./spec/support/reporters/test-coverage.mjs"
```
MartinJJones, line 18: Could we update the file name to include the name of the reporter, for example: `nyc-test-converage.js`

spec/javascripts/helpers/test-coverage.mjs
```diff
@@ -0,0 +1,9 @@
+window.__coverage__ ??= {};
+
+class TestCoverageReporter {
+  static suiteDone(result) {
```
hannalaakso, line 4: Could we add a code comment here, something along the lines of:
`Runs after every top level Jasmine describe-block`

spec/javascripts/helpers/test-coverage.mjs
```diff
@@ -0,0 +1,9 @@
...
+    jasmine.getEnv().setSuiteProperty("__coverage__", window.__coverage__);
```
hannalaakso, line 5: Could we add a code comment here eg. `Transfer the property from the browser to Node`

spec/javascripts/helpers/test-coverage.mjs
```diff
@@ -0,0 +1,9 @@
+window.__coverage__ ??= {};
```
hannalaakso, line 1: Could we add a code comment here at the top of the file to say:
`This file runs in the browser`

spec/support/reporters/test-coverage.mjs
```diff
@@ -0,0 +1,24 @@
+import fsPromises from 'node:fs/promises';
+import {URL} from 'url';
+import path from 'node:path';
+
+export default class TestCoverageReporter {
+  suiteDone({properties}) {
```
hannalaakso, line 6: Could we add a code comment here, something along the lines of:
`Runs after every top level Jasmine describe-block`

spec/support/reporters/test-coverage.mjs
hannalaakso: As we discussed, having this as a code comment probably makes the most sense right now. Although it's possible that in the future we would have some designated place for docs once we share the functionality between repos.

Maybe we could shorten the code comment to say something like: 

```
// Remove the previous report and the nyc-instrumented assets as they are not needed after the run has completed.
```

unoduetre: Done

spec/support/jasmine-browser.json
unoduetre: Done

spec/javascripts/helpers/test-coverage.mjs
unoduetre: Done

spec/javascripts/helpers/test-coverage.mjs
unoduetre: Done.

spec/javascripts/helpers/test-coverage.mjs
unoduetre: Done.

spec/support/reporters/test-coverage.mjs
unoduetre: Done. I checked and it runs after every Jasmine block, not only top-level ones.

spec/support/reporters/nyc-test-coverage-reporter.mjs
```diff
@@ -0,0 +1,32 @@
...
+    const assetsPath = path.join(nycOutputPath, 'assets');
+    const testCoveragePath = path.join(nycOutputPath, 'out.json');
+    return Promise.all([
+      /* Remove the previous report and the nyc-instrumented assets
+       * as they are not needed after the run has completed,
+       * and if not removed, they would accumulate and potentially
+       * confuse the developer.
+       */
+      fsPromises.rm(reportPath, {recursive: true, force: true}),
+      fsPromises.rm(assetsPath, {recursive: true, force: true}),
```
MartinJJones, line 28: After looking at this further we should be able to use the `delete` option in nyc to remove the output directory - 

> The --delete option will remove the existing output directory before instrumenting.

https://github.com/istanbuljs/nyc/blob/main/docs/instrument.md#pre-instrumented-source

For example, we would update this line - https://github.com/alphagov/frontend/pull/4142/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R12

From: `"nyc:instrument": "yarn run nyc instrument public/assets/frontend .nyc_output/assets"`

To: "nyc:instrument": "yarn run nyc instrument public/assets/frontend .nyc_output/assets --delete"

.nycrc
```diff
@@ -0,0 +1,22 @@
...
+    "public/assets/frontend/manifest-*",
+    "public/assets/frontend/application-*",
+    "public/assets/frontend/test-dependencies-*",
+    "public/assets/frontend/dependencies-*",
+    "public/assets/frontend/es6-components-*"
+  ],
+  "report-dir": "coverage/javascript",
+  "reporter": [
+    "text",
+    "html"
```
MartinJJones, line 19: I wonder if we should remove the `html` option from the base level config for now for the reasons below:

- It appears that the HTML report will also be created when running the Continuous Integration checks but is not required here
- We can always add this feature in the future for use in a development environment by creating a custom nyc dev config - https://github.com/istanbuljs/nyc?tab=readme-ov-file#adding-your-overrides
- Would allow us to remove some of the code used in the nyc-test-coverage-reporter file used to delete the reports - https://github.com/alphagov/frontend/pull/4142/files#diff-cad9e4a78533bc33cf9c9b72531808b7be8503ad13528c114c664067d243893cR27

package.json
```diff
@@ -8,10 +8,16 @@
     "lint": "yarn run lint:js && yarn run lint:scss",
     "lint:js": "standardx 'app/assets/javascripts/**/*.js' 'spec/javascripts/**/*.js'",
     "lint:scss": "stylelint app/assets/stylesheets/",
-    "jasmine:prepare": "RAILS_ENV=test bundle exec rails assets:clobber assets:precompile",
-    "jasmine:ci": "yarn run jasmine:prepare && yarn run jasmine-browser-runner runSpecs",
+    "rails:precompile-assets": "RAILS_ENV=test bundle exec rails assets:clobber assets:precompile",
+    "nyc:instrument": "yarn run nyc instrument public/assets/frontend .nyc_output/assets",
+    "jasmine:prepare": "yarn run rails:precompile-assets && yarn run nyc:instrument",
+    "nyc:report": "yarn run nyc report",
```
MartinJJones, line 14: Small nitpick: should we group similar scripts together? or put the scripts in alphabetical order?

Non-blocking as it feels more like something we should have a general consensus on and tooling in place to manage this, ES Lint for example - https://eslint.org/docs/latest/contribute/package-json-conventions#order

package.json
```diff
@@ -8,10 +8,16 @@
...
+    "jasmine:ci": "yarn run jasmine:prepare && yarn run jasmine-browser-runner runSpecs && yarn run nyc:report",
     "jasmine:browser": "yarn run jasmine:prepare && yarn run jasmine-browser-runner"
   },
+  "imports": {
+    "#root/*": "./*"
```
MartinJJones, line 19: Depending on if the other changes are made in [nyc-test-coverage-reporter.mjs](https://github.com/alphagov/frontend/pull/4142/files#diff-cad9e4a78533bc33cf9c9b72531808b7be8503ad13528c114c664067d243893c) we may only need a reference to nyc_output, although I can see the benefit of having a reference to the root directory as well, is it possible to add a reference to the nyc_output folder here as well?

Another approach that we could use is creating a JS file that contains the paths to be used in the application, example from the govuk-frontend (Design System)
https://github.com/alphagov/govuk-frontend/blob/a50eb60c9583aea38e82625889fa7917a2f2fad9/shared/config/paths.js

spec/support/reporters/nyc-test-coverage-reporter.mjs
```diff
@@ -0,0 +1,32 @@
...
+
+export default class NycTestCoverageReporter {
+  // The following method runs after every Jasmine describe block.
+  suiteDone({properties}) {
+    this.coverage = properties?.__coverage__ || this.coverage;
+  }
+
+  jasmineDone() {
+    const rootDirname = path.dirname(
+      new URL(import.meta.resolve('#root/package.json')).pathname
```
MartinJJones, line 15: Looking at the documentation in Node it appears that `import.meta.resolve` is marked as experimental, [Stability: 1.2](https://nodejs.org/docs/latest/api/documentation.html#stability-index) - Release candidate. It appears the minimum node version required to use this is 18.20.4, using a version less than this produces the error `TypeError: (intermediate value).resolve is not a function`. Could there be any issues in using this?

MartinJJones:
Changes look good to me overall 👍, nice work, I just have some further suggestions / questions, nothing else to review from a frontend perspective after that, thanks 😃

spec/support/reporters/nyc-test-coverage-reporter.mjs
unoduetre: A good idea, let me test it.

.nycrc
unoduetre: No, continuous integration runs jasmine:ci

Also, developers need html reports in order to check the coverage, otherwise it's difficult to discover what tests needs to be added.

package.json
unoduetre: Yeah, we can change the ordering after it's decided in another tiny PR.

spec/support/reporters/nyc-test-coverage-reporter.mjs
unoduetre: We control the local node version. `import.meta.resolve` is a part of html standard, not node, so there is no risk of this feature being removed or changed: https://html.spec.whatwg.org/multipage/webappapis.html#integration-with-the-javascript-module-system

.nycrc
MartinJJones: > No, continuous integration runs jasmine:ci

`yarn run jasmine:prepare && yarn run jasmine-browser-runner runSpecs && yarn run nyc:report`

This script also appears to run `nyc:report`, at which point does the HTML report also get generated?

package.json
unoduetre: We can also add a specific path instead of using a pattern.

spec/support/reporters/nyc-test-coverage-reporter.mjs
MartinJJones: I was thinking are not running it in a node context here? and if so does the statement from MDN apply:

> Node.js also implements [the import.meta.resolve() function](https://nodejs.org/docs/latest/api/esm.html#importmetaresolvespecifier), but adds an additional parent parameter if you use the --experimental-import-meta-resolve flag.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve#not_an_ecmascript_feature

unoduetre: This relates to an additional parameter to resolve() named "parent". We don't use this experimental feature in this code.

MartinJJones: Thanks for clarifying 👍

spec/support/reporters/nyc-test-coverage-reporter.mjs
unoduetre: Interestingly this options is not documented when one runs `yarn run nyc instrument --help` but it seems to work.

.nycrc
unoduetre: Yes, but nyc:report only runs the reporters specified in this array, so no html report will be generated. You can check yourself by commenting `"html"`, removing the existing report and running `yarn run jasmine:ci` again.

unoduetre: After the discussion we decided to add a separate package.json script for creating the html report and require the user to manually remove the previous report before generating a new one.

package.json
unoduetre: After the discussion we decided to keep the imports solution. Nevertheless, because of getting rid of file removal from the reporter, the reporter can be simplified.

unoduetre: Done

spec/support/reporters/nyc-test-coverage-reporter.mjs
unoduetre: Done

.nycrc
unoduetre: Done

MartinJJones: Thanks, I think we can always look to add another step / something in the build process to help improve the developer experience in the future to avoid manually deleting the files generated by the HTML report

MartinJJones:
Great work on this!, happy to approve from a frontend perspective 👍

unoduetre:
> Great work on this!, happy to approve from a frontend perspective 👍

Thank you!

spec/support/reporters/test-coverage.mjs
unoduetre: Please, check the discussion in one of the Martin's comments below.

I've also combined the "#root" commit with the main commit.

beccapearce (approved):
This all looks good to me!! 🚀 
nice one!

PR #4162, opened 2024-07-17
Use type=module for application.js

AshGDS:
## What / Why
- Use type=module for this application's JS file
- To coincide with https://github.com/alphagov/govuk_publishing_components/pull/4111
- Has [DO NOT MERGE] as we want the above PR live on `static` first before this is merged.
- We will combine `es6-components.js` again as a separate task unless you disagree with this
- To accurately test this you will need to use local static that is using a local version the `govuk_publishing_components` branch linked above. If you test without, this app's `application.js` seems to crash.
- Trello card: https://trello.com/c/Hoa45SMD/141-turn-off-javascript-in-legacy-browsers, [Jira issue PNP-8523](https://gov-uk.atlassian.net/browse/PNP-8523)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -41,7 +41,7 @@
     %>
 
     <%= javascript_include_tag 'test-dependencies.js' if Rails.env.test? %>
-    <%= javascript_include_tag 'application.js', integrity: false %>
+    <%= javascript_include_tag 'application.js', integrity: false, type: "module" %>
     <%= javascript_include_tag 'es6-components.js', type: "module" %>
     <%= yield :extra_javascript %>
     <%= yield :extra_headers %>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -40,7 +40,7 @@
       render_component_stylesheets
     %>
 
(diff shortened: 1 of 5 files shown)


app/views/layouts/application.html.erb
```diff
@@ -41,7 +41,7 @@
     %>
 
     <%= javascript_include_tag 'test-dependencies.js' if Rails.env.test? %>
```
MartinJJones, line 43: I wonder if we should also add type="module" here as well. It appears that test-dependencies.js is used to pull in [JS used by static to replicate the approach used on dev/production](https://github.com/alphagov/frontend/blob/main/app/assets/javascripts/test-dependencies.js#L1-L2), not 100% sure if this would actually make any difference to the tests, but would just align the approach used (to be used) more closely.

MartinJJones (commented):
The code changes look to me 👍

The JS error in the browser relates to the closest polyfill, further info here - https://github.com/alphagov/smart-answers/pull/6846#pullrequestreview-2198973334

I'm happy to give this another look and approve when the new version of the gem is released and included in this PR as this should then resolve the issue.

MartinJJones (approved):

MartinJJones (approved):

PR #4165, opened 2024-07-18
Remove unused pagePath function

unoduetre:

## What

This function is not used in the codebase and it cannot work as there is no "return" statement, so the function returns undefined and doesn't do anything as there are no side effects.

## Why

In preparation for https://github.com/alphagov/frontend/pull/4142

[Trello card](https://trello.com/c/6cX3Xipx/2664-investigate-adding-javascript-test-coverage-and-implement-it-if-possible-l), [Jira issue NAV-15267](https://gov-uk.atlassian.net/browse/NAV-15267)

### DIFF
--- a/app/assets/javascripts/views/travel-advice.js
+++ b/app/assets/javascripts/views/travel-advice.js
@@ -147,10 +147,6 @@
     }
   }
 
-  CountryFilter.prototype.pagePath = function () {
-    window.location.pathname.split('/').pop()
-  }
-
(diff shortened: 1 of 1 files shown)


MartinJJones (commented):
~Nice work!, would it be OK to merge this change into `main` instead?~

Thanks @unoduetre for explaining this will automatically switch to main 👍

MartinJJones (approved):

PR #4181, opened 2024-07-30
Disable corrections for homepage search field

csutter:
## What
Disable corrections for homepage search field

## Why
As part of our work on site search, we've discovered a pain point where mobile browsers' autocorrection features mess up user queries, especially when it comes to domain specific terminology. For example, searches for "HMRC" are frequently autocorrected to "Hercules", or searches for "SORN" to "sworn".

See: https://github.com/alphagov/govuk_publishing_components/pull/4112

## How

- Bump `govuk_publishing_components` to 40.1.0
- Set `disable_corrections` parameter for `search` component in homepage header

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -162,7 +162,7 @@ GEM
     govuk_personalisation (0.16.0)
       plek (>= 1.9.0)
       rails (>= 6, < 8)
-    govuk_publishing_components (40.0.0)
+    govuk_publishing_components (40.1.0)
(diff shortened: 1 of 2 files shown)


andysellick (approved):
Change looks good to me. Apologies if I'm jumping in with a review here, I'm trying to get all our apps updated to the latest components gem ahead of a new breaking release that I'm about to do.

unoduetre (approved):
Looks good to me

PR #4214, opened 2024-08-30
Upgrade Rails to version 7.2.1

unoduetre:

## What

Upgrade frontend to use the latest version of Rails (7.2.1).

## Why

[Trello card](https://trello.com/c/EfS9DHDw/2784-update-our-applications-to-rails-72-l), [Jira issue NAV-15354](https://gov-uk.atlassian.net/browse/NAV-15354)


bin/bundle
```diff
@@ -1,3 +1,114 @@
 #!/usr/bin/env ruby
```
hannako, line 1: What is this change? It doesn't appear to have come from running `rails app:update`

unoduetre: If I remember correctly, it was suggested while running `rails g` when I tested the generator functionality with rubocop.

The output is the result of running `bundle binstubs bundler --force`.

unoduetre: (You can check my comments in the corresponding PR in government-frontend as well.)

unoduetre: I did a quick check now, and it seems it doesn't produce the warning for 7.2.1, so I think this part can be reverted.

unoduetre: I've reverted the changes to bin/bundle here and in government-frontend as discussed.

bin/setup
```diff
@@ -17,12 +17,13 @@ FileUtils.chdir APP_ROOT do
   system! "gem install bundler --conservative"
   system("bundle check") || system!("bundle install")
 
-  # Install JavaScript dependencies
-  system! 'bin/yarn'
-
```
beccapearce, line 22: Just want to check that this is ok to remove?

unoduetre: This line was added by 2963c7147490bd5e83d44fbc67943a2080c58a40 with the commit message: "Apply rails app:update changes", so it looks like a change suggested by app:update and not a separately introduced thing. The current app:update run suggests to remove it. I followed the suggestion.

It might be related to decoupling Rails from JavaScript in the newer versions, as I think there was a period when the direction of Rails was supposedly to integrate more tightly with JavaScript package managers etc. but nowadays Hotwire seems to be the direction.

What's your opinion on that?

beccapearce: Sounds good to me! I'll approve

beccapearce (approved):

PR #4221, opened 2024-09-04
Convert content_item from simple response to Model

KludgeKML:

## What

When creating the content_item, always instantiate it as a model rather than just passing back the response from GdsApi.

## Why

More complex behaviours for content items will be necessary for the move of document types from government frontend, and it's better to have a single behaviour for content items in frontend rather than then sometimes being an API response and sometimes a model.

https://trello.com/c/0LYU5bSd/318-move-route-document-type-takepart-from-governmentfrontend-to-frontend, [Jira issue PNP-8559](https://gov-uk.atlassian.net/browse/PNP-8559)

## How

A factory class creates the model instances from the API response - in future this will use the document_type to instantiate specific classes. Some aspects of the response become attributes on the model, and small changes in code accommodate this.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -1,0 +1,17 @@
+class ContentItem
+  attr_reader :content_store_response, :body, :image, :description, :document_type, :title, :base_path, :locale
+
+  def initialize(content_store_response)
+    @content_store_response = content_store_response
+    @body = content_store_response.dig("details", "body")
+    @image = content_store_response.dig("details", "image")
+    @description = content_store_response["description"]
+    @document_type = content_store_response["document_type"]
+    @title = content_store_response["title"]
+    @base_path = content_store_response["base_path"]
+    @locale = content_store_response["locale"]
+  end
+
+  delegate :to_h, to: :content_store_response
+  delegate :cache_control, to: :content_store_response
+end
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -22,7 +22,7 @@ private
   end
 
   def content_item
-    @content_item ||= request.env[:content_item] || request_content_item(content_item_slug || "/#{params[:slug]}")
+    @content_item ||= ContentItemFactory.build(request.env[:content_item] || request_content_item(content_item_slug || "/#{params[:slug]}"))
   end
 
   def content_item_slug
@@ -53,8 +53,8 @@ private
   end
 
   def set_locale
-    I18n.locale = if content_item["locale"] && I18n.available_locales.map(&:to_s).include?(content_item["locale"])
-                    content_item["locale"]
+    I18n.locale = if content_item.locale && I18n.available_locales.map(&:to_s).include?(content_item.locale)
+                    content_item.locale
                   else
                     I18n.default_locale
                   end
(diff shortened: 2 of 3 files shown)


app/models/content_item.rb
```diff
@@ -0,0 +1,53 @@
...
+    translations = content_item["links"]["available_translations"] || []
+
+    mapped_locales(sorted_locales(translations))
+  end
+
+  def sorted_locales(translations)
+    translations.sort_by { |t| t["locale"] == I18n.default_locale.to_s ? "" : t["locale"] }
+  end
+
+  def show_default_breadcrumbs?
```
leenagupte, line 25: This feels like a presentation concern rather than something that should be in the model.

app/models/content_item.rb
```diff
@@ -0,0 +1,53 @@
...
+    true
+  end
+
+  def parsed_content_item
```
leenagupte, line 29: What is this expected to be used for?

app/models/content_item.rb
```diff
@@ -0,0 +1,53 @@
...
+    content_item
+  end
+
+  delegate :to_h, to: :content_item
+  delegate :cache_control, to: :content_item
```
leenagupte, line 34: I know this is needed by `content_item_hash`, I'm not clear where `to_h` is being delegated to. Aren't we already in content_item?

app/models/content_item.rb
```diff
@@ -0,0 +1,53 @@
...
+        locale: translation["locale"],
+        base_path: translation["base_path"],
+        text: native_language_name_for(translation["locale"]),
+      }.tap do |h|
+        h[:active] = true if h[:locale] == I18n.locale.to_s
+      end
+    end
+  end
+
+  def native_language_name_for(locale)
```
leenagupte, line 50: I think this too is a presentation concern.

leenagupte (commented):
I really like the idea of having a ContentItem model, but I have a few questions about the implementation.

app/models/content_item.rb
KludgeKML: Removed (not necessary for this small change)

app/models/content_item.rb
KludgeKML: Removed (not necessary for this small change)

app/models/content_item.rb
KludgeKML: Removed (not necessary for this small change)

app/models/content_item.rb
KludgeKML: Have renamed the internal copy of the api response so these lines are less confusing.

app/models/content_item.rb
```diff
@@ -0,0 +1,17 @@
...
+    @image = content_store_response.dig("details", "image")
+    @description = content_store_response["description"]
+    @document_type = content_store_response["document_type"]
+    @title = content_store_response["title"]
+    @base_path = content_store_response["base_path"]
+    @locale = content_store_response["locale"]
+  end
+
+  delegate :to_h, to: :content_store_response
+  delegate :cache_control, to: :content_store_response
```
leenagupte, line 16: Oh! Now I understand! Thanks for renaming this.

leenagupte (approved):
The change looks good to me. This should make it easier to move to using models to model the response from the content-api, rather than presenters 🎉 

I tried out a few routes in the review app and they all look good 🪇

PR #4223, opened 2024-09-09
Update calendar code

sairamya93:

## What

Adds locale values for When do the clock change file, moving them from the JSON files, and remove the hack file that handled returning either a translated key (for keys in the locale files), or the original file (for hardcoded text)

We also remove the task for publishing the calendars from frontend and add the relevant data into special route publisher (here:  https://github.com/alphagov/special-route-publisher/pull/351). In order to do that, we make the description of when-do-the-clocks-change more generic (removing the list of years), and remove the body (since the body is not present for the bank holiday pages, and appears not to be used in search).

- Also sets locale correctly in ICS files, and serves Welsh content in ICS if accessed from welsh page.
- Also removes an unused bank holiday generator task

## Why

[Trello card?] https://trello.com/c/dx543EiX/322-tidy-when-do-the-clocks-change-files, [Jira issue PNP-8592](https://gov-uk.atlassian.net/browse/PNP-8592)

## How

We should handle the when do the clocks change page the same way as the bank holidays one, which would allow us to translate it accordingly in the config file and lets us remove the hack file


app/controllers/calendar_controller.rb
```diff
@@ -13,9 +13,9 @@ class InvalidCalendarScope < StandardError; end
   def show_calendar
     respond_to do |format|
       format.html do
-        @faq_presenter = FaqPresenter.new(scope, calendar, content_item_hash, view_context)
+        @faq_presenter = FaqPresenter.new(Calendar.file_name(params[:scope]), calendar, content_item_hash, view_context)
```
leenagupte, line 16: Why `FaqPresenter` being instantiated with `Calendar.file_name(params[:scope])` now rather than `calendar`? Are they different?

app/models/calendar.rb
```diff
@@ -14,21 +14,34 @@ def self.find(slug)
     end
   end
 
-  attr_reader :slug, :title, :description
+  def self.file_name(slug)
+    slug == "gwyliau-banc" ? "bank-holidays" : slug
```
leenagupte, line 18: It feels like this should return the file name rather than the slug.

app/models/calendar.rb
```diff
@@ -14,21 +14,34 @@ def self.find(slug)
...
+  end
+
+  def years_covered
```
leenagupte, line 21: I think this method is missing a test.

leenagupte (commented):
@KludgeKML, @sairamya93 This looks like a worthwhile change, and thanks for making a start with it.

There's a lot going on in the second commit, so it would be great it could be broken up so that I can follow what's changing. Could you also add the "why" to the commit messages?

app/models/calendar.rb
KludgeKML: Have reengineered to return "scope", which is how the bulk of the code refers to "bank-holidays" vs "when-the-clocks-change" regardless of translation.

leenagupte (approved):
Thanks for breaking up the commits and adding the test. That made these changes much easier to follow.

That was a lot to unpick, so well done for figuring it all out. The use of "scope" is a bit confusing, but you've managed to make it easier to understand 🎉 

There's one really minor comment that's not a blocker, otherwise it all looks really good 👍

app/controllers/calendar_controller.rb
```diff
@@ -31,7 +31,7 @@ def division
 
     respond_to do |format|
       format.json { render json: div }
-      format.ics { render plain: IcsRenderer.new(div.events, request.path).render }
+      format.ics { render plain: IcsRenderer.new(div.events, request.path, I18n.locale).render }
```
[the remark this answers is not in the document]
KludgeKML: set_locale is what makes I18n.locale have the right values, so it's needed (and it does affect the json)

app/models/calendar.rb
KludgeKML: Method no longer exists, so marking this as resolved.

app/controllers/calendar_controller.rb
KludgeKML: Fixed to return calendar.scope

leenagupte (approved):
👍

PR #4225, opened 2024-09-10
Move Travel Advice country pages

leenagupte:
, [Jira issue PNP-8514](https://gov-uk.atlassian.net/browse/PNP-8514)⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

Copy the rendering code for the travel advice pages from government-frontend. 
The rendering code has been modified to fit the best practices for frontend as outlined in the Contributing guide.

## Why

[Trello card](https://trello.com/c/wGMX5KPk)

## How

This routes have a fixed prefix, how a scoped route wasn't added for them as they needed to accommodate the pre-existing index page as well as a print routes and multiple parts.

The download link component has also been ported across, as well as the CSS for the "responsive-bottom-margin" as that is used to add whitespace between the print link and the horizontal rule above the "Explore more" section.

Generic helpers have also been added for atom and print links.

## Screenshots
Rendered version: https://govuk-frontend-app-pr-4225.herokuapp.com/foreign-travel-advice/albania

|government-frontend (before)| frontend (after)|
|-------|--------|
|screencapture-gov-uk-foreign-travel-advice-afghanistan-2024-11-18-15_03_38|screencapture-frontend-dev-gov-uk-foreign-travel-advice-afghanistan-2024-11-18-15_29_20|
|screencapture-gov-uk-foreign-travel-advice-afghanistan-entry-requirements-2024-11-18-15_03_57|screencapture-frontend-dev-gov-uk-foreign-travel-advice-afghanistan-entry-requirements-2024-11-18-15_29_38|
|Screenshot 2024-11-18 at 15 04 24|Screenshot 2024-11-18 at 15 29 54|

### DIFF
--- a/app/views/travel_advice/_first_part.html.erb
+++ b/app/views/travel_advice/_first_part.html.erb
@@ -1,0 +1,11 @@
+<%= render 'govuk_publishing_components/components/metadata', content_item.metadata %>
+<% if content_item.map %>
+  <figure class="map">
+    <img src="<%= content_item.map["url"] %>" alt="<%= content_item.map["alt_text"] %>" class="map-image">
+    <% if content_item.map_download_url %>
+      <figcaption>
+        <%= render 'components/download_link', href: content_item.map_download_url, link_text: "Download a more detailed map (PDF)" %>
+      </figcaption>
+    <% end %>
+  </figure>
+<% end %>
--- a/app/views/travel_advice/_first_part.html.erb
+++ b/app/views/travel_advice/_first_part.html.erb
@@ -1,4 +1,14 @@
-<%= render 'govuk_publishing_components/components/metadata', content_item.metadata %>
+<%
+  latest_update = simple_format(@travel_advice_presenter.latest_update, { class: "metadata__update" }, wrapper_tag: "span") if @travel_advice_presenter.latest_update.present?
+%>
+<%= render 'govuk_publishing_components/components/metadata', {
+    other: {
+      I18n.t("formats.travel_advice.still_current_at") => I18n.l(Time.zone.now, format: "%-d %B %Y"),
+      I18n.t("formats.travel_advice.updated") => display_date(content_item.reviewed_at || content_item.updated_at),
+      "Latest update" => latest_update
+    }
+  } %>
+
 <% if content_item.map %>
   <figure class="map">
     <img src="<%= content_item.map["url"] %>" alt="<%= content_item.map["alt_text"] %>" class="map-image">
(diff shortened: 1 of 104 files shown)


config/routes.rb
```diff
@@ -28,6 +28,7 @@
   get "/find-local-council/:authority_slug" => "find_local_council#result"
 
   get "/foreign-travel-advice", to: "travel_advice#index", as: :travel_advice
+  get "/foreign-travel-advice/:country/*slug", to: "travel_advice#show"
```
KludgeKML: Hmm, could we call the part something other than slug? Slug's already got too many meanings, and I think in this case we're going to end up overloading it.

app/views/travel_advice/_first_part.html.erb
```diff
@@ -1,4 +1,14 @@
-<%= render 'govuk_publishing_components/components/metadata', content_item.metadata %>
+<%
+  latest_update = simple_format(@travel_advice_presenter.latest_update, { class: "metadata__update" }, wrapper_tag: "span") if @travel_advice_presenter.latest_update.present?
```
KludgeKML, line 2: I'm mildly suspicious about this, because it feels like a slightly complex bit of code to have in the partial where we don't have coverage info, but I think it's fine.

config/locales/ar.yml
```diff
@@ -103,6 +103,8 @@ ar:
   components:
     figure:
       image_credit: 'حقوق ملكية الصورة: %{credit}'
+    print_link:
```
KludgeKML, line 106: Hmm, this commit has all the locale changes (rather than separate commits for the component locales and the travel advice locales, which we have with the other move PRs)

KludgeKML (commented):
I think the PR in general is good, just a few small comments. I think we also need to think a bit about the mix of uses of content_item and @content_item in views, but that's a wider thing we already know about.

config/routes.rb
leenagupte: Looking through the code I think it's best to keep this as "slug". It's called "slug" in the content item, and I think the part and part helper code will become very confusing if we're using two different terms to refer to the same thing. I think it makes more sense in the context of travel-advice pages for the "slug" to refer to the sub-page of a country, and for it to mean that everywhere.

config/locales/ar.yml
KludgeKML: Ah, are there no translations for the actual component that got moved?

KludgeKML: I see, the component is just a wrapper around an SVG. Okay, no problem then.

KludgeKML (approved):
LGTM

PR #4228, opened 2024-09-11
Improve coverage

KludgeKML:

## What

Various updates to improve coverage, plus bug fixes discovered while improving coverage:
- fix sessions controller always ignoring referer when deciding if a redirect was valid

Current improvement in line coverage: 97.99% -> 98.47%
Current improvement in branch coverage:  86.36% -> 89.39%

## Why

As we move the routes from government-frontend over, the better our existing coverage the easier it will be to spot coverage problems in the new code. Ideally if we have 100% line coverage, any lack of coverage in the new code will stick out like a sore thumb. We should also strive for 100% branch coverage as a secondary goal.

https://trello.com/c/cCAkk9wZ/300-consolidate-4-frontend-apps

### DIFF
--- a/spec/system/sign_in_spec.rb
+++ b/spec/system/sign_in_spec.rb
@@ -2,7 +2,7 @@ RSpec.describe "Sign in" do
   context "rendering the sign-in help page" do
     before do
       payload = {
-        base_path: "/help/sign-in",
+        base_path: "/sign-in",
         format: "special_route",
         title: "Sign in to a service",
         description: "",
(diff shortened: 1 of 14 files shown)


spec/system/sign_in_spec.rb
```diff
@@ -0,0 +1,31 @@
+RSpec.describe "Sign in" do
+  context "rendering the sign-in help page" do
+    before do
+      payload = {
+        base_path: "/help/sign-in",
```
leenagupte, line 5: Non-blocking nitpick. This should be `/sign-in` to match the base_path used in the rest of these tests.

leenagupte (approved):
This is great! ⭐ 

There's only one nitpick in the comments.

spec/system/sign_in_spec.rb
KludgeKML: Oh yeah, d'oh. Thanks!

PR #4241, opened 2024-09-24
Handle root favicon requests

KludgeKML:
This allows us to move the /favicon special route from static to frontend, removing the icon redirect controller from static in preparation for its retirement.

- Remove favicon.ico in public directory as it's not copied to the S3 bucket and it short-circuits matching the route.
- Route high up in the routes file so that we don't have to check for/load a content item to do this redirect.

### DIFF
--- a/app/controllers/favicon_controller.rb
+++ b/app/controllers/favicon_controller.rb
@@ -1,0 +1,12 @@
+# Because favicon has to be present at root, we need a controller
+# to redirect it to the asset. When there's a better solution to this
+# problem we can remove this controller and routes.
+class FaviconController < ApplicationController
+  before_action { expires_in(1.day, public: true) }
+
+  def redirect_to_asset
+    redirect_to(view_context.asset_path("favicon.ico"),
+                status: :moved_permanently,
+                allow_other_host: true)
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,6 +15,9 @@ Rails.application.routes.draw do
   # http://stackoverflow.com/a/3443678
   get "*path.gif", to: proc { |_env| [404, {}, ["Not Found"]] }
 
+  # Favicon redirect
+  get "/favicon.ico", to: "favicon#redirect_to_asset"
+
   unless Rails.env.production?
     get "/development", to: "development#index"
   end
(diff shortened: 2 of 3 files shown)


app/controllers/favicon_controller.rb
```diff
@@ -0,0 +1,12 @@
+# Because favicon has to be present at root, we need a controller
+# to redirect it to the asset. When there's a better solution to this
+# problem we can remove this controller and routes.
+class FaviconController < ApplicationController
+  before_action { expires_in(1.day, public: true) }
+
+  def redirect_to_asset
+    redirect_to(view_context.asset_path("favicon.ico"),
```
hannako, line 8: Sorry I can't keep this in my head it's so complicated!

How does Frontend know where to get that asset from? I would expect there to be an asset called favicon.ico in the assets directory. But there isn't, and I know that the [favicon](https://github.com/alphagov/govuk_publishing_components/blob/f8d8c29d2dded9729368fe5261aec59480a7d8ff/app/assets/images/govuk_publishing_components/favicon-production.png) is in the [gem](https://github.com/alphagov/govuk_publishing_components/blob/f8d8c29d2dded9729368fe5261aec59480a7d8ff/app/views/govuk_publishing_components/components/_layout_for_public.html.erb#L82). How is this wired up?

KludgeKML: When the app compiles its assets, it includes the assets from the gem. So the app has a sprockets manifest that maps a logical name "favicon.ico" to the fingerprinted asset name ("/assets/frontend/favicon-121ab2(etc).ico"). When you ask for the logical name using .asset_path, it looks up the fingerprinted name in the sprockets manifest.

So here we're saying "if someone asks for /favicon.ico, permanently redirect them to /assets/frontend/favicon-be123a(etc).ico".

Here's the same code in Static https://github.com/alphagov/static/blob/main/app/controllers/icon_redirects_controller.rb (where it's doing exactly the same thing, but obviously to "/assets/static/favicon-(blahblah).ico").

hannako: So will you merge this, and then update the Route in Router API, and then remove the [controller](https://github.com/alphagov/static/blob/main/app/controllers/icon_redirects_controller.rb) from static?

hannako (approved):

app/controllers/favicon_controller.rb
hannako: Thanks for explaining!

KludgeKML: No problem!

PR #4247, opened 2024-10-02
Landing pages backend scaffold

KludgeKML:

## What

Add the scaffolding required to render flexible "blocks", where the type of block, and the order of the blocks can be controlled by the user.

## Why

A new flexible content type called a landing page is being created.
The content type will allow publishers to pick the "blocks" that should be included on the page, e.g. govspeak, tabs, image, video, and decide on the order of these "blocks".

## How

The publishing app doesn't exist for these blocks yet, nor does a content schema. 

Therefore, for the time being dummy content is being hard-coded. This will be replaced by actual hard-coded content when the first landing page is ready to go live. The hard-coded content will be removed when it is possible to publish the content from a CMS.

As a content schema doesn't exist for the landing_page type yet, a fake content item has been added to the controller. This will be removed in a later PR once a content schema exists.

## Screenshots?
[Rendered page](https://govuk-frontend-app-pr-4247.herokuapp.com/landing-page)

### DIFF
--- a/lib/data/landing_page_content_items/landing_page.yaml
+++ b/lib/data/landing_page_content_items/landing_page.yaml
@@ -1,0 +1,43 @@
+blocks:
+- type: govspeak
+  content: |
+    <h2>Here's a heading</h2>
+    <p>Here's some content!</p>
+    <p>Here's some more content</p>
+    <ul>
+      <li>What?</li>
+      <li>When?</li>
+    </ul>
+    <a href="/landing-page/sub-page-1">Visit sub-page 1</a>
+- type: tabs
+  id: landing-pages-item-selector
+  tab_items:
+  - id: tab-1
+    label: Item 1
+    content: Here's the content for item one
+  - id: tab-2
+    label: Item 2
+    content: Here's the content for item two
+- type: govspeak
+  content: <p>Here's some more content!</p>
+- type: two_column_layout
+  theme: two_thirds_one_third
+  blocks:
+  - type: govspeak
+    content: <p>Left content!</p>
+  - type: govspeak
+    content: <p>Right content!</p>
+- type: govspeak
+  content: <h2>Statistics</h2>
+- type: columns_layout
+  blocks:
+  - type: big_number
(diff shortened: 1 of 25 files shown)


lib/data/landing_page_content_items/landing_page.yaml
```diff
@@ -0,0 +1,22 @@
+blocks:
+- type: paragraph
+  content: Here's some content!
```
richardTowers, line 3: I was thinking we could use the [govspeak component](https://components.publishing.service.gov.uk/component-guide/govspeak) for things that are a bit long-form-text-like. That way we can keep our blocks to high level constructs, and avoid reproducing paragraphs / lists / tables etc.

richardTowers: Gives the user a lot of flexibility (possibly too much, but I think it's better to err on that side of things for this project)

KludgeKML: Yeah, that makes sense. The paragraph was just for the proof of concept, I'll replace it with a govspeak block.

KludgeKML: Done

KludgeKML: Only remaining question here is for this version, do we allow govspeak in the content for that item or do we stick closer to what we'll get in the end from content store (HTML fragments converted from govspeak, which is what the component takes). I've chosen the latter path for now, I assume we generally do not want to be converting govspeak to HTML in the frontend.

richardTowers: Hmm, that's a good question. It's a bit annoying to have to hardcode it as HTML, but I think that'll do for now. Options for later:

- Hardcode as HTML (current version)
- Hardcode as govspeak, have a build step to swap it out for HTML
- Use a real content item (publishing-api will render the govspeak)

richardTowers (approved):
Nice work both! Looks good to me!

PR #4250, opened 2024-10-04
Help page document

georges1996:

## What

Handle Help page documents in Frontend, in preparation for them being taken out of Government Frontend

## Why

https://trello.com/c/GP0bkcD4/350-move-document-type-helppage-from-government-frontend-to-frontend, [Jira issue PNP-8507](https://gov-uk.atlassian.net/browse/PNP-8507)

## How

These documents all have a fixed prefix, so add a scoped route for them, a controller, a model and a view. Here we add the initial ContentItem model the document type models will be based on, with the minimum functionality required to get Help page documents working.

We also create a couple of helpers for rendering generic content.

## Screenshots?

From example page: https://www.gov.uk/help/browsers

### Before (running in Government Frontend)
Screenshot 2024-10-04 at 07 54 53

### After (running in Frontend)
Screenshot 2024-10-04 at 07 54 35

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -1,6 +1,6 @@
 class ContentItem
   attr_reader :content_store_response, :content_store_hash, :body, :image, :description,
-              :document_type, :schema_name, :title, :base_path, :locale
+              :document_type, :schema_name, :title, :base_path, :locale, :public_updated_at
 
   # SCAFFOLDING: remove the override_content_store_hash parameter when full landing page
   # content items including block details are available from content-store
@@ -16,6 +16,7 @@ class ContentItem
     @title = content_store_hash["title"]
     @base_path = content_store_hash["base_path"]
     @locale = content_store_hash["locale"]
+    @public_updated_at = content_store_hash["public_updated_at"]
   end
 
   alias_method :to_h, :content_store_hash
--- a/app/views/help_page/show.html.erb
+++ b/app/views/help_page/show.html.erb
@@ -1,0 +1,9 @@
+<% content_for :extra_headers do %>
+	<%= render "govuk_publishing_components/components/machine_readable_metadata", { content_item: @content_item.to_h, schema: :article } %>
+
+  <% if @content_item.base_path == '/help/cookie-details' %>
+    <meta name="robots" content="noindex">
+  <% end %>
+<% end %>
+
+<%= render 'shared/body_with_related_links' %>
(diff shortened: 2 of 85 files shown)


spec/helpers/date_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
+RSpec.describe DateHelper do
```
leenagupte, line 1: Nice tests 🎉

app/views/help_page/show.html.erb
```diff
@@ -0,0 +1,9 @@
+<% content_for :extra_head_content do %>
```
leenagupte, line 1: This should be `extra_headers` not `extra_head_content`. It's something that tripped up both myself and Keith.

You can check if the metadata is being displayed by viewing the page source. If the metaproperty appears it's all ok:

```
<meta property="og:description" content="How to upgrade your web browser - improve safety, speed and features">
<meta property="og:title" content="Browsers">
<meta property="og:url" content="https://www.gov.uk/help/browsers">
<meta property="og:type" content="article">
<meta property="og:site_name" content="GOV.UK">
<meta name="twitter:card" content="summary">
```
To combat this Keith started porting over the [meta tags tests](https://github.com/alphagov/frontend/pull/4218/commits/975dca796779ef0a3ec0b5c3d836488164371e3e) as shared examples that could be added to all of the system tests. Perhaps that could be done in this PR instead?

leenagupte (commented):
This is a really good start. I think not being able to merge the Take Part PR means that this one needs some extra work before it's ready.

We've already spoken IRL about linking to the files in government-frontend directly. I think the only commit that needs the links updated is the one that ports over the components.

I've added inline comments for everything else.

app/models/content_item.rb
```diff
@@ -10,6 +10,8 @@ def initialize(content_store_response)
     @title = content_store_response["title"]
     @base_path = content_store_response["base_path"]
     @locale = content_store_response["locale"]
+    @last_updated = content_store_response["public_updated_at"]
```
KludgeKML, line 13: Hmm. I sort of think here we should be keeping the attributes named the same as they are in the content store - like, this should be `public_updated_at`

KludgeKML (approved):
LGTM (but we need a second opinion because I am now implicated in some of this)

leenagupte (commented):
Looks good! 

There are a couple of inline comments about linking to the original files in government-frontend, otherwise it looks great. 🎉

leenagupte (approved):
Looks great 👍

PR #4251, opened 2024-10-07
Create a hero block

richardTowers:

## What

This adds a new "block" for the hero section, consisting of a responsive image (using the `<picture>` tag to allow for art direction at different screen sizes), and some content (which is a list of blocks, as we want to support action_link components, which aren't available in govspeak).

## Why

[Trello card](https://trello.com/c/jQ8M2kPq/41-create-blocks-for-elements-of-the-designs)

## How

See individual commits

## Screenshots?

See heroku prototype - try resizing your browser to see different images. Should get 2x images on Retina screens, 1x images on "normal" DPI screens.


spec/models/block/layout_base_spec.rb
```diff
@@ -17,5 +17,9 @@
     it "builds blocks of the correct type" do
       expect(described_class.new(blocks_hash).blocks.first.type).to eq("big_number")
     end
+
+    it "sets blocks to full width by default" do
```
leenagupte, line 21: The test description doesn't match the expectation.

app/views/landing_page/show.html.erb
```diff
@@ -1,7 +1,13 @@
 
 <div class="landing-page">
   <% @content_item.blocks.each do |block| %>
-    <%= render "landing_page/blocks/#{block.type}", block: %>
+    <% if block.full_width? %>
+      <%= render "landing_page/blocks/#{block.type}", block: %>
+    <% else %>
+      <div class="govuk-width-container">
```
leenagupte, line 7: Minor, and not a blocker, but as the thing being rendered is the same, and it's only and extra style being applied, this check could probably inline.

app/assets/images/landing_page/placeholder_desktop.png
leenagupte, line 1: I understand that these are placeholder images in general. But how do you envision images to be organised for multiple landing pages?
Does `landing_page` in the images path represent the `base_path` of the page or the `document_type`?

app/views/landing_page/blocks/_govspeak.html.erb
leenagupte, line 1: I'm curious about what this means: "Wrapping the publishing components in partials like this shows a bit of a deficiency with our approach, but tolerating it for now." I think this opinion should either be explained fully in the commit message or removed.

app/models/block/hero.rb
```diff
@@ -0,0 +1,21 @@
...
+  HeroImageSources = Data.define(:desktop, :desktop_2x, :tablet, :tablet_2x, :mobile, :mobile_2x)
+  HeroImage = Data.define(:alt, :sources)
+
+  class Hero < Block::Base
+    attr_reader :image, :hero_content
+
+    def initialize(block_hash)
+      super(block_hash)
+
+      alt, sources = block_hash.fetch("image").values_at("alt", "sources")
```
leenagupte, line 11: Minor, and not a blocker, but `data` could be used from this point onwards rather than `block_hash`.

app/models/block/hero.rb
```diff
@@ -0,0 +1,21 @@
+module Block
+  HeroImageSources = Data.define(:desktop, :desktop_2x, :tablet, :tablet_2x, :mobile, :mobile_2x)
```
leenagupte, line 2: What happens if not all of the image sources are provided, or do you think "we" will be expected to create them?

I'm wondering what happens to the page if the images don't exist.

app/views/landing_page/blocks/_hero.html.erb
```diff
@@ -0,0 +1,25 @@
+<%
+  add_view_stylesheet("landing_page/hero")
+%>
+
+<div class="app-b-hero">
+  <%= picture_tag({ class: "app-b-hero__imagewrapper"}) do %>
```
leenagupte, line 6: One of the things I've struggled with in the "scaffolding" PR is what to add into the system tests so that we can test that the various different blocks are rendering as expected. Perhaps a loop around the blocks in the landing page fixture that checks for specific bits of CSS?

Could something like that be added here?

leenagupte (commented):
This is looking good. 🎉 

The commit that aded the hero especially looks good, though I think someone else should review the CSS.

I have a few other questions and comments inline.

app/views/landing_page/blocks/_govspeak.html.erb
richardTowers: So we've got these components in govuk_publishing_components, which are massively configurable (e.g. supporting `inverse` when they're light-on-dark, but they've got loads of params generally). Then we wrap them in a block partial, and it's not possible to configure them at all. Then we actually do need some configurability (like here with inverse), and we extend the block partial to pass stuff through to the component partial. Just feels like a messy abstraction. But it's not a massive problem, so I think we should tolerate it for now.

I can drop that comment from the commit message though.

app/models/block/hero.rb
richardTowers: I was imagining them as kind of mandatory. At the moment if you leave them out of the YAML, this code will raise and error like:

```
ArgumentError
missing keywords: :desktop, :desktop_2x, :tablet, :tablet_2x, :mobile, :mobile_2x
```

spec/models/block/layout_base_spec.rb
richardTowers: Yep, good catch - addressed in cf11d7fe099025e302c174fe06ab44485999d130

app/views/landing_page/show.html.erb
richardTowers: Yeah, I was trying to avoid the extra wrapping div if it's not needed, but actually it's probably helpful for it to always be there. Addressed in 631c1d54e1eebd15c295f57ef37e9acdf8fb0329.

app/assets/images/landing_page/placeholder_desktop.png
richardTowers: The paths just have to match whatever's in the YAML - the code isn't doing anything clever working out paths. So I haven't actually worked out how we're going to arrange them yet.

I think having a directory per-hero-image with all the resolutions in it is probably cleaner for when we've got lots of images though. Moved 'em in 82131eac84b3c5f86fefd0041e159e78d53d7bd6

app/views/landing_page/blocks/_govspeak.html.erb
richardTowers: Done

app/models/block/hero.rb
richardTowers: 13727cdee3ffa59f128e27af893c6d0429d6482f

app/views/landing_page/blocks/_hero.html.erb
richardTowers: Something like this? 03126f6ff4da76a19dc7b0b9f8d8d144926901cc

app/models/block/hero.rb
leenagupte: Ok. I think this will need to be captured somewhere. Perhaps in some yet-to-be-written docs on adding blocks to pages.

leenagupte: Ideally we'd add some exception handling, but that can wait until post-launch.

app/views/landing_page/show.html.erb
```diff
@@ -1,13 +1,9 @@
...
 <div class="landing-page">
   <% @content_item.blocks.each do |block| %>
-    <% if block.full_width? %>
-      <%= render "landing_page/blocks/#{block.type}", block: %>
-    <% else %>
-      <div class="govuk-width-container">
-        <%= render "landing_page/blocks/#{block.type}", block: %>
-      </div>
-    <% end %>
+    <%= tag.div class: ["govuk-block__#{block.type}", ("govuk-width-container" unless block.full_width?)] do
```
leenagupte, line 4: `["govuk-block__#{block.type}`
This is a great addition and something nice to test for.

leenagupte (approved):
Thanks for making the changes, especially to the commit messages and the system tests.

Normally, I'd recommend fixing-up the later commits, but given the time constraints and that they too have good commit messages I'm happy to let it go 😎 

👍

PR #4259, opened 2024-10-09
Fix content store response type

richardTowers:

## What

The code was sometimes passing GdsApi::Responses into the ContentItem constructor, and sometimes Hashes.

GdsApi::Response behaves enough like Hash to be confusing, but it's not exactly the same. For example, it doesn't have a deep_merge method. It's also missing things which aren't to do with the body, like a cache_control method.

Technically the ContentItem constructor requires a GdsApi::Response, because it delegates the cache_control method to that object. So we shouldn't have been passing it a Hash.

This commit adds a temporary `override_content_store_hash` parameter, which we can use to override the bits we want to without needing to pass the wrong type of object to the constructor. We'll be able to clean all this up once the content store returns the whole object.

## Why

Certain code paths were throwing a NoMethodError for deep_merge, because the code was expecting a Hash but getting a GdsApi::Response

### DIFF
--- a/app/models/content_item_factory.rb
+++ b/app/models/content_item_factory.rb
@@ -1,6 +1,6 @@
 class ContentItemFactory
-  def self.build(content_hash)
-    content_item_class(content_hash["document_type"]).new(content_hash)
+  def self.build(content_store_response)
+    content_item_class(content_store_response["document_type"]).new(content_store_response)
   end
 
   def self.content_item_class(document_type)
--- a/app/models/landing_page.rb
+++ b/app/models/landing_page.rb
@@ -7,21 +7,28 @@ class LandingPage < ContentItem
     if content_store_response.dig("details", "blocks")
       super(content_store_response)
     else
-      super(content_store_response.deep_merge(load_additional_content(content_store_response["base_path"])))
+      super(
+        content_store_response,
+        override_content_store_hash: load_additional_content(content_store_response)
+      )
     end
 
-    @blocks = (@content_store_response.dig("details", "blocks") || []).map { |block_hash| BlockFactory.build(block_hash) }
+    @blocks = (content_store_hash.dig("details", "blocks") || []).map { |block_hash| BlockFactory.build(block_hash) }
   end
 
 private
 
   # SCAFFOLDING: can be removed (and reference above) when full content items
   # including block details are available from content-store
-  def load_additional_content(base_path)
+  def load_additional_content(content_store_response)
+    base_path = content_store_response["base_path"]
     file_slug = base_path.split("/").last.gsub("-", "_")
     filename = Rails.root.join("#{ADDITIONAL_CONTENT_PATH}/#{file_slug}.yaml")
-    return { "details" => {} } unless File.exist?(filename)
+    content_hash = content_store_response.to_hash
+    return content_hash unless File.exist?(filename)
 
-    { "details" => YAML.load(File.read(filename)) }
+    content_hash.deep_merge(
+      "details" => YAML.load_file(filename),
+    )
   end
 end
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -1,17 +1,21 @@
 class ContentItem
-  attr_reader :content_store_response, :body, :image, :description, :document_type, :title, :base_path, :locale
+  attr_reader :content_store_response, :content_store_hash, :body, :image, :description, :document_type, :title, :base_path, :locale
 
-  def initialize(content_store_response)
+  # SCAFFOLDING: remove the override_content_store_hash parameter when full landing page
+  # content items including block details are available from content-store
+  def initialize(content_store_response, override_content_store_hash: nil)
     @content_store_response = content_store_response
-    @body = content_store_response.dig("details", "body")
-    @image = content_store_response.dig("details", "image")
-    @description = content_store_response["description"]
-    @document_type = content_store_response["document_type"]
-    @title = content_store_response["title"]
-    @base_path = content_store_response["base_path"]
-    @locale = content_store_response["locale"]
+    @content_store_hash = override_content_store_hash || content_store_response.to_hash
+
+    @body = content_store_hash.dig("details", "body")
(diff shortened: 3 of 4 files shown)


app/models/content_item_factory.rb
```diff
@@ -1,6 +1,6 @@
 class ContentItemFactory
-  def self.build(content_hash)
-    content_item_class(content_hash["document_type"]).new(content_hash)
+  def self.build(content_store_response)
```
leenagupte, line 2: I don't think this needs to be renamed. What's being passed in was one of three potential things anyway.

app/models/landing_page.rb
```diff
@@ -4,13 +4,16 @@ class LandingPage < ContentItem
   ADDITIONAL_CONTENT_PATH = "lib/data/landing_page_content_items".freeze
 
   def initialize(content_store_response)
-    if content_store_response.dig("details", "blocks")
-      super(content_store_response)
-    else
-      super(content_store_response.deep_merge(load_additional_content(content_store_response["base_path"])))
+    override_content_store_hash = nil
+    if content_store_response.dig("details", "blocks").nil?
+      override_content_store_hash = content_store_response.to_hash.deep_merge(
```
leenagupte, line 9: In the tests a hash converted to a `GdsApi::Response`. Is it possible for the same to be done here?
i.e. after the `content_store_response` is converted to a hash to allow the `deep_merge` for the additional content, could it then be converted back to a `GdsApi::Response` before being passed to the super class?

I'm trying to think of how to limit the blast radius for something that will only temporarily affect Landing Pages and that will be easier to unpick later.

leenagupte (commented):
I can see the problem that's trying to be solved here. I've added a question / suggestion inline about how the blast radius for the change could be reduced.

app/models/content_item_factory.rb
richardTowers: No, I think all three things are GdsApi::Responses. The only time it was actually getting a hash was in our tests, where we were giving it a hash.

The three possible things are:

- `request.env[:content_item]`
- `request_content_item(content_item_slug)`
- `request_content_item("/#{params[:slug]}")`

`request_content_item` always returns a GdsApi::Response, and this is also what's cached on `request.env[:content_item]` [here](https://github.com/alphagov/frontend/blob/main/lib/format_routing_constraint.rb#L17)

leenagupte: I mean it's a passed in parameter and doesn't need to be renamed as part of this change. It could be renamed independently of everything else and nothing else will be affected.

app/models/landing_page.rb
richardTowers: The `GdsApi::Response` constructor takes a `RestClient::Response` as its input. The tests construct a fake one, but that's okay because we can use an `instance_double`. To do it here, we'd need to make a real fake HTTP response by turning the hash back to a JSON string for the body and reassembling the HTTP headers from the original response.

Super gross, basically.

I don't think the blast radius of this change is particularly bad? We're adding an optional parameter to the constructor, and it's only provided in this one case. Other than that, the logic should be unchanged.

richardTowers: I did consider a couple of other options, which you may find more or less gross than this:

richardTowers: 1) Wrap the `GdsApi::Response` with another class which intercepts any method calls which return things from the body and merges in the additional content. Pass this into the ContentItem constructor, and it doesn't need to change.

richardTowers: 2. Monkey patch the `GdsApi::Response` instance 🙈

richardTowers: I felt this was the least magical / most easy to reason about option.

app/models/content_item_factory.rb
richardTowers: True, but the name is wrong - it's not a hash, it's a response.

This change is about clearing up the response / hash confusion, so I think it's legitimate to fix the parameter name here.

richardTowers: Happy to pull it out into a separate commit if that would make it more palatable?

leenagupte: All of the naming in general is confusing. But fine, this isn't a blocker. Separate commit would be better. There is so much going on.

app/models/landing_page.rb
leenagupte: Yes, this is probably the least ugly.

I think it would be a bit nicer if the truthy check `if content_store_response.dig("details", "blocks")` was kept in and it was only the `else` block that had the overhaul. 

i.e.
```ruby
  if content_store_response.dig("details", "blocks")
    super(content_store_response)
  else 
    override_content_store_hash = content_store_response.to_hash
    override_content_store_hash.deep_merge(
      load_additional_content(content_store_response["base_path"]),
    )

    super(content_store_response, override_content_store_hash:)
  end

  @blocks = (content_store_hash.dig("details", "blocks") || []).map { |block_hash| BlockFactory.build(block_hash) }
end
```

But again not a blocker.

app/models/landing_page.rb
```diff
@@ -7,21 +7,28 @@ def initialize(content_store_response)
...
   # SCAFFOLDING: can be removed (and reference above) when full content items
   # including block details are available from content-store
-  def load_additional_content(base_path)
+  def load_additional_content(content_store_response)
+    base_path = content_store_response["base_path"]
     file_slug = base_path.split("/").last.gsub("-", "_")
     filename = Rails.root.join("#{ADDITIONAL_CONTENT_PATH}/#{file_slug}.yaml")
-    return { "details" => {} } unless File.exist?(filename)
+    content_hash = content_store_response.to_hash
+    return content_hash unless File.exist?(filename)
```
leenagupte, line 28: Moving the logic changes to this method is a much more elegant solution. 🎉 
If I wanted to be really pedantic I would say that you could switch these to lines around and return nil if a file doesn't exist. But that's really minor and not blocker.

leenagupte (approved):
Looks good. 👍

PR #4260, opened 2024-10-09
Add featured block

andysellick:

## What
Adds the featured block for the landing page.

Uses code copied from the hero block to provide the text content, including more flexibility than we probably currently need. Might be good to consolidate this code - there's potentially a third block going to use the same approach.

Image handling is also copied from the hero block, using a picture element and screen pixel density detection to display an appropriate image for the current device.

## Why
Required for the landing page.

## Visual changes
Appearance on desktop. The image should be a set height, and it will determine the minimum height of the box and be centred within it. This means that as the screen gets narrower the left and right edges of the image will be deliberately lost.

Note that the image shown below isn't optimised for height, this is just an example.

Full width desktop | Constrained desktop
----------------- | -------------
Screenshot 2024-10-10 at 11 54 25 | Screenshot 2024-10-10 at 11 56 30

This has been built like this so that we never end up with a gap below the image. If the text is longer or the user has a higher font size set, the image will still fill the available height.

Normal | Increased text size
------ | ------
Screenshot 2024-10-10 at 11 56 30 | Screenshot 2024-10-10 at 11 58 24

Appearance on tablet and below.

Screenshot 2024-10-09 at 10 47 44

Trello card: https://trello.com/c/YCUbimh9/50-build-featured

### DIFF
--- a/app/assets/stylesheets/views/_landing_page/featured.scss
+++ b/app/assets/stylesheets/views/_landing_page/featured.scss
@@ -1,0 +1,37 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.featured {
+  display: flex;
+  margin-bottom: govuk-spacing(6);
+  background: govuk-colour("dark-blue");
+
+  @include govuk-media-query($until: desktop) {
+    flex-direction: column;
+  }
+}
+
+.featured__child--content {
+  flex: 1 1 33.333%;
+  padding: govuk-spacing(6);
+  box-sizing: border-box;
+}
+
+.featured__child--image {
+  flex: 1 1 66.666%;
+  display: flex;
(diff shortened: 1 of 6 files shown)


AshGDS (approved):
Looks good to me 👍  

However, if I'm looking at the latest designs correctly, the aspect ratio of the image we will use looks more like a square? I'm wondering if it's worth using more square examples just to ensure it will look correct when content is added. Here's some examples. Looks nice with a large resolution block, but might look a bit odd if the image we get is smaller.

(I used placehold.co for the example images)

image
 
image

andysellick:
@AshGDS I checked with design and they suggested this alternative, which I've implemented and updated the description to explain. Can you take a look?

PR #4262, opened 2024-10-09
Improve the FE for the hero

andysellick:

## What
Improve the frontend code for the hero block. Specifically:

- set padding correctly on the text box
- improve position of text box on mobile
- set a minimum height of the picture element, to reduce CLS and stop the text box from hitting the bottom of the page before the image loads

## Why
FE added with initial block code was prototype.

## Visual changes
The image is now in a container with a fixed height and we'll have to make the images be the right height. This is to avoid cumulative layout shift and also to prevent the negative top margin text block from overflowing into the site header if the image fails to load for some reason.

The image is centred in the space so narrower versions of each breakpoint could lose the left/right of the image. Again this is deliberate.

Mobile | Tablet | Desktop | Desktop (large)
----- | ------- | ------- | ------
Screenshot 2024-10-10 at 13 39 39 | Screenshot 2024-10-10 at 13 39 25 | Screenshot 2024-10-10 at 13 39 12 | Screenshot 2024-10-10 at 13 39 00

Trello card: https://trello.com/c/GYhdvFfi/45-create-hero-block

### DIFF
--- a/app/assets/stylesheets/views/_landing_page/hero.scss
+++ b/app/assets/stylesheets/views/_landing_page/hero.scss
@@ -1,25 +1,32 @@
 @import "govuk_publishing_components/individual_component_support";
 
 .app-b-hero {
-  margin-bottom: govuk-spacing(4);
+  margin-bottom: govuk-spacing(6);
 }
 
+.app-b-hero__imagewrapper {
+  display: block;
+  min-height: 200px;
+}
+
+// assuming the image is the right height and sets the height of the block
 .app-b-hero__image {
   display: block;
   position: relative;
-  max-height: 600px;
+  max-height: 610px;
   max-width: 100%;
   margin: 0 auto;
 }
 
 .app-b-hero__textbox {
   position: relative;
-  padding: govuk-spacing(4);
+  padding: govuk-spacing(6);
   margin-top: -(100px);
   background: govuk-colour("dark-blue");
   color: govuk-colour("white");
 
   @include govuk-media-query($until: tablet) {
-    margin-top: -(govuk-spacing(2));
+    padding: govuk-spacing(4);
+    margin-top: -(govuk-spacing(6));
   }
 }
--- a/app/assets/stylesheets/views/_landing_page/hero.scss
+++ b/app/assets/stylesheets/views/_landing_page/hero.scss
@@ -5,17 +5,21 @@
 }
 
 .app-b-hero__imagewrapper {
-  display: block;
-  min-height: 200px;
+  display: flex;
+  justify-content: center;
+  overflow: hidden;
+  height: 610px;
+
+  @include govuk-media-query($until: desktop) {
+    height: 512px;
+  }
+  @include govuk-media-query($until: tablet) {
+    height: 427px;
+  }
 }
 
-// assuming the image is the right height and sets the height of the block
 .app-b-hero__image {
   display: block;
-  position: relative;
-  max-height: 610px;
-  max-width: 100%;
-  margin: 0 auto;
 }
 
 .app-b-hero__textbox {


app/assets/stylesheets/views/_landing_page/hero.scss
```diff
@@ -1,25 +1,32 @@
 @import "govuk_publishing_components/individual_component_support";
 
 .app-b-hero {
-  margin-bottom: govuk-spacing(4);
+  margin-bottom: govuk-spacing(6);
 }
 
+.app-b-hero__imagewrapper {
+  display: block;
+  min-height: 200px;
```
richardTowers, line 9: We should probably have width / height attributes on each of the sources / the img inside the picture element, so that the browser doesn't have to wait for the stylesheet to make an educated guess about the size of the images.

If we had that, you probably wouldn't need this?

andysellick: That's a good point, but it sort of depends on how we want to handle the image. If we just set the height then it breaks the aspect ratio at a lower width:

Screenshot 2024-10-09 at 16 00 44

However, we can fix the height and centre the image in the space, but that would crop the edges off at a lower width:

Screenshot 2024-10-09 at 16 04 10

Think this might be a question for design.

richardTowers: Can we set the width and the height to the intrinsic size of the image, and then override them in CSS once that's loaded?

I think that way we'd get a flash of the image being not-quite-the-right size (but not stretched or cropped) before the CSS loads and right-sizes it. Which is better than a flash of the image being totally the wrong size.

andysellick: @richardTowers have reworked this block to account for the height suggestion, and updated PR description to explain.

richardTowers: Looks good as far as I can tell. I'll resolve this conversation, but I think I'd want a frontender to actually review the CSS.

matthillco: I appreciate this has already been resolved, but this looks like a candidate for `aspect-ratio` and `object-fit` CSS properties. Those could provide a better overall solution with more flexibility for different sized images. If the aspect-ratio of the image container is determined in the CSS, eg `aspect-ratio: 16 / 9` and we use `object-fit:cover` on the image, any image that is too large will automatically be cropped within the container but retain its aspect ratio. Is that the objective here?

matthillco (commented):
Just a couple of small points to consider:

AshGDS (commented):
Looks good to me, though will allow Matt to handle approving as he has made comments. 👍

app/assets/stylesheets/views/_landing_page/hero.scss
```diff
@@ -1,25 +1,36 @@
...
 .app-b-hero {
-  margin-bottom: govuk-spacing(4);
+  margin-bottom: govuk-spacing(6);
+}
+
+.app-b-hero__imagewrapper {
+  display: flex;
+  justify-content: center;
+  overflow: hidden;
+  height: 610px;
```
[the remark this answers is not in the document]
andysellick: Agreed but as you said it causes layout shift. It also would mean that the blue text box can overlap the site header if the image fails to load. Agree that it's a bit content specific but I think at this point we're being very specific with image requirements so this is okay (and yes, we'll probably end up adjusting these numbers at some point when we get the images, but I think that's okay).

app/assets/stylesheets/views/_landing_page/hero.scss
```diff
@@ -1,25 +1,36 @@
...
-  max-height: 600px;
-  max-width: 100%;
-  margin: 0 auto;
 }
 
 .app-b-hero__textbox {
   position: relative;
-  padding: govuk-spacing(4);
+  padding: govuk-spacing(6);
   margin-top: -(100px);
```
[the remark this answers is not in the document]
andysellick: Have now demonstrated this to design and they seem happy. I expect once we get the pages mocked up there'll be loads of tweaks like this but for now I think it's okay.

matthillco (approved):

PR #4263, opened 2024-10-09
Add a build_all helper method to BlockFactory

richardTowers:

## What

We're finding there are lots of places where we want to take a list of block hashes and turn them into a list of blocks (not all of these have hit main yet, but the need is there)

## Why

Currently we're duplicating the bit mapping code, and it's a bit ugly. This helper method DRYs things up a bit.


leenagupte (approved):
Nice! 🥇

PR #4267, opened 2024-10-10
Use schema_name instead of document_type to look up model classes

richardTowers:
## What / Why

Previously we would look up our content item model class based on its document type, ignoring schema name.

This is a inconsistent with FormatRoutingConstraint, which uses schema name to decide whether the route matches (and ignores document type). That could lead to a confusing situation where the route and controller are picked based on schema name, but the model is picked based on document type.

There's also a problem with using document type in some of the tests. Some of the tests generate "random" content items:

    https://github.com/alphagov/govuk_schemas/blob/c0d74240bafcd2d4ebcf3e8550071028fc6fbbde/lib/govuk_schemas/random_example.rb#L6-L18

 By default, they use the "generic" schema, which can have any document type. This means there's a chance that a random example will have schema_name: generic, document_type: landing_page.

 In that situation, the ContentItemFactory would have instantiated the LandingPage model instead of the ContentItem model, and we end up with a confusing situation where a controller that isn't expecting it gets a LandingPage model instead of a ContentItem model. If we use schema_name instead we don't have that problem, because a generic schema is always going to get the base ContentItem class.

Using schema_name feels more consistent, and it avoids the issue with "Random" examples in the tests.

Supersedes #4265


leenagupte (approved):
Looks good! Thanks for adding a test for ContentItemFactory 🥇

PR #4268, opened 2024-10-10
Add an example video using the govspeak block

richardTowers:
## What

Now that the designs don't call for a fancy video collection block, I think we can just use govspeak for the video.

This is consistent with the approach taken on the current number 10 org page:

https://github.com/alphagov/collections/blob/main/app/views/organisations/_videos.html.erb#L22-L24

If we need more configurability than govspeak gives us we can add a video block. But if we don't need it, let's not implement it!

## Why

[Trello card](https://trello.com/c/J3TRzA4R/51-build-video-collection)

## Screenshots

Check out the preview app.


leenagupte (approved):
I approve the video choice 🤣

PR #4273, opened 2024-10-11
Add the skeleton of a "goals" page, as per the designs

richardTowers:
## What

Adds a goals page, roughly following the designs in Figma.

## Why

[Trello card](https://trello.com/c/eIueNNk9/63-build-a-demo-for-one-of-each-of-the-4-page-types)

## Screenshots?

image

See the preview app for more details


lib/data/landing_page_content_items/goals.yaml
```diff
@@ -103,7 +100,6 @@ blocks:
         <h2>Dorem sit</h2>
         
         <p>TODO - social media links</p>
```
leenagupte, line 102: This should be removed as well as `share_links` block has already been added.

leenagupte (commented):
One small comment, otherwise it looks good.

lib/data/landing_page_content_items/goals.yaml
richardTowers: Ah damnit, I saw that and then got distracted by the dog 🤦🏻  - sorry!

richardTowers: Fixed in [this force push](https://github.com/alphagov/frontend/compare/1815f563ce1d668af074f9f2f68b3a4f0dfef4b4..81b239c3caf5de6cd31b8bc0f2a9c5c55e8373a5)

leenagupte (approved):
I'm going to approve this because the skeleton is there, but this extra title above the `share_links` block should be removed at some point:

Screenshot 2024-10-18 at 11 50 04

PR #4274, opened 2024-10-11
Make the second column optional in two column layouts

richardTowers:
## What

This is convenient for the case where we're only using the left column to constrain the width to two thirds (which is a common gov.uk pattern)

## Why

[Trello card](https://trello.com/c/eIueNNk9/63-build-a-demo-for-one-of-each-of-the-4-page-types)


jon-kirwan (approved):
Looks good to me 👍

PR #4277, opened 2024-10-14
Fix the action link inverse option

andysellick:

## What / why
Fix the appearance of the action link on a dark background on the landing pag.e

- previously had an override to convert the generally used `inverse` option to `light_text`, which the action link component required
- however the action link has now been modified to use `inverse` so this code is no longer needed

## Visual changes

Before | After
----- | ------
Screenshot 2024-10-14 at 09 09 12 | Screenshot 2024-10-14 at 09 09 18

Trello card: https://trello.com/c/GYhdvFfi/45-create-hero-block


leenagupte (approved):
🤩

PR #4286, opened 2024-10-15
Add hero block variant

jon-kirwan:
## What

Add hero block variant

## Why

Design requirements

### Desktop

frontend dev gov uk_landing-page(iPad Pro)


app/views/landing_page/blocks/_hero.html.erb
```diff
@@ -1,8 +1,14 @@
 <%
   add_view_stylesheet("landing_page/hero")
+
+  hero_classes = "app-b-hero"
+  hero_classes << " app-b-hero--mid-page" if block.data["mid_page"]
```
andysellick, line 5: Rails suggestion to avoid needing a leading space on the extra classes:

```Ruby
<%
  hero_classes = %w[app-b-hero]
  hero_classes << "app-b-hero--mid-page" if block.data["mid_page"]

  hero_content_classes = %w[app-b-hero__textbox]
  hero_content_classes << "app-b-hero__textbox--mid-page" if block.data["mid_page"]
%>
# ...

<%= content_tag("div", class: hero_classes) do %>
  # ...
  <%= content_tag("div", class: hero_content_classes) do %>
    # ...
  <% end %>
<% end %>
```

app/assets/stylesheets/views/_landing_page/quote.scss
andysellick, line 1: Design difference: the mobile and tablet versions should have a small negative margin to slightly overlap the quote box over the hero image (see design, below).

Mobile | Tablet
------ | ------
Screenshot 2024-10-15 at 16 20 38 | Screenshot 2024-10-15 at 16 23 19

andysellick, line 1: If the user has a higher than normal font size set in their browser or they're using 'zoom text only' on Firefox, the quote box can exceed the height of the image and escape through the top and bottom of the block. We don't have a design for what should happen in this situation, but I think probably the text box should escape downwards only, and push the content beneath it out of the way.

Screenshot 2024-10-15 at 16 38 42

app/assets/stylesheets/views/_landing_page/hero.scss
```diff
@@ -33,4 +42,16 @@
...
     margin-top: -(govuk-spacing(6));
   }
+
+  &.app-b-hero__textbox--mid-page {
+    margin: 0;
+
+    @include govuk-media-query($from: tablet) {
+      color: govuk-colour("white");
+      position: absolute;
+      max-width: 336px;
```
andysellick, line 52: This seems like a bit of a magic number - I'm worried about the box overlapping the image close to the breakpoints (see below). Could we somehow stick with the govuk-grid widths?

Screenshot 2024-10-15 at 16 41 44

app/assets/stylesheets/views/_landing_page/quote.scss
```diff
@@ -0,0 +1,27 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.app-b-quote {
+  color: govuk-colour("white");
+
+  .app-b-quote__svg {
```
andysellick, line 6: Nitpick: if we're using BEM like this, I don't think we need to nest all these classes.

lib/data/landing_page_content_items/landing_page.yaml
```diff
@@ -77,6 +77,22 @@ blocks:
     content: <p>Right content!</p>
 - type: govspeak
   content: <h2>Statistics</h2>
+- type: hero
+  mid_page: true
```
richardTowers, line 81: I think `theme: middle_left` might be better.

Couple of reasons for that:

1. mid_page makes it sounds like somethings going to be in the middle of the _page_, which isn't quite what this is. The content is in the middle of the hero, not the page.
2. Using booleans for this kind of setting leads to a common problem with proliferation of options. At some point, maybe we'll want one of these with the content on the right, and we'll add `right_page: true`, and then we can do bottom-left, mid-left, bottom-right, and mid-right. Then, later, someone will want to do `top`, and the code will have to decide how to handle invalid combinations like `mid_page: true; top_page: true`
3. Other blocks use `theme` (e.g. the two column layout has a `two_thirds_one_third` "theme")

app/assets/stylesheets/views/_landing_page/quote.scss
```diff
@@ -0,0 +1,27 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.app-b-quote {
+  color: govuk-colour("white");
```
richardTowers, line 4: I guess at some point we might need the concept of `inverse` things here (as obviously this block would render badly if it wasn't on a dark background).

richardTowers (commented):
Couple of thoughts - probably best for a frontender to review the CSS, but that bit looks good to me :)

app/views/landing_page/blocks/_hero.html.erb
jon-kirwan: Great thanks, that's much better

app/assets/stylesheets/views/_landing_page/quote.scss
jon-kirwan: Thanks, good spot. I reviewed with Nik and have updated to add different negative margin for mobile and tablet. Also, I've changed the grid class, for the hero content, to `govuk-grid-column-two-thirds-from-desktop` (from `govuk-grid-column-two-thirds`.

app/assets/stylesheets/views/_landing_page/quote.scss
jon-kirwan: Yep good point. Updated now

lib/data/landing_page_content_items/landing_page.yaml
jon-kirwan: Thanks, yep theme is much better and consistent with other blocks. Updated now.

app/assets/stylesheets/views/_landing_page/hero.scss
jon-kirwan: Thanks @andysellick for all the help in finding a solution here. This approach involves absolutely positioning the hero image and introducing a new text box wrapper element that acts as an overlay over the hero image. The grid classes can then be applied inside the wrapper without needing custom widths.

app/assets/stylesheets/views/_landing_page/quote.scss
jon-kirwan: The reflow issues are now resolved by changes mentioned [here](https://github.com/alphagov/frontend/pull/4286#discussion_r1801469900)

app/assets/stylesheets/views/_landing_page/hero.scss
```diff
@@ -34,3 +40,45 @@
...
+    min-height: $mobile-height;
+  }
+
+  .app-b-hero__image {
+    position: absolute;
+    left: 50%;
+    transform: translate(-50%);
+  }
+
+  .app-b-hero__textbox-wrapper {
```
andysellick, line 64: Whilst this is a nice solution, I'm afraid it reintroduces the original problem of overflow if the text content is too big. Sorry.

Screenshot 2024-10-17 at 12 54 43

My only other thought on how to do this would be if we calculated what the height would be, and set that as the min-height. So if the overall hero is 610px, the min-height would need to be 610px - top margin - bottom margin (of the textbox)?

andysellick (commented):
Looks good, just one small problem.

app/assets/stylesheets/views/_landing_page/hero.scss
jon-kirwan: Thanks. I've added these changes as a separate [commit](https://github.com/alphagov/frontend/pull/4286/commits/1c9242b51bd23f37c9180a3acff12d6644c5560f) changing heights to minimum heights which seems to resolve it.

andysellick (approved):
Well done for sticking through this one - it wasn't easy 👏

PR #4289, opened 2024-10-16
Add header

JamesCGDS:

## What

This PR adds the landing page header. It is dependent on [PR #4284](https://github.com/alphagov/govuk_publishing_components/pull/4284) in the components gem being released.

## Why

[Trello card](https://trello.com/c/W2Jk6Qka/46-build-landing-page-header)

### DIFF
--- a/app/assets/stylesheets/views/_landing_page.scss
+++ b/app/assets/stylesheets/views/_landing_page.scss
@@ -5,6 +5,11 @@
   background-color: #231F20;
 }
 
+.landing-page-header__blue-bar {
+  background-color: govuk-colour("blue");
+  height: govuk-spacing(2);
+}
+
 .landing-page-header__org {
   padding-top: govuk-spacing(2);
   padding-bottom: govuk-spacing(6);
--- a/app/assets/stylesheets/views/_landing_page.scss
+++ b/app/assets/stylesheets/views/_landing_page.scss
@@ -1,0 +1,19 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.landing-page-header {
+  // I've used the hex value for the background colour as it doesn't exist in the colour palette
+  background-color: #231F20;
+}
+
+.landing-page-header__org {
+  padding-top: govuk-spacing(2);
+  padding-bottom: govuk-spacing(6);
+
+  // I've added this styling here rather than in the gem because it's application specific
+  // and it would introduce further work to make this variation work in the gem without
+  // affecting the other variants.
+  // TODO add this as a variant in the components gem.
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/views/_landing_page.scss
```diff
@@ -0,0 +1,19 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.landing-page-header {
+  // I've used the hex value for the background colour as it doesn't exist in the colour palette
+  background-color: #231F20;
+  width: 100%;
```
andysellick, line 6: Width isn't needed here?

app/assets/stylesheets/views/_landing_page.scss
```diff
@@ -0,0 +1,19 @@
...
+
+.landing-page-header__org {
+  padding-top: govuk-spacing(2);
+  padding-bottom: govuk-spacing(6);
+
+  // I've added this styling here rather than in the gem because it's application specific
+  // and it would introduce further work to make this variation work in the gem without
+  // affecting the other variants.
+  .gem-c-organisation-logo__link:link:not(:hover) {
+    text-decoration: none;
```
andysellick, line 22: Why remove the underline on hover? Did I miss a discussion about the design?

andysellick (commented):
Frontend code looks good, just a couple of things. Will wait to approve pending a BE review.

app/assets/stylesheets/views/_landing_page.scss
JamesCGDS: Quite right, think it was left over from me messing around with some things. Will remove

app/assets/stylesheets/views/_landing_page.scss
JamesCGDS: This doesn't remove the underline on hover. According to the designs, the link should not have an underline for its default state but I imagine we still want an underline to appear on hover. Basically the above is a one liner for doing something like:

```
.gem-c-organisation-logo__link:link {
    text-decoration: none;

    &:hover {
      // reinstate hover underline styling
    }
  }
```

If this is more readable and preferable, happy to make the change

andysellick: Oh I see. Actually I was more concerned about styling a component outside of a component, which breaks component isolation. For the purposes of getting this over the line can you just put a comment that we need to add a variant of the organisation logo to do this, rather than styling it here? Then we'll sort that separately.

andysellick (approved):
Frontend looks good to me 👍

PR #4291, opened 2024-10-16
Expose schema name and query

KludgeKML:

## What

Exposes .schema on the content item, and adds (via metaprogramming) methods like content_item.is_a_landing_page? which allow us to query schema tidily in views, etc, without having to put .schema == " whatever" or add instance variables in controllers.

    - returns true if called with the schema name of the
      document, so is_a_landing_page? is true for landing
      pages, but false for anything else.
    - is_an_xxxx? also works, for the nicer feel of being able to
      say is_an_organisation? when we move other document
      types into frontend.


leenagupte (approved):
Thanks for adding the tests. That makes it much easier to follow what's going on.

PR #4294, opened 2024-10-18
Add 'See the latest data on our progress' link to homepage

jon-kirwan:
## What

Add 'See the latest data on our progress' link to homepage

## Why

https://trello.com/c/7Sb6I491/95-see-the-latest-data-on-our-progress

## How

I've added the link as a separate block immediately following the 'latest data' grid container block. I experimented with adding the link to the grid container but it seemed quite fiddly to amend the block to accomodate another row. I don't think there will be any need for the text to be part of the 

### Desktop

frontend dev gov uk_landing-page(iPad Pro) (1)


andysellick (commented):
Looks like the right solution, but could you add it to `/landing-page/homepage`? Check with backend if you're not sure which YML file that should be.

jon-kirwan:
> Looks like the right solution, but could you add it to `/landing-page/homepage`? Check with backend if you're not sure which YML file that should be.

Thanks for the review. Added to homepage.yaml now instead

andysellick (approved):

PR #4296, opened 2024-10-18
Options for adding extra vertical spacing between blocks

matthillco:

## What

Two new ways to add extra spacing between blocks:
- a generalised class for adding a pre-defined responsive bottom margin on any block
- new options for govspeak blocks to add top and bottom static margins

The numbers to pass to the govspeak margin options are taken from the [Design System Spacing guide](https://design-system.service.gov.uk/styles/spacing/) and range from 0-9. It's currently using the static margins, but this could instead use the responsive margins depending on which works best for this use case: might need design involvement to determine static or responsive.

## Why

Blocks for landing pages do not currently match the landing-page designs and need increased breathing space.

[Trello](https://trello.com/c/YAeRSLr4/91-blocks-need-more-spacing)

### DIFF
--- a/app/assets/stylesheets/views/_landing_page.scss
+++ b/app/assets/stylesheets/views/_landing_page.scss
@@ -22,3 +22,23 @@
     text-decoration: none;
   }
 }
+
+// Consistent lower margins for top-level blocks
+// Self-chaining the class name gives higher
+// specificty and avoids the use of !important.
+//
+// Note the desktop size is hardcoded here as the
+// govuk-spacing(x) helper does not include large
+// enough sizes.
+
+.app-margin-bottom-responsive.app-margin-bottom-responsive  {
+  margin-bottom: govuk-spacing(6);
+
+  @include govuk-media-query($from: tablet) {
+    margin-bottom: govuk-spacing(8);
+  }
+
+  @include govuk-media-query($from: desktop) {
+    margin-bottom: 80px;
+  }
+}
--- a/app/views/landing_page/blocks/_govspeak.html.erb
+++ b/app/views/landing_page/blocks/_govspeak.html.erb
@@ -1,3 +1,18 @@
+<%-
+  margin_top = block.data["margin_top"]
+  margin_bot = block.data["margin_bottom"]
+
+  classes = ""
+  classes << "govuk-!-margin-top-#{margin_top} " if margin_top
+  classes << "govuk-!-margin-bottom-#{margin_bot} " if margin_bot
+%>
+
+<% if classes.present? %>
+  <div class="<%= classes %>">
+<% end %>
 <%= render "govuk_publishing_components/components/govspeak", { inverse: block.data["inverse"] } do %>
   <%= block.data["content"].html_safe %>
+<% end %>
+<% if classes.present? %>
+  </div>
 <% end %>
--- a/lib/data/landing_page_content_items/landing_page.yaml
+++ b/lib/data/landing_page_content_items/landing_page.yaml
@@ -85,6 +85,8 @@ blocks:
     content: <p>Right content!</p>
 - type: govspeak
   content: <h2>Statistics</h2>
+  margin_top: 5
+  margin_bottom: 5
 - type: hero
   theme: middle_left
   image:
--- a/app/views/landing_page/blocks/_featured.html.erb
+++ b/app/views/landing_page/blocks/_featured.html.erb
@@ -1,7 +1,7 @@
 <%
   add_view_stylesheet("landing_page/featured")
 %>
-<div class="featured">
+<div class="featured app-margin-bottom-responsive">
   <div class="featured__child featured__child--content">
     <% block.featured_content.each do |subblock| %>
       <%= render_block(subblock) %>
--- a/app/views/landing_page/blocks/_hero.html.erb
+++ b/app/views/landing_page/blocks/_hero.html.erb
@@ -3,6 +3,7 @@
   add_view_stylesheet("landing_page/hero")
 
   hero_classes = %w[app-b-hero]
+  hero_classes << "app-margin-bottom-responsive"
   hero_classes << "app-b-hero--mid-page" if block.data["theme"] == "middle_left"
 
   grid_class = "govuk-grid-column-two-thirds-from-desktop"


lib/data/landing_page_content_items/landing_page.yaml
```diff
@@ -85,6 +85,8 @@ blocks:
     content: <p>Right content!</p>
 - type: govspeak
   content: <h2>Statistics</h2>
+  margin_top: 5
+  margin_bottom: 5
```
andysellick, line 89: I think we want to avoid having margins specified in the content item, because that implies that this is something that will be configured by the user, which I don't think we would want (could potentially break the page design).

Govspeak is a slightly tricky one, but maybe something that could be handled internally to the govspeak component. For the other blocks that need bottom margin, maybe the approach could be to remove their existing margins and rewrite your bottom margin class as a mixin, then just add it to their Sass?

andysellick (commented):
Took me several comments to get to my overall thoughts about this approach - probably best to just skip to the last comment first 👍

app/assets/stylesheets/views/_landing_page.scss
```diff
@@ -22,3 +22,23 @@
...
+
+.app-margin-bottom-responsive.app-margin-bottom-responsive  {
+  margin-bottom: govuk-spacing(6);
+
+  @include govuk-media-query($from: tablet) {
+    margin-bottom: govuk-spacing(8);
+  }
+
+  @include govuk-media-query($from: desktop) {
+    margin-bottom: 80px;
```
maxgds, line 42: Are you sure this figure is right? Didn't look like that much to me on the PDF. 

Also, have you looked at using the responsive rather than the static spacing from the design system - ideally we would let that do the hard work rather than hand coding the break points ourselves.

matthillco: I eyeballed it based on the Figma design, which doesn't seem to provide a way of measuring it exactly. I can find out the exact figure and change it as necessary.

I did look at the responsive scale, but it doesn't provide numbers that match the design. If we want to use the responsive spacing then we'd need the design to be changed to match it.

maxgds: Worth a chat with Jeremy to get his thoughts on both getting an accurate figure and seeing if it can align with the DS scales, even if the answer is no.

app/assets/stylesheets/views/_landing_page.scss
```diff
@@ -22,3 +22,23 @@
...
+
+// Consistent lower margins for top-level blocks
+// Self-chaining the class name gives higher
+// specificty and avoids the use of !important.
+//
+// Note the desktop size is hardcoded here as the
+// govuk-spacing(x) helper does not include large
+// enough sizes.
+
+.app-margin-bottom-responsive.app-margin-bottom-responsive  {
```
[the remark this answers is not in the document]
matthillco: I think keeping `app` in the name is prudent since it helps identify where the style is coming from: that's useful for other developers who might be looking at the markup to obtain clues for where to find associated styles. I don't really want to use the word `block` here because it might cause confusion over whether we're talking about the blocks of content or block level margins. On balance then, I think the original class name I proposed makes more sense.

app/views/landing_page/blocks/_govspeak.html.erb
```diff
@@ -1,3 +1,18 @@
+<%-
+  margin_top = block.data["margin_top"]
```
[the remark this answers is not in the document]
matthillco: It seemed prudent to offer it as an option and as a corollary to having the bottom margin. I understand it's possibly an example of solving a problem we don't yet have, happy to remove it in that case.

app/assets/stylesheets/views/_landing_page.scss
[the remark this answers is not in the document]
maxgds: Using "app" is not a pattern we use anywhere but components, though, let's avoid that.

lib/data/landing_page_content_items/landing_page.yaml
matthillco: Yes I did weigh this up but felt that given possible uncertainly over final content, it might be useful for content editors to have some control of how much spacing to apply around govspeak elements. As you note, there's always the possibility of that control causing design issues. 

For your other suggestion, do you mean avoid the use of the named class `app-margin-bottom-responsive` and instead use a mixin? This seems less flexible to me, but perhaps it's just apples and oranges.

app/assets/stylesheets/views/_landing_page.scss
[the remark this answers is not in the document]
matthillco: OK, understood. I think I'll revisit this once I've nailed the approach: if we end up using a mixin we might not need this named class at all.

lib/data/landing_page_content_items/landing_page.yaml
richardTowers: To a degree, this gets us into the heart of the problem with this whole "blocks" approach.

If the user of the publishing system is allowed to pick from a list of arbitrary blocks, and stack and nest them in any order they like, it's going to be a significant challenge for the system to get the right margins and padding for every block in every context.

In particular, if there are examples like "if a video block follows a hero image, then the margin-bottom on the hero image should be X, but if a govspeak block follows a hero image then the margin-bottom on the hero image should be Y" then we're going to get into a really mucky place.

If it's possible for every block to have the same margins / padding as every other kind of block (at least at the top level), and for them to stack nicely in a way that doesn't ruin the designs, I think that's the ideal outcome.

If on the other hand, blocks need different margins in different contexts, I think it's probable more pragmatic to allow the user to provide their own margins (as we do here), perhaps while holding our noses a bit and telling ourselves "it's okay, someone from GDS will always review this content before it goes live".

app/views/landing_page/blocks/_govspeak.html.erb
```diff
@@ -1,3 +1,18 @@
...
+  margin_bot = block.data["margin_bottom"]
+
+  classes = ""
+  classes << "govuk-!-margin-top-#{margin_top} " if margin_top
+  classes << "govuk-!-margin-bottom-#{margin_bot} " if margin_bot
+%>
+
+<% if classes.present? %>
+  <div class="<%= classes %>">
+<% end %>
```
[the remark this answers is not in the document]
matthillco: OK. Would the div still need a class name to identify it, or would it just end up as unnamed in the markup if there are no margin classes on it? It seems a bit wrong to have unnamed divs in the markup as that doesn't really fit the BEM model.

app/views/landing_page/blocks/_govspeak.html.erb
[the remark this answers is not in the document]
andysellick: In that case let's remove it - we usually try to have a margin down approach.

lib/data/landing_page_content_items/landing_page.yaml
andysellick: Allowing content authors control over spacing could introduce inconsistencies as well as allowing the possibility that they could overlook problems, particularly if margins are different at different breakpoints. I'd vote for the consistent spacing approach.

matthillco: Also, if we use a responsive margin in the selector above, then we can tweak the spacing for mobile/tablet/desktop while still maintaining consistency.

app/views/landing_page/blocks/_govspeak.html.erb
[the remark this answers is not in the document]
andysellick: Sorry, I meant to refer to `capture` there. It's something like...

```Rails
<% markup = capture do %>
  (the markup you're creating)
<% end %>

<% if classes.present? %>
  <div class="<%= classes %>">
    <%= markup %>
  </div>
<% else %>
  <%= markup %>
<% end %>
```

lib/data/landing_page_content_items/landing_page.yaml
matthillco: The exception would be `govspeak` content/headings, which probably would not want the same amount of consistent bottom margin as typically these are related/connect to the following component. So we might need a general rule:

_All components/blocks get `margin-bottom: XX` when used as a direct child of the main content container, **except** govspeak components that are just text and relate to the following component. These should get `margin-bottom: YY`._

app/views/landing_page/blocks/_govspeak.html.erb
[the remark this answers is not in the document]
matthillco: Cool, OK thanks. I'll hold off applying this change as we might not need any of it if the margins will be handled automatically.

matthillco:
We've mostly solved this now in these other PRs:

[Apply responsive bottom margin to blocks #4306](https://github.com/alphagov/frontend/pull/4306)
[Use header block in demos #4308](https://github.com/alphagov/frontend/pull/4308)
[Add shared helper and component wrapper helper to govspeak component #4325](https://github.com/alphagov/govuk_publishing_components/pull/4325)
[Add documentation and test for govspeak margin #4327](https://github.com/alphagov/govuk_publishing_components/pull/4327)

I'll leave this PR open for now in case there's any further discussions to be had, but I suspect we'll be able to close this once we've had a design review of those PRs above.

matthillco:
I'm closing this now as we've effectively solved the issues discussed.

PR #4297, opened 2024-10-18
Add basic heading block

AshGDS:

## What / Why
- When a `govspeak` paragraph only contains a YouTube video, it removes the paragraph tag entirely, and therefore the spacing we gain from the `<p>` styles are lost. Therefore, the heading was right against the video. Both the heading and video were in a govspeak block. Instead, we can create a basic heading block that can be used here, instead of relying on a `govspeak` paragraph to give us spacing. Our gem heading component has a `padding` option which will  be used to give us the spacing from the video instead.
- Alternatively, I could make a `video` block that then renders `govspeak` and a `heading` gem component, if that's preferred. However I thought a generic heading block might be useful in other situations.
- Trello https://trello.com/c/7RFX4SjL/94-video-block-adjustments

### Before
image

### After

image

### DIFF
--- a/app/views/landing_page/blocks/_heading.html.erb
+++ b/app/views/landing_page/blocks/_heading.html.erb
@@ -1,0 +1,9 @@
+<%
+  heading_level = block.data["heading_level"] || 2
+%>
+
+<%= render "govuk_publishing_components/components/heading", {
+  text: block.data["text"],
+  heading_level: heading_level,
+  padding: true
+} %>
--- a/lib/data/landing_page_content_items/homepage.yaml
+++ b/lib/data/landing_page_content_items/homepage.yaml
@@ -52,9 +52,10 @@ blocks:
         content: |
           <h2><a href="http://gov.uk">Lorem ipsum dolor sit</a></h2>
           <p>Lorem ipsum dolor sit amet. In voluptas dolorum vel veniam nisi et voluptate dolores id voluptatem distinctio. Et quia accusantium At ducimus quis aut voluptates iusto aut esse suscipit.</p>
+- type: heading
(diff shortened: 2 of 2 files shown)


app/views/landing_page/blocks/_heading.html.erb
```diff
@@ -0,0 +1,10 @@
+<%
+  heading_level = block.data["heading_level"] || 2
+  padding = block.data["padding"] || false
+%>
+
+<%= render "govuk_publishing_components/components/heading", {
+  text: block.data["text"],
+  heading_level: heading_level,
+  padding: true
```
leenagupte, line 9: I think this should be `padding:` to pick up the value in `block.data["padding"]`, otherwise it will always be set to `true`.

Is there a time when the padding shouldn't be set? I'm wondering if it should always be true in the heading block?

AshGDS: @leenagupte Good spot, thanks. I will remove the `.yaml` option for now.

app/views/landing_page/blocks/_heading.html.erb
```diff
@@ -0,0 +1,10 @@
+<%
+  heading_level = block.data["heading_level"] || 2
+  padding = block.data["padding"] || false
```
leenagupte, line 3: If the setting is no longer in the YAML, and I assume that means that the padding should always be "true", then this line can be removed too.

AshGDS: @leenagupte Whoops, Monday morning brain - have removed now!

leenagupte (approved):
Great addition!

PR #4301, opened 2024-10-21
Use a blocks container to keep blocks to one column

leenagupte:

## What

Use a blocks container to keep blocks to one column

## Why

This is rather than defining multiple `two_column_layout` blocks


richardTowers (approved):
Nice

PR #4305, opened 2024-10-21
Move get involved document

sairamya93:

## What

To handle the rendering of get_involved document type from government_frontend to frontend as part of [RFC 175](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-175-frontend-fewer-apps.md)

## Why

https://trello.com/c/mXj8oTI3/349-move-document-type-getinvolved-from-government-frontend-to-frontend, [Jira issue PNP-8568](https://gov-uk.atlassian.net/browse/PNP-8568)

## How

To add a scoped route for, a controller, a model and a view. Here we split up the methods in the original presenter file into Presenter and Model logic, with the minimum functionality required to get Get Involved document working.

## Screenshots?

Get Involved page from 
http://frontend.dev.gov.uk/government/get-involved and https://www.gov.uk/government/get_involved

Screenshot 2024-10-21 at 15 33 26

Screenshot 2024-10-21 at 15 59 28

Screenshot 2024-10-21 at 16 02 16

Screenshot 2024-10-21 at 16 03 03

Screenshot 2024-10-21 at 16 08 10

### DIFF
--- a/app/models/get_involved.rb
+++ b/app/models/get_involved.rb
@@ -1,0 +1,62 @@
+class GetInvolved < ContentItem
+  attr_reader :take_part_pages
+
+  def initialize(content_store_response, override_content_store_hash: nil)
+    super
+    @take_part_pages = content_store_hash.dig("links", "take_part_pages")
+  end
+
+  def open_consultation_count
+    GdsApi.search.search({ filter_content_store_document_type: "open_consultation", count: 0 })["total"]
+  end
+
+  def closed_consultation_count
+    query = {
+      filter_content_store_document_type: "closed_consultation",
+      filter_end_date: "from: #{1.year.ago}",
+      count: 0,
+    }
+
+    GdsApi.search.search(query)["total"]
+  end
+
+  def next_closing_consultation
+    query = {
+      filter_content_store_document_type: "open_consultation",
+      filter_end_date: "from: #{Time.zone.now.to_date}",
+      fields: "end_date,title,link",
+      order: "end_date",
+      count: 1,
+    }
+
+    GdsApi.search.search(query)["results"].first
+  end
+
+  def recently_opened_consultations
+    query = {
+      filter_content_store_document_type: "open_consultation",
+      fields: "end_date,title,link,organisations",
+      order: "-start_date",
+      count: 3,
+    }
+
+    GdsApi.search.search(query)["results"]
+  end
+
+  def recent_consultation_outcomes
+    query = {
+      filter_content_store_document_type: "consultation_outcome",
+      filter_end_date: "to: #{Time.zone.now.to_date}",
+      fields: "end_date,title,link,organisations",
+      order: "-end_date",
+      count: 3,
+    }
+
+    GdsApi.search.search(query)["results"]
+  end
+
+  def consultations_link
+    filters = %w[open_consultations closed_consultations]
+    "/search/policy-papers-and-consultations?#{filters.to_query('content_store_document_type')}"
+  end
+end
--- a/app/presenters/get_involved_presenter.rb
+++ b/app/presenters/get_involved_presenter.rb
@@ -1,0 +1,58 @@
+class GetInvolvedPresenter
+  attr_reader :content_item
+
+  def initialize(content_item)
+    @content_item = content_item
+  end
+
+  def recently_opened
+    filtered_links(content_item.recently_opened_consultations, I18n.t("formats.get_involved.closes"))
+  end
+
+  def recent_outcomes
+    filtered_links(content_item.recent_consultation_outcomes, I18n.t("formats.get_involved.closed"))
+  end
+
+  def time_until_next_closure
+    days_left = (content_item.next_closing_consultation["end_date"].to_date - Time.zone.now.to_date).to_i
+    if days_left.negative?
+      I18n.t("formats.get_involved.closed")
+    elsif days_left.zero?
+      I18n.t("formats.get_involved.closing_today")
+    elsif days_left == 1
+      I18n.t("formats.get_involved.closing_tomorrow")
+    else
+      I18n.t("formats.get_involved.days_left", number_of_days: days_left)
+    end
+  end
+
+private
+
+  def filtered_links(array, close_status)
+    array.map do |item|
+      {
+        link: {
+          text: item["title"],
+          path: item["link"],
+          description: "#{close_status} #{item['end_date'].to_date.strftime('%d %B %Y')}",
+        },
+        metadata: {
+          public_updated_at: Time.zone.parse(org_time(item)),
+          document_type: org_acronym(item),
+        },
+      }
+    end
+  end
+
+  def org_time(item)
+    item["organisations"].map { |org|
+      org["public_timestamp"]
+    }.join(", ")
+  end
+
+  def org_acronym(item)
+    item["organisations"].map { |org|
+      org["acronym"]
+    }.join(", ")
+  end
+end
--- a/config/locales/az.yml
+++ b/config/locales/az.yml
@@ -388,6 +388,47 @@ az:
   formats:
     find_my_nearest:
       valid_postcode_no_locations:
+    get_involved:
+      civil_service:
+      civil_service_quarterly:
+      closed:
+      closed_consultations:
+      closes:
+      closing_today:
+      closing_tomorrow:
+      days_left:
+      engage_with_gov:
+      fcdo_bloggers:
+      follow:
+      follow_links:
+      follow_paragraph:
+      gds_blog:
+      gov_past:
+      intro_paragraph:
+        body_html:
+        engage_with_government:
+        take_part:
+      join_ics:
+      make_donation:
+      more_ways:
+      neighbourhood_watch:
+      open_consultations:
+      page_heading:
+      page_title:
+      petition_paragraph:
+        body_html:
+        create_a_petition:
+      read_respond:
+      recent_outcomes:
+      recently_opened:
+      respond_to_consultations:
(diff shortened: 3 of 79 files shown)


app/models/get_involved.rb
```diff
@@ -0,0 +1,48 @@
+class GetInvolved < ContentItem
+  attr_reader :take_part_pages
+
+  def initialize(content_store_response, override_content_store_hash: nil)
+    super(content_store_response, override_content_store_hash:)
```
KludgeKML, line 5: For the sake of simplicity, if you call super without any parameters it takes the parameters from the calling method. 

So you could replace `super(content_store_response, override_content_store_hash:)` with just `super`

app/models/content_item.rb
```diff
@@ -1,6 +1,6 @@
 class ContentItem
```
KludgeKML: You could cut out this whole commit, since you're overwriting it later.

app/models/get_involved.rb
```diff
@@ -5,4 +5,44 @@ def initialize(content_store_response, override_content_store_hash: nil)
...
+      filter_end_date: "from: #{Time.zone.now.to_date}",
+      fields: "end_date,title,link",
+      order: "end_date",
+      count: 1,
+    }
+
+    GdsApi.search.search(query)["results"].first
+  end
+
+  def time_until_closure(consultation)
```
leenagupte, line 35: This method has moved about a bit. Considering that's looking up specific "get_involved" translation keys it should be in the presenter.

.byebug_history
leenagupte, line 1: This file shouldn't have been commited.

app/presenters/get_involved_presenter.rb
```diff
@@ -1,9 +1,5 @@
 class GetInvolvedPresenter < ContentItemPresenter
-  # def initialize(content_item)
-  #   super(content_item)
-  #   @content_item = content_item
-  # end
-
+
```
leenagupte: It's good that this has been deleted, but this should be fixed up to the previous commit.

.gitignore
```diff
@@ -18,6 +18,7 @@ spec/reports/pacts
 .yarn/cache/
 .yarn/install-state.gz
 .nyc_output
+.byebug_history
```
leenagupte, line 21: I think that as this change is not related to Get Involved pages, it should either be in it's own commit, or even better, added in a separate PR.

spec/requests/get_involved_spec.rb
```diff
@@ -34,7 +34,7 @@
     GetInvolved.new(content_item)
   end
 
-  context "when the consultation is closed" do        
+  context "when the consultation is closed" do
```
leenagupte, line 37: Lint errors should be fixed up to the commit they first occurred in.

spec/requests/get_involved_spec.rb
```diff
@@ -15,6 +16,29 @@
   end
 end
 
+RSpec.describe "Time until closure" do
```
leenagupte, line 18: I don't think we've documented this well, but the convention we have is for the top-level descriptor to use `RSpec.describe` and all of the other blocks to in the test file to either use `describe` if describing a method name, or `context` if describing a scenario. I think for this test only the "when the consultation is closed" context is needed.

leenagupte (commented):
I think this is a really good start.

The Get Involved pages turned out to much more complicated then at first realised due to the need to integrate with search. That means that the original presenter in government-frontend has a lot of tightly-coupled model and presentation logic.

You've made a valiant attempt to sort this out, but the to-ing and fro-ing about where a method should go has left the commit history for this PR in a bit of a mess. With these consolidating app PRs the commit history is more important than usual because we need to sign post future developers to where the changes they're looking at originated from.

I think the request and system tests are in a good place, that will help with sorting out the model and presenter.

I think the next steps should be to sort out the nesting in the locale files, and fix up all of the lint errors, and accidentally commited files.

For the model and presenter, I think the best option is to squash those changes into one commit, and then it'll be easier to separate out the methods into their individual parts. I do think those methods can't just be lifted and shifted from the original presenter. which is a real shame.

A last thing to check is the meta tags and make sure that they have been added correctly. The easiest way to do that is to include the shared example in the [system tests](https://github.com/alphagov/frontend/blob/890c15a940fd98ba3b2fd069aae882376eaff064/spec/system/help_page_spec.rb#L2).

I think rebasing and sorting out the model and presenter is going to be quite large job and I'm happy to pair with you on this.

.byebug_history
KludgeKML: Gone now

spec/requests/get_involved_spec.rb
leenagupte: Good spot! Thanks

app/presenters/get_involved_presenter.rb
```diff
@@ -0,0 +1,107 @@
+class GetInvolvedPresenter < ContentItemPresenter
```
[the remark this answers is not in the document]
sairamya93: Updated the presenter and added the link for the original get_involved presenter

spec/requests/get_involved_spec.rb
```diff
@@ -0,0 +1,38 @@
+RSpec.describe "Get Involved" do
+  before do
+    stub_content_store_has_item("/government/get-involved", GovukSchemas::Example.find("get_involved", example_name: "get_involved"))
+    stub_search_query(query: hash_including(filter_content_store_document_type: "open_consultation"), response: { "results" => [], "total" => 83 })
+    stub_search_query(query: hash_including(filter_content_store_document_type: "closed_consultation"), response: { "results" => [], "total" => 110 })
+    stub_search_query(query: hash_including(filter_content_store_document_type: "consultation_outcome"), response: { "results" => [consultation_result] })
```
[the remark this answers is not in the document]
KludgeKML: I think we want the actual results in the response, which stub_any_search_post doesn't do?

KludgeKML (approved):
LGTM

spec/requests/get_involved_spec.rb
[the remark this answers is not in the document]
leenagupte: This request test isn't testing the search results, so they're not really needed. I don't think all of that search stubbing is required when just the response and the page title is being tested.

spec/presenter/get_involved_spec.rb
```diff
@@ -0,0 +1,111 @@
+RSpec.describe GetInvolved do
+  ## Mock data for testing
```
leenagupte, line 2: Minor: This comment isn't needed.

spec/models/get_involved_spec.rb
leenagupte, line 1: Minor, but could this tests be moved to the commit following the addition of the model and presenter?

app/presenters/get_involved_presenter.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def time_until_next_closure
+    days_left = (content_item.next_closing_consultation["end_date"].to_date - Time.zone.now.to_date).to_i
+    case days_left
+    when :negative?.to_proc
+      I18n.t("formats.get_involved.closed")
+    when :zero?.to_proc
+      I18n.t("formats.get_involved.closing_today")
+    when 1
```
leenagupte, line 23: This case statement is pretty hideous. I know you've copied it verbatim from government-frontend, which was the correct thing to do, but perhaps it could be refactored in another commit to use "if/else"?

i.e.

```ruby
if days_left.negative?
  ...
elsif days_left.zero?
  ...
elsif days_left == 1
 ...
else
 ...
end
```

config/locales/az.yml
```diff
@@ -388,6 +388,47 @@ az:
   formats:
     find_my_nearest:
       valid_postcode_no_locations:
+    get_involved:
```
leenagupte, line 391: Thanks for moving these ⭐

spec/system/get_involved_spec.rb
```diff
@@ -0,0 +1,60 @@
+RSpec.describe "Get Involved" do
+  it_behaves_like "it has meta tags", "get_involved", "/government/get-involved"
```
leenagupte, line 2: This is great!

leenagupte (commented):
The changes look great! 🎉 

I've added a couple of inline requests for a bit of tidy-up, but it's nothing major. You've done the lions share and it looks good. 

The commit audit trails back to government-frontend looks good too.

spec/requests/get_involved_spec.rb
```diff
@@ -0,0 +1,15 @@
+RSpec.describe "Get Involved" do
+  before do
+    stub_content_store_has_item("/government/get-involved", GovukSchemas::Example.find("get_involved", example_name: "get_involved"))
+    stub_request(:get, /\A#{Plek.new.find('search-api')}\/search.json/).to_return(body: { "results" => [], "total" => 83 }.to_json)
```
leenagupte, line 4: This is much better! Thanks for getting rid of the extraneous stubbing in this test 🎉

leenagupte (approved):
This looks great! 👍 

Thanks for persevering with all of the rebasing.

I've got the page running on Heroku too and it looks good: https://govuk-frontend-app-pr-4305.herokuapp.com/government/get-involved

The commit structure is great, and the audit trail back to government-frontend looks complete. 🎉 

There are two follow-up tasks to do as part of making these changes live:

1. Get the pages looked over by a frontend dev to make sure nothing has been missed. Last time they caught that the `main` tag was missing.
2. Now that frontend relies on `search-api` that dependency should be added to the govuk-docker config, startup script, and app.json (for Heroku). These can be updated in follow-up PRs.

PR #4307, opened 2024-10-21
Fix calendar param names

KludgeKML:

## What

Improve / Simplify bank holiday routing

## Why

Routing for bank holiday pages currently relies on a side-effect of other (correctly written) FormatRoutingConstraint blocks above it, has unnecessary explicit routes for the Gwyliau-banc pages, and always loads the content item twice from the API.

## How

Fix the parameter from scope to slug and translate to scope only in the controller, remove the explicit gwyliau-banc routes that existed only to set the locale to Welsh on those pages and simply use the locale from the relevant content item. Finally, remove the overridden content_item method to use the ContentItemController default method, which utilises the data cached during routing, preventing a second call to the API for the same data.

## Screenshots?

No visual differences

### DIFF
--- a/app/controllers/calendar_controller.rb
+++ b/app/controllers/calendar_controller.rb
@@ -40,10 +40,6 @@ private
 
   helper_method :calendar
 
-  def content_item
-    @content_item ||= GdsApi.content_store.content_item("/#{params[:slug]}")
-  end
-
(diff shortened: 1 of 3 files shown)


leenagupte (approved):
Nice! 🎉

PR #4312, opened 2024-10-22
Tweak main_navigation block to align with designs

AshGDS:

## What / Why
- Makes the margin between the `Service name` (black link) and the menu items (blue links) the same.
- Adds more spacing to the mobile menu, and adds left padding to the list items so that they are staggered away from the menu button, as per the designs.
- Adds a `border_bottom` to the block. This required turning it into a proper block, so that `full_width?` could be used, otherwise the border would only go up to the width limits of the `govuk-width-container` class. 
- Renames the `children:` variable for child links to `items:`. I tried to get it work with `children:` as thats more readable, but for some reason it didn't work. Might be a backend thing going over my head.
- Trello: https://trello.com/c/BmgLrgZD / https://trello.com/c/hzfHN1fI

### Before
image
image

### After

image

image

### DIFF
--- a/app/assets/stylesheets/views/_landing_page/main-navigation.scss
+++ b/app/assets/stylesheets/views/_landing_page/main-navigation.scss
@@ -1,12 +1,20 @@
 @import "govuk_publishing_components/individual_component_support";
 
 .app-b-main-nav {
+  border-bottom: 2px solid $govuk-border-colour;
+}
+
+.app-b-main-nav__container {
   padding-bottom: govuk-spacing(1);
 
   @include govuk-media-query($from: "desktop") {
     display: flex;
     flex-wrap: wrap;
     align-items: center;
+  }
+  @include govuk-media-query($until: "desktop") {
+    padding-top: govuk-spacing(4);
+    padding-bottom: govuk-spacing(4);
   }
 }
 
@@ -68,7 +76,7 @@
 .app-b-main-nav__heading-p {
   display: inline-block;
   margin-bottom: 0;
-  margin-right: govuk-spacing(9);
+  margin-right: govuk-spacing(7);
 }
 
 .app-b-main-nav__heading-link {
@@ -83,6 +91,7 @@
 
 .app-b-main-nav__childlist {
   display: none;
+  padding-left: govuk-spacing(2);
   @include govuk-media-query($until: "desktop") {
     display: block;
   }
@@ -99,13 +108,19 @@
 
   @include govuk-media-query($until: "desktop") {
     display: block;
-    margin-top: govuk-spacing(3);
+    margin-top: govuk-spacing(4);
+    margin-left: govuk-spacing(2);
+  }
+}
+
+.app-b-main-nav__listitem:last-of-type {
+  @include govuk-media-query($until: "desktop") {
+    margin-bottom: 0;
   }
 }
 
 .app-b-main-nav__link {
   @include govuk-media-query($from: "desktop") {
-    padding-top: govuk-spacing(3);
     padding-bottom: govuk-spacing(3);
   }
 }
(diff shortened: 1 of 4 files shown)


app/views/landing_page/blocks/_main_navigation.html.erb
```diff
@@ -1,45 +1,46 @@
...
+              </ul>
+            <% end %>
           <% end %>
         <% end %>
-      <% end %>
-    </ul>
-  </nav>
+      </ul>
+    </nav>
+  </div>
```
AshGDS, line 45: For ease of review as the diff looks a bit crazy, all I've done in this file is:
- added `<div class="app-b-main-nav--border">` around the block
- added `govuk-width-container` to the first child div
- switched `block.data["title"]` to `block.title`, `block.data["children"] to `block.items`, etc.

app/views/landing_page/blocks/_main_navigation.html.erb
```diff
@@ -1,45 +1,46 @@
 <%
   add_view_stylesheet("landing_page/main-navigation")
 %>
+<div class="app-b-main-nav--border">
+  <div class="app-b-main-nav govuk-width-container" data-module="app-b-main-navigation">
```
andysellick, line 5: Seems like the BEM's got a bit confusing here... `app-b-main-nav--border` is a modifier class, but it's being used as the parent class. I think the first class that wraps the whole block should be `app-b-main-nav` and then the inner div (with the govuk width container) should be an element class, like `app-b-main-nav__wrapper` or something. Will need to juggle the actual styles, obviously.

andysellick, line 5: I'd also put the `data-module` on the parent, if that's not going to cause a problem.

app/assets/stylesheets/views/_landing_page/main-navigation.scss
andysellick, line 1: The focus state for desktop links seems to be offset up by about a pixel? Not sure if that's intentional?

Screenshot 2024-10-23 at 09 40 55

AshGDS: Thanks, fixed.

app/views/landing_page/blocks/_main_navigation.html.erb
AshGDS: Thanks, done 👍

spec/models/block/main_navigation_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+            {
+              "text" => "Child b",
+              "href" => "/b",
+            },
+          ],
+        },
+      ] }
+  end
+
+  describe "#title" do
```
leenagupte, line 28: I love that you've added model tests ⭐

app/models/block/main_navigation.rb
```diff
@@ -0,0 +1,17 @@
+module Block
+  class MainNavigation < Block::Base
+    attr_reader :title, :title_link, :links
+
+    def initialize(block_hash)
+      super(block_hash)
+
+      @title = data.fetch("title")
```
leenagupte, line 8: To be honest, I'm not sure if this model is really needed as the you can just use `block.data["title"]` etc in the views. However being able to call `block.title` is tidier.

leenagupte (commented):
This looks good from a backend perspective. 👍 

As there's a lot of frontendy changes, I'll let Andy approve.

AshGDS:
Thanks both 👍 

@andysellick - the frontend should be ready for re-review

app/assets/stylesheets/views/_landing_page/main-navigation.scss
andysellick: The issue's fixed although I notice it's changed the appearance of the focus state so that it no longer reaches to the top of the element. I think this is fine and shouldn't block this PR, but I've realised that I can't see any focus states in the designs so we should get this checked separately by design.

andysellick (approved):
Frontend looks good to me 👍

PR #4314, opened 2024-10-23
Add small charts

andysellick:

## What

- add small charts to /landing-page/homepage
- updates the card block to have the image be optional (we might want to remove the image option entirely at some point, but leaving it for now)
- modifies the data for the card blocks to have statistics as an extra block after heading
- make some changes to how the card block outputs these blocks for style purposes by wrapping each block in an appropriate wrapper (heading needs to be in a blue box, chart needs to be in a box with a border round it)
- change the heading block to accept a margin parameter, so that 0 can be passed when it's inside a card text box (spacing around the heading is already included) and include that option in the data files
- change the statistics block to accept options to output a minimal chart component and include those options in the data files
- tweak some of the styles in the card block to fix the background colour and add a border for a chart

## Why

Existing charts were image placeholders.

## Visual changes

Charts are now actual charts, in minimal mode (using placeholder data):

Screenshot 2024-10-23 at 11 34 52

Trello card: https://trello.com/c/DFT05NWN

### DIFF
--- a/app/views/landing_page/blocks/_card.html.erb
+++ b/app/views/landing_page/blocks/_card.html.erb
@@ -1,16 +1,24 @@
 <%
   add_view_stylesheet("landing_page/card")
-
-  link_classes = %w(app-b-card__link govuk-link gem-print-link)
 %>
-
 <div class="app-b-card">
-  <div class="app-b-card__textbox">
-    <% block.card_content.each do |subblock| %>
+  <% block.card_content.each do |subblock| %>
+    <% if subblock.data["type"] == "heading" %>
+      <div class="app-b-card__textbox">
+        <%= render_block(subblock)  %>
+      </div>
+    <% elsif subblock.data["type"] == "statistics" %>
+      <div class="app-b-card__chart">
+        <%= render_block(subblock)  %>
+      </div>
+    <% else %>
       <%= render_block(subblock)  %>
     <% end %>
-  </div>
-  <figure class="app-b-card__figure">
-    <%= image_tag(block.image.source, alt: block.image.alt, class: "app-b-card__image") %>
-  </figure>
+  <% end %>
+
+  <% if block.image %>
+    <figure class="app-b-card__figure">
+      <%= image_tag(block.image.source, alt: block.image.alt, class: "app-b-card__image") %>
+    </figure>
+  <% end %>
 </div>
--- a/app/views/landing_page/blocks/_card.html.erb
+++ b/app/views/landing_page/blocks/_card.html.erb
@@ -5,11 +5,13 @@
   <% block.card_content.each do |subblock| %>
     <% if subblock.data["type"] == "heading" %>
       <div class="app-b-card__textbox">
-        <%= render_block(subblock)  %>
+        <% options = { margin_bottom: 0, inverse: true } %>
+        <%= render_block(subblock, options:)  %>
       </div>
     <% elsif subblock.data["type"] == "statistics" %>
       <div class="app-b-card__chart">
-        <%= render_block(subblock)  %>
+        <% options = { margin_bottom: 0, height: 200 } %>
+        <%= render_block(subblock, options:)  %>
       </div>
     <% else %>
       <%= render_block(subblock)  %>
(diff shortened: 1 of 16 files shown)


app/views/landing_page/blocks/_card.html.erb
```diff
@@ -1,16 +1,24 @@
...
   add_view_stylesheet("landing_page/card")
-
-  link_classes = %w(app-b-card__link govuk-link gem-print-link)
 %>
-
 <div class="app-b-card">
-  <div class="app-b-card__textbox">
-    <% block.card_content.each do |subblock| %>
+  <% block.card_content.each do |subblock| %>
+    <% if subblock.data["type"] == "heading" %>
```
leenagupte, line 6: I like this. I think this logic will make it easier to move the image logic here as well if we wanted, so it too could be defined as a block.

lib/data/landing_page_content_items/homepage.yaml
```diff
@@ -67,35 +67,59 @@ blocks:
...
+          margin_bottom: 0
+        - type: statistics
+          title: "Chart to visually represent some data"
+          x_axis_label: "X Axis"
+          y_axis_label: "Y Axis"
+          csv_file: "data_one.csv"
+          data_source_link: https://www.ons.gov.uk/economy/grossdomesticproductgdp/timeseries/ihyq/qna
+          minimal: true
+          minimal_link: /landing-page/task
+          margin_bottom: 0
```
leenagupte, line 85: It feels odd having this level of presentation data in what will ultimately become the content item.

Should this always be `0` if `minimal` is true? Could `minimal` be used to determine the `margin_bottom` instead?

leenagupte (commented):
This looks great and works like a charm.

I have one inline question about how / where `margin_bottom` is defined.

lib/data/landing_page_content_items/homepage.yaml
andysellick: Yes, and that's much tidier, so I've made that change. I've also applied this to the `height` parameter, as that also makes sense.

However, I think this only works for this exact design, where we want to use these small charts in a three column layout - this wouldn't work for a more flexible system where the chart might appear elsewhere. I think at that point we could look at putting those overrides (for height, margin_bottom etc.) into the cards block, for that specific chart block.

leenagupte: I've been having a think about this, as needing to set the `margin_bottom` affects the `heading` block too, and that doesn't have an `minimal` attribute that can be hijacked.

I think instead, the `card` view can be used to control how the blocks that appear inside it are rendered.

e.g.
Update `card` to pass an `options` hash to `render_block`

```ruby
<% block.card_content.each do |subblock| %>
    styling = { margin_bottom: 0 }

    <% if subblock.data["type"] == "heading" %>
      <div class="app-b-card__textbox">
        <%= render_block(subblock, styling:)  %>
      </div>
    <% elsif subblock.data["type"] == "statistics" %>
      <div class="app-b-card__chart">
        <%= render_block(subblock, styling:)  %>
      </div>
    <% else %>
      <%= render_block(subblock)  %>
    <% end %>
  <% end %>
```

Update the `render_block` method to accept an `options` hash:

```ruby
  def render_block(block, options: {})
    render("landing_page/blocks/#{block.type}", block:, options:)
  rescue ActionView::MissingTemplate
    Rails.logger.warn("Missing template for block #{block.type}")
    ""
  end
``` 

Lastly, update the views to read from the `options` hash

e.g. in the `statistics` partial add:

```ruby
<%
  add_gem_component_stylesheet("chart")

  margin_bottom = options[:margin_bottom] || 8
  height = block.data["height"]
%>
```

I wonder if something similar could be applied for the height?

The `heading` partial would have something similar

```ruby
<%
.
.
  margin_bottom = styling[:margin_bottom] || 6
%>
```

andysellick: @leenagupte great shout, have implemented this. It's spread a bit wider than this - turns out this is a great way for blocks to control `inverse` for subblocks.

PR #4315, opened 2024-10-23
Create dummy tasks pages

andysellick:

## What

- create some copies of the 'task' page and give them dummy names and add them to the side navigation, so that we can reasonably navigate around the pages
- also update the main nav header to point to /landing-page/homepage, which is our actual homepage, instead of /landing-page/, which is a kind of placeholder test page that we'll need to delete fairly soon
- also delete the old `/landing-page/task` page as these new ones replace it

## Why
Aiming to be at the stage where we can navigate around these pages as they are intended, even if we don't have any content yet.

## Visual changes
New pages are just copies of the existing page at `/landing-page/task`.

Trello card: https://trello.com/c/t0DlqDkW


lib/data/landing_page_content_items/be_kinder.yaml
```diff
@@ -5,13 +5,8 @@ blocks:
     href: /landing-page/goals
   - text: Tasks
     href: /landing-page/tasks
-    children:
-      - text: Child 1
-        href: /a
-      - text: Child 2
-        href: /b
```
AshGDS, line 12: Haven't done a full code review as I assume this is more of a backend PR, but are we able to keep the child links in so that the pages match the designs on mobile?  

The key was renamed to `items` as part of https://github.com/alphagov/frontend/pull/4285, so if you change the key to `items` you should see the child links on mobile.

(I've also renamed `children` to `items` in https://github.com/alphagov/frontend/pull/4312)

Thanks!

leenagupte (commented):
I've added one inline comment, similar to Ash's about main navigation, otherwise it all looks good. The side navigation works like a charm. 

There should definitely be a task about eating more cake though 😉

lib/data/landing_page_content_items/be_kinder.yaml
andysellick: Restored.

lib/data/landing_page_content_items/goals.yaml
```diff
@@ -5,13 +5,8 @@ blocks:
       href: /landing-page/goals
     - text: Tasks
       href: /landing-page/tasks
-      children:
-        - text: Child 1
-          href: /a
-        - text: Child 2
-          href: /b
```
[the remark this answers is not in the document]
andysellick: I've put the sub links into all of the pages now. I'm not convinced they should be there (as they only appear on mobile, which means users could rotate their device and be confused about suddenly missing navigation) but save that conversation for the accessibility testing.

lib/data/landing_page_content_items/be_kinder.yaml
AshGDS: Thanks!

lib/data/landing_page_content_items/learn_something_new.yaml
```diff
@@ -0,0 +1,146 @@
...
+        path: https://www.gov.uk
+        document_type: News article
+        public_updated_at: "2021-01-16 11:34:12 +0000"
+      - text: A third example link
+        path: https://www.gov.uk
+        document_type: Consultation
+        public_updated_at: "2024-02-01 09:00:11 +0000"
+
+
+#### Curated links go here ###
```
leenagupte, line 129: These comment can be removed as you've added a `document_list` block. Sorry!

andysellick: Removed, thanks!

PR #4320, opened 2024-10-23
Update routing constraints

KludgeKML:

## What

Update routing constraints so that we can put landing pages on arbitrary paths.

## Why

Landing pages may be on any path, so we need a way to match them by routing constraint. The existing FormatRoutingConstraint was designed for items where an initial slug served multiple paths, and cached at 
the slug level, meaning it could never match a full path.

## How

Add a new class to handle format constraints from full paths, which does not cache its calls to the content store or
rely on preexisting cached calls. This isn't optimal, but acceptable for the moment given that there is currently only one 
schema_name being matched this way (landing_page).


leenagupte (approved):
Looks great! 🥇

PR #4330, opened 2024-10-24
[WIP] Create guidance for building flexible blocks

leenagupte:

## What

Documents what each of the blocks is, when they should be used and how they are configured.

## Why

[Trello card](https://trello.com/c/eoF6iYLM)


docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,566 @@
...
+#### Card
+
+A card block renders its `card_content:` value as a [Heading component](https://components.publishing.service.gov.uk/component-guide/heading), followed by the subblocks laid out vertically, followed by an image (if specified - at time of writing the image in the card block can not be taken from the content item and must be present in frontend).
+
+##### Example
+
+```yaml
+- type: card
+  card_content:
+  blocks:
```
leenagupte, line 256: TO DO: Blocks need to be nested under `card_content`,

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,566 @@
...
+##### Example
+
+```yaml
+- type: heading
+  content: Porem ipsum dolor
+```
+
+#### Image
+
+A simple image.
```
leenagupte, line 127: TO DO: should mention that 6 versions of an image need to be supplied and that this also applies to blocks that contain images.

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,566 @@
...
+```yaml
+extends: /landing-page/homepage
+blocks:
+- type: main_navigation
+  navigation_group_id: Top Menu
+- type: govspeak
+  content: <p>Hello from the child!</p>
+```
+
+...the navigation_groups element will be copied into the child page when it is saved (allowing you to specify a shared menu in one page and simply republish all the child pages with a standard Whitehall rake task).
```
leenagupte, line 566: TO DO: Should mention whether the `extends` attribute copies over everything from the parent or just `navigation_groups`.

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,566 @@
...
+#### Share links
+
+A wrapper around a [Heading component](https://components.publishing.service.gov.uk/component-guide/heading) followed by a [Share links component](https://components.publishing.service.gov.uk/component-guide/share_links)
+
+##### Example
+
+```yaml
+- type: share_links
+  links:
+    - href: /twitter-profile
```
leenagupte, line 176: TO DO: check the YAML indentation in all of the examples as they are likely to be copy and pasted into Whitehall.

leenagupte, line 176: TO DO: check the YAML indentation in all of the examples as they are likely to be copy and pasted into Whitehall.

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
+# Building blocks for flexible content
+
+The "landing-page" document type will allow content designers to build pages with a flexible design by creating a "block" that renders certain parts of the page.
+
+Blocks can be nested within each other. A top-level block is a horizontal chunk of the page.
+
+Blocks are rendered in the order that they appear in the content item.
+
+Blocks are added to the content item in the details hash. The landing page editor is currently a work in progress. At time of writing, the config for blocks need to be added in YAML format in Whitehall. This YAML config is then merged into `details` hash when the landing page is saved and published.
```
maxgds, line 9: grammar nitpick: either "the details hash" or just "details" without "hash"

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+- type: block_type
+```
+
+...and may include a top-level `navigation_groups` element for use by blocks (See [Navigation Groups](#navigation-groups))
+
+Block configuration is designed to be as flexible as possible, so blocks can be nested inside other blocks. An example configuration for each type of block is shown below.
+
+## Types of Block
+
+Each block rendered using its own partial template. The full list of available types of block available can be found by looking at the contents of: `app/views/landing_page/blocks`
```
maxgds, line 24: Grammar nitpick:
Each block is rendered
OR
Each block renders

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+
+### Simple blocks
+
+Simple blocks general render one component or "thing". They can either be rendered directly or as the children of compound blocks.
```
maxgds, line 28: Nitpick:
generally, not general

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+  - text: A third example link
+    path: https://www.gov.uk
+    document_type: Consultation
+    public_updated_at: "2024-02-01 09:00:11 +0000"
+```
+
+#### Govspeak
+
+A wrapper around the [Govspeak content component](https://components.publishing.service.gov.uk/component-guide/govspeak). Note that the content item data for the component is not Govspeak, but HTML that has been converted from Govspeak. If you're editing the block data in Whitehall, though, you can use Govspeak directly by taking advantage of the publishing-api automatic conversion (see second example)
+
```
maxgds, line 92: This reads as a bit confusing. This makes it sound like I need to go and find a tool to produce some HTML by converting it whereas in fact I can just supply any supported HTML. I don't think there's any need to mention publishing-api, just that there are two options for input: Any HTML which only uses the govspeak approved subset of tags, or govspeak. Then call out the syntax difference needed for each option as illustrated.

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+    content: |
+      ### Lorem heading
+
+      Korem ipsum dolor sit amet, consectetur adipiscing elit. Etiam eu turpis
+      molestie, dictum esta, mattis tellus.
+```
+
+#### Heading
+
+A wrapper around the [Heading component](https://components.publishing.service.gov.uk/component-guide/heading). Note, though, that the block uses `content` where the component uses `text` for the title text key. This is so that headings can be indexed (only values inside a `content` key will be indexed when being published from Whitehall - see [Indexing block content in search](#indexing-block-content-in-search))
```
maxgds, line 116: This contextless reference to indexing threw me. The link to indexing helped, but I could have made more sense of this if you specified search here - I was thinking we were numbering headings for some reason as per an array index. Suggestion:
"This is so that headings can appear in search (only values..." etc

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+
+```yaml
+- type: quote
+  text: "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nunc vulputate libero et velit interdum, ac aliquet odio mattis class"
+  cite: "Lorem ipsum dolor sit, 28 September 2024"
+```
+
+#### Share links
+
+A wrapper around a [Heading component](https://components.publishing.service.gov.uk/component-guide/heading) followed by a [Share links component](https://components.publishing.service.gov.uk/component-guide/share_links)
```
maxgds, line 171: The example doesn't seem to include a heading.

maxgds: Unless that's the hidden text? In which case might be worth mentioning that explicitly.

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+##### Example
+
+```yaml
+- type: side_navigation
+  navigation_group_id: Sidebar
+```
+
+#### Statistics
+
+A wrapper around the [Chart](https://components.publishing.service.gov.uk/component-guide/chart) component. The `csv_file` key can be a hardcoded file (from /lib/data/landing_page_content_items/statistics), or (preferably) an attachment from the content item, in which case the filename of the attachment will be matched to the `csv_file` key.
```
maxgds, line 213: suggestion:
"uploaded file" rather than hardcoded

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+
+- [Blocks container](#blocks-container)
+- [Columns layout](#columns-layout)
+- [Grid Container](#grid-container)
+- [Two column layout](#two-column-layout)
+
+#### Blocks container
+
+A blocks container is used as an empty unstyled parent container to
+hold other elements. It is used when we don't want to want to create
```
maxgds, line 339: Why/when would we not want to create a row/grid layout?

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+      theme: full_width
+      src: "landing_page/placeholder/960x640.png"
+      alt: "Placeholder image"
+    - type: govspeak
+      content: |
+        <p>Korem ipsum dolor sit amet, consectetur adipiscing elit. Etiam eu turpis molestie, dictum est a, mattis tellus.</p>
+```
+
+The left column is always populated if only one column is defined.
+A blocks container applied to a column so that everying in the column can be styled relative to one and other.
```
maxgds, line 489: I think this sentence is missing some words, I don't quite follow. Also typo in "everying".

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+
+## Indexing block content in search
+
+Only content in blocks that has a label of `content` will be indexed in site search.
```
maxgds, line 493: grammar: "have a label" not has

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+        href: "/landing-page/donate-to-charity"
+      - text: Learn something new
+        href: "/landing-page/learn-something-new"
+      - text: Be thankful
+        href: "/landing-page/be-thankful"
+```
+
+## Extending a page (Whitehall Only)
+
+When editing a page in Whitehall, you may wish to use data from another page. Adding the top-level key `extends`, and the path of another landing page, will cause the data from the current block to be merged over the data from the extended block before publishing. This is a shallow merge (ie only top-level values are checked), and is currently used so that navigation groups can be specified in one page and used by sub-pages.
```
maxgds, line 536: Suggestion: For clarity change "you may wish to use data from another page" to "you may wish to use data from another flexible content page" as there are many things on GOV.UK called pages.

docs/building_blocks_for_flexible_content.md
```diff
@@ -0,0 +1,571 @@
...
+blocks:
+- type: main_navigation
+  navigation_group_id: Top Menu
+- type: govspeak
+  content: <p>Hello from the child!</p>
+```
+
+...the navigation_groups element will be copied into the child page when it is saved (allowing you to specify a shared menu in one page and simply republish all the child pages with a standard Whitehall rake task).
+
+If the child page didn't specify any blocks, those too would be merged over from the parent. That's because `extends` does a shallow merge of everything in the `details` element of the parent. This means that it is possible to add more top-level elements that can be shared across child pages.
```
maxgds, line 571: I'm a bit confused by this bit about copying blocks over and why they are copied if empty and not copied otherwise.

I also don't really understand what I'll see in Whitehall - if I create a child page as above and save it, when I load it again will I see what I originally wrote or will it have expanded out all the links? If a navigation item changes do I need to hand edit every child or run a rake task to do so?`

Links out to running relevant rake tasks would be helpful if that's necessary - details on how to specify the right content items for the rake task etc

maxgds (changes requested):
A few changes requested and a bit of clarity sought. Some quite nitpicky so it's not as bad as it might seem at a glance.

docs/building_blocks_for_flexible_content.md
leenagupte: At the moment the heading is hard-coded in the share links block and is not configurable.

docs/building_blocks_for_flexible_content.md
leenagupte: I pilfered that explanation from the [commit message](https://github.com/alphagov/frontend/commit/f15d1e35ed7140b4eac00497ccca7ffd88fbd735). Does the comment in the PR make sense? If so I can replicate it wholesale.

docs/building_blocks_for_flexible_content.md
leenagupte: Does this read better?

---

A wrapper around the [Govspeak content component](https://components.publishing.service.gov.uk/component-guide/govspeak). 

Content for this block can either be written as govspeak or HTML. If you are adding a Govspeak block in config in the `frontend` application, it must be added as HTML. This is because, despite the name, the Govspeak component it calls actually renders HTML content.

If you are adding the block in config in Whitehall you have a choice of writing it in either govspeak or HTML. That is because govspeak content added in Whitehall is automatically converted to HTML by publishing-api. However, if you use govspeak, you'll need to supply the `content-type` (see second example).

maxgds: Yes, much better.

I know that govspeak doesn't support all HTML and will drop some tags, would that affect stuff input here? I'm not sure what tags are unsupported. If it would affect though it's worth calling out.

docs/building_blocks_for_flexible_content.md
maxgds: Worth stating that, then, I think.

docs/building_blocks_for_flexible_content.md
maxgds: Yes, that helps. I think the key is knowing that it's for situations where you've already defined your grid and columns and you're nesting things within that space.

docs/building_blocks_for_flexible_content.md
leenagupte: I didn't know that! The best I can do is link to the govspeak docs.

docs/building_blocks_for_flexible_content.md
leenagupte: Oh that's a good explanation. I'll add it.

leenagupte:
@maxgds Thanks for your review. I have applied all of the recommendations in your comments and fixed up. Can you give it another review?

docs/building_blocks_for_flexible_content.md
maxgds: I think it's fine as is.

PR #4331, opened 2024-10-24
Read menu items from document collections

KludgeKML:
Work in progress - assumes that all documents that have main_navigation (and side_nav, but haven't tested that yet) blocks in them are either:
1) a landing page with a document collection called "Top menu" and another called "Side navigation", or
2) a landing page that's part of those document collections.

Important notice - the default block initializer is updated as part of this work to include the landing page content item as a parameter, so that these navigation blocks (and others in the future, perhaps image and chart blocks) will be able to read data from the parent landing page.

When a landing page is created, it lazily initialises a collection_group map. This is generated either from the existing details.collection_group + links.documents info in the content item hash (in case 1), or by loading the first linked item from links.document_collections and copying the collection_group info from that.

TODO:
- [ ] stop hardcoding the "Top menu" / "Side navigation" in the blocks, and take the relevant collection groups from a data item in those blocks.
- [ ] work out how to handle the nested submenu in main_navigation.
- [ ] test suite

https://trello.com/c/2xR6z6o4/86-consider-better-way-of-managing-sidebar-links-than-hardcoding-them

### DIFF
--- a/app/models/block/base.rb
+++ b/app/models/block/base.rb
@@ -1,11 +1,12 @@
 module Block
   class Base
-    attr_reader :id, :type, :data
+    attr_reader :id, :type, :data, :landing_page
 
-    def initialize(block_hash)
+    def initialize(block_hash, landing_page)
       @data = block_hash
       @id = data["id"]
       @type = data["type"]
+      @landing_page = landing_page
     end
 
     def full_width?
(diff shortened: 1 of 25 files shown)


KludgeKML:
Decided against this approach for the moment.

PR #4333, opened 2024-10-24
Add bottom margin to image blocks

JamesCGDS:

## What
This PR adds some bottom margin to the image block via a govuk spacing override class.

I opted for `govuk-!-margin-bottom-6` as that's what the `h2` above the image has.

## Why
As per design review.

### DIFF
--- a/app/views/landing_page/blocks/_image.html.erb
+++ b/app/views/landing_page/blocks/_image.html.erb
@@ -1,5 +1,10 @@
+<% 
+  img_class = ("govuk-!-width-full" if block.data["theme"] == "full_width")
+  img_class << " govuk-!-margin-bottom-6"
+%>
+
 <%= image_tag(
       block.data["src"],
       alt: block.data["alt"],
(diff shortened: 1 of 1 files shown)


app/views/landing_page/blocks/_image.html.erb
```diff
@@ -1,5 +1,10 @@
+<% 
+  img_class = ("govuk-!-width-full" if block.data["theme"] == "full_width")
+  img_class << " govuk-!-margin-bottom-6"
```
andysellick, line 3: Apparently these should be used for overriding only, but I think considering the time pressures here I'm okay with it, but we might circle back and redo later.

andysellick (approved):

app/views/landing_page/blocks/_image.html.erb
JamesCGDS: Sounds good, thanks Andy 👍

PR #4346, opened 2024-10-28
Homepage: Fix search form path

csutter:
## What
Fix the URL for the homepage search form

## Why
The form points to the wrong path for search - `/search` is a landing page, not the actual all content finder (which is `/search/all`). This means users searching from the homepage have to go through an unnecessary redirection:
image

See also: https://github.com/alphagov/govuk_publishing_components/pull/4341

## How
- Changes URL in the form
- Cleans up the form code by using Rails's tag helper for neater data attributes


leenagupte (approved):
🎉

PR #4347, opened 2024-10-28
Add main html tag

AshGDS:
## What / Why
- Add `<main>` HTML element
- For accessibility reasons
- https://trello.com/c/ZIIk6smn


KludgeKML (approved):
LGTM

PR #4350, opened 2024-10-28
Statistics from attachments

KludgeKML:
Allow landing_page statistics blocks to use attachments linked from the content item.

https://trello.com/c/bUVC9hYD/123-csv-data-as-attachments-in-whitehall

### DIFF
--- a/app/models/landing_page/block/statistics.rb
+++ b/app/models/landing_page/block/statistics.rb
@@ -1,7 +1,9 @@
 require "csv"
+require "open-uri"
 
 module LandingPage::Block
   class Statistics < Base
+    # SCAFFOLDING
     STATISTICS_DATA_PATH = "lib/data/landing_page_content_items/statistics".freeze
 
     def x_axis_keys
@@ -30,15 +32,25 @@ module LandingPage::Block
       rows
     end
 
+    def attachment
+      @attachment ||= landing_page.attachments.find { |att| att.filename == data["csv_file"] }
+    end
+
   private
 
     def csv_rows
-      rows = CSV.read(csv_file_path, headers: true).map(&:to_h)
-      rows.each(&:deep_symbolize_keys!)
+      @csv_rows ||= load_csv
     end
 
-    def csv_file_path
-      @csv_file_path ||= Rails.root.join("#{STATISTICS_DATA_PATH}/#{data['csv_file']}")
+    def load_csv
+      rows = if attachment
+               CSV.new(URI.parse(attachment.url).open, headers: true).map(&:to_h)
+             else
+               # SCAFFOLDING
+               csv_file_path = Rails.root.join("#{STATISTICS_DATA_PATH}/#{data['csv_file']}")
+               CSV.read(csv_file_path, headers: true).map(&:to_h)
+             end
+      rows.each(&:deep_symbolize_keys!)
     end
   end
 end
--- a/spec/requests/landing_page_spec.rb
+++ b/spec/requests/landing_page_spec.rb
@@ -5,7 +5,21 @@ RSpec.describe "Landing Page" do
         stub_const("LandingPage::ADDITIONAL_CONTENT_PATH", "spec/fixtures")
         @content_item = GovukSchemas::Example.find("landing_page", example_name: "landing_page")
         @base_path = @content_item.fetch("base_path")
+        @content_item["details"]["attachments"] = [
+          {
+            "accessible" => false,
+            "attachment_type" => "document",
+            "content_type" => "text/csv",
+            "file_size" => 123,
+            "filename" => "data_one.csv",
+            "id" => 12_345,
+            "preview_url" => "https://www.asset.test.gov.uk/data_one.csv/preview",
+            "title" => "Data One",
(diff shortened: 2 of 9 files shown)


app/models/landing_page/block/statistics.rb
```diff
@@ -30,15 +29,15 @@ def rows
       rows
     end
 
+    def attachment
+      @attachment ||= landing_page.attachments.find { |att| att.title == data["attachment_csv"] }
```
leenagupte, line 33: Using title feels a bit brittle. Should this account for typos? i.e errors in case or extra whitespace, so we allow "Data One" to be the same as "Data one" and "Data  One" etc.

I do think title is better than using an id as I'm not sure how easy it would be to find the id in the publishing tool to configure the block correctly.

app/models/landing_page/block/statistics.rb
```diff
@@ -30,15 +29,15 @@ def rows
...
   private
 
     def csv_rows
-      rows = CSV.read(csv_file_path, headers: true).map(&:to_h)
-      rows.each(&:deep_symbolize_keys!)
-    end
-
-    def csv_file_path
-      @csv_file_path ||= Rails.root.join("#{STATISTICS_DATA_PATH}/#{data['csv_file']}")
+      rows = CSV.new(URI.parse(attachment.url).open, headers: true).map(&:to_h)
```
leenagupte, line 39: Could we also allow the CSVs to be read in from an csv file in whilst we're relying on local demos for testing?

I'm concerned that until one of these pages go live, it will be super difficult to make frontend changes to the statistics block.

spec/requests/landing_page_spec.rb
```diff
@@ -20,26 +34,5 @@
         expect(response).to render_template(:show)
       end
     end
-
-    context "when a content item does not exist" do
```
leenagupte, line 24: I think this should be left in for now, until the scaffold is completely removed.

leenagupte (commented):
On the whole I think this looks good.

I've added a few inline questions. One about the attachments lookup. The others concern testing locally.

app/models/landing_page/block/statistics.rb
KludgeKML: Yeah, title is the one thing you have control of in the Whitehall UI, so it's possible to match up. ID is harder to find. We could use filename, that might be preferable?

KludgeKML: Changed to filename

app/models/landing_page/block/statistics.rb
KludgeKML: I think there's a longer conversation to be had about this, for the moment I've reinstated scaffolding code.

spec/requests/landing_page_spec.rb
KludgeKML: As above.

app/models/landing_page/block/statistics.rb
```diff
@@ -30,15 +32,21 @@ def rows
...
-    def csv_file_path
-      @csv_file_path ||= Rails.root.join("#{STATISTICS_DATA_PATH}/#{data['csv_file']}")
+      rows = if attachment
+               CSV.new(URI.parse(attachment.url).open, headers: true).map(&:to_h)
+             else
+               # SCAFFOLDING
+               csv_file_path = Rails.root.join("#{STATISTICS_DATA_PATH}/#{data['csv_file']}")
+               CSV.read(csv_file_path, headers: true).map(&:to_h)
+             end
+      @csv_rows = rows.each(&:deep_symbolize_keys!)
```
leenagupte, line 49: Minor, but I don't think you need to assign a `@csv_rows` instance variable here.

app/models/landing_page/block/statistics.rb
```diff
@@ -30,15 +32,21 @@ def rows
       rows
     end
 
+    def attachment
+      @attachment ||= landing_page.attachments.find { |att| att.filename == data["csv_file"] }
```
leenagupte, line 36: Using the filename is more robust than the title. 👍

leenagupte (approved):
One minor comment that's not a blocker. Everything else looks good. 👍

app/models/landing_page/block/statistics.rb
KludgeKML: I've adjusted this code slightly to memoize the calls to load/fetch the csv.

app/models/landing_page/block/statistics.rb
```diff
@@ -30,15 +32,25 @@ def rows
...
     def csv_rows
-      rows = CSV.read(csv_file_path, headers: true).map(&:to_h)
-      rows.each(&:deep_symbolize_keys!)
+      @csv_rows ||= load_csv
     end
 
-    def csv_file_path
-      @csv_file_path ||= Rails.root.join("#{STATISTICS_DATA_PATH}/#{data['csv_file']}")
+    def load_csv
+      rows = if attachment
```
leenagupte, line 46: I don't think you need two methods. I think you can use `begin` and `end`
e.g.

```ruby
@csv_rows ||= begin
  rows = if attachment
  .
  .
  .
  rows.each(&:deep_symbolize_keys!)
end

PR #4354, opened 2024-10-29
Navigation from details menus

KludgeKML:
Update side and main navigation so that they take their structure from the `menus:` key

https://trello.com/c/2xR6z6o4/86-consider-better-way-of-managing-sidebar-links-than-hardcoding-them


app/models/landing_page.rb
```diff
@@ -13,6 +13,7 @@ def initialize(content_store_response)
       )
     end
 
+    @navigation_groups = (content_store_hash.dig("details", "navigation_groups") || []).map { [_1["id"], _1] }.to_h
```
leenagupte, line 16: Why does `to_h` need to be called here?

spec/factories/landing_pages.rb
```diff
@@ -1,11 +1,43 @@
 FactoryBot.define do
-  factory :landing_page, parent: :content_item do
+  factory :landing_page, class: LandingPage, parent: :content_item do
```
leenagupte, line 2: Why does the class need to be specified now?

spec/models/landing_page_spec.rb
```diff
@@ -42,4 +45,11 @@
       expect(described_class.new(content_item).blocks.first.type).to eq(expected_type)
     end
   end
+
+  describe "#menus" do
```
leenagupte, line 49: This test needs to be updated to use `navigation_groups`, which I think is a much better name than "menus". 🎉

spec/models/landing_page_spec.rb
```diff
@@ -42,4 +45,11 @@
...
+    it "returns a map of menus" do
+      expect(described_class.new(content_item).navigation_groups["Top Menu"]["id"]).to eq("Top Menu")
+      expect(described_class.new(content_item).navigation_groups["Top Menu"]["links"].count).to eq(2)
```
leenagupte, line 52: Shouldn't this be 4?

app/models/landing_page/block/main_navigation.rb
```diff
@@ -5,9 +5,10 @@ class MainNavigation < Base
     def initialize(block_hash, landing_page)
       super
 
-      @title = data.fetch("title")
-      @title_link = data.fetch("title_link")
-      @links = data.fetch("links")
+      navigation_group = landing_page.navigation_groups[data["navigation_group_id"]]
+      @title = navigation_group["links"].select { |l| l["theme"] == "homepage" }.first["text"]
+      @title_link = navigation_group["links"].select { |l| l["theme"] == "homepage" }.first["href"]
+      @links = navigation_group["links"].reject { |l| l["theme"] == "homepage" }
```
leenagupte, line 11: Could these be methods? It would make them easier to read.
It would also be good to add tests for them.

Could also add something to the commit message about what the "homepage" theme is, (have we seen it before?) and why it's being excluded.

lib/data/landing_page_content_items/links/side_navigation.yaml
leenagupte, line 1: I think a `navigation_group` block should be added to the at least one of the hard-coded demos so that it's easy to do frontend testing. At least until a landing page goes live.

spec/models/landing_page/block/side_navigation_spec.rb
```diff
@@ -1,24 +1,17 @@
...
+    {
+      "type" => "side_navigation",
+      "navigation_group_id" => "Submenu",
+    }
   end
+  let(:subject) { described_class.new(blocks_hash, build(:landing_page_with_navigation_groups)) }
 
   describe "#links" do
-    it "returns all of the side navigation links" do
+    it "returns all of the navigation links from the requested menu" do
```
leenagupte, line 11: I think "from the requested menu" should be removed from the test name. or changed to "from the requested group".

leenagupte (commented):
On the while this is looking good. 🎉 

I have a few inline comments. The main thing is explaining the "why" for this change in the commit messages. Adding another top-level block is quite a large change, so it would be good to record for posterity (and future "us") why this approach was decided upon.

I think we need to also consider the implications for frontend development before all of the scaffolding is completely removed.

spec/factories/landing_pages.rb
KludgeKML: (It should have been all along, we just got lucky that the previous tests we used it in didn't mind that they were actually getting a ContentItem)

spec/models/landing_page_spec.rb
KludgeKML: This is just the direct links (of which there are two).

leenagupte: Ah, perhaps the test should name should be updated to reflect that.
Is there ever a case when all 4 are returned?

lib/data/landing_page_content_items/links/side_navigation.yaml
KludgeKML: It's in all the hardcoded demo pages now (in the 4th commit)

spec/models/landing_page_spec.rb
KludgeKML: I... don't understand what's going on here. On my local machine, with these commits, this method is called `#navigation_groups`, and the comment below also says that.

spec/models/landing_page/block/side_navigation_spec.rb
KludgeKML: Done.

app/models/landing_page/block/main_navigation.rb
KludgeKML: I've removed the concept of theme and the idea it was supposed to support (that the first link in the main navigation was somehow special and needed to be extracted at this point rather than in the partial)

spec/models/landing_page_spec.rb
KludgeKML: Ah, sneaky fix commit got merged in later on where I couldn't spot it easily. Fixed.

spec/fixtures/landing_page.yaml
```diff
@@ -1,12 +1,25 @@
+navigation_groups:
+- id: "Top Menu"
+  links:
+  - href: "/homepage"
+    text: "Homepage"
+    theme: "homepage"
```
leenagupte, line 6: Is theme still being used? I think a comment said that it wasn't? If not, then this should be removed.

spec/factories/landing_pages.rb
```diff
@@ -20,7 +20,7 @@
             {
               id: "Top Menu",
               links: [
-                { href: "https:/www.gov.uk", text: "Service Name", theme: "homepage" },
+                { href: "https://www.gov.uk", text: "Service Name" },
```
leenagupte, line 23: Can you fix this up to where it was added.

lib/data/landing_page_content_items/be_kinder.yaml
```diff
@@ -1,11 +1,13 @@
-blocks:
-- type: main_navigation
+navigation_groups:
```
leenagupte, line 1: Thanks for updating all of these 🎉

spec/factories/landing_pages.rb
KludgeKML: Fixed

spec/fixtures/landing_page.yaml
KludgeKML: Gone now!

leenagupte (approved):

PR #4362, opened 2024-10-31
Use OpenStruct rather than Data

KludgeKML:
- Some content items rendered by frontend (eg. licences) turn out to have non-standard structures for attachments, so fail if you try to make them as a data item (which requires the inputs to match the structure exactly).

Related:
https://govuk.sentry.io/issues/6030737250/events/f86bf91ae0ee41e28002c43d7e1b5de4/

Example failing page:
https://www.gov.uk/find-licences/the-beef-carcase-classification-scheme-work-as-a-classifier

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -1,3 +1,5 @@
+require "ostruct"
+
 class ContentItem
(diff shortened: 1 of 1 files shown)


unoduetre (approved):
Approved as a part of a temporary fix of an ongoing 2nd line issue.

PR #4366, opened 2024-11-01
Local content item support

KludgeKML:

## What

Adds more structured support for loading content items from the local machine. By setting an environment variable and putting a content item in a particular directory, it will be loaded from that file instead of from a call to the content store.

## Why

There's currently a restricted version of this code working for Landing Pages (it only loads the blocks from the fixture, but includes fake values for the rest of the content item). This scaffolding has proved useful during Landing Pages development, but it's limited in scope and only works for Landing Pages. By replacing it with a more robust option, we can improve developer experience in the absence of a local version of content store, and make it easier to demo experimental content items on Heroku.

https://trello.com/c/FytCkByy/383-improve-contentitem-offline-loading, [Jira issue PNP-8577](https://gov-uk.atlassian.net/browse/PNP-8577)

## How

We add a new singleton class (ContentItemLoader) that is responsible for calls to GdsApi.content_store.get_content - this allows us to have a cache of items (simplifying the current Format Constraint code, which caches responses in the request env field), and to have a single point where those calls can be intercepted and swapped out for local code if the required environment variable is set. Setting ALLOW_LOCAL_CONTENT_ITEM_OVERRIDE to true will cause the loader to look in /config/local-content-items/path/to/the/item for any call to the content store (falling back to an actual call if the file isn't present).

## See Also:

* https://github.com/alphagov/govuk-docker/pull/793. (adds the env var to govuk-docker frontend apps)

### DIFF
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -5,9 +5,16 @@ private
 
   # SCAFFOLDING: can be removed when basic content items are available
   # from content-store
-  def request_content_item(_base_path)
-    GdsApi.content_store.content_item(request.path).to_h
-  rescue StandardError
+  def content_item
+    @content_item ||= ContentItemFactory.build(old_scaffolding_content_item)
+  end
+
+  # SCAFFOLDING: can be removed when basic content items are available
+  # from content-store
+  def old_scaffolding_content_item
+    result = ContentItemLoader.load(request.path)
+    return result.to_h if result.is_a?(GdsApi::Response)
+
     fake_data
   end
 
--- a/docs/local-content-items.md
+++ b/docs/local-content-items.md
@@ -1,0 +1,11 @@
+# Local Content Items
+
+For local development and preview apps, it is sometimes desirable to have the ability to load content items that are specified locally rather than pointing to production/integration or having the content store running locally.
+
+To support this, set:
+
+`ALLOW_LOCAL_CONTENT_ITEM_OVERRIDE=true`
+
+...in your environment.
+
+With that environment variable set, the ContentItemLoader will add an additional check when loading a content item from the content store. Before making the GdsApi content store call, it will look in config/local-content-items/ for a JSON or YAML file matching the path of the content item (for instance, if you're looking for /find-licences/my-licence, it will look for config/local-content-items/find-licences/my-licence.json, then config/local-content-items/find-licences/my-licence.yml)
--- a/lib/content_item_loader.rb
+++ b/lib/content_item_loader.rb
@@ -1,0 +1,53 @@
+require "ostruct"
+
+class ContentItemLoader
+  LOCAL_ITEMS_PATH = "lib/data/local-content-items".freeze
+
+  @cache = {}
+
+  class << self
+    attr_reader :cache
+
+    def load(base_path)
+      cache[base_path] ||= if use_local_file? && File.exist?(yaml_filename(base_path))
+                             Rails.logger.debug("Loading content item #{base_path} from #{yaml_filename(base_path)}")
+                             load_yaml_file(base_path)
+                           elsif use_local_file? && File.exist?(json_filename(base_path))
+                             Rails.logger.debug("Loading content item #{base_path} from #{json_filename(base_path)}")
+                             load_json_file(base_path)
+                           else
+                             begin
+                               GdsApi.content_store.content_item(base_path)
+                             rescue GdsApi::HTTPErrorResponse, GdsApi::InvalidUrl => e
+                               e
+                             end
+                           end
+    end
+
+  private
+
+    def use_local_file?
+      ENV["ALLOW_LOCAL_CONTENT_ITEM_OVERRIDE"] == "true"
+    end
+
+    def yaml_filename(base_path)
+      Rails.root.join("#{LOCAL_ITEMS_PATH}#{base_path}.yml")
+    end
+
+    def load_yaml_file(base_path)
+      GdsApi::Response.new(OpenStruct.new(code: 200, body: YAML.load(File.read(yaml_filename(base_path))).to_json, headers:))
+    end
+
+    def json_filename(base_path)
+      Rails.root.join("#{LOCAL_ITEMS_PATH}#{base_path}.json")
+    end
+
+    def load_json_file(base_path)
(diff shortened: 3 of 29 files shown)


docs/local-content-items.md
```diff
@@ -0,0 +1,11 @@
+# Local Content Items
+
+For local development and preview apps, it is sometimes desirable to have the ability to load content items that are specified locally rather than pointing to production/integration or having the content store running locally.
+
+To support this, set:
+
+`ALLOW_LOCAL_CONTENT_ITEM_OVERRIDE=true`
```
leenagupte, line 7: It'd be good if there was an accompanying change in govuk-docker that sets this.

app/controllers/landing_page_controller.rb
```diff
@@ -5,8 +5,14 @@ class LandingPageController < ContentItemsController
...
-  def request_content_item(_base_path)
-    GdsApi.content_store.content_item(request.path).to_h
+  def content_item
+    @content_item ||= ContentItemFactory.build(old_scaffolding_content_item)
+  end
+
+  # SCAFFOLDING: can be removed when basic content items are available
+  # from content-store
+  def old_scaffolding_content_item
+    ContentItemLoaderGdsApi.content_store.content_item(request.path).to_h
```
leenagupte, line 15: This doesn't look right. There isn't actually a class called `ContentItemLoaderGdsApi` is there?

The tests are passing, so either this is fine somehow, or this bit of code is not being tested.

leenagupte (commented):
I like the idea in principle. I have a few inline comments that I think need to be fixed before this will work.

docs/local-content-items.md
KludgeKML: https://github.com/alphagov/govuk-docker/pull/793

app/controllers/landing_page_controller.rb
KludgeKML: Yeah, not sure what was going on here. I've fixed it now!

lib/content_item_loader.rb
```diff
@@ -0,0 +1,47 @@
...
+      File.exist?(local_json_filename(base_path)) || File.exist?(local_yaml_filename(base_path))
+    end
+
+    def local_json_filename(base_path)
+      Rails.root.join("#{LOCAL_ITEMS_PATH}#{base_path}.json")
+    end
+
+    def local_yaml_filename(base_path)
+      Rails.root.join("#{LOCAL_ITEMS_PATH}#{base_path}.yml")
+    end
```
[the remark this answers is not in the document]
KludgeKML: We actually don't really need to do that - dashes are okay in file names, and it allows us to be simpler in matching the path. You can see an example here: https://github.com/alphagov/frontend/pull/4387/files

lib/content_item_loader.rb
```diff
@@ -0,0 +1,47 @@
...
+                             begin
+                               GdsApi.content_store.content_item(base_path)
+                             rescue GdsApi::HTTPErrorResponse, GdsApi::InvalidUrl => e
+                               e
+                             end
+                           end
+    end
+
+    def local_file?(base_path)
+      File.exist?(local_json_filename(base_path)) || File.exist?(local_yaml_filename(base_path))
```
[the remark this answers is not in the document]
leenagupte: All of the methods apart from `load` should be made private as they're not being used or tested elsewhere.

app/controllers/landing_page_controller.rb
```diff
@@ -5,9 +5,16 @@ class LandingPageController < ContentItemsController
...
-    GdsApi.content_store.content_item(request.path).to_h
-  rescue StandardError
+  def content_item
+    @content_item ||= ContentItemFactory.build(old_scaffolding_content_item)
+  end
+
+  # SCAFFOLDING: can be removed when basic content items are available
+  # from content-store
+  def old_scaffolding_content_item
+    result = ContentItemLoader.load(request.path)
```
leenagupte, line 15: It's great that you've fixed this. I think it would be good to add a test for this scaffolding too.

leenagupte (commented):
I had a bit more time to review, so had a closer look at the ContentItemLoader class and added a few inline suggestions for refactoring it.

There's also a request for some tests around the scaffolding if this is going to become a permanent feature.

There's nothing major though and the overall approach is great 🎉

docs/local-content-items.md
KludgeKML: Done (and added it to app.json)

app/controllers/landing_page_controller.rb
leenagupte: They can't have been thorough enough because nothing caught the attempted call to `ContentItemLoaderGdsApi.content_store.content_item(request.path).to_h` which doesn't exist!

spec/unit/content_item_loader_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      after do
+        ENV["ALLOW_LOCAL_CONTENT_ITEM_OVERRIDE"] = nil
+      end
+
+      context "with a local JSON file" do
+        let!(:item_request) { stub_content_store_has_item("/my-json-item") }
+
+        it "loads from that" do
```
leenagupte, line 37: Not a blocker, but this isn't a great test name! 

It should be probably be something like `it "loads content from the json file not content store". Or something.

spec/unit/content_item_loader_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+          expect(item_request).not_to have_been_made
+          expect(ContentItemFactory.build(response).schema_name).to eq("json_page")
+        end
+      end
+
+      context "with a local YAML file" do
+        let!(:item_request) { stub_content_store_has_item("/my-yaml-item") }
+
+        it "loads from that file" do
```
leenagupte, line 48: Ditto here about the test name, but again, not a blocker.

leenagupte (approved):
There are a couple of non-blocking comments about the test names, otherwise it looks good. 🎉

spec/unit/content_item_loader_spec.rb
KludgeKML: Ta! Fixed.

spec/unit/content_item_loader_spec.rb
KludgeKML: Ta! Fixed.

PR #4372, opened 2024-11-04
Move case studies routes 

georges1996:

This PR https://github.com/alphagov/frontend/pull/4379 needs to be merged in and this will need to be rebased to fix the case studies which have translations e.g https://www.gov.uk/government/case-studies/doing-business-in-spain
## What
Handle Case Studies documents in Frontend, in preparation for them being taken out of Government Frontend

## Why

Ticket: https://trello.com/c/ybpXwp34/368-move-document-type-casestudy-from-government-frontend-to-frontend, [Jira issue PNP-9253](https://gov-uk.atlassian.net/browse/PNP-9253)

## How

By moving the routes and necessary code to get the pages rendering in the frontend application 

## Screenshots?
# Frontend 
Screenshot 2024-11-06 at 08 22 40
Screenshot 2024-11-06 at 08 23 02

# Gov-Frontend
Screenshot 2024-11-06 at 08 22 34
Screenshot 2024-11-06 at 08 22 54

Example case studies check: 
- https://govuk-frontend-app-pr-4372.herokuapp.com/government/case-studies/doing-business-in-spain
- https://govuk-frontend-app-pr-4372.herokuapp.com/government/case-studies/terence-age-33-stoke-on-trent
- https://govuk-frontend-app-pr-4372.herokuapp.comgovernment/case-studies/vdec-is-supporting-a-gbs-vaccine-to-prevent-newborn-deaths
- https://govuk-frontend-app-pr-4372.herokuapp.com/government/case-studies/2013-elections-in-swaziland

### DIFF
--- a/app/helpers/link_helper.rb
+++ b/app/helpers/link_helper.rb
@@ -13,6 +13,10 @@ module LinkHelper
     classes = "govuk-link"
     classes << " govuk-link--inverse" if inverse
 
-    "<a href='#{path}' class='#{classes}'>#{text}</a>".html_safe
+    "<a href='#{path}' class='#{classes}'>".html_safe + text + "</a>".html_safe
+  end
+
+  def govuk_styled_links_list(links)
+    links.map { |link| govuk_styled_link(link["title"], path: link["base_path"]) }
   end
 end
--- a/app/models/concerns/updatable.rb
+++ b/app/models/concerns/updatable.rb
@@ -1,0 +1,44 @@
+module Updatable
+  extend ActiveSupport::Concern
+
+  included do
+    def updated
+      public_updated_at if any_updates?
+    end
+
+    def history
+      return [] unless any_updates?
+
+      reverse_chronological_change_history
+    end
+  end
+
+private
+
+  def any_updates?
+    if public_updated_at && initial_publication_date
+      Time.zone.parse(public_updated_at) != Time.zone.parse(initial_publication_date)
+    else
+      false
+    end
+  end
+
+  def initial_publication_date
+    first_public_at || first_published_at
+  end
+
+  def reverse_chronological_change_history
+    change_history.sort_by { |item| Time.zone.parse(item[:timestamp]) }.reverse
+  end
+
+  def change_history
+    changes = content_store_hash.dig("details", "change_history") || []
+    changes.map do |item|
+      {
+        display_time: item["public_timestamp"],
+        note: item["note"],
+        timestamp: item["public_timestamp"],
+      }
+    end
+  end
+end
--- a/app/models/concerns/withdrawable.rb
+++ b/app/models/concerns/withdrawable.rb
@@ -5,6 +5,14 @@ module Withdrawable
     def withdrawn?
       withdrawal_notice.present?
     end
+
+    def withdrawn_at
+      withdrawal_notice["withdrawn_at"]
+    end
+
+    def withdrawn_explanation
+      withdrawal_notice["explanation"]
+    end
   end
 
 private
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -2,10 +2,12 @@ require "ostruct"
 
 class ContentItem
   include Withdrawable
+  include Organisations
 
   attr_reader :attachments, :base_path, :body, :content_store_hash,
-              :content_store_response, :description, :document_type, :image,
-              :links, :locale, :phase, :public_updated_at, :schema_name, :title
+              :content_store_response, :description, :document_type, :first_public_at,
+              :first_published_at, :image, :links, :locale, :phase, :public_updated_at,
+              :schema_name, :title
 
   # SCAFFOLDING: remove the override_content_store_hash parameter when full landing page
   # content items including block details are available from content-store
@@ -24,6 +26,8 @@ class ContentItem
     @public_updated_at = content_store_hash["public_updated_at"]
     @links = content_store_hash["links"]
     @phase = content_store_hash["phase"]
+    @first_published_at = content_store_hash["first_published_at"]
+    @first_public_at = content_store_hash.dig("details", "first_public_at")
 
     @attachments = get_attachments(content_store_hash.dig("details", "attachments"))
 
@@ -57,6 +61,10 @@ class ContentItem
     @meta_section ||= content_store_hash.dig(
       "links", "parent", 0, "links", "parent", 0, "title"
     )&.downcase
+  end
+
+  def contributors
+    organisations
   end
 
 private
--- a/app/views/case_study/show.html.erb
+++ b/app/views/case_study/show.html.erb
@@ -1,0 +1,72 @@
+<% content_for :title, "#{@case_study_presenter.page_title} - #{I18n.t("formats.#{content_item.document_type}", count: 1)} - GOV.UK" %>
+
+<% content_for :extra_headers do %>
+  <meta name="description" content="<%= @content_item.description %>">
+
+  <%= render "govuk_publishing_components/components/machine_readable_metadata", { content_item: content_item.to_h, schema: :article } %>
+<% end %>
+
+<main role="main" id="content" class="case-study" <%= lang_attribute(content_item.locale) %>>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds responsive-top-margin">
+      <%= render "govuk_publishing_components/components/title", {
+        title: content_item.title,
+        context: I18n.t("formats.#{content_item.document_type}", count: 1),
+        context_locale: t_locale_fallback("formats.#{content_item.document_type}", count: 1),
+        average_title_length: "long",
+      } %>
+    </div>
+    <%= render "shared/translations" %>
+    <div class="govuk-grid-column-two-thirds">
+      <%= render "govuk_publishing_components/components/lead_paragraph", text: content_item.description %>
+    </div>
+  </div>
+
+  <%= render "shared/publisher_metadata", locals: {
+    from: govuk_styled_links_list(content_item.contributors),
+    first_published: display_date(content_item.first_public_at || content_item.first_published_at),
+    last_updated: display_date(content_item.updated),
+    see_updates_link: true,
+    } %>
+
+  <% if content_item.withdrawn? %>
+    <% withdrawn_time_tag = tag.time(display_date(content_item.withdrawn_at), datetime: content_item.withdrawn_at) %>
+
+    <%= render "govuk_publishing_components/components/notice", {
+      title: I18n.t("withdrawn_notice.title", schema_name: I18n.t("formats.#{content_item.schema_name}", count: 1, locale: :en).downcase, withdrawn_time: withdrawn_time_tag).html_safe,
+      description_govspeak: content_item.withdrawn_explanation&.html_safe,
+      time: withdrawn_time_tag,
+      lang: "en",
+    } %>
+  <% end %>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="content-bottom-margin">
+        <div class="responsive-bottom-margin">
+          <%= render "components/figure",
+            src: content_item.image["url"],
+            alt: content_item.image["alt_text"],
+            credit: content_item.image["credit"],
+            caption: content_item.image["caption"] if content_item.image %>
+
+          <%= render "govuk_publishing_components/components/govspeak", {
+            direction: page_text_direction,
+          } do %>
+            <%= raw(content_item.body) %>
+          <% end %>
+        </div>
+
+        <%= render "components/published_dates", {
+          published: display_date(content_item.first_public_at),
+          last_updated: display_date(content_item.updated),
+          history: content_item.history,
+        } %>
+      </div>
+    </div>
+
+    <%= render "shared/sidebar_navigation" %>
+  </div>
+</main>
+
+<%= render "shared/footer_navigation" %>
--- a/app/views/components/_published_dates.html.erb
+++ b/app/views/components/_published_dates.html.erb
@@ -37,7 +37,7 @@
           <ol class="app-c-published-dates__list">
             <% history.each do |change| %>
               <li class="app-c-published-dates__change-item">
-                <time class="app-c-published-dates__change-date timestamp" datetime="<%= change[:timestamp] %>"><%= change[:display_time] %></time>
+                <time class="app-c-published-dates__change-date timestamp" datetime="<%= change[:timestamp] %>"><%= display_date(change[:display_time]) %></time>
                 <p class="app-c-published-dates__change-note"><%= change[:note].strip %></p>
               </li>
             <% end %>
--- a/app/views/shared/_publisher_metadata.html.erb
+++ b/app/views/shared/_publisher_metadata.html.erb
@@ -1,0 +1,16 @@
+<%
+  add_view_stylesheet("publisher_metadata")
+%>
+
+<div class="govuk-grid-row">
+  <div class="metadata-wrapper">
+    <div class="govuk-grid-column-two-thirds metadata-column">
+      <%= render "govuk_publishing_components/components/metadata", {
+        from: locals[:from],
+        first_published: locals[:first_published],
+        last_updated: locals[:last_updated],
+        see_updates_link: locals[:see_updates_link],
+      } %>
+    </div>
+  </div>
+</div>
--- a/spec/helpers/locale_helper_spec.rb
+++ b/spec/helpers/locale_helper_spec.rb
@@ -43,4 +43,11 @@ RSpec.describe LocaleHelper do
       end
     end
   end
+
+  describe "#native_language_name_for" do
+    it "returns the native language name for the given locale" do
+      expect(native_language_name_for(:en)).to eq("English")
+      expect(native_language_name_for(:cy)).to eq("Cymraeg")
+    end
+  end
 end
--- a/spec/models/case_study_spec.rb
+++ b/spec/models/case_study_spec.rb
@@ -1,0 +1,27 @@
+RSpec.describe CaseStudy do
+  let(:content_store_response) do
+    GovukSchemas::Example.find("case_study", example_name: "doing-business-in-spain")
+  end
+
+  it_behaves_like "it has updates", "case_study", "has-updates"
+  it_behaves_like "it has no updates", "case_study", "doing-business-in-spain"
+  it_behaves_like "it can have worldwide organisations", "case_study", "doing-business-in-spain"
+  it_behaves_like "it can have emphasised organisations", "case_study", "doing-business-in-spain"
+
+  describe "#contributors" do
+    it "returns the organisations ordered by emphasis followed by worldwide organisations" do
+      content_item = described_class.new(content_store_response)
+
+      worldwide_organisations = content_store_response.dig("links", "worldwide_organisations")
+      organisations = content_store_response.dig("links", "organisations")
+
+      expected_contributors = [
+        { "title" => organisations[1]["title"], "base_path" => organisations[1]["base_path"], "content_id" => organisations[1]["content_id"] },
+        { "title" => organisations[0]["title"], "base_path" => organisations[0]["base_path"], "content_id" => organisations[0]["content_id"] },
+        { "title" => worldwide_organisations[0]["title"], "base_path" => worldwide_organisations[0]["base_path"], "content_id" => worldwide_organisations[0]["content_id"] },
+      ]
+
+      expect(content_item.contributors).to eq(expected_contributors)
+    end
+  end
+end
--- a/spec/support/concerns/organisations.rb
+++ b/spec/support/concerns/organisations.rb
@@ -1,0 +1,19 @@
+RSpec.shared_examples "it can have emphasised organisations" do |schema|
+  let(:content_store_response) { GovukSchemas::Example.find(schema, example_name: schema) }
+
+  it "knows it has emphasised organisations" do
+    expect(described_class.new(content_store_response).contributors.first["content_id"]).to eq(content_store_response["details"]["emphasised_organisations"].first)
+  end
+end
+
+RSpec.shared_examples "it can have organisations" do |schema|
+  let(:content_store_response) { GovukSchemas::Example.find(schema, example_name: schema) }
+
+  it "knows it has an organisations base_path, title and content_id" do
+    expect(described_class.new(content_store_response).organisations).to eq(
+      content_store_response["links"]["organisations"].map do |org|
+        { "base_path" => org["base_path"], "title" => org["title"], "content_id" => org["content_id"] }
+      end,
+    )
+  end
+end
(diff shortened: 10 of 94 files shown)


spec/models/concerns/linkable_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def initialize
+    @content_item = {
+      "base_path" => "/a/base/path",
+      "links" => {},
+    }
+  end
+end
+
+RSpec.describe Linkable, type: :model do
```
KludgeKML, line 17: You shouldn't need `type: model` here - it's inferred from the file location.

app/models/concerns/withdrawable.rb
```diff
@@ -0,0 +1,45 @@
...
+  def withdrawal_notice_time
+    ActionController::Base.helpers.tag.time(
+      english_display_date(withdrawal_notice["withdrawn_at"]),
+      datetime: withdrawal_notice["withdrawn_at"],
+    )
+  end
+
+  def english_display_date(timestamp, format = "%-d %B %Y")
+    I18n.l(Time.zone.parse(timestamp), format:, locale: :en) if timestamp
+  end
```
KludgeKML, line 44: A bit suspicious of these presentational items being in a model concern, might be better as helpers?

app/helpers/locale_helper.rb
```diff
@@ -7,6 +7,16 @@ def page_text_direction
...
+  def t_locale_fallback(key, options = {})
+    options[:locale] = I18n.locale
+    options[:fallback] = nil
+    translation = I18n.t(key, **options)
+
+    if translation.nil? || translation.downcase.include?("translation missing")
+      I18n.default_locale
+    end
+  end
+
```
KludgeKML, line 19: This is just... moving the method? Is it necessary?

app/models/content_item.rb
```diff
@@ -1,9 +1,11 @@
 require "ostruct"
 
 class ContentItem
+  include Withdrawable
```
KludgeKML, line 4: Is Withdrawable definitely something that all ContentItems (or, let's say, the overwhelming majority of them) are going to use?

app/controllers/case_studies_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+  def show
+    @case_study = CaseStudiesPresenter.new(@content_item, view_context)
+  end
+
+private
+
+  def content_item_slug
+    request.path
+  end
+end
```
KludgeKML, line 11: I think we can remove this now (all ContentItems use request.path)

app/presenters/case_studies_presenter.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def title_and_context
+    {
+      title: content_item.title,
+      context: I18n.t("formats.#{content_item.document_type}", count: 1),
+      context_locale: view_context.t_locale_fallback("formats.#{content_item.document_type}", count: 1),
+      average_title_length: "long",
+    }
+  end
+end
```
KludgeKML, line 18: It might be worth just putting this straight into the view (take a look at the take_part controller), which would allow you to cut out the entire presenter.

app/views/case_studies/show.html.erb
```diff
@@ -0,0 +1,52 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds responsive-top-margin">
+    <%= render 'govuk_publishing_components/components/title', @case_study.title_and_context %>
+  </div>
+  <%= render 'shared/translations' %>
+  <div class="govuk-grid-column-two-thirds">
+    <%= render 'govuk_publishing_components/components/lead_paragraph', text: @content_item.description %>
+  </div>
+</div>
+
```
KludgeKML, line 16: You might need to look at the take part PR around adding a main class here (this might also affect help pages, FYI)

app/models/concerns/withdrawable.rb
leenagupte: Travel Advice and Guides need to know the withdrawn state too, but I've tackled it a bit differently by splitting out the model and presentational concerns:

model concern: https://github.com/alphagov/frontend/pull/4225/commits/0dfd81189601511ab3d468001b35098ffec32cad

generic presenter: https://github.com/alphagov/frontend/pull/4225/commits/0590ed72bf4543ff55feae6b1539f817fb54b3b5

travel advice presenter: https://github.com/alphagov/frontend/pull/4225/commits/53ee7c310524d77dc7c83fe86ed10d5d23b2cf8d

I'm not completely happy with the approach I've taken either, so I think it would be good to have a discussion about this.

spec/helpers/locale_helper_spec.rb
```diff
@@ -33,7 +33,7 @@
 
     it "returns default locale for a string with no locale translation" do
       I18n.with_locale(:cy) do
-        expect(t_locale_fallback("formats.big_decision", count: 1, locale: :cy)).to eq(:en)
+        expect(t_locale_fallback("content_item.schema_name.case_study.no_locale", count: 1, locale: :cy)).to eq(:en)
```
KludgeKML, line 36: Is this right? I think we want the case_study locales to go into `formats.case_study.<whatever>`

app/models/content_item.rb
```diff
@@ -20,6 +22,8 @@ def initialize(content_store_response, override_content_store_hash: nil)
     @base_path = content_store_hash["base_path"]
     @locale = content_store_hash["locale"]
     @public_updated_at = content_store_hash["public_updated_at"]
+    @first_published_at = content_store_hash["first_published_at"]
+    @first_public_at = content_store_hash.dig("details", "first_public_at") || first_published_at
```
KludgeKML, line 26: Do we know how many different schemas have a `first_public_at`? (is it only case studies?)

app/presenters/case_study_presenter.rb
KludgeKML, line 1: I don't know if we should make empty presenters (if it needs a presenter but it's just the standard ContentItemModelPresenter, let's use that in the controller).

app/models/concerns/metadata.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  included do
+    def publisher_metadata
+      metadata = {
+        from: linkable_organisations,
+        first_published: first_published_at,
+        last_updated: updated,
+      }
+
+      unless pending_stats_announcement?
```
KludgeKML, line 12: Can we add a test that covers this? (I think we're not testing when pending_stats_announcements exist)

app/models/concerns/updatable.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+  end
+
+private
+
+  def any_updates?
+    if public_updated_at && first_public_at
+      Time.zone.parse(public_updated_at) != Time.zone.parse(first_public_at)
+    else
+      false
```
KludgeKML, line 22: We're not testing this branch, could we add a test for it?

app/presenters/withdrawable_presenter.rb
```diff
@@ -0,0 +1,37 @@
...
+  attr_reader :content_item
+
+  def initialize(content_item)
+    @content_item = content_item
+  end
+
+  def withdrawal_notice_component
+    {
+      title: withdrawal_notice_title,
+      description_govspeak: content_item.withdrawal_notice["explanation"]&.html_safe,
```
KludgeKML, line 11: We're not testing if the explanation key doesn't exist.

app/presenters/withdrawable_presenter.rb
```diff
@@ -0,0 +1,37 @@
...
+      time: withdrawal_notice_time,
+      lang: I18n.locale.to_s == "en" ? false : "en",
```
KludgeKML, line 13: We're not testing the "en" branch (I think? we're definitely not testing one of the branches here)

leenagupte (changes requested):
I think this has made some good progress.

I have a few concerns about the split between presentational and model code, and think some of the model concerns need to be looked again. I think it would get good to talk about this in person.

app/views/case_studies/show.html.erb
```diff
@@ -0,0 +1,59 @@
...
+        average_title_length: "long",
+      } %>
+    </div>
+    <%= render 'shared/translations' %>
+    <div class="govuk-grid-column-two-thirds">
+      <%= render 'govuk_publishing_components/components/lead_paragraph', text: @content_item.description %>
+    </div>
+  </div>
+
+  <hr class="govuk-section-break govuk-section-break--s govuk-section-break--visible" aria-hidden="true">
```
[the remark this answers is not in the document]
deborahchua: The border is defined in https://github.com/alphagov/government-frontend/blob/main/app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss so I think once it's added in https://github.com/alphagov/frontend/pull/4372/commits/26a483dd6224a6d6edbda1925b80c8eeb07d16c1 this can be removed.

spec/support/concerns/updatable.rb
```diff
@@ -0,0 +1,51 @@
...
+        "public_timestamp" => "2024-05-23T00:00:00.000+01:00",
+      },
+    ]
+  end
+
+  it "knows it's updated" do
+    expect(described_class.new(@example_doc).updated).to eq "2024-07-16T12:36:44Z"
+  end
+
+  it "knows it's history" do
```
KludgeKML, line 22: type: it's -> its

app/assets/stylesheets/components/_publisher-metadata-with-logo.scss
```diff
@@ -0,0 +1,24 @@
+@import "govuk_publishing_components/individual_component_support";
```
KludgeKML, line 1: I think we might have to find somewhere different for this file - if something's a component, it has to meet a set of criteria (it has to have a specific set of component tests, docs, the class names for stylesheet matching have to have an app-c- prefix, etc). This wasn't a component in government-frontend, so perhaps we can find a better place for it in frontend. (maybe we can just have a stylesheets/views/shared folder?)

app/views/components/_published_dates.html.erb
```diff
@@ -34,13 +34,14 @@
...
             <li class="app-c-published-dates__change-item">
-              <time class="app-c-published-dates__change-date timestamp" datetime="<%= change[:timestamp] %>"><%= change[:display_time] %></time>
+              <time class="app-c-published-dates__change-date timestamp" datetime="<%= change[:timestamp] %>"><%= display_date(change[:display_time]) %></time>
               <p class="app-c-published-dates__change-note"><%= change[:note].strip %></p>
             </li>
           <% end %>
         </ol>
       </div>
     <% end %>
+    <hr class="govuk-section-break govuk-section-break--l govuk-section-break">
```
leenagupte, line 44: Why did this HR need to be added to the component?

leenagupte (changes requested):
I've added some inline comments.

My biggest area for concern is still the Metadata concern and how it's used. It would be good to talk about it in person and come up with a plan.

spec/support/concerns/organisations.rb
```diff
@@ -0,0 +1,55 @@
+RSpec.shared_examples "it can have organisations" do |schema|
+  before do
+    @example_doc = GovukSchemas::Example.find(schema, example_name: schema)
```
deborahchua, line 3: I think `@example_doc` should be renamed to `@content_store_response` for consistency with the other RSpec files.

spec/support/concerns/organisations.rb
```diff
@@ -0,0 +1,55 @@
...
+    @example_doc["details"]["emphasised_organisations"] = %w[11234500]
```
deborahchua, line 4: Do we want to hardcode an example here, when I think we want to test the data in the schema that's passed into it?

spec/support/concerns/organisations.rb
KludgeKML: FYI, there's a BIG PR about changing the existing spec files - @georges1996 it would probably be useful if you temporarily added the rspec rubocop config and just run it over the files you're touching/adding in this PR.

```# .rubocop.yml
inherit_gem:
  rubocop-govuk:
    ...
    - config/rspec.yml

File: app/models/concerns/updatable.rb [hunk 20]
```diff
@@ -0,0 +1,44 @@
...
+
+  def reverse_chronological_change_history
+    change_history.sort_by { |item| Time.zone.parse(item[:timestamp]) }.reverse
+  end
+
+  def change_history
+    changes = content_store_hash.dig("details", "change_history") || []
+    changes.map do |item|
+      {
+        display_time: item["public_timestamp"],
```
deborahchua, line 38: This is missing the `display_date` method from https://github.com/alphagov/government-frontend/blob/a938e4bee78b794e4e5303a1bde1dfd05e94b287/app/presenters/content_item/updatable.rb#L31, or is the technical approach going forward to use it in the view? e.g. we should be including it [here](https://github.com/alphagov/government-frontend/blob/af54031948650448b965ac274c3248d23658c4f6/app/views/components/_published_dates.html.erb#L37) instead

spec/support/concerns/updatable.rb
```diff
@@ -0,0 +1,52 @@
+RSpec.shared_examples "it can be updated" do |schema|
```
deborahchua, line 1: Could we update this test so that it accepts an example name that differs to the schema name as an optional argument?

app/views/case_study/show.html.erb
```diff
@@ -0,0 +1,71 @@
...
+      title: I18n.t("formats.case_study.withdrawn_notice.title", schema_name: I18n.t("formats.#{content_item.schema_name}", count: 1, locale: :en).downcase, withdrawn_time: withdrawn_time_tag).html_safe,
+      description_govspeak: content_item.withdrawn_explanation&.html_safe,
+      time: withdrawn_time_tag,
+      lang: "en",
+    }  %>
+  <% end %>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="content-bottom-margin">
```
deborahchua, line 44: I think we need to copy over the [content-bottom-margin helper stylesheet](https://github.com/alphagov/government-frontend/blob/main/app/assets/stylesheets/helpers/_content-bottom-margin.scss) and import in [application.scss](https://github.com/alphagov/government-frontend/blob/main/app/assets/stylesheets/application.scss#L17) which will add the missing spacing here (I just spotted this for call for evidence docs and also in your screenshots)

app/views/shared/_publisher_metadata.html.erb
```diff
@@ -0,0 +1,16 @@
+<%
+  add_stylesheet_path("helpers/_publisher-metadata-with-logo.css")
+%>
+
+<div class="govuk-grid-row govuk-!-margin-top-3">
```
deborahchua, line 5: Do we need the class `govuk-!-margin-top-3` here? It adds some additional spacing that the page doesn't have in government-frontend

leenagupte (commented):
This is making progress. 

I've added some inline comments. They're predominantly about the information contained in the commit messages, the hard-coded elements in the shared examples and some naming concerns.

app/models/content_item.rb
```diff
@@ -51,6 +54,10 @@ def available_translations
     translations.sort_by { |t| t["locale"] == I18n.default_locale.to_s ? "" : t["locale"] }
   end
 
+  def related_entities
```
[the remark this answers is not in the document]
KludgeKML: (I'm not 100% sure what the actual use is here, and the code in this comment is outdated, but I'd suggest we have some kind of prinicipal where by default we don't refer to the same thing multiple times. If this method is - as implied by the link above - taking an array of organisations, perhaps it should be called **organisation_whatever** - like `organisation_links_list`, say, if it's exclusively being used to pass into `govuk_styled_links_list`)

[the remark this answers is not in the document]
leenagupte: I think, and I hope that this is correct, that `related_entities` is the list of contributors / content creators rendered in the "from" section of the page.

I assume that "organisations" is the default list, and that's why it's been added to the the content item model. But other document types will need a slightly different list, for example a speech will need to add the speaker (person) as well as the organisation, and in that case the `related_entities` will be overridden in the child model.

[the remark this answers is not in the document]
leenagupte: I do think that the whole object doesn't need to be passed to the helper though, it should just be a list of titles and urls.

leenagupte (commented):
Thanks for making the updates requested in the previous review.

I think there's still some work left to be done on how the Updatable concern is tested. 
I've added inline comments with detail, as well as a few other small things that should be fixed.

app/helpers/link_helper.rb
```diff
@@ -13,6 +13,10 @@ def govuk_styled_link(text, path: nil, inverse: false)
     classes = "govuk-link"
     classes << " govuk-link--inverse" if inverse
 
-    "<a href='#{path}' class='#{classes}'>#{text}</a>".html_safe
+    "<a href='#{path}' class='#{classes}'>".html_safe + text + "</a>".html_safe
```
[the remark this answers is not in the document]
leenagupte: Was it in a separate PR?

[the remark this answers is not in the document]
leenagupte: I'm talking about a test for "link titles containing an &"

leenagupte (commented):
Thank for making the changes requested in the previous reviews.

There are only 3 inline comments this time, all of them related to tests. The most important one is that there are no tests for the `contributors` method in the case study the model.

app/views/case_study/show.html.erb
```diff
@@ -0,0 +1,70 @@
+<% content_for :title, "#{@case_study_presenter.page_title} - #{I18n.t("formats.#{content_item.document_type}", count: 1)} - GOV.UK" %>
+
+<% content_for :extra_headers do %>
+  <%= render "govuk_publishing_components/components/machine_readable_metadata", { content_item: content_item.to_h, schema: :article } %>
```
deborahchua, line 4: @sairamya93 and I were just checking the metadata and noticed the meta `description` tag needs to be included here, similar to https://github.com/alphagov/frontend/pull/4509/files#diff-ee902e589d40510716bed23cb80d0a765570d0867944a3a9ea98cffcf34fa1afR10

spec/models/case_study_spec.rb
```diff
@@ -0,0 +1,8 @@
+RSpec.describe CaseStudy do
+  it_behaves_like "it has updates", "case_study", "has-updates"
+  it_behaves_like "it has updates", "case_study", "only-has-first-published-at"
```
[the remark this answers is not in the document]
leenagupte: I see that you have updated the "only-has-first-published-at" example so that the `published_at` date is later than `first_published_at`, but doesn't have any updates, so the example is still incorrect.

If you update the example to have a change history, then it becomes the same test as for "has-updates". 

If you change `published_at` to equal `first_published_at`, then it becomes the same test as "doing-business-in-spain".

I think perhaps the "only-has-first-published-at" example and this line are not needed and should be removed.

spec/models/case_study_spec.rb
```diff
@@ -0,0 +1,29 @@
+RSpec.describe CaseStudy do
+  let(:content_item) do
+    GovukSchemas::Example.find("case_study", example_name: "doing-business-in-spain")
+  end
+
+  it_behaves_like "it has updates", "case_study", "has-updates"
+  it_behaves_like "it has updates", "case_study", "only-has-first-published-at"
+  it_behaves_like "it has no updates", "case_study", "doing-business-in-spain"
+  it_behaves_like "it has no updates", "case_study", "no-public-updated-at"
```
leenagupte, line 9: The "no-public-updated-at" is bogus as there should always be a value for `public_updated_at` so I think this test and the associated example can be removed.

spec/models/case_study_spec.rb
```diff
@@ -0,0 +1,29 @@
+RSpec.describe CaseStudy do
+  let(:content_item) do
```
leenagupte, line 2: To be consistent with other tests this should be called `content_store_response`.

spec/models/case_study_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  it_behaves_like "it has updates", "case_study", "has-updates"
+  it_behaves_like "it has updates", "case_study", "only-has-first-published-at"
+  it_behaves_like "it has no updates", "case_study", "doing-business-in-spain"
+  it_behaves_like "it has no updates", "case_study", "no-public-updated-at"
+  it_behaves_like "it can have worldwide organisations", "case_study", "doing-business-in-spain"
+  it_behaves_like "it can have emphasised organisations", "case_study", "doing-business-in-spain"
+
+  describe "#contributors" do
+    it "returns the organisations ordered by emphasis" do
+      model_instance = described_class.new(content_item)
```
leenagupte, line 15: To be consistent with other tests this variable should be called `content_item` rather than `model_instance`

e.g. 
```ruby
content_item = described_class.new(content_store_response)
```

spec/models/case_study_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  it_behaves_like "it has updates", "case_study", "has-updates"
+  it_behaves_like "it has updates", "case_study", "only-has-first-published-at"
+  it_behaves_like "it has no updates", "case_study", "doing-business-in-spain"
+  it_behaves_like "it has no updates", "case_study", "no-public-updated-at"
+  it_behaves_like "it can have worldwide organisations", "case_study", "doing-business-in-spain"
+  it_behaves_like "it can have emphasised organisations", "case_study", "doing-business-in-spain"
+
+  describe "#contributors" do
+    it "returns the organisations ordered by emphasis" do
```
leenagupte, line 14: This should be renamed to include "worldwide organisations". Perhaps something like:

`it "returns the organisations ordered by emphasis followed by worldwide organisations"`

spec/system/case_study_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  context "when visiting a Case Study page" do
+    it "displays the case_study page" do
+      visit "/government/case-studies/get-britain-building-carlisle-park"
+
+      expect(page).to have_title("Get Britain Building: Carlisle Park - Case study - GOV.UK")
+
+      expect(page).to have_css("h1", text: "Get Britain Building: Carlisle Park")
+      expect(page).to have_text("Nearly 400 homes are set to be built on the site of a former tar distillery thanks to Gleeson Homes and HCA investment.")
+
+      expect(page).to have_css(".gem-c-translation-nav")
```
leenagupte, line 19: I think this should be extracted into it's own context, rather than being tested by stealth or being shoe-horned into the withdrawn context. Perhaps something like `context "when the page has translations"`

spec/system/case_study_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    context "when visiting a Withdrawn Case Study page which is also translatable" do
```
leenagupte, line 22: As mentioned above, the "translatable" part of this context should be extracted into a separate context, along with the expectation.

leenagupte (commented):
Thanks for adding a test for the `contributors` method 🎉 

There are couple of other minor things in the tests that should be tidied up.

leenagupte (approved):
Thanks for making the requested changes. 👍 

I've checked a few case studies, and the pages seem to be rendering as expected and with the correct metadata.

A frontender should check this over as well before merging.

matthillco (approved):
I've looked at the visual changes and the frontend code: all looks good to me. Nice work.

PR #4374, opened 2024-11-05
Build statistics block data from new format of csv files

leenagupte:

## What

Builds the statistics block from csv data in an updated format. 

## Why

[Trello card](https://trello.com/c/E856h55m)

Previously we though we'd be receiving csv data in the format:

```csv
Date,variable,value
2024-01-01,variable_name,10
2024-02-01,variable_name,11
2024-03-01,variable_name,12
```

We've since learned that the data will be in the more sensible format:

```csv
,Percentage of people being kinder to one and other
1/6/2013,0.500
1/6/2014,0.600
1/6/2015,0.700
```

It's likely that the x-axis label could be missing, and as it's not currently needed, it has been omitted from the test data.

## How

The format of the expected data in tests didn't change, though the values have been expanded to include floats and missing data points as well as whole numbers.

## Screenshots?
N/A - There are no visual changes.

https://govuk-frontend-app-pr-4374.herokuapp.com/landing-page/homepage
https://govuk-frontend-app-pr-4374.herokuapp.com/landing-page/tasks


app/models/landing_page/block/statistics.rb
```diff
@@ -40,15 +39,28 @@ def attachment
...
+    def csv_headers
+      opened_csv.headers
+    end
+
+    def opened_csv
+      return csv_from_url if attachment
+
+      csv_from_file
+    end
+
```
KludgeKML, line 45: Is it worth memoizing opened_csv? Otherwise a call to rows potentially reads / fetches it twice (once when it sets the variable name, then again when it calls csv_rows)

app/models/landing_page/block/statistics.rb
```diff
@@ -12,19 +12,18 @@ def x_axis_keys
 
     def rows
       rows = []
+      variable_name = csv_headers.last
```
KludgeKML: Are we sure we're only ever going to have one variable per chart?

KludgeKML: Oh I see, it's in the next commit. Uh, maybe smash the last two commits?

leenagupte:
@KludgeKML Thanks for your review. I've made the changes and fixed up. All of the data file changes are in the first commit now.

KludgeKML (approved):
LGTM!

PR #4375, opened 2024-11-05
Landing page images from content item

KludgeKML:
Hero and Featured blocks will now take their image sets from the content item, and can understand the difference between hardcoded images that have to be pathed and remote images where the url can be used directly.

Related to: https://github.com/alphagov/whitehall/pull/9579

https://trello.com/c/00bwdpYL/124-images-from-whitehall

### DIFF
--- a/app/models/landing_page/block/concerns/has_image_set.rb
+++ b/app/models/landing_page/block/concerns/has_image_set.rb
@@ -1,0 +1,15 @@
+require "ostruct"
+
+module LandingPage::Block::Concerns
+  module HasImageSet
+    FeaturedImage = Data.define(:alt, :sources)
+
+    def image
+      @image ||= load_image
+    end
+
+    def load_image
+      FeaturedImage.new(alt: data["image"]["alt"], sources: OpenStruct.new(data["image"]["sources"]))
+    end
+  end
+end
--- a/app/views/landing_page/blocks/_hero.html.erb
+++ b/app/views/landing_page/blocks/_hero.html.erb
@@ -11,10 +11,10 @@
 
 <%= content_tag("div", class: hero_classes) do %>
   <%= picture_tag(class: "app-b-hero__imagewrapper") do %>
-    <%= tag.source srcset: "#{image_path(block.image.sources.desktop)}, #{image_path(block.image.sources.desktop_2x)} 2x", media: "(min-width: 769px)" %>
-    <%= tag.source srcset: "#{image_path(block.image.sources.tablet)}, #{image_path(block.image.sources.tablet_2x)} 2x", media: "(min-width: 641px) and (max-width: 768px)" %>
-    <%= tag.source srcset: "#{image_path(block.image.sources.mobile)}, #{image_path(block.image.sources.mobile_2x)} 2x", media: "(max-width: 640px)" %>
-    <%= image_tag(block.image.sources.desktop, alt: block.image.alt, class: "app-b-hero__image") %>
+    <%= tag.source srcset: "#{block_image_path(block.image.sources.hero_desktop_1x)}, #{block_image_path(block.image.sources.hero_desktop_2x)} 2x", media: "(min-width: 769px)" %>
+    <%= tag.source srcset: "#{block_image_path(block.image.sources.hero_tablet_1x)}, #{block_image_path(block.image.sources.hero_tablet_2x)} 2x", media: "(min-width: 641px) and (max-width: 768px)" %>
+    <%= tag.source srcset: "#{block_image_path(block.image.sources.hero_mobile_1x)}, #{block_image_path(block.image.sources.hero_mobile_2x)} 2x", media: "(max-width: 640px)" %>
+    <%= image_tag(block.image.sources.hero_desktop_1x, alt: block.image.alt, class: "app-b-hero__image") %>
   <% end %>
 
   <div class="govuk-width-container app-b-hero__textbox-wrapper">
--- a/lib/data/landing_page_content_items/be_kinder.yaml
+++ b/lib/data/landing_page_content_items/be_kinder.yaml
@@ -35,14 +35,14 @@ blocks:
   navigation_group_id: Top Menu
 - type: hero
   image:
-    alt: "Placeholder alt text"
+    alt_text: "Placeholder alt text"
     sources:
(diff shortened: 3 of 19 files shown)


spec/models/landing_page/block/hero_spec.rb
```diff
@@ -3,14 +3,62 @@
...
         "alt" => "some alt text",
-        "sources" => {
-          "desktop" => "landing_page/desktop.jpeg",
-          "desktop_2x" => "landing_page/desktop_2x.jpeg",
-          "mobile" => "landing_page/mobile.jpeg",
-          "mobile_2x" => "landing_page/mobile_2x.jpeg",
-          "tablet" => "landing_page/tablet.jpeg",
-          "tablet_2x" => "landing_page/tablet_2x.jpeg",
-        },
+        "sources" => [
```
richardTowers, line 6: I could probably do a much better job of structuring this payload on the whitehall side. All the way down to this if we wanted:

```
      "image" => {
        "alt" => "some alt text",
        "sources" => {
          "hero_desktop_1x" => "landing_page/desktop.jpeg",
          "hero_desktop_2x" => "landing_page/desktop_2x.jpeg",
          "hero_tablet_1x" => "landing_page/tablet.jpeg",
          "hero_tablet_2x" => "landing_page/tablet_2x.jpeg",
          "hero_mobile_1x" => "landing_page/mobile.jpeg",
          "hero_mobile_2x" => "landing_page/mobile_2x.jpeg",
        },
```

... and whitehall could validate that the alt text is nil or the same for all of the images etc.

The other metadata (created_at, caption etc.) is always going to be a bit confusing while we're uploading three separate images (mobile, tablet, desktop) with 2 variants each.

https://github.com/alphagov/whitehall/pull/9579 does produce the payload in the current format - I haven't attempted to improve it yet. I'm happy to stick with the current format if you like, but I'd probably vote to improve it before the two PRs merge.

KludgeKML: Okay, let's simplify it. I was hoping that maybe leaving the code the way it is would give us some more general support in the future, but I'm forgetting YAGNI. Let's make the code simpler now, and we can complicate it later if we need to.

richardTowers: I'll confirm what the structure will be once I've updated my second PR

richardTowers: Current proposed structure:

https://github.com/alphagov/whitehall/pull/9579/files#diff-bec4c63fd7b93e9d9e92a09e0efaaa528c3c3ee2c1a09e5b689970a3d870b1f8R158-R192

KludgeKML: Hmm, are we losing the alt text entirely?

KludgeKML: I've updated, but kept the top-level alt.

app/models/landing_page/block/concerns/has_image_set.rb
```diff
@@ -0,0 +1,15 @@
...
+module LandingPage::Block::Concerns
+  module HasImageSet
+    FeaturedImage = Data.define(:alt, :sources)
+
+    def image
+      @image ||= load_image
+    end
+
+    def load_image
+      FeaturedImage.new(alt: data["image"]["alt"], sources: OpenStruct.new(data["image"]["sources"]))
```
richardTowers, line 12:
```suggestion
      FeaturedImage.new(alt: data.dig("image", "alt"), sources: OpenStruct.new(data.dig("image", "sources")))
```

app/views/landing_page/blocks/_hero.html.erb
```diff
@@ -11,10 +11,10 @@
 
 <%= content_tag("div", class: hero_classes) do %>
   <%= picture_tag(class: "app-b-hero__imagewrapper") do %>
-    <%= tag.source srcset: "#{image_path(block.image.sources.desktop)}, #{image_path(block.image.sources.desktop_2x)} 2x", media: "(min-width: 769px)" %>
-    <%= tag.source srcset: "#{image_path(block.image.sources.tablet)}, #{image_path(block.image.sources.tablet_2x)} 2x", media: "(min-width: 641px) and (max-width: 768px)" %>
-    <%= tag.source srcset: "#{image_path(block.image.sources.mobile)}, #{image_path(block.image.sources.mobile_2x)} 2x", media: "(max-width: 640px)" %>
-    <%= image_tag(block.image.sources.desktop, alt: block.image.alt, class: "app-b-hero__image") %>
+    <%= tag.source srcset: "#{block_image_path(block.image.sources.hero_desktop_1x)}, #{block_image_path(block.image.sources.hero_desktop_2x)} 2x", media: "(min-width: 769px)" %>
```
richardTowers, line 14: Currently [my branch drops the `hero_` prefix](https://github.com/alphagov/whitehall/pull/9579/files#diff-bec4c63fd7b93e9d9e92a09e0efaaa528c3c3ee2c1a09e5b689970a3d870b1f8R158-R192) from the sources hash, so they're just `desktop_2x` etc. That was just because it felt like duplication to be doing `- type: "hero"` and then reiterating `hero_` a bunch of times. But it does introduce an inconsistency with the version names configured in whitehall, as those have to be globally unique, so they do need prefixes.

I don't mind either way really - I'd be happy to update my branch to put back the `hero_`s if you think its more consistent to have them. Or we could update them here.

lib/data/landing_page_content_items/be_kinder.yaml
```diff
@@ -35,14 +35,14 @@ blocks:
   navigation_group_id: Top Menu
 - type: hero
   image:
-    alt: "Placeholder alt text"
+    alt_text: "Placeholder alt text"
```
richardTowers, line 38: Ah, I didn't mean to change this from `alt` to `alt_text` 😓

app/views/landing_page/blocks/_hero.html.erb
richardTowers: 🤔  actually maybe I can get it to match whats currently there exactly, and we can lose this PR 🤔

KludgeKML:
Closing because changes to Whitehall have made this PR obsolete (Whitehall is now giving us the images in the format we were already expecting).

PR #4382, opened 2024-11-11
Static error pages

KludgeKML:

## What

Allow Frontend to render static error pages - this is currently being done by Static, but will still need to be done after Static is retired. We include specific CSS/JS files just for the error pages files, since the application css/js files don't currently contain all the required styles/codes. This mirrors the current situation of the files provided by Static, but doesn't rely on Static being there.

These pages aren't published, the routes exist only to be called by govuk-helm-charts - at deploy time they're called, and the files are pushed up to an S3 bucket to be served. (see here: https://github.com/alphagov/govuk-helm-charts/blob/5b8186099067549179a824c157d44d78c70f1f00/charts/generic-govuk-app/upload-frontend-error-pages.sh )

## Why

https://trello.com/c/bzUFBHkX/382-move-error-pages-from-static-to-frontend, [Jira issue PNP-9252](https://gov-uk.atlassian.net/browse/PNP-9252)

### DIFF
--- a/app/assets/stylesheets/static-error-pages.scss
+++ b/app/assets/stylesheets/static-error-pages.scss
@@ -1,0 +1,13 @@
+
+@import "govuk_publishing_components/govuk_frontend_support";
+@import "govuk_publishing_components/component_support";
+
+@import "govuk_publishing_components/components/button";
+@import "govuk_publishing_components/components/cookie-banner";
+@import "govuk_publishing_components/components/cross-service-header";
+@import "govuk_publishing_components/components/feedback";
+@import "govuk_publishing_components/components/label";
+@import "govuk_publishing_components/components/layout-footer";
+@import "govuk_publishing_components/components/layout-for-public";
+@import "govuk_publishing_components/components/layout-header";
+@import "govuk_publishing_components/components/layout-super-navigation-header";
--- a/app/views/static_error_pages/400.html.erb
+++ b/app/views/static_error_pages/400.html.erb
@@ -1,0 +1,5 @@
+<%= render partial: "error_page", locals: {
+    heading: "Sorry, there is a problem",
+    intro: '<p class="govuk-body">Go back and change the information you entered.</p>',
+    status_code: 400,
+} %>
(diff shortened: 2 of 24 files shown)


app/assets/stylesheets/static-error-pages.scss
leenagupte, line 1: I think that as with the other consolidation work, it would be great to add an audit trail of the files moved to the commit message.

I think in this case `application.js`, `application.scss` and `_error_page.html.erb` are the source for some of these files?

app/views/static_error_pages/_error_page.html.erb
```diff
@@ -0,0 +1,30 @@
+<% content_for :head do %>
```
leenagupte, line 1: I think the emergency banner also needs to be added here.

app/views/static_error_pages/400.html.erb
leenagupte, line 1: It's great that these have been added in a separate commit. 🎉 

Saying that, could you add the audit trail for these files to the commit message.

leenagupte (commented):
These look great! 🎉 

I've added some inline comments about keeping the audit trail upto date, and adding the emergency banner.

Could you also add the information from the description in the Trello card to the first commit or the PR description. There's a lot of useful information in there that future developers won't be able to access if they're trying to understand these changes.

app/views/static_error_pages/_error_page.html.erb
KludgeKML: Ah, no - that was specifically removed, but I do need to mention that in the commit somehow (https://gds.slack.com/archives/C06SVL1GC2Y/p1732628784914559)

app/views/static_error_pages/400.html.erb
KludgeKML: Audit trail to directory added.

app/assets/stylesheets/static-error-pages.scss
KludgeKML: Audit trail added for controller and template. I didn't add the JS and SCSS ones because they don't relate exactly to the ones in Static. Static has "this is everything every page needs!" JS/CSS (not in fact true, because there's a lot in there that these pages don't appear to need), these ones are just the component styles and scripts needed for the error pages

app/views/static_error_pages/_error_page.html.erb
KludgeKML: I've added the reasoning into the commit. I didn't bother putting the slack link in because it'll expire, but I'll add a transcript of it into the trello card so there's at least some way of getting to it.

leenagupte (approved):
Looks great! Thanks for updating the commit messages and adding an audit trail.

PR #4392, opened 2024-11-13
Add metadata block

AshGDS:

## What / Why
- Adds a `metadata` block for the landing page
- These meta tags make social share previews look better
- It grabs the `title` and `description` from the content item by default, but these can be overridden in the `yml` body.
- You can use the [Social share preview](https://chromewebstore.google.com/detail/social-share-preview/ggnikicjfklimmffbkhknndafpdlabib?hl=en) plugin to see how it will look to users
- The [Machine readable data](https://components.publishing.service.gov.uk/component-guide/machine_readable_metadata) component crashes when I import it into PfG, and doesn't allow featured image customisation, so I've just added the metatags using a block. [We've done similar for the GOVUK Roadmap](https://github.com/alphagov/frontend/blob/27f1a1f07cc08282daecd7e7dd8ad05834da460a/app/views/roadmap/index.html.erb#L3)
- [Trello card](https://trello.com/c/RolYznuu/132-make-pages-shareable)

## Screenshots

image

### DIFF
--- a/app/views/landing_page/show.html.erb
+++ b/app/views/landing_page/show.html.erb
@@ -2,6 +2,13 @@
   add_view_stylesheet("landing_page")
 %>
 
+<% content_for :extra_headers do %>
+  <%= render "govuk_publishing_components/components/machine_readable_metadata", {
+    content_item: @content_item.to_h,
+    schema: :article
+  } %>
+<% end %>
+
 <div class="landing-page-header">
   <div class="govuk-width-container">
     <div class="landing-page-header__blue-bar">


andysellick (commented):
Looks fine to me, my only question is should this be a block that we need to remember to add, or should it be built into the page template somehow?

andysellick:
@AshGDS I'm prepared to be wrong on this one, so maybe just start with a chat with a BE dev and see what they think.

andysellick (approved):
Approving for now but may find a different solution later.

AshGDS:
@andysellick Thanks for reviewing the initial version of this 👍 Keith has helped me make it a lot simpler fortunately. 

@KludgeKML I've got this code as simple as possible (thanks for the help.) Does it look OK to you? We can't use the machine readable metadata block and override the content item in the `.yml`, as then we would need to define an absolute image URL which we can't do with our `.yml` files. Therefore what I've done here is the best I think we can do. Are you able to review it?

Thanks!

andysellick:
Nice work @AshGDS. I'll leave to a backender to review. Only one thought - should we bother trying to set a specific image for these pages or just use the default GOV.UK one? I imagine sourcing these images might be time consuming...

AshGDS:
@andysellick Ah OK, in that case we can use the `machine_readable_metadata` block as the image was what was causing the problems 👍

AshGDS:
Thanks @andysellick - the PR is very simple now, so are you able to quickly re-review?

KludgeKML (approved):
LGTM!

PR #4393, opened 2024-11-13
Adjust hero images

andysellick:

## What
Adjust hero image behaviour on large desktop by adding a new breakpoint, so the image will always be full width. We retain the height though to prevent the page from being pushed down, so image content will be lost at large screen sizes.

Note: I was going to more on this but other stuff is calling, so I'm passing it on. Currently the heights for the different breakpoints aren't finalised (you'll notice that the desktop and large desktop are the same) but that will be addressed shortly.

## Why
Trying to accommodate feedback.

## Visual changes

Trello card: https://trello.com/c/F6n5lNH7


AshGDS (approved):
LGTM 👍

PR #4395, opened 2024-11-14
Stop converting nil graph values to zero

leenagupte:

## What

Stop converting nil graph values to zero.

## Why

Calling `to_f` or `to_i` on a nil value converts it zero. This is not the desired behaviour as we'd rather show a gap in the graph when the data is missing.

### Before

Screenshot 2024-11-14 at 14 47 13

### After
Screenshot 2024-11-14 at 14 46 42


KludgeKML (approved):
LGTM

PR #4400, opened 2024-11-15
Add full width background to blocks

andysellick:

## What
- adds an option on all blocks to have a full width grey background while maintaining the block’s normal width in relation to the page
- have had to account for some blocks with this needing to have padding and some not, so there’s two options on a block now
- have had to put some slightly gnarly stuff in to deal with situations where we want two of these to be next to each other, for example where a heading block has a govspeak video block after it, but they need to look like they’re in the same full width background
- so the heading and govspeak blocks now allow a margin bottom to be passed to the component inside them from the YAML

Slightly nervous about this approach as it feels like there's a lot of complexity now - options on blocks, options within blocks that are passed to other blocks, and spacing based on best guesses rather than a consistent overall strategy.

Things to do:

- update blocks documentation
- add tests?

## Why
New requirement.

## Visual changes
Example on a video. This is a heading block followed by a govspeak block (containing a video). Both blocks have the full width background, both have the full width background padding. The heading has been given a margin_bottom of 0 from the YAML, and the govspeak has been given a margin_bottom of 2, to balance out the spacing.

Screenshot 2024-11-15 at 09 07 04

Same showing if cookies haven't been accepted. I was concerned there'd be too much grey but I think it's alright.

Screenshot 2024-11-15 at 09 06 48

Trello card: https://trello.com/c/iSQk5ZWA

### DIFF
--- a/app/assets/stylesheets/views/_landing_page.scss
+++ b/app/assets/stylesheets/views/_landing_page.scss
@@ -22,3 +22,20 @@
     text-decoration: none;
   }
 }
+
+.landing-page__full-background {
+  background: govuk-colour("light-grey");
+  margin-bottom: govuk-spacing(9);
+
+  &:has(+ .landing-page__full-background) {
+    margin-bottom: 0;
+  }
+}
+
+.landing-page__full-background--padding {
+  padding: govuk-spacing(4) 0;
+
+  &:has(+ .landing-page__full-background) {
+    padding-bottom: 0.05px;
+  }
+}
--- a/app/models/landing_page/block/base.rb
+++ b/app/models/landing_page/block/base.rb
@@ -12,5 +12,13 @@ module LandingPage::Block
     def full_width?
       false
     end
+
+    def full_background?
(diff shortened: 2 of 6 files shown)


leenagupte:
@andysellick I've tried out using `full_background` as the single setting here: https://github.com/alphagov/frontend/compare/full-width-background...full-width-background-leena-spike?expand=1

leenagupte (commented):
Sorry I thought I'd posted my review 🤦‍♀️ 

I'm not convinced by the approach to add margin-bottom fields to the content item. I've spiked out adding the full_background to the models and using that to determine the margin-bottom.

I do think we need to really look at how all of this stuff works, because I think some of the confusion is based on where the rendering code is.

For example, the `card` component passes `options` to the statistics block when it shouldn't need to. The statistics block should know whether it's minimal or not and behave accordingly.

andysellick:
Closing in favour of https://github.com/alphagov/frontend/pull/4406

PR #4401, opened 2024-11-15
Remove block options

andysellick:

## What
- some blocks have options, which are passed to them to control things like inverse and margin when they're inside other blocks
- for example, all blocks within a hero block text box need to be inverse because the background colour is dark blue, otherwise the text would be unreadable

Things to do:

- update blocks documentation

## Why
- I came up with this system, but I now think it's confusing as blocks do 'magic' things to make them appear correctly and these things are effectively hard coded, removing control
- so I'm removing this, and moving all the options into the YAML so they can be controlled from there
- it makes the YAML a bit more verbose, but I think makes adding and configuring blocks clearer

## Visual changes
None, but we might need to update some of the data currently in draft otherwise the pages are going to have visual/layout problems.

### DIFF
--- a/app/views/landing_page/blocks/_action_link.html.erb
+++ b/app/views/landing_page/blocks/_action_link.html.erb
@@ -1,6 +1,5 @@
-<% inverse = options[:inverse] || false %>
 <%= render "govuk_publishing_components/components/action_link", {
-  inverse: inverse,
+  inverse: block.data["inverse"],
   text: block.data["text"],
   href: block.data["href"]
 } %>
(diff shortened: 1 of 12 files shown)


andysellick:
@leenagupte does this PR seem sensible to you? Options seemed like a good idea at the time, but they're starting to get in the way, particularly in light of https://github.com/alphagov/frontend/pull/4400

leenagupte:
> @leenagupte does this PR seem sensible to you? Options seemed like a good idea at the time, but they're starting to get in the way, particularly in light of #4400

So reason that block options were added in the first place, is that everything that's configured in the blocks will appear in the content item. And it feels very odd to define such purely low-level presentational things like "margin-bottom" in the content item. It felt like the type of block should be able to handle this stuff. i.e. the block should be able to figure out the margin it needs based on information it has about itself. i.e. if it's minimal, or is full-width.

andysellick:
@leenagupte that's fair. I think I'm going to go with a different approach, but good to have your input.

PR #4406, opened 2024-11-18
Add block spacing model and background

andysellick:

## What
- adds a general spacing model for all blocks, with a margin bottom of 60px
- some blocks don't want this, so we have specific overrides e.g. share links and main navigation
- this allows us to add a coloured background to blocks a bit more easily, although there are still slightly complex CSS rules around backgrounded blocks following others, where we need to do that (a heading preceding a govspeak block, for example, where they need to look like they're part of the same thing)
- also changes the main block layout, so that all blocks can now have a full width background, and any blocks that should be page width will have the govuk-width-container element added inside them, not as part of the main block div

## Why
Two reasons. We've not had consistent/general rules for block spacing, and the need for some blocks to have a full page width background is forcing this.

## Visual changes
Blocks with and without the background, to show the difference.

First, the video 'block' (actually two blocks, including the heading)

Without background | With background
------ | ------
Screenshot 2024-11-18 at 11 38 08 | Screenshot 2024-11-18 at 11 38 24

Progress area:

Without background | With background
------ | ------
Screenshot 2024-11-18 at 11 42 16 | Screenshot 2024-11-18 at 11 42 27

Trello card: https://trello.com/c/iSQk5ZWA

### DIFF
--- a/app/views/landing_page/show.html.erb
+++ b/app/views/landing_page/show.html.erb
@@ -6,7 +6,7 @@
   <div class="govuk-width-container">
     <div class="landing-page-header__blue-bar">
     </div>
-    
+
     <%= render 'govuk_publishing_components/components/contextual_breadcrumbs', content_item: content_item_hash, inverse: true %>
 
     <div class="landing-page-header__org">
@@ -26,9 +26,15 @@
 
 <main class="landing-page" id="content">
   <% @content_item.blocks.each do |block| %>
-    <%= tag.div class: ["govuk-block__#{block.type}", ("govuk-width-container" unless block.full_width?)] do
-      render_block(block)
-    end %>
+    <%= tag.div class: ["govuk-block govuk-block__#{block.type}", ("govuk-block--background" if block.full_width_background?)] do %>
+      <% if block.full_width? %>
+        <%= render_block(block) %>
+      <% else %>
+        <div class="govuk-width-container">
+          <%= render_block(block) %>
+        </div>
+      <% end %>
+    <% end %>
   <% end %>
   <% if @content_item.blocks.empty? %>
     Warning: No blocks specified for this page
--- a/app/assets/stylesheets/views/_landing_page.scss
+++ b/app/assets/stylesheets/views/_landing_page.scss
@@ -22,3 +22,29 @@
     text-decoration: none;
   }
 }
+
+.govuk-block {
+  margin-bottom: govuk-spacing(9);
+}
+
+.govuk-block--background {
+  background: govuk-colour("light-grey");
+  padding: govuk-spacing(8) 0;
+
+  &:has(+ .govuk-block--background) {
+    padding-bottom: 0.05px; // stop margin escaping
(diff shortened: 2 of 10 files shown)


andysellick:
@leenagupte I've had a rethink of this and I think this is a cleaner approach. Interested in your thoughts.

AshGDS (commented):
Looks good, though I have a question. 

I added the `full_width_background` to the `heading` / `grid_container` elements below the video and it resulted in this spacing between the video and the heading (see attached screenshot.) Is this what you were referring to by needing `slightly complex CSS rules around backgrounded blocks following others` ? Or is this a bug, or me using this functionality in the wrong way?
image

Thanks

andysellick:
@AshGDS good spot, it's because of margin escape from the govspeak (have pushed a fix).

app/views/landing_page/show.html.erb
```diff
@@ -26,9 +26,15 @@
...
 <main class="landing-page" id="content">
   <% @content_item.blocks.each do |block| %>
-    <%= tag.div class: ["govuk-block__#{block.type}", ("govuk-width-container" unless block.full_width?)] do
-      render_block(block)
-    end %>
+    <%= tag.div class: ["govuk-block govuk-block__#{block.type}", ("govuk-block--background" if block.full_width_background?)] do %>
+      <% if block.full_width? %>
+        <%= render_block(block) %>
+      <% else %>
+        <div class="govuk-width-container">
```
leenagupte, line 33: We can probably refactor this a little so that the if / else is inline, but that is really really minor.

leenagupte (commented):
This is a much nicer approach! I am really glad that none of the specific styling details have encroached into the content item.

It's a ✅ from the backend.

app/views/landing_page/show.html.erb
andysellick: Agreed, I was just writing out to be clear.

andysellick: Also can't think of a shorter way of writing it, if you have a suggestion would be happy to add.

leenagupte: Yeah, I can't either. Perhaps one to worry about later?

andysellick:
> Thanks @andysellick - when I add `full_width_background: true` to the heading below the video, the spacing between them reduces - is that intentional? I'm assuming it is so the elements all look like they're all a part of the same grey square?

Yes, that's correct. It's because some sections are made up of multiple blocks and we need to account for that. For example, the heading above the video is part of a separate block, but we want to include it inside the grey background, which without the tweaks I've included would leave a gap between them and have wildly wrong padding.

I think the general takeaway from this is that we need some rules around the usage of this thing, basically no two consecutive sections should have this background applied, but it is designed to allow headings to fit closer together with their content. 

So yes, we wouldn't give the same background to both the video 'section' and whatever follows it.

AshGDS (approved):
LGTM 👍

PR #4407, opened 2024-11-18
Rework hero image block

jon-kirwan:
## What

Rework hero image block

See https://trello.com/c/bdr8UDPd/147-rework-hero-image-block

## Why

Design requirements.


app/views/landing_page/blocks/_heading.html.erb
```diff
@@ -1,12 +1,14 @@
 <%
   heading_level = block.data["heading_level"] || 2
+  font_size = block.data["font_size"] || 'm'
```
andysellick, line 3: You've clearly worked hard on this, but I think maybe it would be best if we just set the text size based on whether it's a top of the page or mid page hero - it means it looks correct out of the box, and we don't have to worry about people accidentally setting the wrong font size on some pages and making them inconsistent.

app/helpers/mission_type_helper.rb
```diff
@@ -0,0 +1,20 @@
+module MissionTypeHelper
+  def self.border_style(mission_type)
```
andysellick, line 2: I think it should be possible to write this shorter, something like...

```Ruby
def self.border_style(mission_type)
  valid_numbers = [1..6]
  return "mission-1" unless valid_numbers.include? mission_type

  "mission_#{mission_type}"
end
```

app/assets/stylesheets/views/_landing_page/helpers/_border-style.scss
```diff
@@ -0,0 +1,23 @@
+.mission-1 {
+  border-top: 20px solid #e60067;
```
andysellick, line 2: I think you're the first to add these colours so it might help if they were declared as variables for wider use e.g. `$mission-1-colour: #e60067;`

app/assets/stylesheets/views/_landing_page/hero.scss
```diff
@@ -34,7 +36,7 @@ $large-desktop-height: 610px;
 .app-b-hero__image {
   display: block;
 
-  @media (min-width: 1281px) {
+  @include govuk-media-query($from: desktop) {
```
andysellick, line 39: This has changed the behaviour slightly, just need to check. 

Previously the hero on desktop was a fixed height and didn't scale, so image content could be lost left/right. That was fine until it passed the width of the image at 1280, and a gap appeared.

To fix this, we added a new breakpoint above desktop (> 1280) to make the image absolute and truly full width, but retaining a fixed height, with image content potentially being lost off the bottom.

I think the requirement was only to set the image to scale at an aspect ratio for tablet and mobile, with the behaviour on desktop being acceptable. Shall we catch up about this?

app/views/landing_page/blocks/_govspeak.html.erb
```diff
@@ -1,7 +1,11 @@
-<% inverse = options[:inverse] || false %>
+<% 
+  inverse = options[:inverse] || false
+  margin_bottom = block.data["margin_bottom"] || 9
```
andysellick, line 3: We don't want to include margins in the YAML because then content authors could break the spacing all over the page. Instead, if this has to be set, use an option inside the hero block, to set margin on govspeak only inside that block. See the card block for an example.

andysellick (commented):
Looking good! A few comments, happy to chat about them.

andysellick:
One additional thought - the changes in height for the hero on desktop (400px) aren't yet applied to the mid page variant. Is it possible to make the mid page consistent with the top of page? Otherwise we'll need to specify different image sizes for each, which will get complicated.

app/helpers/mission_type_helper.rb
jon-kirwan: I'm separating these into a new PR to make the helper available for other components requiring mission specific styles https://github.com/alphagov/frontend/pull/4416.

app/assets/stylesheets/views/_landing_page/helpers/_border-style.scss
jon-kirwan: Same as above, I'm separating these into a new PR to make the helper available for other components requiring mission specific styles https://github.com/alphagov/frontend/pull/4416.

app/views/landing_page/blocks/_heading.html.erb
jon-kirwan: Thanks, that makes sense. I've updated it to keep 'm' as the default and set it to 'l' only when the position is 'top'.

app/assets/stylesheets/views/_landing_page/hero.scss
```diff
@@ -1,40 +1,36 @@
...
-$tablet-height: 512px;
-$desktop-height: 610px;
-$large-desktop-height: 610px;
-
-$desktop-height-top: 400px;
-
-.app-b-hero {
-  position: relative;
-}
+// $tablet-height: 512px;
```
andysellick, line 4: This line is commented - does it need uncommenting or removing?

app/assets/stylesheets/views/_landing_page/hero.scss
```diff
@@ -75,19 +58,13 @@ $desktop-height-top: 400px;
   @include govuk-media-query($until: desktop) {
     display: flex;
     flex-direction: column-reverse;
-    min-height: $tablet-height;
+    // min-height: $tablet-height;
```
andysellick, line 61: This line is also commented?

lib/data/landing_page_content_items/goals.yaml
```diff
@@ -93,12 +93,12 @@ blocks:
   image:
     alt: "Placeholder alt text"
     sources:
-      desktop: "landing_page/placeholder/desktop.png"
-      desktop_2x: "landing_page/placeholder/desktop_2x.png"
-      mobile: "landing_page/placeholder/mobile.png"
-      mobile_2x: "landing_page/placeholder/mobile_2x.png"
-      tablet: "landing_page/placeholder/tablet.png"
-      tablet_2x: "landing_page/placeholder/tablet_2x.png"
+      desktop: "landing_page/placeholder/missions_hero_desktop.jpg"
```
andysellick, line 96: Fine to keep these changes in - these images are already in `frontend`.

andysellick (approved):
All looks good, if you can squash some of these commits together that'd great, and just a couple of commented lines needing attention. Thanks!

app/assets/stylesheets/views/_landing_page/hero.scss
jon-kirwan: Also removed.

PR #4414, opened 2024-11-19
Main element

KludgeKML:

## What

Add the main element to the layout, and remove it from individual views.

> [!IMPORTANT]
> To make this easier to review, in most of the commits I have not updated indenting after removing the main elements
> Removing indenting confuses Github's comparison viewer, making it a lot harder to see what I've changed. I've saved
> all the indenting updates to the last commit, which is almost entirely whitespace (when viewing it, select Hide
> Whitespace from the settings and you should see almost all files are whitespace only, and the handful that aren't 
> are just additional lines added for readability (ie also whitespace, but one that GH for some reason doesn't ignore).

## Why

https://trello.com/c/T0Yt3VCD/384-make-main-element-default-in-frontend, [Jira issue PNP-7358](https://gov-uk.atlassian.net/browse/PNP-7358)

## Screenshots

Almost all screens do not change at all, there are two minor changes which fix spacing issues caused by a combination of the move from government-frontend and the correct implementation of govuk-main-wrapper.

#### Before

Screenshot 2025-01-28 at 15 27 35

#### After

Screenshot 2025-01-28 at 15 27 29

#### Before

Screenshot 2025-01-28 at 15 24 45

#### After

Screenshot 2025-01-28 at 15 24 52

### DIFF
--- a/app/views/calendar/when_do_the_clocks_change.html.erb
+++ b/app/views/calendar/when_do_the_clocks_change.html.erb
@@ -5,59 +5,59 @@
 
 <%= render :partial => "calendar_head" %>
 
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <%= render "govuk_publishing_components/components/heading", {
-          text: @calendar.title,
-          heading_level: 1,
-          font_size: "xl",
-          margin_bottom: 8,
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/heading", {
+      text: @calendar.title,
+      heading_level: 1,
+      font_size: "xl",
+      margin_bottom: 8,
+    } %>
+
+    <article role="article">
+      <%- division = @calendar.divisions.first -%>
+
+      <% if division.upcoming_event %>
+        <%= render "govuk_publishing_components/components/panel", {
+          prepend: "The #{division.upcoming_event.notes.gsub(' one hour', '').downcase}",
+          title: division.upcoming_event.date.strftime("%e %B"),
         } %>
+      <% end %>
 
-        <article role="article">
-          <%- division = @calendar.divisions.first -%>
-
-          <% if division.upcoming_event %>
-            <%= render "govuk_publishing_components/components/panel", {
-              prepend: "The #{division.upcoming_event.notes.gsub(' one hour', '').downcase}",
-              title: division.upcoming_event.date.strftime("%e %B"),
-            } %>
-          <% end %>
-
-          <div class="app-c-calendar app-c-calendar--clocks">
-            <%= render "govuk_publishing_components/components/table", {
-              first_cell_is_header: true,
-              head: [
-                { text: t("bank_holidays.year") },
-                { text: t("bank_holidays.clocks_forward") },
-                { text: t("bank_holidays.clocks_backward") },
-              ],
-              rows: division.years.each.map do |year| [
-                { :text => year },
-                { :text => year.events[0].date.strftime("%e %B") },
-                { :text => year.events[1].date.strftime("%e %B") },
-              ] end,
-            } %>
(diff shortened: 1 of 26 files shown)


matthillco (approved):
Looks great, it's good to see this consistency in the views now. Nice job.

app/views/calendar/when_do_the_clocks_change.html.erb
```diff
@@ -5,8 +5,6 @@
 
 <%= render :partial => "calendar_head" %>
 
-<div class="govuk-width-container">
```
leenagupte, line 8: Should this "govuk-width-container" class be being removed as well?

app/views/shared/_base_page.html.erb
```diff
@@ -25,12 +24,8 @@
         content_item: content_item.to_h %>
     </div>
 
-    <% content_for :after_content do %>
```
leenagupte, line 28: Shouldn't the footer content be after the "main" block, i.e. in the `:after_content` block?

leenagupte (commented):
I have a couple of inline questions, otherwise I think it looks good.

Could you restart the Heroku review so that I can double-check that "When do the clocks change" and "Licence transactions" are rendering ok?

app/helpers/application_helper.rb
```diff
@@ -17,6 +17,10 @@ def wrapper_class(content_item = nil)
     html_classes.join(" ")
   end
 
+  def content_item_class
+    content_item&.schema_name ? content_item.schema_name.dasherize : ""
```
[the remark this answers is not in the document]
KludgeKML: There's no current styling based on this element, so for the moment I'd be inclined to leave it as is (let's have this be the schema name, not "schema name or sometimes document type, who knows?"). It makes sense in the factory because you genuinely want to treat the document types differently. But I don't think we need it here yet. (maybe later, so let's keep the thought in mind, but not for this change).

app/views/calendar/when_do_the_clocks_change.html.erb
KludgeKML: I've checked with Andy, and yes, we do want to remove this (the root of these views should be the bit just inside the main element, and we already have a govuk-width-container outside the main elements - the wrapper).

app/views/shared/_base_page.html.erb
KludgeKML: Uh, it's tricky! The contextual footer _should_ be in the main (it's related to the main content). Shared footers at the bottom of the page shouldn't be. I'll run through again just to make sure the footers are all in the right place.

app/helpers/application_helper.rb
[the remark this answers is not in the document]
leenagupte: If there's no current styling, do we need this class at all? I know that [government-frontend adds the schema name](https://github.com/alphagov/government-frontend/blob/7148e79bd5d4efba78a60b792455c822e6d6588e/app/views/content_items/field_of_operation.html.erb#L42) as a class, but if there's nothing backing it up, why bother?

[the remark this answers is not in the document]
KludgeKML: Well, my thought was: there's no styling for licence transactions, but there might be for others. But in fact, I don't think either styling or JS uses these classes, so yeah, it might be possible to remove entirely. Let me try that against the test suite.

app/views/shared/_base_page.html.erb
KludgeKML: Have updated the contextual footers for case studies, travel advice, and take part (all of which were outside the main element but should have been inside).

app/helpers/application_helper.rb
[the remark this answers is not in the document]
KludgeKML: Removed.

leenagupte (approved):
Thanks for responding to the previous comments. It looks good to me 👍

PR #4416, opened 2024-11-19
Add mission type style helpers

jon-kirwan:
## What

Add mission type style helpers

## Why

To create a shared helper for generating border and background style colors that highlight mission specific content

## Anything else

To use, add the helper within a block view where needed to select the mission name, and then concatenate it with the style name e.g.
```rb
class: "border-top--#{MissionTypeHelper.style(4)}" if block.data["mission_type"]"
```


andysellick (approved):
Looks good! 👏

PR #4421, opened 2024-11-20
Change design for cards

andysellick:

## What
- first bit, changing cards in 3 columns on the homepage
- use new branding styles for boxes
- remove link cover as no longer needed
- second bit, reworking progress page
- mainly just adjusting example blocks to match new layout

## Why
New design.

## Visual changes
Before | After
------ | ------
Screenshot 2024-11-20 at 09 55 17 | Screenshot 2024-11-20 at 09 53 01
Screenshot 2024-11-20 at 10 45 39 | Screenshot 2024-11-20 at 10 45 18

### DIFF
--- a/app/assets/stylesheets/views/_landing_page/card.scss
+++ b/app/assets/stylesheets/views/_landing_page/card.scss
@@ -7,15 +7,14 @@
   justify-content: space-between;
   flex: 1 1;
   min-width: 0;
-  margin-bottom: govuk-spacing(6);
+
+  @include govuk-media-query($until: desktop) {
+    margin-bottom: govuk-spacing(6);
+  }
 }
 
 .app-b-card__textbox {
   padding: govuk-spacing(4) 0;
-
-  @include govuk-media-query($until: tablet) {
-    padding: govuk-spacing(4);
-  }
 }
 
 .app-b-card__figure {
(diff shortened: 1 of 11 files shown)


JamesCGDS:
Looks good! Just a quick couple of things (and apologies if these are getting a bit too granular for this stage):

- Should the link beneath the cards be styled as bold?
- Do the card headings need some left padding removed on mobile for consistent alignment? (screenshot below)

image

AshGDS (commented):
Looks good to me, apart the things that @JamesCGDS noticed 👍

matthillco:
Looking at the mission helper in [Jon's PR](https://github.com/alphagov/frontend/pull/4416), I think the property names should be made consistent. He's using `mission_type` but you're using `mission` -- that's a bit confusing. Also, since we have the helper and it's merged, would it not be better to use that here? @andysellick @JamesCGDS

andysellick:
Thanks @JamesCGDS.

> * Should the link beneath the homepage cards be styled as bold?

Yes, probably. I also need to fix the too-large gap between this and the graphs. Mind if I raise that in a separate PR? Currently not sure how to fix that, but this feels like a step in the right direction.

> * Do the headings on the homepage cards need some left padding removed on mobile for consistent alignment? 

Fixed now, thanks.

> * The non-JS content for the charts on the homepage cards might need updating as they currently say the data is available in a below table

Yes, good shout. I've fixed this, but it's probably dependent on some changes in the component itself, so I'll raise a separate PR for that.

JamesCGDS (approved):
All looks good to me, nice work! And sounds good re fixing the other stuff in a separate PR

PR #4423, opened 2024-11-20
Add Press Notices block

matthillco:

## What

Add a new block to handle Press Notices. This combines two gem components, `heading` and `document_list`. [Trello card?](https://trello.com/c/Mi5AaGpo/156-change-our-progress-page)

## Why

The Press Notices design has some custom spacing and a top border, which isn't covered by any of the current blocks. Creating a new block therefore allows us to add those custom elements and keep them self contained.

## Screenshot
image

### DIFF
--- a/app/views/landing_page/blocks/_press_notices.html.erb
+++ b/app/views/landing_page/blocks/_press_notices.html.erb
@@ -1,0 +1,13 @@
+<% if block.items.any? %>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Press Notices",
+    padding: true,
+    border_top: 5,
+    margin_bottom: 3
+  } %>
+
+  <%= render "govuk_publishing_components/components/document_list", {
+    margin_bottom: 0,
+    items: block.items,
+  } %>
+<% end %>
--- a/lib/data/landing_page_content_items/tasks.yaml
+++ b/lib/data/landing_page_content_items/tasks.yaml
@@ -170,6 +170,23 @@ blocks:
         y_axis_label: "Y Axis"
         csv_file: "data_one.csv"
         data_source_link: https://www.ons.gov.uk/economy/grossdomesticproductgdp/timeseries/ihyq/qna
+- type: two_column_layout
+  theme: two_thirds_one_third
+  blocks:
+  - type: press_notices
+    items:
+    - text: An example title for this press notice 1
+      href: "https://www.gov.uk"
+      document_type: Press release
+      public_updated_at: "2024-10-27 10:29:44 +0000"
+    - text: An example title for this press notice 2
+      href: "https://www.gov.uk"
+      document_type: Press release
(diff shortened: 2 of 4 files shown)


app/views/landing_page/blocks/_press_notices.html.erb
```diff
@@ -0,0 +1,13 @@
+<% if block.items.any? %>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Press Notices",
+    padding: true,
+    border_top: 5,
+    margin_bottom: 3
+  } %>
+
+  <%= render "govuk_publishing_components/components/document_list", {
+    margin_bottom: 4,
```
andysellick, line 10: I think you could take this margin off (or probably set it to zero) as we now have a general spacing model for blocks, which will put in the required space beneath this block.

lib/data/landing_page_content_items/tasks.yaml
```diff
@@ -149,6 +149,23 @@ blocks:
             csv_file: "data_one.csv"
             data_source_link: https://www.ons.gov.uk/economy/grossdomesticproductgdp/timeseries/ihyq/qna
             padding: true
+- type: two_column_layout
+  theme: two_thirds_one_third
+  blocks:
+  - type: press_notices
+    items:
+    - text: An example title for this press notice 1
```
andysellick, line 157: Might need to check with @richardTowers on his plans for how content is pulled into this space, although this looks fine for now.

lib/data/landing_page_content_items/tasks.yaml
```diff
@@ -149,6 +149,23 @@ blocks:
...
+      path: https://www.gov.uk
+      document_type: Press release
+      public_updated_at: "2024-10-27 10:29:44 +0000"
+    - text: An example title for this press notice 2
+      path: https://www.gov.uk
+      document_type: Press release
+      public_updated_at: "2024-10-16 11:34:12 +0000"
+    - text: An example title for this press notice 3
+      path: https://www.gov.uk
```
andysellick, line 166: Nitpick: can we use `href` instead of `path`? Just to make it consistent with (hopefully all of) the other blocks.

matthillco: Sure, although this one's using `document_list` under the hood which uses `path` instead. Should that one also change to `href`?

andysellick: You mean the component itself? No, that's fine, just here in the YML.

app/views/landing_page/blocks/_press_notices.html.erb
matthillco: Margin is now set to 0 to work with the general spacing.

lib/data/landing_page_content_items/tasks.yaml
matthillco: OK this is is done.

andysellick (approved):

PR #4424, opened 2024-11-20
Add AB test for search autocomplete

csutter:
## What
This adds an AB test to the homepage, replacing the `search` component with the new `search_with_autocomplete` one for the B variant.

## Why
The same AB test will be running across Frontend and Finder Frontend to allow us to measure whether providing users with autocomplete suggestions as they type will lead to better outcomes from interacting with site search.

## How
Following the usual conventions for setting up AB tests in GOV.UK frontend apps. The AB test is already live on Fastly (with 100% traffic to Z).

### DIFF
--- a/app/controllers/concerns/search_autocomplete_ab_testable.rb
+++ b/app/controllers/concerns/search_autocomplete_ab_testable.rb
@@ -1,0 +1,28 @@
+module SearchAutocompleteAbTestable
+  ALLOWED_VARIANTS = %w[A B Z].freeze
+
+  def self.included(base)
+    base.helper_method :search_autocomplete_ab_test_variant
+    base.after_action :set_search_autocomplete_ab_test_response_header
+  end
+
+  def search_autocomplete_ab_test
+    @search_autocomplete_ab_test ||= GovukAbTesting::AbTest.new(
+      "SearchAutocomplete",
+      allowed_variants: ALLOWED_VARIANTS,
+      control_variant: "Z",
+    )
+  end
+
+  def search_autocomplete_ab_test_variant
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/search_autocomplete_ab_testable.rb
```diff
@@ -0,0 +1,31 @@
...
+      :show_search_autocomplete?,
+    )
+    base.after_action :set_search_autocomplete_ab_test_response_header
+  end
+
+  def search_autocomplete_ab_test
+    @search_autocomplete_ab_test ||= GovukAbTesting::AbTest.new(
+      "SearchAutocomplete",
+      allowed_variants: ALLOWED_VARIANTS,
+      control_variant: "Z",
```
leenagupte, line 16: Minor: I thought the control variant was A? 
https://github.com/alphagov/govuk-fastly/blob/main/dictionaries.yaml#L17

csutter: Ah. This is the old debate from the other week around my usage of the word "control" 🥲 

_Technically_, A is the control and Z is the baseline, however, the terminology in the GOV.UK AB testing library is different. In Finder Frontend, we also [declare Z as the control](https://github.com/alphagov/finder-frontend/blob/main/app/controllers/ab_test/search_autocomplete_ab_testable.rb#L10) for this test.

leenagupte (approved):
I'm approving based this as the the comments in the [ab test dictionary are being updated to match](https://github.com/alphagov/govuk-fastly/pull/126).

PR #4426, opened 2024-11-20
Fix mid page hero container sizing issue

jon-kirwan:
## What

Fix "mid page" hero container sizing issue

## Why

To fix alignment of hero content.

### Before

127 0 0 1_3005_landing-page_goals(iPad Pro) (2)

### After

127 0 0 1_3005_landing-page_goals(iPad Pro) (1)

### DIFF
--- a/app/assets/stylesheets/views/_landing_page/hero.scss
+++ b/app/assets/stylesheets/views/_landing_page/hero.scss
@@ -70,7 +70,6 @@ $large-desktop-height: 400px;
       display: flex;
       min-height: $desktop-height;
       margin-top: -$desktop-height;
-      padding: govuk-spacing(6) 0;
     }
   }
 
--- a/app/assets/stylesheets/views/_landing_page/hero.scss
+++ b/app/assets/stylesheets/views/_landing_page/hero.scss
@@ -1,6 +1,5 @@
 @import "govuk_publishing_components/individual_component_support";
 
-$mobile-height: 427px;
 $desktop-height: 400px;
 $large-desktop-height: 400px;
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/views/_landing_page/hero.scss
```diff
@@ -52,23 +51,18 @@ $large-desktop-height: 400px;
...
-    min-height: $mobile-height;
-  }
-
   .app-b-hero__textbox-wrapper {
     @include govuk-media-query($from: desktop) {
       box-sizing: border-box;
       align-items: center;
       display: flex;
-      min-height: $desktop-height;
+      height: $desktop-height;
```
andysellick, line 65: I'm not 100% sure what the problem being fixed here is but this change introduces an old problem, which is that the text box can get trapped in the wrapper. Before:

Screenshot 2024-11-20 at 15 11 19

With this change:

Screenshot 2024-11-20 at 15 11 28

jon-kirwan: Thanks for the review and good catch (I'd forgotten about that issue). The problem I'm trying to fix is that hero content seems to be slightly misaligned. I now think the cause is too much padding in the `app-b-hero__textbox-wrapper` container, which causes the hero content to overflow slightly

Screenshot 2024-11-20 at 16 21 29

I think reducing the padding to 20px should resolve the issue

Screenshot 2024-11-20 at 16 21 14

andysellick: I think you could take the padding off entirely for this variant - it doesn't seem to do anything.

andysellick (approved):

PR #4427, opened 2024-11-20
Allow main navigation button text to be edited

AshGDS:

## What / Why
- Makes the first `text` key in `navigation_groups` become the text for the menu button
- I was trying to make the button text editable with minimal backend changes, but appreciate this may not be the best solution.

[Trello card](https://trello.com/c/JZcLx7JF/165-menu-tweaks)

### DIFF
--- a/app/views/landing_page/blocks/_main_navigation.html.erb
+++ b/app/views/landing_page/blocks/_main_navigation.html.erb
@@ -1,18 +1,20 @@
 <%
   add_view_stylesheet("landing_page/main-navigation")
+  menu_items = contents_list(request.path, block.links)
+  button_label = menu_items.shift()
 %>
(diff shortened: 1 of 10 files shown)


leenagupte (commented):
Does the button label have to be in the links list if it's only being used once?

I see that all of the existing "heading" elements have been renamed to "text", so perhaps the button label can be the only element in the block called "heading"? 

That way rather than having to use `shift` and setting `<%= button_label[:text] %>` you could use `<%= block.heading %>` directly.

maxgds:
Richard did this, which I saw before this one: https://github.com/alphagov/frontend/pull/4433 - sorry @AshGDS!

AshGDS:
No worries @maxgds 👍

PR #4430, opened 2024-11-20
Use breadcrumbs from YAML instead of contextual breadcrumbs

richardTowers:

## What

Now that our landing pages are part of the taxonomy, it's reasonably clear that the contextual breadcrumbs won't work well enough for our stakeholders. In particular, showing the "Government efficiency and transparency" taxon is too noisy / confusing.

We'll want to have full control of the breadcrumbs, so let's source them from the YAML.

I think we'll need to set these on every page, rather than using the extends hack, because the individual missions pages will need one more level of depth compared to the missions homepage.

I've also removed the landing_page.yaml file, as it wasn't well formed YAML and it was duplicating the one in spec/fixtures.

## Why

[Trello card](https://trello.com/c/V3iZuTt0/164-customise-or-remove-breadcrumbs)

## Screenshots

image


KludgeKML (approved):
LGTM

PR #4431, opened 2024-11-20
[WIP ] Populate document list from a taxon

leenagupte:

## Why

[Trello card?](url)


app/models/landing_page/block/document_list.rb
```diff
@@ -1,15 +1,73 @@
...
+      @taxon_base_path = data["taxon_base_path"]
+    end
+
     def full_width?
       false
     end
 
     def items
-      hardcoded_items
+      items_from_taxon.presence || hardcoded_items
```
richardTowers, line 22: This will return the hardcoded items if there are legitimagely no items in the taxon (i.e. taxon_base_path is configured, the taxon exits, but there are no search results).

Is that what we want?

richardTowers: From reading the tests and subsequent commits: yes

I was confused by the term "hardcoded_items" - I'll do a separate comment to explain myself there.

app/models/landing_page/block/document_list.rb
```diff
@@ -5,6 +5,12 @@ def full_width?
     end
 
     def items
+      hardcoded_items
+    end
+
+  private
+
+    def hardcoded_items
```
richardTowers, line 71: I don't like this method name - in this context I expected "hardcoded" to mean "hardcoded in the frontend application", but it actually means "specified explicitly in the block configured in the content item".

This confused me in the next commit, because I thought we were falling back to a hardcoded list, when actually we're falling back to the items in the block.

Maybe `explicitly_configured_items` would be better?

richardTowers (approved):

app/models/landing_page/block/document_list.rb
leenagupte: Yes, I thought it would be prudent to have a "fallback" option just in case.

richardTowers (approved):

PR #4432, opened 2024-11-21
Do not cache content items in content item loader

richardTowers:
# What / Why

The @cache field was being set on the ContentItemLoader class object, not on an instance of it.

The class object will be created once per process, and reused by every request made to that process.

Because our web server (puma) is multithreaded, each process will handle many HTTP requests. In practice, we set WEB_CONCURRENCY=2 and RAILS_MAX_THREADS=5 by default, and run between 2 and 4 replicas.

This combination means each instance of frontend has two content item caches (one for each process), and since there's no mechanism to clear the caches, once they're set there's no way to update them (other than restarting the process, which we only do during deployments / scaling events).

This leads to very confusing behaviour. Content items don't update as expected, but even the same frontend pod can have two different versions of the content item cached. This means that page refreshes can flip flop between multiple versions of the content item.

We could look at using the thread ID or something as a cache key (so that we don't reuse the cache across requests), or using a cache like ActiveSupport::Cache::MemoryStore which supports expiry.

However, I think doing any kind of caching of content items is optimisation we don't strictly need, and in the spirit of making the smallest change I can to fix the behaviour, initially I'm just removing the cache.

[Trello card](https://trello.com/c/FcLNTLWU/167-fix-buggy-caching-of-drafts)
[Slack thread](https://gds.slack.com/archives/C07Q150R69G/p1732115521513839)

### DIFF
--- a/lib/content_item_loader.rb
+++ b/lib/content_item_loader.rb
@@ -3,25 +3,21 @@ require "ostruct"
 class ContentItemLoader
   LOCAL_ITEMS_PATH = "lib/data/local-content-items".freeze
 
-  @cache = {}
-
   class << self
-    attr_reader :cache
-
     def load(base_path)
-      cache[base_path] ||= if use_local_file? && File.exist?(yaml_filename(base_path))
-                             Rails.logger.debug("Loading content item #{base_path} from #{yaml_filename(base_path)}")
-                             load_yaml_file(base_path)
-                           elsif use_local_file? && File.exist?(json_filename(base_path))
-                             Rails.logger.debug("Loading content item #{base_path} from #{json_filename(base_path)}")
-                             load_json_file(base_path)
-                           else
-                             begin
-                               GdsApi.content_store.content_item(base_path)
-                             rescue GdsApi::HTTPErrorResponse, GdsApi::InvalidUrl => e
-                               e
-                             end
-                           end
+      if use_local_file? && File.exist?(yaml_filename(base_path))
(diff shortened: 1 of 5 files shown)


richardTowers:
This issue looks like it was introduced here - https://github.com/alphagov/frontend/pull/4366/files#diff-8f720467cd4cfd6b9a611598b9491f0ffa8582cd2d260b733a36e16c0179bfd0 - and that prior to that change there was no caching of content items, other than in `request[:content_item]`

kevindew (approved):
This looks good and makes sense. 

Does seem danger having it as a class ivar. I assume the hope was for a cache that was only of the duration of a request so we don't query the content store twice during a request. I imagine the way we want to achieve this is to use a cache that is set up as part of rack middleware and lasts for the duration of the request, I believe this is how Rails achieves some of it's singleton interfaces.

Edit: I thought I knew of a prior art of this and found it here: https://github.com/alphagov/content-publisher/blob/8829183a7f4dec74995e7667344ae40f2bddef2a/config/initializers/bulk_data_rack_middleware.rb#L15-L24

kevindew:
Nice catch by the way - I imagine this was quite the head scratcher

richardTowers:
I suspect we're also accidentally overloading the `Class.load` method by doing `class << self; def load` which is probably inadvisable, but I didn't want to go too far down the rabbit hole, so this is just fixing the cache issue.

PR #4434, opened 2024-11-21
Rework featured block

JamesCGDS:

## What
Rework feature block.

## Why
[Trello card](https://trello.com/c/3qI5JERP/152-rework-featured-block)

### Before (featured block with image - desktop)
image

### After (featured block with image - desktop)
image

### Before (featured block with image - mobile)
image

### After (featured block with image - mobile)
image

### Newly added (featured block without image - desktop)
image

### Newly added (featured block without image - mobile)
image

### DIFF
--- a/app/assets/stylesheets/views/_landing_page/featured.scss
+++ b/app/assets/stylesheets/views/_landing_page/featured.scss
@@ -10,6 +10,14 @@
 }
 
 .featured__child--content {
+  width: 66.666%;
+
+  @include govuk-media-query($until: desktop) {
+    width: 100%;
+  }
+}
+
+.featured__child--content-with-image {
   flex: 1 1 33.333%;
   padding: govuk-spacing(6);
   box-sizing: border-box;
--- a/app/assets/stylesheets/views/_landing_page/featured.scss
+++ b/app/assets/stylesheets/views/_landing_page/featured.scss
@@ -18,11 +18,10 @@
 
 .featured__child--content-with-image {
   flex: 1 1 33.333%;
-  padding: 0 govuk-spacing(6) govuk-spacing(6) 0;
   box-sizing: border-box;
 
-  @include govuk-media-query($until: desktop) {
-    padding-bottom: 0;
+  @include govuk-media-query($from: desktop) {
+    padding: 0 govuk-spacing(6) govuk-spacing(6) 0;
   }
 }
 
--- a/app/assets/stylesheets/views/_landing_page/featured.scss
+++ b/app/assets/stylesheets/views/_landing_page/featured.scss
@@ -2,7 +2,6 @@
 
 .featured {
   display: flex;
-  background: govuk-colour("dark-blue");
 
   @include govuk-media-query($until: desktop) {
     flex-direction: column;
@@ -19,8 +18,12 @@
 
 .featured__child--content-with-image {
   flex: 1 1 33.333%;
-  padding: govuk-spacing(6);
+  padding: 0 govuk-spacing(6) govuk-spacing(6) 0;
   box-sizing: border-box;
+
+  @include govuk-media-query($until: desktop) {
+    padding-bottom: 0;
+  }
 }
 
 .featured__child--image {
--- a/app/views/landing_page/blocks/_featured.html.erb
+++ b/app/views/landing_page/blocks/_featured.html.erb
@@ -7,7 +7,10 @@
 <div class="featured">
   <%= tag.div class: featured_child_content_classes do %>
     <% block.featured_content.each do |subblock| %>
-      <% options = { inverse: true } %>
+      <%
+        options = {}
+        options[:position] = "top" if subblock.type == "heading"
+      %>
       <%= render_block(subblock, options:) %>
     <% end %>
   <% end %>
--- a/app/views/landing_page/blocks/_featured.html.erb
+++ b/app/views/landing_page/blocks/_featured.html.erb
@@ -1,19 +1,24 @@
 <%
   add_view_stylesheet("landing_page/featured")
+  featured_child_content_classes = "featured_child"
+  block.image ? featured_child_content_classes << " featured__child--content-with-image" : featured_child_content_classes << " featured__child--content"
 %>
+
 <div class="featured">
-  <div class="featured__child featured__child--content">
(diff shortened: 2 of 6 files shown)


lib/data/landing_page_content_items/homepage.yaml
```diff
@@ -81,10 +81,12 @@ blocks:
     blocks:
       - type: heading
         content: Our tasks
-        path: /landing-page/tasks
       - type: govspeak
         content: |
           <p>Lorem ipsum dolor sit amet. In voluptas dolorum vel veniam nisi et voluptate dolores id voluptatem distinctio. Et quia accusantium At ducimus quis aut voluptates iusto aut esse suscipit.</p>
+      - type: govspeak
+        content: |
+          <p><a href="/landing-page/tasks" class="govuk-link govuk-!-font-weight-bold">Find out more about our tasks</a></p>
```
jon-kirwan, line 89: A small suggestion: we can remove the `govuk-link` formatting since it's within a Govspeak component and already inherits the link style formatting.

app/assets/stylesheets/views/_landing_page/featured.scss
```diff
@@ -2,17 +2,28 @@
...
-  padding: govuk-spacing(6);
+  padding: 0 govuk-spacing(6) govuk-spacing(6) 0;
   box-sizing: border-box;
+
+  @include govuk-media-query($until: desktop) {
+    padding-bottom: 0;
+  }
 }
 
 .featured__child--image {
```
jon-kirwan, line 29: I see that the image scales within the container, and any overflow is clipped. Just wanted to confirm if that's okay or if we want the image to maintain its aspect ratio instead?

app/views/landing_page/blocks/_featured.html.erb
```diff
@@ -1,19 +1,27 @@
...
-    <% end %>
-  </div>
+  <% end %>
+  <% if block.image %>
+    <div class="featured__child featured__child--image">
+      <%= picture_tag do %>
+        <%= tag.source srcset: "#{image_path(block.image.sources.desktop)}, #{image_path(block.image.sources.desktop_2x)} 2x", media: "(min-width: 769px)" %>
+        <%= tag.source srcset: "#{image_path(block.image.sources.tablet)}, #{image_path(block.image.sources.tablet_2x)} 2x", media: "(min-width: 641px) and (max-width: 768px)" %>
+        <%= tag.source srcset: "#{image_path(block.image.sources.mobile)}, #{image_path(block.image.sources.mobile_2x)} 2x", media: "(max-width: 640px)" %>
+        <%= image_tag(block.image.sources.desktop, alt: block.image.alt, class: "featured__image") %>
```
jon-kirwan, line 23: Do these images need alt text, or are they decorative like the hero block? If they're decorative, we should set a default empty string for the `alt` attribute, such as `image_alt_text = image.alt || ""`.

lib/data/landing_page_content_items/homepage.yaml
jon-kirwan: There's quite a bit of spacing between the two Govspeak blocks (due to the `govuk-!-margin-bottom-9` spacing). Maybe the text and link could be combined into a single block?

jon-kirwan (commented):
Looking good! I just asked one question and added a few minor style related comments

app/assets/stylesheets/views/_landing_page/featured.scss
```diff
@@ -2,17 +2,28 @@
...
+
+  @include govuk-media-query($until: desktop) {
+    width: 100%;
+  }
+}
+
+.featured__child--content-with-image {
   flex: 1 1 33.333%;
-  padding: govuk-spacing(6);
+  padding: 0 govuk-spacing(6) govuk-spacing(6) 0;
```
jon-kirwan, line 21: I think this causes excess padding when viewed on mobile.

Screenshot 2024-11-21 at 13 39 30

We might need to set it like `padding: 0 0 30px;` initially, and then use a (desktop?) media query to add the `padding-right`.

lib/data/landing_page_content_items/homepage.yaml
AshGDS: You can also override the margin bottom if needed:

https://github.com/alphagov/frontend/blob/7b9dd43935d12345fb87bbcc122a5a04b1a26c3e/app/assets/stylesheets/views/_landing_page.scss#L47-L50

AshGDS (commented):
Looks good to me as well, apart from the stuff Jon has already mentioned 👍

lib/data/landing_page_content_items/homepage.yaml
JamesCGDS: Great spots - I've removed `govuk-link` although I've kept it in a separate govspeak block since it looks like the designs call for a bit more spacing between the content and the link. Happy to discuss an alternative approach if preferable though!

app/assets/stylesheets/views/_landing_page/featured.scss
JamesCGDS: Good question, I'm not sure - it's looking like it currently works like that on `main` (i.e. gets clipped) atm, I can check with folks to confirm

app/views/landing_page/blocks/_featured.html.erb
JamesCGDS: Have pinged you a quick message on Slack about this one

JamesCGDS: Thanks @jon-kirwan, have updated 👍 I've had a look at what happens when no alt text is provided and it renders `alt` rather than `alt=""`. At first I was a bit concerned that we'd need to have `alt=""` but it looks like just having [empty attribute syntax implicitly sets the value to ""](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2)

AshGDS (commented):
I forgot to mention - are you able to update the Featured block docs in `docs/building_blocks_for_flexible_content.md`? Thanks

JamesCGDS:
Thanks for the reviews @jon-kirwan @AshGDS - I've made those changes now 👍

jon-kirwan (approved):
Me too thanks @JamesCGDS

PR #4435, opened 2024-11-21
Add default empty alt text when no placeholder text provided

jon-kirwan:
## What

Add default empty alt text when no placeholder text provided

## Why

For accessibility, hero images are decorative and assigned null (empty) alt text using `alt=""`. However, if no placeholder text is provided in the YAML file, the `alt` attribute is currently omitted


richardTowers (approved):
Ah thanks for this. It will actually always be nil (when published from whitehall), so I guess we could also have hardcoded it to "". This is fine though.

PR #4438, opened 2024-11-21
Support draft CSVs

richardTowers:

## What

URI.open doesn't work on draft CSVs, because you need to provide authorisation headers of some kind to prove you're allowed to access the draft.

Instead, we need to use the asset manager API, as we do with CSV previews.

Unfortunately Whitehall doesn't provide the ID we need to call asset manager in a convenient form. Instead, we're parsing the URL and extracting it from there. There's a TODO in the code because it would be much nicer if whitehall provided the ID so we didn't have to do this parsing.

I've been strict with the format of the ID (24 hexadecimal digits) because I think it's better for the parsing to fail than to proceed with a bad ID and fail later with a bad request. This does mean that if the format of asset manager IDs changes this will break, but hopefully we will have updates Whitehall to provide the ID directly long before that happens.

## Why

[Trello card](https://trello.com/c/GHwrVbI7/169-investigate-whether-attached-csvs-work-correctly)


KludgeKML (approved):
LGTM

app/models/landing_page/block/statistics.rb
```diff
@@ -42,7 +42,14 @@ def opened_csv
     end
 
     def csv_from_url
-      CSV.parse(URI.parse(attachment.url).open, headers: true)
+      # TODO: asset_id is a 24 char hexadecimal number which doesn't appear elsewhere in the attachment hash.
+      #       We should get Whitehall to provide it rather than parsing the URL.
+      match = attachment.url.match(%r{/media/(?<asset_id>[[:xdigit:]]{24})/(?<filename>[^/]+[.]csv)\z})
+      raise "Unexpected URL format #{attachment.url} - cannot load CSV" unless match
```
leenagupte, line 48: If an error is raised does that mean that the page will fail?

At the moment the code in main has the chart fails silently. The page still loads, but omits the chart. Will that still happen here?

This PR is missing the most up to date code. The PR needs to be rebased to make sure that the tests still pass.

leenagupte (changes requested):
One comment. This PR needs to be rebased against main and double check that the statistics block fails silently and doesn't stop the rest of the page from loading.

app/models/landing_page/block/statistics.rb
KludgeKML: Tests still work after a rebase, it looks like.

I'm in two minds about the chart omission. On the one hand, it's nice and safe. On the other hand, it leads to a bit of a difficult to diagnose problem (if a chart is missing, what's causing that? missing attachment? broken attachment? chart block typo in YAML?) If safety is our priority, then we can catch this. Perhaps we should always catch blocks that raise an error in the BlockFactory and just safely drop them?

KludgeKML: We could add:
```    def opened_csv
      @opened_csv ||= attachment ? csv_from_url : csv_from_file
    rescue StandardError => e
      Rails.logger.error(e)
      nil
    end```

Comment (leenagupte) [reply app/models/landing_page/block/statistics.rb:60]: Sounds good.

Comment (leenagupte) [reply app/models/landing_page/block/statistics.rb:60]: The main thing is that the rest of the page loads and doesn't bomb.

File: spec/models/landing_page/block/statistics_spec.rb [hunk 2]
```diff
@@ -129,4 +129,14 @@
       expect(subject.rows).to eq([])
     end
   end
+
+  context "with an unparseable attachment URL" do
```
leenagupte, line 133: Minor: I think this should be inside the `describe #rows` block.

richardTowers (commented):
Keith's changes look good to me, and I think resolve Leena's comment. I can't approve because it's my PR, but I think we should be good to go now.

app/models/landing_page/block/statistics.rb
richardTowers: Looks like Keith's resolved this 👇🏻

leenagupte (approved):
There's one non-blocking in-line comment.

spec/models/landing_page/block/statistics_spec.rb
richardTowers: Sorted

PR #4440, opened 2024-11-21
Content item loader cache

KludgeKML:

## What

Re-add the cache that was removed in a previous fix. The original cache was too aggressive (it mistakenly relied on the ContentItemLoader being reloaded per request, but the threaded nature of the server in fact causes that not to happen, and the loader's cache was then preserved across requests causing instability). Reinstate it as an object-level cache, and attach that object to the request object (in its `.env` map) so that it's explicitly rebuilt for each request.

### DIFF
--- a/spec/unit/content_item_loader_spec.rb
+++ b/spec/unit/content_item_loader_spec.rb
@@ -1,12 +1,47 @@
 RSpec.describe ContentItemLoader do
   include ContentStoreHelpers
 
+  let(:subject) { described_class.new }
   let!(:item_request) { stub_content_store_has_item("/my-random-item") }
 
-  describe ".load" do
-    it "does not cache calls to the content store" do
-      ContentItemLoader.load("/my-random-item")
-      ContentItemLoader.load("/my-random-item")
+  describe ".for_request" do
+    it "returns a new object per request" do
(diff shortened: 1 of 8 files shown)


sihugh (approved):

spec/unit/content_item_loader_spec.rb
```diff
@@ -1,12 +1,38 @@
...
 
+  let(:subject) { described_class.new }
   let!(:item_request) { stub_content_store_has_item("/my-random-item") }
 
-  describe ".load" do
-    it "does not cache calls to the content store" do
-      ContentItemLoader.load("/my-random-item")
-      ContentItemLoader.load("/my-random-item")
+  describe ".for_request" do
+    it "returns a new object per request" do
```
richardTowers, line 8: Is it worth another test that `it "returns the same object for the same request"` or something?

richardTowers (approved):

spec/unit/content_item_loader_spec.rb
KludgeKML: It's very low cost to add / run the test, so yes. I'll add it in.

richardTowers (approved):

PR #4442, opened 2024-11-21
Landing page better columns

matthillco:

## What

Improved `columns_layout` block that uses CSS grid to create 2 or 3 columns. We now use a single container class for basic styling, with a `data-columns` attribute containing the user specified number of columns. This avoids the need for multiple class names. The old behaviour using design system rows/columns was removed. [Trello](https://trello.com/c/VGg9r3b6/151-create-missions-layout-blocks)

## Why

To support the new design for landing pages that use 2 or 3 column sections.

## Visuals

| Two columns | Three columns | Four columns |
| -------- | ------- | ------- |
| image | image | image |

### DIFF
--- a/app/views/landing_page/blocks/_columns_layout.html.erb
+++ b/app/views/landing_page/blocks/_columns_layout.html.erb
@@ -1,5 +1,8 @@
-<div class="govuk-grid-row">
-  <% block.columns.each do |column| %>
-    <div class="<%= column_class_for_equal_columns(block.columns.count) %>"><%= render "landing_page/blocks/#{column.type}", block: column %></div>
+<%
+  add_view_stylesheet("landing_page/columns_layout")
+%>
+<div class="app-b-columns-layout" data-columns="<%= block.data["columns"] || 3 %>">
+  <% block.blocks.each do |child| %>
+    <%= render "landing_page/blocks/#{child.type}", block: child %>
   <% end %>
 </div>
--- a/docs/building_blocks_for_flexible_content.md
+++ b/docs/building_blocks_for_flexible_content.md
@@ -440,12 +440,20 @@ This example uses the `blocks_container` for the element type in the second [two
 
 #### Columns layout
 
-Columns layout spreads its blocks out horizontally. Depending on the number of child blocks, each column will take all (1 block), 1/2 (2 blocks), 1/3 (3 blocks), or 1/4 (4 or more blocks) of the width. If there are more than 4 blocks, blocks will wrap around after every fourth block.
+Columns layout spreads its blocks out horizontally. The `columns` property determines how many columns will be in each row, and can accept the following values:
+- `columns: 2` Two columns, each taking up half of the available space.
+- `columns: 3` Three columns, each taking up one-third of the available space.
+- any other value for `columns` will have no effect: all child elements will render the full width of this container.
+
+If `columns` is missing entirely, three columns will be rendered by default.
 
 ##### Example
 
+In the following example, the first two `big_number` blocks will be side-by-side on the first row and the third block will drop down onto the next row, taking up the left hand column.
+
 ```yaml
 - type: columns_layout
+  columns: 2
   blocks:
   - type: big_number
     number: £75m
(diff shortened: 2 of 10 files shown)


app/views/landing_page/blocks/_columns_layout.html.erb
```diff
@@ -1,5 +1,8 @@
-<div class="govuk-grid-row">
-  <% block.columns.each do |column| %>
-    <div class="<%= column_class_for_equal_columns(block.columns.count) %>"><%= render "landing_page/blocks/#{column.type}", block: column %></div>
+<%
+  add_view_stylesheet("landing_page/columns_layout")
+%>
+<div class="app-b-columns-layout" data-column-count="<%= block.data["column_count"] || 3 %>">
```
leenagupte, line 4: I'd consider renaming `column_count`. I thought it meant the total number of columns, and that the number of configured columns had to match the total. 

Perhaps it should be called `number_of_columns_per_row`?

docs/building_blocks_for_flexible_content.md
```diff
@@ -418,12 +418,13 @@ This example uses the `blocks_container` for the element type in the second [two
 
 #### Columns layout
 
-Columns layout spreads its blocks out horizontally. Depending on the number of child blocks, each column will take all (1 block), 1/2 (2 blocks), 1/3 (3 blocks), or 1/4 (4 or more blocks) of the width. If there are more than 4 blocks, blocks will wrap around after every fourth block.
+Columns layout spreads its blocks out horizontally. Depending on the value of `column_count`, each column will take all (1 block), 1/2 (2 blocks), 1/3 (3 blocks), or 1/4 (4 or more blocks) of the width. If there are more than `column_count` blocks, blocks will wrap to a new row. `column_count` can be a value between 2-4 inclusive, any other value will render full-width columns. If `column_count` is missing entirely, it will default to 3.
```
leenagupte, line 421: Thanks for updating the docs! ⭐

leenagupte (commented):
One minor comment from the backend.

app/views/landing_page/blocks/_columns_layout.html.erb
KludgeKML: The key point here is that we separate columns from blocks - so this _is_ a column count. You don't have configured columns, you configure blocks that are layed out into columns.

matthillco: I can see both perspectives :) Another 'naming things' problem! I'm personally OK with `column_count` but can see that it's potentially confusing.  What about just `columns` or is that even more ambiguous?

docs/building_blocks_for_flexible_content.md
andysellick, line 421: This is a really nice solution and I think it works well. My only comment would be that I can't imagine a situation where we'd want four columns, unless I'm overlooking something?

matthillco: It's really just there for future proofing 'just in case'. Happy to remove it but I don't see the harm in keeping it as an option.

app/views/landing_page/blocks/_columns_layout.html.erb
andysellick: `number_of_columns`?

docs/building_blocks_for_flexible_content.md
andysellick: I think unless we have a need for it we should remove it. It looks fairly easy to extend if we do require it in future, but I don't want to have this as an option for laying out pages if we don't do this kind of layout.

app/views/landing_page/blocks/_columns_layout.html.erb
matthillco: I'd prefer to keep these shorter if possible. Easier to read. What's wrong with just `columns`?

matthillco: I think `columns` is fine, so unless anyone has an objection, I'll go with that.

docs/building_blocks_for_flexible_content.md
matthillco: OK, this only supports 2 or 3 columns now.

AshGDS (approved):
LGTM, nice use of the for loop 👍

PR #4447, opened 2024-11-22
Allow statistics block to set further options

andysellick:

## What / why
Expand the statistics block to allow it to use new options on the chart component, to set the format of axes to "date" to force correct axis labels on some of the data we're using. We're not using this option yet but we will shortly.

Also includes the `hide_legend` option, as legend may not be required on graphs with a single line.

## Visual changes
None.

Trello card: https://trello.com/c/cqG0SoNT

### DIFF
--- a/app/views/landing_page/blocks/_statistics.html.erb
+++ b/app/views/landing_page/blocks/_statistics.html.erb
@@ -18,4 +18,7 @@
   height: height,
   table_direction: "vertical",
   point_size: block.data["point_size"],
+  h_axis_format: block.data["x_axis_format"],
+  v_axis_format: block.data["y_axis_format"],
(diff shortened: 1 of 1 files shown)


app/views/landing_page/blocks/_statistics.html.erb
```diff
@@ -18,4 +18,7 @@
   height: height,
   table_direction: "vertical",
   point_size: block.data["point_size"],
+  h_axis_format: block.data["h_axis_format"],
+  v_axis_format: block.data["v_axis_format"],
```
leenagupte, line 22: I wonder if we should make these `x_axis_format` and `y_axis_format` to match `x_axis_label` and `y_axis_label`?

richardTowers (approved):

PR #4455, opened 2024-11-26
Fix non-format-constraint routes error behaviour

KludgeKML:
- Controllers derived from ContentItemController which are on fixed routes (ie which aren't format-constraint-based routes) can be routed to directly, bypassing the error constraint. This means that the content item loader might return an error (for instance, a forbidden if a draft page is accessed), and we're currently not checking for one. This change adds a raise if the response for the loader is a standard error, and adds a test for it.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -22,7 +22,10 @@ private
   end
 
   def content_item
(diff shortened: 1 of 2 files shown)


hannako (approved):
LTGM

PR #4457, opened 2024-11-26
Fallback to a default brand colour on blocks

AshGDS:

## What / Why
- Fallback to a black border when `theme_colour` is not passed to a block, or the value passed is incorrect. 
- Also removes a hardcoded brand style in the `hero` blocks.
- Previously, the styles would default to one of the brand colours, which made it more difficult to see when the theme colour value was incorrect.
- https://trello.com/c/gf86I9Kh/175-elements-with-branding-shouldnt-have-a-default

## Screenshots?
image


matthillco (approved):
Looks good. Nice use of the default value in the colours map.

PR #4473, opened 2024-11-28
Make image blocks responsive

richardTowers:

## What

The block documentation incorrectly stated that image blocks took a list of sources in the same way that hero / featured blocks do.

This is the format that whitehall is going to send, so frontend needs to support it.

Since whitehall is providing lots of different image resolutions, we may as well provide markup so that the browser can work out what to do.

I believe this combination of srcset / sizes attributes on images in new to GOV.UK. I'm following the MDN guidance here:

https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#resolution_switching_different_sizes

Translating this into this specific circumstance, what this HTML is saying is:

- We have six image sources, with intrinsic widths of 630, 1260, 708, 1416 etc.
- When the (max-width: 640px) breakpoint is true, we expect the element to be about 610px wide
- When the (max-width: 768px) breakpoint is true, we expect the element to be about 707px wide
- Otherwise, we expect the element to be about 630px wide

The browser will then use its knowledge of pixel density etc. to pick the best image source from the srcset.

This is a better approach than using the picture element when art-direction isn't a requirement, which for plain image blocks it isn't.

## Screenshots?

image


maxgds:
There's an option in the image card to support in a similar way, so not entirely new, for what it's worth: https://components.publishing.service.gov.uk/component-guide/image_card#with_sizes_attribute

AshGDS (approved):
Frontend looks OK to me so I can approve that aspect. I tested with all the placeholder image resolutions. I'm guessing now the benefit is reducing bandwidth for users, whereas previously this was used on our pages to account for different cropping/design requirements.

KludgeKML (approved):
LGTM

richardTowers:
> There's an option in the image card to support in a similar way, so not entirely new, for what it's worth: https://components.publishing.service.gov.uk/component-guide/image_card#with_sizes_attribute

Oh cool! I didn't know.

> I'm guessing now the benefit is reducing bandwidth for users, whereas previously this was used on our pages to account for different cropping/design requirements.

Yes @AshGDS, exactly that. And also supporting high-dpi screens (without serving 2x resolution images to users who don't need them).

PR #4474, opened 2024-11-28
Add theme_colour option to image block

AshGDS:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What / Why
- Adds the `theme_colour` option to the image blocks
- So we can put the border top colours on the images

### Before

image

### After

image

PR #4475, opened 2024-11-28
Make image blocks width: 100% by default

richardTowers:
Otherwise the image size will just be the file size, which at small resolutions can push the content outside the edge of the viewport leading to a horizontal scrollbar.

We might want some other themes in future (e.g. two-thirds width images), but we haven't got a use case for those yet, and it feels like taking the whole width of the container is a good default.

(Note: draft while we're building on top of https://github.com/alphagov/frontend/pull/4474)


KludgeKML (approved):
LGTM

PR #4476, opened 2024-11-28
Homepage: Finish search autocomplete AB test

csutter:
This removes the AB testing code and makes the new search autocomplete component permanent.


spec/system/homepage_spec.rb
```diff
@@ -12,31 +10,11 @@
...
-
-        expect(page).to have_css(".gem-c-search-with-autocomplete")
-        expect(page).to have_css("[data-source-url='http://www.dev.gov.uk/api/search/autocomplete.json']")
-      end
-    end
-
-    it "does not render the search autocomplete on the Z variant" do
-      with_variant(SearchAutocomplete: "Z") do
-        visit "/"
+  it "renders the search with autocomplete component with the correct source URL" do
```
leenagupte, line 13: Thanks for keeping this. 😍

leenagupte (approved):

spec/system/homepage_spec.rb
```diff
@@ -10,11 +10,23 @@
...
+      end
+    end
 
-    expect(page).to have_css(".gem-c-search-with-autocomplete")
-    expect(page).to have_css("[data-source-url='http://www.dev.gov.uk/api/search/autocomplete.json']")
+    it "does not render the search with autocomplete component when autocomplete is disabled" do
+      ClimateControl.modify GOVUK_DISABLE_SEARCH_AUTOCOMPLETE: "1" do
+        visit "/"
+
+        expect(page).to_not have_css(".gem-c-search-with-autocomplete")
```
kevindew, line 27: I imagine it'd be good to also assert that we have `.gem-c-search` to flag that it renders search component instead.

spec/system/homepage_spec.rb
```diff
@@ -10,11 +10,23 @@
     expect(page).not_to have_css(".homepage-inverse-header__title")
   end
 
-  it "renders the search with autocomplete component with the correct source URL" do
-    visit "/"
+  context "search autocomplete" do
```
kevindew, line 13: This should be failing our RSpec linting (turns out this project doesn't have our rspec rules installed 😢 https://github.com/alphagov/frontend/blob/63d532445eb2ebf15993f7de9850184acd231ea6/.rubocop.yml#L4) by not having a "when" or "with" at the start - https://rspec.rubystyle.guide/#context-descriptions

I'd imagine to make it more idiomatic RSpec you might want to refactor this two contexts:

context "when search with autocomplete is enabled"
context "when search with autocomplete is disabled"

kevindew (commented):
Code looks great - just a couple of comments on the specs.

kevindew (approved):

PR #4478, opened 2024-11-28
Use large headings for default hero and featured blocks

richardTowers:

## What

This will cause all headings which are children of default hero or featured blocks to get font_size = "l"

Previously this only happened to headings which were children of heros with position: top, which isn't configured everywhere. Now it happens for all heros with the default (non-middle_left) theme.

It's a bit of a breach of block isolation for hero / featured blocks to be saying "if there's a child heading, it should be large". It might be nicer just to provide the heading size in the YAML.

For now though, heading_font_size is a much clearer name than position, and the heading block no longer has to work out "if position is top then I should be large", which really felt like a lot more knowledge about what its parents get up to at night than any block should have to have.

I've also moved a bit of configuration into the model so it's tested, and a bit less verbose.

## Why

Noted as issue 6 in the list of things we spotted while testing the site.

### DIFF
--- a/app/models/landing_page/block/hero.rb
+++ b/app/models/landing_page/block/hero.rb
@@ -3,15 +3,18 @@ module LandingPage::Block
   HeroImage = Data.define(:alt, :sources)
 
   class Hero < Base
-    attr_reader :image, :hero_content
+    attr_reader :image, :hero_content, :theme, :theme_colour
 
     def initialize(block_hash, landing_page)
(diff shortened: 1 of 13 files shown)


richardTowers:
Moving to draft while I think about the text underneath the heading, which I think will need a different approach.

richardTowers:
Closing in favour of a slightly different approach.

PR #4484, opened 2024-11-29
Big hero headings

richardTowers:

## What

Updates the hero component so that in its default state, any child heading blocks will have large text.

It does this by passing params down to the heading component.

This removes the need for `position: top` on the hero block YAML, as we just use the fact that they have the default theme to infer they're at the top. That means we can replace the `position` param in headings with a `font_size` param, which is much clearer. The featured block may not need big headings, and we're not using it currently, so I'm just removing the option from that.

When Andy and I first worked on this, we discussed setting the heading level of these to `<h1>`, because they're basically the top heading of the page. I've decided not to do that by default though, because in theory a default hero block could be used further down the page, and we shouldn't have multiple h1s. If we want to use h1s for these, we should update the YAML to set `heading_level` on the headings.

## Why

[Trello card](https://trello.com/c/vxoiwFba/191-make-headings-and-text-in-main-hero-blocks-bigger)

## Screenshots

(Note: out of date - this PR no longer covers big govspeak, just big headings)

| Before | Big Hero Content | Big Hero 6 |
|------|------|------|
| image | image | image |

### DIFF
--- a/app/views/landing_page/blocks/_featured.html.erb
+++ b/app/views/landing_page/blocks/_featured.html.erb
@@ -9,7 +9,7 @@
     <% block.featured_content.each do |subblock| %>
       <%
         options = {}
-        options[:position] = "top" if subblock.type == "heading"
+        options[:font_size] = "l" if subblock.type == "heading"
       %>
       <%= render_block(subblock, options:) %>
     <% end %>
--- a/app/views/landing_page/blocks/_featured.html.erb
+++ b/app/views/landing_page/blocks/_featured.html.erb
@@ -7,11 +7,7 @@
 <div class="featured">
   <%= tag.div class: featured_child_content_classes do %>
     <% block.featured_content.each do |subblock| %>
-      <%
-        options = {}
-        options[:font_size] = "l" if subblock.type == "heading"
-      %>
-      <%= render_block(subblock, options:) %>
+      <%= render_block(subblock) %>
     <% end %>
   <% end %>
   <% if block.image %>
--- a/app/views/landing_page/blocks/_hero.html.erb
+++ b/app/views/landing_page/blocks/_hero.html.erb
@@ -2,16 +2,16 @@
 <%
   add_view_stylesheet("landing_page/hero")
 
-  hero_classes = %W[app-b-hero app-b-hero--#{block.data.fetch("theme", "default")}]
+  hero_classes = %W[app-b-hero app-b-hero--#{block.theme}]
 
   grid_class = "govuk-grid-column-two-thirds-from-desktop"
-  grid_class = "govuk-grid-column-one-third-from-desktop" if block.data["theme"] == "middle_left"
+  grid_class = "govuk-grid-column-one-third-from-desktop" if block.theme == "middle_left"
 
   hero_textbox_classes = %w[app-b-hero__textbox]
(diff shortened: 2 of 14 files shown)


app/views/landing_page/blocks/_featured.html.erb
```diff
@@ -9,7 +9,7 @@
     <% block.featured_content.each do |subblock| %>
       <%
         options = {}
-        options[:position] = "top" if subblock.type == "heading"
```
andysellick, line 12: Do we want to have a large heading size in the Featured block as well? I'm guessing this was already here and you're just updating it to match, but I think if the hero already has a large heading, we probably don't need another one?

app/views/landing_page/blocks/_hero.html.erb
```diff
@@ -26,9 +30,10 @@
       <%= content_tag("div", class: grid_class) do %>
         <%= content_tag("div", class: hero_textbox_classes) do %>
           <% block.hero_content.each_with_index do |subblock, index| %>
-            <% options = {
-              position: block.data["position"],
-              margin_bottom: (0 if block.hero_content.length - 1 == index) } %>
+            <%
+              options = { margin_bottom: (0 if block.hero_content.length - 1 == index) }
```
andysellick, line 34: Can you clarify what this line is doing?

app/views/landing_page/blocks/_featured.html.erb
richardTowers: Yeah, we're not actually using the featured block (I don't think?), so it's a bit of a moot point. I'd be happy to go either way on it. Shall I just remove the option?

app/views/landing_page/blocks/_hero.html.erb
richardTowers: It would be clearer if I did:

```ruby
is_final_subblock = block.hero_content.length - 1 == index
options = { margin_bottom: (0 if is_final_subblock) }
```

app/views/landing_page/blocks/_featured.html.erb
andysellick: Yeah let's remove it entirely from the featured block, thanks 👍

app/views/landing_page/blocks/_hero.html.erb
andysellick: Ah thanks I get it now. Maybe keep it as it is but add a comment? `# give last block zero margin bottom`?

richardTowers: Done in b25bcdb

richardTowers: Ah, sorry, I didn't see your comment until after I added that commit. I think it's better with the variable rather than a comment though - less likely to get left in place and become misleading if we change the code later.

app/views/landing_page/blocks/_featured.html.erb
richardTowers: Done in d44c687

andysellick (approved):
Looks fine to me, but maybe worth getting a pair of BE eyes on it as well.

richardTowers:
I'm just going to press ahead and click merge - sooner we deploy it, the more time we have to revert it if there's a problem.

PR #4485, opened 2024-11-29
Consolidate document list and press notice blocks

leenagupte:

## What

Consolidates the document list and press notice blocks and allow the title of the document list block to be configurable.

## Why

[Trello](https://trello.com/c/P1MWaE2G)

There are two competing blocks that do the almost the same thing. We don't actually need both blocks as only one is likely to ever be used on the page, and if more than one is needed we should add options to the existing block.

Document list has been chosen as the one that survives at is has the more generic name.

## Screenshots

Screenshot 2024-11-29 at 14 08 19


app/models/landing_page/block/document_list.rb
```diff
@@ -12,6 +12,7 @@ def initialize(block_hash, landing_page)
       super
 
       @taxon_base_path = data["taxon_base_path"]
+      @heading = data["heading"]
```
richardTowers, line 15: You could set the default value here, and save a line of ERB in the view:

```ruby
      @heading = data.fetch("heading", "Latest Updates")
```

(Also makes it easier to write a test for the default, if you're into that)

richardTowers (approved):
Approved with one non-blocking suggestion

app/models/landing_page/block/document_list.rb
leenagupte: Really, these defaults should be in the locales.

PR #4503, opened 2024-12-04
Make hero content optional

leenagupte:

## What

Make content in the hero blocks optional.

## Why

[Trello card](https://trello.com/c/TRJoQsvt)

We can't guarantee that there will always be content for the hero blocks. In those cases we should just display the full width image.

## Screenshots?

|Before|After|
|------|-----|
|Screenshot 2024-12-04 at 11 54 05|Screenshot 2024-12-04 at 11 53 43|


app/models/landing_page/block/hero.rb
```diff
@@ -3,21 +3,27 @@ module LandingPage::Block
...
     def full_width?
       true
     end
+
+    def hero_content
+      block_content = data.dig("hero_content", "blocks")
+      return if block_content.blank?
+
+      @hero_content = LandingPage::BlockFactory.build_all(block_content, landing_page)
+    end
```
KludgeKML, line 27: Probably not super-critical, but is there ever a chance that hero_content might be called more than once in the view - do we want to memoize this (or make it a private method that's called to set the old attribute during initialisation?)

KludgeKML (approved):
One non-blocking note, otherwise LGTM

app/models/landing_page/block/hero.rb
```diff
@@ -3,21 +3,27 @@ module LandingPage::Block
...
-    attr_reader :image, :hero_content, :theme, :theme_colour
+    attr_reader :image, :theme, :theme_colour
 
     def initialize(block_hash, landing_page)
       super
 
       alt, sources = data.fetch("image").values_at("alt", "sources")
       sources = HeroImageSources.new(**sources)
       @image = HeroImage.new(alt:, sources:)
-      @hero_content = LandingPage::BlockFactory.build_all(data.dig("hero_content", "blocks"), landing_page)
```
richardTowers, line 14: If you wanted to still do this in the constructor / with an attr_reader, you could use my new favourite ruby idiom:

```
@hero_content = data.dig("hero_content", "blocks")&.then do |blocks|
    LandingPage::BlockFactory.build_all(blocks, landing_page)
  end
```

(The & means it will be nil if the result of dig is nil, and the .then means it will be the result of the block if dig returns something other than nil)

app/models/landing_page/block/hero.rb
leenagupte: Memoizing is a good call. I'll update and fix up.

app/models/landing_page/block/hero.rb
richardTowers: Although I guess this doesn't cover the case where `.dig("hero_content", "blocks")` returns `[]`.

If you wanted to do that you'd have to do:

```
@hero_content = data.dig("hero_content", "blocks").then do |blocks|
    return if blocks.blank?
    
    LandingPage::BlockFactory.build_all(blocks, landing_page)
  end
```

(Basically like you have in your method below)

leenagupte: I'm not a big fan of multi-line assignments in the constructor. I'd put it in a method.

AshGDS (approved):
Frontend looks OK to me 👍

KludgeKML (approved):
LGTM

PR #4504, opened 2024-12-04
CSS for large govspeak in default hero images

richardTowers:

## What

We're overriding the rules set in markdown-typography in govuk_publishing_components for the tags there's a realistinc chance we might use.

I haven't overridden the styles for h1s or h4-6s, because they're not likely to be used in hero_content.

Note I've used govuk-font-size instead of govuk-font because it's only the size we're changing - we don't need to re-include the CSS for bold etc.

## Why

[Trello card](https://trello.com/c/vxoiwFba/191-make-headings-and-text-in-main-hero-blocks-bigger)

## How

By the awesome power of CSS

## Screenshots?

image


hannako:
@richardTowers Could this PR be closed?

richardTowers:
Yeah, we can reopen it if the fact that text size in hero content doesn't exactly match the designs is a problem we want to solve.

PR #4517, opened 2024-12-10
Replace HomepagePresenter with a model

leenagupte:

## What

Remove HomepagePresenter and move the code to a Homepage model

## Why

## How

In the HomepagePresenter all of the methods look like they could be moved to a model. The only exception was the call to a locale file to get the hardcoded popular links.
Normally a call like this should be in a helper or presenter, however in this case the locale is being used as data. Ideally data like this should either  be removed if it's no longer needed or links moved to a data file and a lookup to content store used to get the title of the link. 
The second and third commits experiment with removing this hard-coded data.

## Screenshots?


leenagupte:
@hannako It would be great if you could review this. Especially commits 2 and 3, which can easily be dropped if there's still a need for the hard-coded links. The more important aim is the moving of Homepage model code to a model.

app/controllers/homepage_controller.rb
```diff
@@ -19,7 +19,7 @@ def search_component
   helper_method :search_component
 
   def publication_class
-    HomepagePresenter
```
KludgeKML, line 22: I don't think we should be putting this into publication class. Can we not update the views so that they don't use publication (since this is the content item, they can use that).

KludgeKML:
I like this, but I think we should start as we mean to go on (not using the publication structure).

app/controllers/homepage_controller.rb
leenagupte: That's a good shout. It turns out that only the popular links partial was calling publication and the homepage doesn't need to rely on it for a breadcrumb.

I've pushed up another commit to remove the `publication_class`

leenagupte:
@KludgeKML Thanks for your review. I've pushed up another two commits that should resolve your comment.

KludgeKML (approved):
LGTM

app/controllers/homepage_controller.rb
KludgeKML: Nice

spec/support/content_store_helpers.rb
```diff
@@ -48,4 +48,15 @@ def content_store_throws_exception_for(path, exception)
   def honours_content_store_ttl
     expect(response.headers["Cache-Control"]).to eq("max-age=#{15.minutes.to_i}, public")
   end
+
+  def stub_homepage_content_item(links: [])
```
hannako, line 52: Minor not blocking:

I can't imagine why we'd ever have a test in this repo where we needed to check what happens if the content of popular links is a certain thing - there's no presentation logic that depends on the type of link etc. So maybe we should just use the existing `content_store_has_example_item`

hannako (approved):
Looks great Leena thank you! I made a couple of minor not blocking comments.

spec/system/sessions_spec.rb
```diff
@@ -9,7 +9,9 @@
   before do
     # We add a fake content item pointed at homepage, which allows us to check we've
     # been redirected
-    stub_content_store_has_item("/", schema: "special_route", links: {})
+    content_item = GovukSchemas::Example.find("homepage", example_name: "homepage_with_popular_links_on_govuk")
+    base_path = content_item.fetch("base_path")
+    stub_content_store_has_item(base_path, content_item)
```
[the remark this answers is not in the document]
leenagupte: I spent ages trying to wrangle with these session tests.

It looks like these tests rely on a route to exist in the routes file for the redirect to work: https://github.com/alphagov/frontend/commit/64e1b6c4b61b67b7876d6ac795139d1cfff815af

That means we can't just rely on the routing constraint, we'd have to specify something that exists, like the travel advice index page. In that case it doesn't really matter what's being explicitly stubbed. At that point I thought it was simplest to leave them pointing to the homepage.

spec/support/content_store_helpers.rb
leenagupte: The only test that really depends on the links existing is a request test, which thinking about it should probably be a system test.

hannako (approved):
LGTM!

PR #4521, opened 2024-12-11
Landing page themes

KludgeKML:

## What

Adds theme support to landing pages. 

## Why

Currently all landing pages have the Number 10 header. We don't want that on all landing pages, so we add a top-level `theme` element, which can be omitted or set to `default` to get a landing page without that header. The existing mission pages will have to be updated with the theme `prime-ministers-office-10-downing-street`, but since it's a top-level element we can do that once in the homepage and republish all, and the others will pick it up due to their `extends` values)

## Do not merge until
- [x] PfC mission page is published with the top level element `theme: prime-ministers-office-10-downing-street` added.
- [x] PfC pages are all republished (so that the child pages pick up the inherited element).

### Before

Screenshot 2024-12-12 at 14 01 25

### After (with no theme set, ie `default`) Note no No. 10 header, and the breadcrumbs are in the normal state rather than inverted.

Screenshot 2024-12-12 at 14 12 02

### After (with `prime-ministers-office-10-downing-street` theme set)

Screenshot 2024-12-12 at 14 12 46

### DIFF
--- a/app/views/landing_page/themes/_prime-ministers-office-10-downing-street.html.erb
+++ b/app/views/landing_page/themes/_prime-ministers-office-10-downing-street.html.erb
@@ -1,0 +1,35 @@
+<%
+  add_view_stylesheet("landing_page/themes/prime-ministers-office-10-downing-street")
+%>
+
+<div class="landing-page-header">
+  <div class="govuk-width-container">
+    <div class="landing-page-header__blue-bar">
+    </div>
+
+    <% if @content_item.breadcrumbs.present? %>
+      <%= render 'govuk_publishing_components/components/breadcrumbs', {
+        collapse_on_mobile: true,
+        inverse: true,
+        breadcrumbs: @content_item.breadcrumbs
+      } %>
+    <% end %>
+
+    <div class="landing-page-header__org">
+      <%= render "govuk_publishing_components/components/organisation_logo", {
+        organisation: {
+          name: sanitize("Prime Minister's Office<br>10 Downing Street"),
+          url: "/government/organisations/prime-ministers-office-10-downing-street",
+          brand: "prime-ministers-office-10-downing-street",
+          crest: "eo",
+        },
+        inverse: true,
+        inline: true,
+      } %>
+    </div>
+  </div>
+</div>
+
+<main class="landing-page" id="content">
+  <%= yield :landing_page_blocks %>
+</main>
--- a/docs/building_blocks_for_flexible_content.md
+++ b/docs/building_blocks_for_flexible_content.md
@@ -15,7 +15,13 @@ blocks:
 - type: block_type
 ```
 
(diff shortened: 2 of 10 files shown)


docs/building_blocks_for_flexible_content.md
```diff
@@ -15,7 +15,13 @@ blocks:
...
 ```
 
-This may include a top-level `navigation_groups` element for use by blocks (See [Navigation Groups](#navigation-groups)), or a top-level `extends` element to inherit configuration from another landing page (See [Extending a page (Whitehall Only)](#extending-a-page-whitehall-only)).
+It may optionally include other top-level elements:
+
+* a `breadcrumbs` element, an array of `title/href` pairs which will be used to create a breadcrumb list.
+* an `extends` element to inherit configuration from another landing page (See [Extending a page (Whitehall Only)](#extending-a-page-whitehall-only)).
+* an `image` element. This has one sub-element (`url`) which should be the url for an image to use with social media metadata.
+* a `navigation_groups` element for use by blocks (See [Navigation Groups](#navigation-groups))
+* a `theme` element which will set the general appearance of the page. Valid themes are `default` or `prime-ministers-office-10-downing-street`) - if no theme is specified, `default` will be used.
```
leenagupte, line 24: This update to the documentation is great!

I think though that a section on themes is needed. The themes described here apply to the whole page, whereas there are themes that only apply to a block, e.g. hero blocks.

leenagupte (commented):
This is looking great!

I'd like to see a couple of things before approving:

1. What the difference between the default theme and the prime ministers theme looks like. I think example pages with the default theme and prime ministers theme applied would be good
2. A frontend person to take a look and check that the prime ministers pages especially are still rendering as expected.

docs/building_blocks_for_flexible_content.md
KludgeKML: Good call, done!

KludgeKML:
> 1. What the difference between the default theme and the prime ministers theme looks like. I think example pages with the default theme and prime ministers theme applied would be good

I've added example images from integration in the description.

> 2. A frontend person to take a look and check that the prime ministers pages especially are still rendering as expected.

👍

app/views/landing_page/themes/_prime-ministers-office-10-downing-street.html.erb
```diff
@@ -0,0 +1,35 @@
+<%
+  add_view_stylesheet("landing_page/themes/prime-ministers-office-10-downing-street")
+%>
+
+<div class="landing-page-header">
+  <div class="govuk-width-container">
+    <div class="landing-page-header__blue-bar">
+    </div>
+
+      <% if @content_item.breadcrumbs.present? %>
```
andysellick, line 10: Nitpick: indentation too much?

KludgeKML: Yes, indentation too much. Thanks! Fixed.

leenagupte (approved):
Thanks for updating the documentation ⭐ 

LGTM! 👍

PR #4527, opened 2024-12-12
Add generic component block

KludgeKML:
- Lots of our blocks are wrappers around components, so adding a general component block allows us to use a wider variety of components without adding a lot more blocks (and potentially we could clean up existing blocks, but that should wait a little).
- The allowlist of components is intended to include all components that are not: Layouts, Banners, Form Elements, Page Furniture that does not work out of a specific place, or components marked as in an Experimental state.
- Adding this unlocks some possibilities for replacing history pages with landing_page document types, since it allows us to start using the Page title component used in those pages.

### DIFF
--- a/app/models/landing_page/block/component.rb
+++ b/app/models/landing_page/block/component.rb
@@ -1,0 +1,45 @@
+module LandingPage::Block
+  class Component < Base
+    ALLOWED_COMPONENTS = %w[
+      action_link
+      attachment
+      back_link
+      big_number
+      button
+      cards
+      chat_entry
+      contents_list
+      copy_to_clipboard
+      details
+      document_list
+      glance_metric
+      govspeak
+      heading
+      image_card
+      inset_text
+      inverse_header
+      lead_paragraph
+      list
+      notice
+      organisation_logo
+      page_title
+      previous_and_next_navigation
+      print_link
+      share_links
+      warning_text
+    ].freeze
+
+    attr_reader :component_name
+
+    def initialize(block_hash, landing_page)
+      super
+
+      @component_name = data["component_name"]
+      raise "Component #{component_name} is not in the allowed list of components for this block" unless ALLOWED_COMPONENTS.include?(component_name)
+    end
+
+    def component_attributes
+      data.except("type", "component_name").symbolize_keys
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/models/landing_page/block/component.rb
```diff
@@ -0,0 +1,45 @@
+module LandingPage::Block
+  class Component < Base
+    ALLOWED_COMPONENTS = %w[
```
andysellick, line 3: I think I'd start very small with this list and expand it later if needed, rather than start big and then whittle down. There's a lot of components in this list that I'd have questions around how/if they'd be used on these pages.

For a starting list I'd suggest only action_link, big_number, cards, document_list, govspeak, heading, lead_paragraph, share_links - basically the ones we've implemented already.

docs/building_blocks_for_flexible_content.md
```diff
@@ -61,6 +62,17 @@ A wrapper around the [Big number component](https://components.publishing.servic
   label: amount of money that looks big
 ```
 
+#### Component
+
+Renders whatever component from the publishing components gem is specified with the `component_name` value. All attributes _except_ `component_name` and `type` are passed directly to the component. Note that `component_name` should be the component in snake case (ie as it appears at the end of the url in the component guide). The example here will render identically to the Big Number example directly above, for instance.
```
andysellick, line 67: Nitpick: no need to emphasise `except`.

spec/models/landing_page/block/component_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  let(:blocks_hash) do
+    { "type" => "component",
+      "component_name" => "big_number",
+      "number" => 123,
+      "label" => "Number of things" }
+  end
+  let(:subject) { described_class.new(blocks_hash, build(:landing_page)) }
+
+  describe "allowed list of components" do
```
leenagupte, line 12: Should this be `describe "#component_name"` then have the following two nested contexts/tests:

* `context "when the component_name is in the allowed list"`
* * `it "sets the  components name"`
* `context "when the component_name is not in the allowed list"`
* * `it "raises an error"`

leenagupte (commented):
I like this idea in principle, but wonder if it will come undone if we find that we need to be able to add different styling depending on how the component is used. Or if it needs a specific theme applied, so we need to create a block for it, but now, as with big number, we have two different ways of calling the component.

I'm also concerned about how a CMS will "know" which components are allowed. Will the list of allowed components have to be maintained in two places?

KludgeKML:
Going to close this PR, not immediately necessary, and we're going to review flexible pages.

PR #4528, opened 2024-12-12
Add simple-smart-answer model

leenagupte:

## What

Move modelling code from SimpleSmartAnswerPresenter to a SimpleSmartAnswer model.

## Why

## How

The code that was in the original SimpleSmartAnswerPresenter has been split between the SimpleSmartAnswerPresenter and the new SimpleSmartAnswer model.
SimpleSmartAnswerPresenter has also been updated so that it's now initialized with an instance of the model.

## Screenshots?

Example simple smart answer: https://govuk-frontend-app-pr-4528.herokuapp.com/pension-credit-calculator

### DIFF
--- a/app/controllers/simple_smart_answers_controller.rb
+++ b/app/controllers/simple_smart_answers_controller.rb
@@ -3,7 +3,9 @@ require "simple_smart_answers/flow"
 class SimpleSmartAnswersController < ContentItemsController
   include Cacheable
 
-  def show; end
+  def show
+    @presenter = SimpleSmartAnswerPresenter.new(content_item)
+  end
 
   def flow
     responses = params[:responses].to_s.split("/")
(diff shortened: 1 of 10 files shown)


app/models/content_item.rb
```diff
@@ -51,6 +51,10 @@ def available_translations
     translations.sort_by { |t| t["locale"] == I18n.default_locale.to_s ? "" : t["locale"] }
   end
 
+  def slug
```
KludgeKML, line 54: Hmm, I'm very suspicious of anything related to slug, because it's so variable between models. If it's currently only being used in SimpleSmartAnswers, I think I'd advise adding it to that model and then we can bring it back here when we have a few other things that use it and we're sure that it is worth making a base attribute.

leenagupte: Fair enough. 👍

leenagupte: @KludgeKML I've made the change and fixed up.

KludgeKML (approved):
LGTM

PR #4511, opened 2024-12-05
Fix travel advice breadcrumb rendering

leenagupte:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What's changed?

Rely on the application layout to render breadcrumbs

## Why?

This follows the same pattern as other pages in frontend

## Screenshots?

N/A - No visible changes.

PR #4531, opened 2024-12-13
Show a breadcrumb on the travel advice index page

leenagupte:

## What

Show a breadcrumb on the travel advice index page.

## Why

Follows on from: https://github.com/alphagov/frontend/pull/4511

It's not clear if this was purposefully left off, or was overlooked because the travel advice controller didn't set the `@publication` variable, however it is odd that the travel advice index page doesn't have a breadcrumb. It is the only page in the travel advice flow that doesn't have a breadcrumb. The other comparable page is the list of departments, and that page does have a breadcrumb.

## Screenshots

|Before|After|
|-------|-----|
|Screenshot 2024-12-13 at 17 20 20|Screenshot 2024-12-13 at 17 21 04|

### DIFF
--- a/app/controllers/travel_advice_controller.rb
+++ b/app/controllers/travel_advice_controller.rb
@@ -37,7 +37,7 @@ private
   # TODO: Controllers should provide a presenter or a publication.
   # These objects duplicate roles (see `presenter || @publication`) in views.
   def publication
-    content_item if country_page?
+    content_item
   end
 
   def content_item_path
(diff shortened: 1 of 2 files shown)


app/controllers/travel_advice_controller.rb
```diff
@@ -37,7 +37,7 @@ def show
   # TODO: Controllers should provide a presenter or a publication.
   # These objects duplicate roles (see `presenter || @publication`) in views.
   def publication
-    content_item if country_page?
```
KludgeKML, line 40: I'm still very suspicious about a publication method that returns a content_item, but I think that's a fix outside of the scope of this change.

KludgeKML (approved):
LGTM

PR #4157, opened 2024-07-12
Prepare Frontend to render popular tasks from the homepage content item

unoduetre:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

Mainstream Publishing Experience (MPE) is working on a homepage publishing tool. As part of this work they will be adding popular tasks to the existing homepage content_item.

This card is about updating Frontend so that it can read the links from the content item if they are present, and read them from the existing locale file if they are not.

## Why

To ensure that we are not blocking MPE. If we ship this work they will able to:

* Put their branch of publisher on integration
* Publish the homepage content item
* Check to see if the homepage on integration shows their new links.

At some point in the future we can remove the code that falls back to the locale file. But there’s no urgency on this.

[Trello card](https://trello.com/c/GmDR8nkf/2686-prepare-frontend-to-render-popular-tasks-from-the-homepage-content-item-m), [Jira issue NAV-12228](https://gov-uk.atlassian.net/browse/NAV-12228)

PR #4533, opened 2024-12-16
Update tests related to homepage's popular links so they use examples from publishing-api

unoduetre:

## What

While [https://github.com/alphagov/frontend/pull/4157](https://github.com/alphagov/frontend/pull/4157 "smartCard-inline")  was developed, the test examples in publishing-api were still not ready, so the tests were prepared using a different approach. The [example](https://github.com/alphagov/publishing-api/pull/2816/commits/e824b2380708eeb063454c9e42c8ba61f9b8f61c "‌") has now been added, so we need to update the tests to use it.

## Why

In order to keep all api examples in a single place and to allow the same examples to be shared between different applications, so they can be easily updated at the same time api is updated.

[Trello card](https://trello.com/c/rC6DUre7/3093-update-tests-related-to-homepages-popular-links-so-they-use-examples-from-publishing-api)

### DIFF
--- a/spec/requests/homepage_spec.rb
+++ b/spec/requests/homepage_spec.rb
@@ -2,7 +2,16 @@ RSpec.describe "Homepage" do
   include GovukAbTesting::RspecHelpers
 
   context "loading the homepage" do
-    before { stub_content_store_has_item("/", schema: "special_route", links: {}) }
+    let(:overrides) { {} }
+
+    before do
+      content_store_has_example_item(
+        "/",
+        schema: "homepage",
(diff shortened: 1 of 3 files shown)


catalinailie (approved):
Looks good to me ⭐

hannako:
@unoduetre 

I think this PR can be closed as the changes are about to be made as part of Leena's work [here](https://github.com/alphagov/frontend/pull/4517/files#diff-cd770bd872f91f591d19af78a3c27935c380da2be3645070ef2fa02ff0c55d7f)

PR #4534, opened 2024-12-16
Add GA4 tracking to some missions page blocks

AshGDS:
## What / Why
- Adds GA4 tracking to some missions page blocks as this was missing in the initial release
- Adds the link tracker for the Menu, Boxes and Org logo links
- Adds the event tracker for the Menu button
- See the [implementation guide](https://docs.google.com/document/d/1LGLRb6zMvJBjzMYPgLAcBWbS-GsKGUzLcumCsZgiUhM/edit?tab=t.0#heading=h.gtmw0pgvmi0n) if you'd like to verify the JSON values.
- I had to create a new container block for the boxes to give me a nice way to track them. This is because they're rendered separately from each other in the code, and for the purposes of GA4 indexes I needed to capture all of them in one `<div>` so the indexes could be set. Therefore there's a new `box_container` block. (I didn't want to add the tracking on to the existing `columns_layout` `<div>` that surrounds all of them already, as I don't think `columns_layout` should have any tracking on it.)
- Trello cards:
    - https://trello.com/c/EGZOPdxC/215-add-ga4-tracking-to-blocks
    - https://trello.com/c/lumWLkCq
    - https://trello.com/c/ccYUfF9Q
    - https://trello.com/c/EQPR2GYw

### DIFF
--- a/app/views/landing_page/blocks/_main_navigation.html.erb
+++ b/app/views/landing_page/blocks/_main_navigation.html.erb
@@ -5,19 +5,48 @@
   <h2 class="govuk-visually-hidden" id="main-navigation-heading">Secondary navigation menu</h2>
   <div class="main-nav__button-container main-nav__button-container--collapsed js-main-nav__button-container">
     <div class="govuk-width-container">
-      <button type="button" class="main-nav__button main-nav__button--no-js govuk-link govuk-link--no-underline" aria-expanded="false" aria-controls="main-nav-container">
-        <%= block.name || "Menu" %>
-      </button>
+      <% button_text = block.name || "Menu" %>
+      <%= tag.button button_text,
+        type: "button",
+        class: "main-nav__button main-nav__button--no-js govuk-link govuk-link--no-underline",
+        aria: { expanded: "false", controls: "main-nav-container" },
+        data: {
+          module: "ga4-event-tracker",
+          ga4_expandable: "",
+          ga4_event: {
+            event_name: "select_content",
+            index_section: 1,
+            index_section_count: 1,
+            type: "secondary header",
+            section: button_text
+          }.to_json
+        } %>
     </div>
   </div>
   <div class="main-nav__nav-container main-nav__nav-container--js-hidden govuk-width-container js-main-nav__nav-container" id="main-nav-container">
-    <nav aria-labelledby="main-navigation-heading">
-      <% contents_list(request.path, block.links).each do |link_group| %>
+    <nav aria-labelledby="main-navigation-heading" data-module="ga4-link-tracker">
+      <% ga4_index_section_count = block.links.length %>
+      <% contents_list(request.path, block.links).each_with_index do |link_group, nav_index| %>
+        <% nav_section_heading = link_group[:text] %>
         <h3 class="main-nav__heading govuk-heading-s"><%= link_group[:text] %></h3>
         <ul class="main-nav__list">
-          <% link_group[:links].each do |link| %>
+          <% ga4_index_total = link_group[:links].length %>
+          <% link_group[:links].each_with_index do |link, link_index| %>
           <li class="main-nav__list-item">
-            <%= link_to link[:text], link[:href], class: "govuk-link govuk-link--no-underline govuk-link--no-visited-state" %>
+            <%= link_to link[:text],
+              link[:href],
+              class: "govuk-link govuk-link--no-underline govuk-link--no-visited-state",
+              data: {
+                ga4_link: {
+                  event_name: "navigation",
+                  type: "secondary header",
+                  index_link: link_index + 1,
+                  index_total: ga4_index_total,
+                  index_section: nav_index + 1,
+                  index_section_count: ga4_index_section_count,
+                  section: nav_section_heading,
+                }.to_json
+              } %>
           </li>
           <% end %>
         </ul>
--- a/app/views/landing_page/themes/_prime-ministers-office-10-downing-street.html.erb
+++ b/app/views/landing_page/themes/_prime-ministers-office-10-downing-street.html.erb
@@ -25,6 +25,15 @@
         },
         inverse: true,
         inline: true,
+        data_attributes: {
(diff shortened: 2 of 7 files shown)


app/views/landing_page/blocks/_box_container.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    ga4_set_indexes: "",
+    ga4_track_links_only: "",
+    ga4_link: {
+      event_name: "navigation",
+      type: "box",
+    }.to_json
+  }
+%>
+<%= tag.div data: data_attributes do %>
+  <% block.blocks.each do |child| %>
```
andysellick, line 13: Instead of creating a new block and wrapping the whole page in it, could these tracking attributes be applied to the show template? https://github.com/alphagov/frontend/blob/main/app/views/landing_page/show.html.erb#L25

app/views/landing_page/blocks/_main_navigation.html.erb
```diff
@@ -5,20 +5,59 @@
...
+      <% button_text = block.name || "Menu" %>
+      <%= tag.button button_text,
+        type: "button",
+        class: "app-b-main-nav__button app-b-main-nav__button--no-js govuk-link govuk-link--no-underline",
+        aria: { expanded: "false", controls: "app-b-main-nav-container" },
+        data: {
+          module: "ga4-event-tracker",
+          ga4_expandable: "",
+          ga4_event: {
+            event_name: "select content",
```
andysellick, line 17: Should this be `select_content`?

app/views/landing_page/blocks/_main_navigation.html.erb
```diff
@@ -5,20 +5,59 @@
...
+    <nav aria-labelledby="app-b-main-navigation-heading" data-module="ga4-link-tracker">
+      <%
+        ga4_index_section = 0
+        ga4_index_section_count = block.links.length
+      %>
       <% contents_list(request.path, block.links).each do |link_group| %>
-        <h3 class="app-b-main-nav__heading govuk-heading-s"><%= link_group[:text] %></h3>
+        <%
+          nav_section_heading = link_group[:text]
+          ga4_index_section = ga4_index_section + 1
```
andysellick, line 35: Rather than having variables to keep track of this value, would it be possible to use `each_with_index` instead? Haven't tried it out so if it's too complicated don't worry, just a thought.

andysellick (commented):
One question about the approach and a couple of small things.

app/views/landing_page/blocks/_box_container.html.erb
AshGDS: Ah no this is different - the box container I've created is only wrapped around all the block we called "box", not the whole page:

image

AshGDS:
Thanks @andysellick  - should be ready to re-review. Also had to rebase as the organisation logo's render statement moved to a new template.

app/views/landing_page/blocks/_box_container.html.erb
andysellick: My concern with this approach is that we have to add an extra layer of blocks just for tracking, which seems like it could be over complicating a publishing interface. Did you consider other options? Perhaps as an option on the columns_layout block?

AshGDS: Happy to move it to the columns layout 👍  - I think I was worried that adding tracking to a "layout" block would be confusing from an architectural perspective, but I think what you said makes sense. We'll need to pass `ga4_tracking: true` and `ga4_type: box` through on the column layout we want to track so that it doesn't affect anything else using the column layout, and so the tracking isn't tied to boxes specifically.

andysellick: I think that sounds sensible. It would be great if it could automatically figure out the kind of tracking it needs based on the blocks themselves (and likewise for any future tracking, so there's just one parameter to turn it on) but don't worry if that's too complex.

AshGDS:
Thanks @andysellick - should be ready to review again 👍

app/views/landing_page/blocks/_box_container.html.erb
AshGDS: Thanks, yeah that would be good - though sometimes the PAs use a different name to what we've called the block, which might complicate things!

app/views/landing_page/themes/_prime-ministers-office-10-downing-street.html.erb
```diff
@@ -25,6 +25,15 @@
         },
         inverse: true,
         inline: true,
+        data_attributes: {
+          module: "ga4-link-tracker",
+          ga4_track_links_only: "",
+          ga4_link: {
+            event_name: "navigation",
+            seciton: "Header",
```
andysellick, line 33: Typo: `seciton`

docs/building_blocks_for_flexible_content.md
```diff
@@ -382,6 +382,7 @@ Note that the `[Image: desktop.png]` syntax must be wrapped in quotes (`"[Image:
 Layout blocks are similar to compound blocks in that they contain nested blocks. They are used to arrange blocks on the page, and to make it easier to apply styling to the blocks.
 
 - [Blocks Container](#blocks-container)
+- [Box Container](#box-container)
```
andysellick, line 385: Presumably this documentation can be removed now?

andysellick (commented):
Looks good, and cleaner too. Just a couple of minor comments - also are there any tests that should be included here? Maybe one to check that if the `ga4_tracking` is applied to a block that the expected attributes appear?

app/views/landing_page/blocks/_box_container.html.erb
andysellick: Fair enough, what you've done now looks good 👍

AshGDS:
Thanks @andysellick - should be ready for another review (the 2 latest commits are the new changes)

andysellick (approved):

PR #4538, opened 2024-12-17
Hide side navigation on mobile

matthillco:

## What
Hide the side navigation on landing pages when viewing on mobile devices.

## Why
- The side navigation on mobile has large spacing above it, and the heading doesn’t look like a heading
- The content flow is also disrupted by the side navigation being right at the top of the content
[Trello](https://trello.com/c/AfLGkyEs/213-post-launch-design-change-hide-side-navigation-on-mobile)

## How
Note that the large space between the hero block and the text won't disappear with this change. That will need to be considered as part of the block spacing refactor.

## Screenshots?

| Before | After |
| -------- | ------- |
| image | image |


app/views/landing_page/blocks/_side_navigation.html.erb
```diff
@@ -1,6 +1,11 @@
+<%
+  add_view_stylesheet("landing_page/side-navigation")
+%>
+<div class="side-nav">
 <%= render "govuk_publishing_components/components/contents_list", {
```
AshGDS, line 5: This component has the component wrapper on it, so you should just be able to pass `classes: "side-nav"` through to the component, instead of adding another div element. 

https://components.publishing.service.gov.uk/component-guide/contents_list/#standard-options

app/assets/stylesheets/views/_landing_page/side-navigation.scss
```diff
@@ -0,0 +1,7 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.side-nav {
+  @include govuk-media-query($until: desktop) {
+    display: none;
+  }
+}
```
AshGDS, line 7: Nitpick: GitHub doesn't seem to like it when there's no empty line at the end of files, if you'd like to fix that.

AshGDS (approved):
Looks good to me, have added a couple minor comments 👍

app/views/landing_page/blocks/_side_navigation.html.erb
AshGDS: Ah yeah I forgot about that - ignore me!

PR #4539, opened 2024-12-17
Remove `app-b` prefixes from all components

matthillco:

## What
Remove the `app-b-` prefix from all components.

## Why
This class naming convention was introduced early in the project as a way of namespacing blocks, but it serves no purpose and we haven’t consistently applied it to blocks, so it should be removed.
[Trello](https://trello.com/c/NBsPCcRL/217-remove-instances-of-app-b-class-names)


spec/javascripts/unit/modules/main-navigation.spec.js
```diff
@@ -5,32 +5,32 @@ describe('Main Menu Block module', function () {
 
   beforeEach(function () {
     var DOM =
-      `<div class="app-b-main-nav" data-module="app-b-main-nav">
```
andysellick, line 8: Changes look fine, but the commit structure could be better - individually, each of these commits currently represents a breaking change. In future if the change warrants multiple commits I'd go block by block, but for simplicity now I'd suggest merging all the commits together.

andysellick (approved):

PR #4544, opened 2024-12-18
Previews check content type

KludgeKML:

## What

Return a 404 instead of the current broken page if a preview link is requested for a non-csv atttachment.

## Why

Although we don't offer preview links to non-csv pages in gov.uk, some are in google search. We should return 404 so that they don't get into search / behave correctly when queried.

https://trello.com/c/bOxuUfBN/51-previews-for-non-csv-files, [Jira issue NAV-15556](https://gov-uk.atlassian.net/browse/NAV-15556)

## How

Add a check for content_type after getting the attachment metadata, then return a cacheable_404 if the content_type isn't "text/csv"

### Before

Screenshot 2024-12-18 at 16 35 19

### After

Screenshot 2025-01-13 at 09 42 06


unoduetre (approved):
Looks good to me.

PR #4548, opened 2024-12-19
Add transaction model

leenagupte:

## What

Move modelling code from TransactionPresenter to a  Transaction model.

## Why

## How

The code that was in the original TransactionPresenter has been split between the TransactionPresenter and the new Transaction model.
TransactionPresenter has also been updated so that it's now initialized with an instance of the model.

## Screenshots?

Example transaction: https://govuk-frontend-app-pr-4548.herokuapp.com/register-to-vote


KludgeKML (approved):
LGTM!

PR #4562, opened 2025-01-06
Remove unneeded blocks from landing page

andysellick:

## What
Remove unneeded blocks from landing page layout. This is part of a larger piece of work that will remove all component specific blocks and replace them with the generic component block, but for now these two blocks (statistics and big number) can be safely removed as they are not in use.

## Why
Blocks are not needed for the landing page.

## Visual changes
None.

Trello card: https://trello.com/c/HvbNIYXA/218-remove-unnecessary-blocks


leenagupte (approved):
It's sad that the statistics block is going. Ah well.

PR #4567, opened 2025-01-07
Allow heading blocks to receive font-sizes from data

matthillco:

## What
Allow heading blocks to receive font-sizes from block config data. [Trello](https://trello.com/c/7h3VaqOI/214-post-launch-design-change-ensure-h2-and-h3-elements-look-different-from-each-other)

## Why
Headings appearing in landing pages have mismatched sizes. The H2 and H3 for example all appear at 27px, which is not our normal standard. Heading blocks therefore needed to accept a `font_size` parameter.

## Screenshots?

A before and after screenshot showing `font_size` parameters of "m" and "s" applied to the H2 and H3.

image

### DIFF
--- a/app/views/landing_page/blocks/_heading.html.erb
+++ b/app/views/landing_page/blocks/_heading.html.erb
@@ -1,6 +1,6 @@
 <%
   heading_level = block.data["heading_level"] || 2
-  font_size = options[:font_size]
+  font_size = block.data["font_size"] || options[:font_size]
   text = block.data["content"]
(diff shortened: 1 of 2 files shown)


richardTowers (approved):
Matt and I discussed whether defaulting to "m" was a good idea or not on slack. We decided not to do it, because changing the default behaviour would have an immediate effect on heading size when this is deployed. Instead, we'll be explicit about heading sizes in the block config, and might revisit changing the defaults in a subsequent PR.

PR #4571, opened 2025-01-08
Add licence transaction model

leenagupte:

## What

Move modelling code from LicenceTransactionPresenter to a  LicenceTransaction model.

## Why

## How

The functionality that was in the original LicenceTransactionPresenter has been moved to the new LicenceTransaction model and old LicenceTransactionPresenter has been removed.

## Screenshots?

Example licence transaction: https://govuk-frontend-app-pr-4571.herokuapp.com/find-licences/aircraft-maintenance-engineering-licence


KludgeKML (approved):
LGTM

PR #4573, opened 2025-01-09
Move speech document

sairamya93:

## What

To handle the rendering of Speech document type from government_frontend to frontend as part of [RFC 175]

## Why

https://trello.com/c/IKDY5W9u/371-move-document-type-speech-from-government-frontend-to-frontend, [Jira issue PNP-7351](https://gov-uk.atlassian.net/browse/PNP-7351)

## How

To add a scoped route for, a controller, a model and a view. Here we split up the methods in the original file into presenter, model and its related concerns, with the minimum functionality required to get Speech document working.

### Speaker without profile

Speech page in Government-frontend and frontend

Screenshot 2025-03-04 at 16 39 13
 
### Frontend

Screenshot 2025-01-09 at 10 39 43
Screenshot 2025-01-09 at 10 40 22
Screenshot 2025-01-09 at 10 40 53

### Government-Frontend

Screenshot 2025-01-09 at 10 44 31
Screenshot 2025-01-09 at 10 44 56
Screenshot 2025-01-09 at 10 45 59

### Metadata

Screenshot 2025-01-22 at 15 02 21


app/controllers/speech_controller.rb
KludgeKML, line 1: The indenting on this file looks a bit off (4 rather than 2 spaces) - rubocop is probably complaining about this?

app/models/speech.rb
deborahchua, line 1: I think this commit should come later once you've committed the concerns included here.

app/controllers/speech_controller.rb
```diff
@@ -0,0 +1,5 @@
+class SpeechController < ContentItemsController
+    def show
+        @presenter =  SpeechPresenter.new(@content_item)
```
deborahchua: The presenter and view hasn't been included yet, so if you were to load a speech page whilst on this commit it would error.
You could update this commit so that the `show` method is empty, and also include an empty view (app/views/speech/show.html.erb).

app/models/speech.rb
```diff
@@ -0,0 +1,8 @@
+class Speech < ContentItem
+  include Body
```
deborahchua: I think this commit should come later once you've committed the concerns included here.

app/models/concerns/news_image.rb
```diff
@@ -0,0 +1,23 @@
+module NewsImage
```
deborahchua, line 1: I think the concerns should be put in to separate commits and the corresponding tests included too e.g. for this file you can include https://github.com/alphagov/frontend/pull/4573/commits/70b58e793e21872674dac097e78959138b962816 in the same commit.

app/models/concerns/political.rb
```diff
@@ -0,0 +1,32 @@
+module Political
```
deborahchua, line 1: Similarly here, you can put this concern in a separate commit and combine it with https://github.com/alphagov/frontend/pull/4573/commits/5996012547f39f30f6bf34674e18656574efebb7

app/views/shared/_history_notice.html.erb
```diff
@@ -0,0 +1,8 @@
+<% if content_item.historically_political? %>
```
deborahchua, line 1: I think this shared partial should also be in a separate commit, which is then followed by the generation of the locale keys in https://github.com/alphagov/frontend/pull/4573/commits/c08cd5b92c591f63772f35078862ba2687e23b57

app/presenters/speech_presenter.rb
```diff
@@ -0,0 +1,41 @@
...
+      I18n.t("formats.speech.written_on")
+    else
+      I18n.t("formats.speech.delivered_on")
+    end
+  end
+
+  def delivered_on_metadata
+    "#{delivered_on}#{speech_type_explanation}"
+  end
+
```
deborahchua, line 13: Could you update the commit message if the [from](https://github.com/alphagov/government-frontend/blob/122ae292fa540059ee165a94f8129d873c5205d5/app/presenters/speech_presenter.rb#L22-L26) and [speaker_without_profile](https://github.com/alphagov/government-frontend/blob/122ae292fa540059ee165a94f8129d873c5205d5/app/presenters/speech_presenter.rb#L50-L52) methods are no longer needed or will live somewhere else, otherwise at first glance it looks like the methods were accidentally missed.

app/presenters/speech_presenter.rb
```diff
@@ -0,0 +1,41 @@
...
+  def important_metadata
+    { I18n.t("formats.speech.location") => location, delivery_type => delivered_on_metadata }
+  end
+
+  def view_context
+    @view_context = ApplicationController.new.view_context
+  end
+
+  def display_date(timestamp, format = "%-d %B %Y")
```
deborahchua, line 22: You can remove this method as the helper method is already available in [date_helper.rb](https://github.com/alphagov/frontend/blob/main/app/helpers/date_helper.rb)

app/views/speech/show.html.erb
```diff
@@ -0,0 +1,58 @@
+
+<% content_for :extra_headers do %>
+   <%= render "govuk_publishing_components/components/machine_readable_metadata", { content_item: @content_item.to_h, schema: :article } %>
+<% end %>
+
```
deborahchua, line 8: I think you need to include the `<main>` tag here, similar to https://github.com/alphagov/frontend/pull/4225/commits/c501ca1d9ec9c2263ea7c3f1903582bca2333f59#diff-2c4760c884914b406ca821f19956295c799447ffe9a1ae9e780322d1d0c256b4R9

app/views/speech/show.html.erb
```diff
@@ -0,0 +1,58 @@
...
+        average_title_length: "long",
+      } %>
+  </div>
+  <%= render 'shared/translations' %>
+  <div class="govuk-grid-column-two-thirds">
+    <%= render 'govuk_publishing_components/components/lead_paragraph', text: @content_item.description %>
+  </div>
+</div>
+
+<%= render 'shared/publisher_metadata_with_logo' %>
```
deborahchua: This has been renamed to `publisher_metadata` in https://github.com/alphagov/frontend/pull/4372/commits/c4dceb3647efdbd076884ab221886c7f51bd74d0

app/views/speech/show.html.erb
```diff
@@ -0,0 +1,58 @@
...
+<%= render 'shared/history_notice', content_item: @content_item %>
+<% if @content_item.withdrawn? %>
+  <%= render 'govuk_publishing_components/components/notice', @content_item.withdrawal_notice_component  %>
```
deborahchua: The way the withdrawal notice works has changed slightly, see https://github.com/alphagov/frontend/pull/4372/commits/6c5ed23d73e9103cb723e51d7e08bdfd29b64089#diff-de0d946d1ce699f8a4070b4d6a67cc44726747222c15044f2bc7e00e3a314743R31-R40

app/views/speech/show.html.erb
```diff
@@ -0,0 +1,58 @@
...
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <div class="content-bottom-margin">
+      <div class="responsive-bottom-margin">
+        <%= render 'components/important_metadata',
```
deborahchua: This is missing the condition which will only render the important metadata component if the information is available see - https://github.com/alphagov/government-frontend/blob/3c8d12788047ec9ebd5c7f044f4e3f3fce0480ca/app/views/content_items/speech.html.erb#L27
Does `inverse: true` and `margin_bottom: 0` no longer need to be included? https://github.com/alphagov/government-frontend/blob/3c8d12788047ec9ebd5c7f044f4e3f3fce0480ca/app/views/content_items/speech.html.erb#L30-L32

app/views/speech/show.html.erb
```diff
@@ -0,0 +1,58 @@
...
+
+        <%= render "govuk_publishing_components/components/govspeak", {
+          direction: page_text_direction,
+        } do %>
+          <%= raw(@content_item.body) %>
+        <% end %>
+      </div>
+
+      <%= render 'components/published_dates', {
+          published: display_date(@content_item.first_published_at),
```
deborahchua, line 72: The variable used here is now `first_public_at`, similar to https://github.com/alphagov/frontend/pull/4372/commits/6c5ed23d73e9103cb723e51d7e08bdfd29b64089#diff-de0d946d1ce699f8a4070b4d6a67cc44726747222c15044f2bc7e00e3a314743R60

app/views/components/_important_metadata.html.erb
```diff
@@ -0,0 +1,21 @@
+<% add_app_component_stylesheet("important-metadata") %>
```
deborahchua: This component was recently removed in https://github.com/alphagov/government-frontend/pull/3414 so you can drop this commit, and update the view

spec/components/important_metadata_spec.rb
```diff
@@ -37,18 +37,18 @@ def component_name
     expect(rendered).to have_css(".app-c-important-metadata dt", text: "Case type:")
```
deborahchua: I think this file crept into this commit during a rebase

spec/system/speech_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      visit "/government/speeches/speech"
+
+      within("[class*='metadata-column']") do
+        expect(page).to have_text("Department of Energy & Climate Change and The Rt Hon Andrea Leadsom MP")
+        expect(page).to have_link("Department of Energy", href: "/government/organisations/department-of-energy-climate-change")
+        expect(page).to have_link("The Rt Hon Andrea Leadsom MP", href: "/government/people/andrea-leadsom")
+        expect(page).to have_text("Published 8 March 2016")
+      end
+
+      # Check important metadata
```
deborahchua, line 30: This test will need to be updated to use the metadata block component instead

spec/system/speech_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    it "translated speech" do
+      visit "/government/speeches/speech-translated"
+
+      expect(page).to have_title("هذا القرار بداية لنهاية برنامج الأسلحة الكيميائية في ليبيا")
+      expect(page).to have_css(".gem-c-translation-nav")
+      expect(page).to have_text("بوريس جونسون: يمنح القرار الترخيص اللازم لمنظمة حظر الأسلحة الكيميائية لإزالة سلائف تلك الأسلحة من ليبيا تمهيدا لإتلافها في بلد ثالث. وبذلك نكون قد خففنا خطر وقوع هذه الأسلحة في أيدي الإرهابيين ")
+    end
+  end
```
deborahchua, line 59: I think an additional test can be added to test a withdrawn speech page, where we would expect the notice is included on the page.

app/views/speech/show.html.erb
deborahchua: I've just seen the tag has been added in a later commit

app/models/concerns/withdrawable.rb
```diff
@@ -9,5 +9,38 @@ def withdrawn?
       def withdrawal_notice
         content_store_hash["withdrawn_notice"]
       end
+
```
deborahchua: I think this accidentally crept in during a rebase.

leenagupte (commented):
Ramya, thanks for tagging me as a reviewer. Looking though the existing review comments, I think that @deborahchua  has it covered, and that if you action all of her recommendations, your PR will be in really good shape.

When it comes to constructing a PR, I was told to think of the PR as telling the story of the change. 
Some people like to add changes from "outside in" i.e. from the frontend (view) to the backend (model). 
I think as long as I can follow the steps you've taken, I don't mind whether you go "outside in" or "inside out" (backend to frontend). But I think something along the lines of "add a thing (model, method etc" then "use the thing" works well.

As Deb mentioned, when you're adding something brand new, as in this case, the tests should pass in every commit, and that should help to order the commits.

I think your code is there, so that's half the battle won. The question you should ask yourself is "will I be able to understand what I've changed 6 months down the road?"

config/locales/ar.yml
deborahchua, line 1: I'm not sure where the `oral_statement` key is being used, is it still needed?

sairamya93: It was there in one of the examples for Speeches in Content_schemas. One of the test cases (withdrawn_speech) was failing due to the missing translation for oral_statement and hence I included that.

app/models/speech.rb
```diff
@@ -0,0 +1,7 @@
+class Speech < ContentItem  
+  include Organisations
```
KludgeKML, line 2: These could probably be alphabetised.

app/presenters/speech_presenter.rb
```diff
@@ -0,0 +1,39 @@
+class SpeechPresenter < ContentItemPresenter
+  include DateHelper
+
+  def delivery_type
```
KludgeKML, line 4: You could potentially refactor this with a guard clause? (not a blocking comment, just a suggestion):

```
def delivery_type
  return I18n.t("formats.speech.written_on") if content_item.document_type == "authored_article"

  I18n.t("formats.speech.delivered_on")
end
```

KludgeKML: (Also this is only ever used in the important metadata method, you might make it private)

app/presenters/speech_presenter.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+
+  def delivered_on_metadata
+    "#{delivered_on}#{speech_type_explanation}"
+  end
+
+  def important_metadata
+    { I18n.t("formats.speech.location") => location, delivery_type => delivered_on_metadata }
+  end
```
KludgeKML, line 18: You could probably accept view_context as a parameter to this method (and pass it on to delivered_on_metadata) which would allow you to not have the view_context method. (What do you think, @leenagupte ?)

app/models/concerns/political.rb
```diff
@@ -0,0 +1,32 @@
...
+    def publishing_government
+      content_store_hash.dig(
+        "links", "government", 0, "title"
+      )
+    end
+  end
+
+private
+
+  def political?
```
KludgeKML, line 18: This construct is probably unnecessary - you're only looking for a true/false on whether the key exists, so it can be

`content_store_hash["details"].key?("political")`

app/presenters/speech_presenter.rb
leenagupte: Is `view_context` really needed? Can `tag.time` not be called in the view?

I think perhaps that the `delivered_on` method needs to be looked at again.

It's modelling `delivered_on_date` which could be moved to the model, and then calling a helper method on it, where as I the functionality in the `important_metadata` and `delivered_on_metadata` could just be moved to the view.

app/models/concerns/updatable.rb
```diff
@@ -0,0 +1,40 @@
...
+    def history
+      return [] unless any_updates?
+
+      reverse_chronological_change_history
+    end
+  end
+
+private
+
+  def any_updates?
```
KludgeKML, line 18: Again, you could reverse this with a guard clause:

```
def any_updates?
  return false unless public_updated_at && first_public_at

  Time.zone.parse(public_updated_at) != Time.zone.parse(first_public_at)
end
```

app/models/concerns/political.rb
sairamya93: Hmm we might have to check the value of the key "political" which could be either true/false? Like in 
https://www.gov.uk/api/content/government/speeches/orlando-frasers-speech-to-the-annual-public-meeting

I can modify as below?
content_store_hash["details"].key?("political") && content_store_hash.dig("details", "political")

KludgeKML: Oh, in that case just do the dig, I think? If political is there, you'll get true/false. If it's not, you'll get nil (which evaluates to false)

KludgeKML: Basically, I think my original key check was too simple, but you don't need to both check the key exists _and_ dig it. Dig will handle if the key doesn't exist.

sairamya93: Ah right. So the method will return false if key is not present or key is either nil/false. Otherwise it will return true.
I can just have this - 
content_store_hash.dig("details", "political")

app/presenters/speech_presenter.rb
KludgeKML: important_metadata is only called in one place (a view) - it could potentially be inlined there?

app/views/speech/show.html.erb
```diff
@@ -1,37 +1,61 @@
...
-        title: content_item.title,
-        context: I18n.t("csv_preview.document_type.#{content_item.document_type}", count: 1),
-        context_locale: t_locale_fallback("csv_preview.document_type.#{content_item.document_type}", count: 1),
-        average_title_length: "long",
-      } %>
-  </div>
-  <%= render 'shared/translations' %>
-  <div class="govuk-grid-column-two-thirds">
-    <%= render 'govuk_publishing_components/components/lead_paragraph', text: @content_item.description %>
+<main role="main" id="content" class="speech" <%= lang_attribute(@content_item.locale) %>>
```
deborahchua: The <main> tag has been globally included now (https://github.com/alphagov/frontend/pull/4414) so you don't need to include it here in the view.

app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
deborahchua, line 1: This file got renamed and was already merged in https://github.com/alphagov/frontend/pull/4372/commits/ad511e05ad4d44aea1a7b6c2aec510647502605e so I think don't think you need this commit anymore.

spec/support/concerns/political.rb
```diff
@@ -0,0 +1,23 @@
...
+    content_store_response["links"]["government"][0]["details"]["current"] = true
+    expect(described_class.new(content_store_response).historically_political?).to be(false)
+  end
+
+  it "knows it would be part of historical government if there is no government" do
+    content_store_response["links"]["government"][0]["details"]["current"] = nil
+    expect(described_class.new(content_store_response).historically_political?).to be(false)
+  end
+
+  it "knows it's publishing government" do
```
KludgeKML, line 19: typo (it's -> its)

spec/system/speech_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      expect(page).not_to have_css(".gem-c-single-page-notification-button")
+    end
+  end
+
+  context "when visiting a speech translation page" do
+    before do
+      content_store_has_example_item("/government/speeches/speech-translated", schema: :speech, example: "speech-translated")
+    end
+
+    it "translated speech" do
```
KludgeKML, line 49: Missing words? (`it "displays the translated speech" do`, maybe?)

spec/system/speech_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      expect(page).to have_text("بوريس جونسون: يمنح القرار الترخيص اللازم لمنظمة حظر الأسلحة الكيميائية لإزالة سلائف تلك الأسلحة من ليبيا تمهيدا لإتلافها في بلد ثالث. وبذلك نكون قد خففنا خطر وقوع هذه الأسلحة في أيدي الإرهابيين ")
+    end
+  end
+
+  context "when visiting a speech that is withdrawn" do
+    before do
+      content_store_has_example_item("/government/speeches/withdrawn-speech", schema: :speech, example: "withdrawn-speech")
+    end
+
+    it "withdrawn speech" do
```
KludgeKML, line 63: As above - (maybe `it "displays the withdrawn speech notice" do` ?)

app/views/speech/show.html.erb
```diff
@@ -1,37 +1,60 @@
+<% content_for :title do %>
```
KludgeKML, line 1: We could maybe squash this commit into "Add views for speeches"?

spec/support/concerns/political.rb
leenagupte: Is it a typo? "it knows it is publishing government" 🤔

app/assets/stylesheets/application.scss
```diff
@@ -65,3 +65,8 @@ $govuk-include-default-font-face: false;
 .bank-hols .govuk-panel {
   margin-bottom: govuk-spacing(6);
 }
+
```
KludgeKML, line 68: Ideally we'd add this in a view stylesheet rather than in application.scss. I'm also a bit suspicious of this as it's just used to set background for a component that should probably understand it's own background:

https://github.com/alphagov/frontend/pull/4573/commits/5643a6c9750fc02bb1fab93ae10db91d399359ca#diff-97f20b0578ce81eb243573451c18ee5b961816e94b2cab6146d48aa497cc1a1fR47-R55

@andysellick could you give a bit of guidance here?

spec/models/speech_spec.rb
```diff
@@ -0,0 +1,48 @@
+RSpec.describe Speech do
```
KludgeKML, line 1: Maybe a good idea to squash this into the commit where you add the implementation of the Content Model? (this is non-blocking, though. I appreciate you might want this here after all the concerns are added)

app/views/speech/show.html.erb
```diff
@@ -9,11 +9,14 @@
 
   <div class="govuk-grid-row">
```
KludgeKML, line 10: Again, you could squash this commit into: Add Views for Speeches

app/assets/stylesheets/application.scss
sairamya93: For now, added `inverse-background` in stylesheets/helpers and imported that in application.scss

app/assets/stylesheets/helpers/_inverse-background.scss
```diff
@@ -0,0 +1,4 @@
+.inverse-background {
+    background: $govuk-brand-colour;
+    color: govuk-colour("white");
```
KludgeKML, line 3: Reminder - I don't think you need to specify the color (just the background) here.

app/assets/stylesheets/application.scss
```diff
@@ -66,3 +66,8 @@ $govuk-include-default-font-face: false;
...
+.inverse-background {
+  background: $govuk-brand-colour;
+  color: govuk-colour("white");
+}
```
KludgeKML, line 73: I think since this has been moved into the helper you can remove it here.

leenagupte (commented):
This is looking great! 🎉  I've added some inline comments, but they're really all quite minor.

spec/requests/speech_spec.rb
```diff
@@ -0,0 +1,13 @@
+RSpec.describe "Speech" do
+  before do
+    content_store_has_example_item("/government/speeches/vehicle-regulations", schema: :speech)
+  end
+
+  describe "GET index" do
+    it "returns 200" do
```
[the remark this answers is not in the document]
sairamya93: Thanks for the reference test. I have added it

config/locales/ar.yml
deborahchua: Should there also be additional system tests for [each schema](https://github.com/alphagov/publishing-api/blob/main/content_schemas/dist/formats/speech/frontend/schema.json#L34-L42), to test it is displaying any variations of content? I think this key was only spotted missing because that document type happened to be used for the withdrawn speech schema example.

spec/support/concerns/political.rb
```diff
@@ -0,0 +1,23 @@
+RSpec.shared_examples "it has historical government information" do |document_type, example_name|
+  let(:content_store_response) { GovukSchemas::Example.find(document_type, example_name:) }
+
+  it "knows it is part of historical government if it is not the current government" do
```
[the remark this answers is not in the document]
sairamya93: That was helpful to know the context.

app/models/concerns/political.rb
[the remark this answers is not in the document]
sairamya93: Yes it was but I have now updated the commit message

spec/support/concerns/news_image.rb
```diff
@@ -0,0 +1,50 @@
+RSpec.shared_examples "it has news image" do |schema|
+  let(:content_store_response) { GovukSchemas::Example.find(schema, example_name: schema) }
+
+  before do
+    {
+      "base_path" => "/a/base/path",
+      "details" => {},
+      "links" => {},
+    }
+  end
```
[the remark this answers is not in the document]
sairamya93: Hmm yes don't think it is required.

spec/support/concerns/news_image.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  it "presents the document's organisation's default_news_image if document's image is not present" do
+    default_news_image = { "url" => "http://www.test.dev.gov.uk/default_news_image.jpg" }
+    content_store_response = {
+      "links" => {
+        "primary_publishing_organisation" => [
+          { "details" => { "default_news_image" => default_news_image } },
+        ],
+      },
+    }
```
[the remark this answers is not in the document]
sairamya93: Ok I have updated the `before` block with the code and the individual tests will modify the values where required.

app/models/speech.rb
```diff
@@ -0,0 +1,31 @@
+class Speech < ContentItem
+  include NewsImage
```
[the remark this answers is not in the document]
sairamya93: Yes but I have now rebased to include the concern and model tests in their corresponding commits

leenagupte (dismissed):
This looks great Ramya!  
Thanks for making all of the requested changes.
The commit history and the audit trail are great and easy to follow too. 🥇

app/models/speech.rb
```diff
@@ -1,11 +1,12 @@
 class Speech < ContentItem
   include NewsImage
   include Organisations
+  include People
   include Political
   include Updatable
 
   def contributors
-    (organisations_ordered_by_emphasis + [speaker].flatten + [speaker_without_profile].flatten).compact
+    (organisations_ordered_by_emphasis + people + [speaker].flatten + [speaker_without_profile].flatten).compact.uniq { |contributor| contributor["content_id"] }
```
leenagupte, line 9: > `{ |contributor| contributor["content_id"] }

What is this block supposed to do?

leenagupte: Is `speaker` or `speaker_without_profile` an array? If they are, then they don't need to be made into an array by wrapping them in `[]`. If they aren't arrays, then `flatten` isn't required because the array will only contain one item.

leenagupte (commented):
Thanks for adding `organisations_with_emphasis` and people to the contributors. 🎉 

There a couple of inline comments.

spec/support/concerns/political.rb
```diff
@@ -0,0 +1,28 @@
...
+    content_store_response["links"]["government"][0]["details"]["current"] = nil
+    expect(described_class.new(content_store_response).historically_political?).to be(false)
+  end
+
+  it "knows it is not political" do
+    content_store_response["details"]["political"] = false
+    expect(described_class.new(content_store_response).historically_political?).to be(false)
+  end
+
+  it "knows it is publishing government" do
```
KludgeKML, line 24: grammar:  `it "knows its publishing government" do`

KludgeKML (approved):
LGTM! (one tiny note about grammer in a test, but otherwise good)

PR #4579, opened 2025-01-13
Add local transaction model

leenagupte:

## What

Move modelling code from LocalTransactionPresenter to a LocalTransaction model.

## Why

- ContentItem models the response from content store
- ContentItemModelPresenter takes a ContentItem object and adds a presentation layer to it.
- ContentItemPresenter does both, it models the response from content store and adds presentation to it.

## How

The functionality that was in the original LocalTransactionPresenter has been moved to the new LocalTransaction model and old LocalTransactionPresenter has been removed.

## Screenshots?

[Contact your Electoral Registration Office](https://govuk-frontend-app-pr-4579.herokuapp.com/contact-electoral-registration-office)


spec/models/local_transaction_spec.rb
```diff
@@ -0,0 +1,105 @@
+RSpec.describe LocalTransaction do
+  let(:content_store_response) do
+    GovukSchemas::Example.find("local_transaction", example_name: "local_transaction_with_devolved_administration_availability")
+  end
+
+  it "gets the introduction" do
```
KludgeKML, line 6: Could we do this and the following two methods like this: 

```
describe "#introduction" do
  it "returns the data from the content item" do (etc)
```

leenagupte: Sure. Will update and fix up.

leenagupte: I've gone with

```ruby
describe "#introduction" do
    it "gets the introduction from the content store response" do
```
As `content_item` is being used in the tests to refer to the model.

PR #4587, opened 2025-01-14
Replace content item presenter and remove publication object

leenagupte:

## What

Replace ContentItemPresenter and remove the publication object.

## Why

We have two ways to model content items and two content item presenters.

- ContentItem models the response from content store
- ContentItemModelPresenter takes a ContentItem object and adds a presentation layer to it.
- ContentItemPresenter does both, it models the response from content store and adds presentation to it.
 
There should only be content item models that inherit from ContentItem and content item presenters that inherit from ContentItemModelPresenter.

Now that all of the document types have been updated to have a ContentItem model we can remove the old ContentItemPresenter.

## Screenshots?

N/A - There shouldn't be any visual changes.


app/views/shared/_base_page.html.erb
```diff
@@ -17,7 +17,7 @@
           <%= yield %>
         </div>
       </div>
-      <%= render 'shared/publication_metadata', :publication => publication %>
+      <%= render 'shared/publication_metadata', :publication => content_item %>
```
KludgeKML, line 20: Is this going to work? We're passing it in as the local "publication" but then using it in the file below as the local "content_item".

app/helpers/application_helper.rb
```diff
@@ -1,6 +1,6 @@
 module ApplicationHelper
```
KludgeKML, line 1: Typo in commit message: except - > accept

app/views/shared/_base_page.html.erb
leenagupte: Good spot! It works because local "publication" is not being used. I'll remove the local as it's not needed.

KludgeKML (approved):
LGTM!

PR #4588, opened 2025-01-14
Tidy calendar

KludgeKML:

## What

Replace all references to scope in calendars with type, and remove scope validation.

## Why

Scope is just calendar type, there's no need for an additional term for it. And the scope validation in the controller is never going to catch anything (the validation has already happened as a part of loading the content item).


leenagupte (approved):
I love this! ⭐

PR #4596, opened 2025-01-17
Gracefully handle errors from Elections API

unoduetre:

## What

The postcode lookup tool on [https://www.gov.uk/contact-electoral-registration-office](https://www.gov.uk/contact-electoral-registration-office "smartCard-inline")  uses the Electoral Commisions' [Elections API](https://api.electoralcommission.org.uk/docs/ "‌") to match postcodes to Registration Offices. Sometimes their API goes down, and as a result when a user searches for a postcode on this page, instead of them seeing a result, they will be show a standard [gov.uk](http://gov.uk "‌") 500 error page.

This card is about adding some error handling to Frontend so that:

- the error is sent to sentry
- a more useful message is shown to the user instead

## Why

[Trello card](https://trello.com/c/8kaDJBx6/3159-gracefully-handle-errors-from-elections-api), [Jira issue NAV-15410](https://gov-uk.atlassian.net/browse/NAV-15410)

## Screenshots

Screenshot 2025-01-28 at 10-05-45 Error Contact your Electoral Registration Office - GOV UK

Zrzut ekranu z 2025-01-29 o 10 06 44

### DIFF
--- a/app/models/location_error.rb
+++ b/app/models/location_error.rb
@@ -7,6 +7,7 @@ class LocationError
     "invalidUprnFormat" => "formats.local_transaction.invalid_uprn",
     "validPostcodeNoElectionsMatch" => "formats.local_transaction.valid_postcode_no_match",
     "validUprnNoElectionsMatch" => "formats.local_transaction.valid_uprn_no_match",
+    "electoralServiceNotAvailable" => "formats.local_transaction.electoral_service_not_available",
   }.freeze
 
   SUB_MESSAGES = {
--- a/app/models/location_error.rb
+++ b/app/models/location_error.rb
@@ -18,6 +18,16 @@ class LocationError
     "validUprnNoElectionsMatch" => "formats.local_transaction.valid_uprn_no_match_sub_html",
   }.freeze
 
+  DATA_RELATED = %w[
+    noLaMatch
+    fullPostcodeNoLocationsApiMatch
+    validPostcodeNoLocation
+    invalidPostcodeFormat
+    invalidUprnFormat
+    validPostcodeNoElectionsMatch
+    validUprnNoElectionsMatch
+  ].freeze
+
   attr_reader :postcode_error, :message, :sub_message, :message_args
 
   def initialize(postcode_error = nil, message_args = {})
@@ -31,5 +41,9 @@ class LocationError
       @message = "formats.local_transaction.invalid_postcode"
       @sub_message = "formats.local_transaction.invalid_postcode_sub"
     end
+  end
+
+  def data_related?
+    DATA_RELATED.include?(@postcode_error)
   end
 end
--- a/app/services/electoral_service.rb
+++ b/app/services/electoral_service.rb
@@ -38,6 +38,9 @@ private
       else
         "validUprnNoElectionsMatch"
       end
+    elsif [500, 502, 503, 504].include? exception.http_code
+      GovukError.notify(exception)
+      "electoralServiceNotAvailable"
     else
       raise
     end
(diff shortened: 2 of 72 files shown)


app/services/electoral_service.rb
```diff
@@ -38,14 +38,21 @@ def assemble_error(exception)
...
+      "electoralServiceNotAvailable"
     else
       raise
     end
   end
 
   def request_url
     endpoint = postcode.present? ? "postcode/#{postcode}" : "address/#{uprn}"
-    "#{api_base_path}/#{endpoint}?token=#{ENV['ELECTIONS_API_KEY']}"
+    "#{api_base_path}/#{endpoint}?token=#{api_key}"
```
unoduetre, line 51: Small refactoring.

spec/models/location_error_spec.rb
```diff
@@ -3,31 +3,43 @@
   # not the values of them. This is because LocationError is a bit odd,
   # and it might be worth refactoring the actual code after the RSpec
   # conversion is finished.
+
```
unoduetre, line 6: I did some refactoring here as well.

app/models/location_error.rb
```diff
@@ -31,4 +42,8 @@ def initialize(postcode_error = nil, message_args = {})
       @sub_message = "formats.local_transaction.invalid_postcode_sub"
     end
   end
+
+  def postcode_related?
```
unoduetre, line 46: This is used in the view to decide whether to display the message next to the postcode entered by the user (if it's postcode related), or not (e.g if it relates to the service itself and not the entered postcode). The message at the top is always displayed.

hannako: Ah gotcha - thanks for explaining. Maybe it should be called user_input_related_error? or something like that then.

unoduetre: I like the general idea of making it more clear. but I'm not sure, if we should mention something like "user input", as this class is very general and it doesn't have anything to do with user input directly e.g. it could be reused for postcode errors coming from elsewhere.

If we can find a better name that distinguishes between errors which mean there is something wrong with the provided data etc. vs errors which mean there is something wrong that is not directly related to the provided data. It's like 400 vs 500 errors in http. They called it client errors vs server errors, but that won't work here.

Maybe `data_related?`. Maybe you can find something better? Or we can go with `user_input_related_error?` if we assume this class will never be used for anything else, but user input? What do you think?

unoduetre:
I've updated the message to make it shorter, as discussed in the ticket.

app/models/location_error.rb
hannako: data_related? sounds good

app/views/application/_location_form.html.erb
```diff
@@ -68,9 +68,9 @@
       name: "postcode",
       id: "postcode",
       hint: t("formats.local_transaction.postcode_hint"),
-      invalid: @location_error ? "true" : "false",
+      invalid: @location_error&.data_related? ? "true" : "false",
```
unoduetre, line 71: I've added a small fix here as well.

app/models/location_error.rb
unoduetre: I've updated the name.

app/models/location_error.rb
```diff
@@ -31,4 +42,8 @@ def initialize(postcode_error = nil, message_args = {})
       @sub_message = "formats.local_transaction.invalid_postcode_sub"
     end
   end
+
+  def data_related?
+    POSTCODE_RELATED.include?(@postcode_error)
+  end
```
hannako, line 48: Could you make the constant name and the variable name the same, ie `DATA_RELATED`

unoduetre: FIxed.

hannako (approved):
Thanks Mat

PR #4599, opened 2025-01-17
Use component wrapper on figure component

AshGDS:
## What
- Adds the component wrapper helper to the `figure` component.

## Why
As the [trello card](https://trello.com/c/t1x49bym/440-use-component-wrapper-helper-in-application-components), [Jira issue PNP-7343](https://gov-uk.atlassian.net/browse/PNP-7343) states:

> Standardising our components to use the component wrapper helper will reduce code, increase standardisation, and improve future feature implementation speed.

## Visual changes

None.

### DIFF
--- a/app/views/components/_figure.html.erb
+++ b/app/views/components/_figure.html.erb
@@ -1,10 +1,13 @@
-<% add_app_component_stylesheet("figure") %>
 <%
+  add_app_component_stylesheet("figure")
   alt ||= ""
   caption ||= ""
   credit ||= ""
+  local_assigns[:lang] ||= "en"
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("app-c-figure")
 %>
-<figure class="app-c-figure" lang="en">
+<%= tag.figure(**component_helper.all_attributes) do %>
   <% if src.present? %>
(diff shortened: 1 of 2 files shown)


app/views/components/_figure.html.erb
```diff
@@ -1,10 +1,14 @@
-<% add_app_component_stylesheet("figure") %>
 <%
+  add_app_component_stylesheet("figure")
   alt ||= ""
   caption ||= ""
   credit ||= ""
+  lang ||= "en"
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("app-c-figure")
+  component_helper.set_lang(lang)
```
andysellick, line 9: Non-blocking: I think you might not need this line, as the wrapper should set lang automatically. If it doesn't, try initialising it as `local_assigns[:lang] ||= "en"` (might be worth double checking with a test as there currently isn't one).

andysellick (approved):

app/views/components/_figure.html.erb
andysellick: Alternatively since lang wasn't an option on the component previously, you could remove that first line and instead just have `component_helper.set_lang("en")`

AshGDS: Good idea, I've used `local_assigns[:lang] ||= "en"` now, thanks.

PR #4600, opened 2025-01-17
Move specialist document pages

leenagupte:
, [Jira issue PNP-6373](https://gov-uk.atlassian.net/browse/PNP-6373)⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

Copy the rendering code for specialist document pages from government-frontend.
The rendering code has been modified to fit the best practices for frontend as outlined in the Contributing guide.

## Why

[Trello card](https://trello.com/c/RrUeLTKc)

## Examples

aaib-report:
* https://www.gov.uk/aaib-reports/aaib-investigation-to-rotorsport-uk-calidus-g-pcpc
* https://govuk-frontend-app-pr-4600.herokuapp.com/aaib-reports/aaib-investigation-to-rotorsport-uk-calidus-g-pcpc

business-finance-support:
* https://www.gov.uk/business-finance-support/rural-business-growth-fund
* https://govuk-frontend-app-pr-4600.herokuapp.com/business-finance-support/rural-business-growth-fund

cma-case:
* https://www.gov.uk/cma-cases/topps-tiles-slash-ctd-tiles-certain-assets-merger-inquiry
* https://govuk-frontend-app-pr-4600.herokuapp.com/cma-cases/topps-tiles-slash-ctd-tiles-certain-assets-merger-inquiry

maib-report:
* https://www.gov.uk/maib-reports/capsize-of-recreational-craft-wheelyboat-123-with-the-loss-of-2-lives
* https://govuk-frontend-app-pr-4600.herokuapp.com/maib-reports/capsize-of-recreational-craft-wheelyboat-123-with-the-loss-of-2-lives

eu-withdrawal-act-2018-statutory-instrument:
* https://www.gov.uk/eu-withdrawal-act-2018-statutory-instruments/the-drivers-hours-and-tachographs-amendment-regulations-2021
* https://govuk-frontend-app-pr-4600.herokuapp.com/the-drivers-hours-and-tachographs-amendment-regulations-2021

N.B. These have a link to the finder at the bottom of the page.

## Screenshots

|government-frontend (before) | frontend (after)|
|-------------------------------|----------------|
|screencapture-gov-uk-business-finance-support-access-to-finance-advice-north-west-england-2025-02-17-13_55_13|screencapture-frontend-dev-gov-uk-business-finance-support-access-to-finance-advice-north-west-england-2025-02-17-14_20_09|
|Screenshot 2025-02-17 at 13 56 58|Screenshot 2025-02-17 at 13 57 35|

## Testing in integration

|Rendered view|Content item|
|---------------|-------------|
|Screenshot 2025-02-24 at 17 38 32|Screenshot 2025-02-24 at 17 37 58|


app/views/specialist_document/show.html.erb
```diff
@@ -10,7 +10,10 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds responsive-top-margin">
-    <%= render "govuk_publishing_components/components/title", @content_item.title_and_context %>
+    <%= render "govuk_publishing_components/components/title", {
```
KludgeKML: I don't know if this is covered in a later commit, but we don't want to be using the title component - we want to be using the [heading component[(https://components.publishing.service.gov.uk/component-guide/heading) instead. (title is being removed very soon).

app/views/specialist_document/show.html.erb
```diff
@@ -10,10 +10,13 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds responsive-top-margin">
-    <%= render "govuk_publishing_components/components/title", {
-        title: content_item.title,
-        average_title_length: "long",
-      } %>
+    <%= render "govuk_publishing_components/components/heading", {
```
KludgeKML, line 13: Ah, there we are. I think we could just do this in one commit.

app/presenters/specialist_document_presenter.rb
```diff
@@ -0,0 +1,19 @@
...
+private
+
+  attr_reader :view_context
+
+  def show_contents_list?
+    content_item.headers.present? && level_two_headings?
+  end
+
+  def level_two_headings?
+    content_item.headers.any? { |header| header[:level] == 2 }
```
KludgeKML, line 54: Can we make any assumptions about the header levels? If the roots are always level 2, then we could just replace this with `.any?`, couldn't we? (since we're nesting the lower level ones). Or if the roots are sometimes level 1, this isn't going to find the level 2 ones if they're nested.

app/models/specialist_document.rb
```diff
@@ -19,6 +19,7 @@ def headers_list(headers)
       h = {
         href: "##{header['id']}",
         text: header["text"].gsub(/:$/, ""),
+        level: header["level"],
```
KludgeKML, line 70: Maybe put this line in the previous commit?

KludgeKML (commented):
Looks pretty good, I have a few inline notes.

app/presenters/specialist_document_presenter.rb
leenagupte: The `headers` are created by parsing markdown added by editors. So I don't think we can make any assumptions about the header levels. 

The code that this logic is based on checks for the existence of ["h2" elements in the body text](https://github.com/alphagov/government-frontend/blob/main/app/presenters/content_item/contents_list.rb#L43). I've tried to do the same thing, but in a nicer way and use the `headers` object we access to.

leenagupte: > roots are sometimes level 1, this isn't going to find the level 2 ones if they're nested

If there are multiple H1s there are bigger problems with the docs. But it does mean that a contents list won't be created if there are just H3s.

leenagupte:
@KludgeKML Thanks for your review. I've amended the commits as suggested, and responded to your question about the headers.

app/views/specialist_document/show.html.erb
```diff
@@ -38,7 +38,7 @@
       <% end %>
     <% end %>
 
-    <%= render "components/contents_list_with_body", contents: @content_item.contents do %>
+      <%= render "components/contents_list_with_body", contents: @specialist_document_presenter.contents do %>
```
deborahchua, line 46: Minor: an additional space here

leenagupte: Good spot! I've updated and fixed up.

deborahchua (approved):
This looks good! I appreciate the small commits, it's made reviewing it really easy 🎉

leenagupte:
@KludgeKML @deborahchua Thank you both for your reviews. After our conversations about modelling organisations I have changed this PR slightly to create a Finder content item model, and use that to model the `finder` from `links`. It doesn't change the code very much as I refrained from going one step further and left `facets` (from `details`) as an array of hashes rather than an array of open structs.

I think this approach is better, but I'd appreciate a re-review.

app/views/specialist_document/show.html.erb
```diff
@@ -10,7 +10,13 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds responsive-top-margin">
-    <%= render "govuk_publishing_components/components/title", @content_item.title_and_context %>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: content_item.title,
+      average_title_length: "long",
```
KludgeKML, line 15: I think you still want to remove this? (this is valid for the title component, not for the heading one)

app/controllers/specialist_document_controller.rb
```diff
@@ -1,5 +1,7 @@
 class SpecialistDocumentController < ContentItemsController
   include Cacheable
 
-  def show; end
+  def show
+    @specialist_document_presenter = SpecialistDocumentPresenter.new(content_item)
```
KludgeKML: :+1:

(a bunch of existing examples of this code have content_item.to_h, we should probably clean them up)

app/models/specialist_document.rb
```diff
@@ -1,10 +1,12 @@
 class SpecialistDocument < ContentItem
-  attr_reader :continuation_link, :headers, :protection_type, :will_continue_on
+  attr_reader :continuation_link, :headers, :metadata,
+              :protection_type, :will_continue_on
 
   def initialize(content_store_response)
     super(content_store_response)
 
     @continuation_link = content_store_hash.dig("details", "metadata", "continuation_link")
```
KludgeKML, line 10: Note - content_store_hash may or may not exist depending on if this is merged before or after the PfG scaffolding PR (Nothing necessarily to do here, just need to note that these two PRs will clash one way or the other, and we have to bear in mind that some rebasing will be needed)

app/presenters/specialist_document_presenter.rb
```diff
@@ -40,4 +61,23 @@ def protected_food_drink_name?
...
+
+  def format_facet_value(type, value, key, filterable)
+    if type == "date"
+      view_context.display_date(value)
+    elsif type == "text" && value.is_a?(Array) && filterable == true
+      value.map { |v| format_filterable_links(key, v[:value], v[:label]) }
+    else
+      value
+    end
+  end
```
KludgeKML, line 73: I wonder if we could do this mapping in the view? (Just because this is the first time in the new consolidated routes that we've passed a view_context to a presenter rather than it just being available in views and helpers, so want to double-check whether this is a legit precedent to set).

Current precedent for it being in view: https://github.com/alphagov/frontend/pull/4225/files#diff-0b7fa2198812d14b54b691c4fde2f55cab0ba0109aaf95efaffba7e58e587d39R4-R10

leenagupte: I went back and forth about this. I'm not happy about passing in a `view_context` but then I didn't want to duplicate the helper methods either. 

The reason I didn't put it directly in the view is because of the amount of logic in it, especially in the `elsif`. That `format_facet_value` method is called inside a loop and it would get messy and hard to debug.

I also felt it would be easier to test in a unit test rather than a system test.

leenagupte: I'd be happy to pair on a better way to do this, as I'm sure there must be one.

app/presenters/specialist_document_presenter.rb
```diff
@@ -40,4 +61,23 @@ def protected_food_drink_name?
...
+
+  def format_filterable_links(key, value, label)
```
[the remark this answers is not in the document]
leenagupte: Do you mean `_link` singular? i.e. `format_filterable_link`?

app/views/specialist_document/show.html.erb
leenagupte: Good spot! Thanks. I copy and pasted from the case study view 🙈 , which will also need to be updated (in a separate change)

leenagupte: I've made the change and fixed up.

app/helpers/link_helper.rb
```diff
@@ -16,7 +16,7 @@ def govuk_styled_link(text, path: nil, inverse: false)
     "<a href='#{path}' class='#{classes}'>".html_safe + text + "</a>".html_safe
   end
```
KludgeKML, line 17: Isn't this last commit now covered in https://github.com/alphagov/frontend/pull/4660/files ?

leenagupte: Yes, I added it to get the tests to pass, but will remove it and rebase.

KludgeKML (approved):
LGTM!

AshGDS (commented):
Thanks for providing page examples across live/the PR 👍 

I've done a HTML code compare rather than a visual check as a first pass, and noticed some stuff:

1. It seems the description meta tag has gone missing between the live site and the preview app? e.g. `<meta name="description" content="Location: Roadford Lake, Devon, England.">`

2. There seems to be an empty div just below the h1 on the pages - `<div class="govuk-grid-column-one-third"></div>` - is this needed?

3. There's a `direction-ltr` class on the `cma-case` page which looks like it doesn't do anything (though I'm not sure you've added this in the code, I can't actually find where it's coming from)

4. Having the language set on the `<main>` element is used for GA4 tracking purposes. Is it possible for the language to be included on the main tag? Our tracking does fall back to the `<html>` language value if it can't find it on the `<main>` element, but that one isn't always reflective of the content language.

JamesCGDS (commented):
I’ve checked the following pages:

- https://www.gov.uk/business-finance-support/access-to-finance-advice-north-west-england
- https://www.gov.uk/cma-cases/review-of-banking-for-small-and-medium-sized-businesses-smes-in-the-uk
- https://www.gov.uk/international-development-funding/uk-aid-direct
- https://www.gov.uk/european-structural-investment-funds/business-development
- https://www.gov.uk/service-standard-reports/apply-for-a-budgeting-loan

I only noticed one visual difference on < tablet screen sizes where additional spacing has been introduced above the “Related Content” section (screenshot below). This is because in this app, the [shared sidebar navigation component](https://github.com/alphagov/frontend/blob/main/app/views/shared/_sidebar_navigation.html.erb) (which contains “Related Content”) is wrapped in a `div` with a `sidebar-navigation` class which adds `margin-top: govuk-spacing(7);` on smaller screens.

Personally I don’t I think this is a problem as I feel the extra spacing gives things a bit more room to breathe but happy to discuss if others feel differently.

| ##Before | ## After |
|-----------|---------|
| image | image |

leenagupte:
@AshGDS Thanks for being so thorough in your review. ⭐ 

I'll address each of the points you raised:

1. Good spot on the meta description. I've added it in this [commit](https://github.com/alphagov/frontend/pull/4600/commits/890f28786fa3c0a43e1c01441b8adbd7938c857c).
2. I've added a [PR](https://github.com/alphagov/frontend/pull/4661) to remove the empty `<div>` tag if there aren't any available translations.
3. The `direction-ltr` is on all pages, even the homepage 😱 so I shall leave it as it is. I suspect it's being pulled in from the layout somewhere.
4. I thought the language was being added to the `<main>` tag? `<main>` is being set in the [application layout](https://github.com/alphagov/frontend/blob/main/app/views/layouts/application.html.erb#L16) so fixing it probably falls outside of the scope of this change.

leenagupte:
@JamesCGDS Thanks for your review 🪇 

That's a good spot about the `side-navigation`. As it's something that's applied globally, it's probably something that should be discussed, and then decide whether it should be changed.

AshGDS (approved):
LGTM 👍 

I guess with the language thing, since it's populated by the content item I assume it will work OK once deployed then?

And James is now on leave, so with the spacing thing he noticed it may be worth running it by Nikin for approval.

leenagupte:
> LGTM 👍
> 
> I guess with the language thing, since it's populated by the content item I assume it will work OK once deployed then?
> 
> And James is now on leave, so with the spacing thing he noticed it may be worth running it by Nikin for approval.

Thanks for re-reviewing!

It looks like the language thing isn't working in frontend at all. As it's supposed to be added in the layout, I think it should be fixed in a separate PR. The same goes for the spacing issue James found. That's also something that's being set in the layout, so should also have a separate discussion and fix. I can raise them as issues.

PR #4608, opened 2025-01-22
Move Corporate information pages from government-frontend

deborahchua:

## What

We want to handle the rendering of `corporate_information_page` documents from government-frontend to frontend.

## Why

[Trello card](https://trello.com/c/EKDXTDsm/398-move-document-type-corporateinformationpage-from-government-frontend-to-frontend), [Jira issue PNP-6452](https://gov-uk.atlassian.net/browse/PNP-6452)

## How

By moving the routes and necessary code to get the pages rendering in Frontend.

## Page examples
- /government/organisations/forensic-science-regulator/about (contents list with additional corporate information heading)
- /government/organisations/forensic-science-regulator/about/accessible-documents-policy (contents list without additional corporate information heading)
- /government/organisations/forensic-science-regulator/about/personal-information-charter (no contents list)
- /government/organisations/companies-house/about.cy (has Welsh translation)
- /government/organisations/department-for-environment-food-rural-affairs/about/modern-slavery-statement (no contents list)

## Screenshots

|government-frontend (before)| frontend (after)|
|-------|--------|
|Screenshot 2025-01-22 at 11 38 56|Screenshot 2025-01-22 at 11 39 11|
|Screenshot 2025-01-22 at 11 43 36|Screenshot 2025-01-22 at 11 43 44|
|Screenshot 2025-01-22 at 11 45 42|Screenshot 2025-01-22 at 11 45 52|
|Screenshot 2025-01-22 at 14 47 07|Screenshot 2025-01-22 at 14 46 32|
|Screenshot 2025-02-06 at 13 31 14|Screenshot 2025-02-06 at 13 31 05|

### DIFF
--- a/app/controllers/corporate_information_page_controller.rb
+++ b/app/controllers/corporate_information_page_controller.rb
@@ -1,0 +1,3 @@
+class CorporateInformationPageController < ContentItemsController
+  def show; end
+end
--- a/app/controllers/corporate_information_page_controller.rb
+++ b/app/controllers/corporate_information_page_controller.rb
@@ -1,3 +1,5 @@
 class CorporateInformationPageController < ContentItemsController
-  def show; end
+  def show
+    @presenter = CorporateInformationPagePresenter.new(@content_item)
+  end
 end
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -4,4 +4,50 @@ class CorporateInformationPage < ContentItem
 
     @default_organisation ||= organisations.detect { |org| org.content_id == organisation_content_id }
   end
+
+  def corporate_information?
+    corporate_information_groups.any?
+  end
+
+  def corporate_information
+    corporate_information_groups.map do |group|
+      {
+        title: group["name"],
+        id: group["name"].tr(" ", "-").downcase,
+        links: normalised_group_links(group),
+      }
+    end
+  end
+
+  def corporate_information_pages
+    linked("corporate_information_pages")
+  end
+
+private
+
+  def normalised_group_links(group)
+    group["contents"].map { |group_item| normalised_group_item_link(group_item) }.compact
+  end
+
+  def normalised_group_item_link(group_item)
+    if group_item.is_a?(String)
+      group_item_link = corporate_information_pages.find { |l| l.content_id == group_item }
+      # it's possible for corporation_information_groups in details and links hashes to be out of sync.
+      if group_item_link
+        {
+          title: group_item_link.title,
+          path: group_item_link.base_path,
+        }
+      end
+    else
+      {
+        title: group_item["title"],
+        path: group_item["path"] || group_item["url"],
+      }
+    end
+  end
+
+  def corporate_information_groups
+    content_store_response.dig("details", "corporate_information_groups") || []
+  end
 end
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -1,2 +1,7 @@
 class CorporateInformationPage < ContentItem
+  def default_organisation
+    organisation_content_id = content_store_response.dig("details", "organisation")
+
+    @default_organisation ||= organisations.detect { |org| org.content_id == organisation_content_id }
+  end
 end
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -1,0 +1,2 @@
+class CorporateInformationPage < ContentItem
+end
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -1,4 +1,12 @@
 class CorporateInformationPage < ContentItem
+  attr_reader :headers
+
+  def initialize(content_store_response)
+    super(content_store_response)
+
+    @headers = content_store_response.dig("details", "headers") || []
+  end
+
   def default_organisation
     organisation_content_id = content_store_response.dig("details", "organisation")
 
--- a/app/presenters/corporate_information_page_presenter.rb
+++ b/app/presenters/corporate_information_page_presenter.rb
@@ -1,0 +1,2 @@
+class CorporateInformationPagePresenter < ContentItemPresenter
+end
--- a/app/presenters/corporate_information_page_presenter.rb
+++ b/app/presenters/corporate_information_page_presenter.rb
@@ -1,2 +1,46 @@
 class CorporateInformationPagePresenter < ContentItemPresenter
+  include LinkHelper
+
+  def corporate_information_heading
+    I18n.t("formats.corporate_information_page.corporate_information")
+  end
+
+  def corporate_information_heading_id
+    corporate_information_heading.tr(" ", "-").downcase
+  end
+
+  def further_information
+    return unless corporate_information_pages?
+
+    further_information_links
+  end
+
+private
+
+  def corporate_information_pages?
+    content_item.corporate_information_pages.any?
+  end
+
+  def further_information_links
+    links = further_information_types.map do |further_information_type|
+      page = content_item.corporate_information_pages.find { |content_item| content_item.document_type == further_information_type }
+      next unless page
+
+      link = govuk_styled_link(page.title, path: page.base_path)
+
+      I18n.t("formats.corporate_information_page.#{further_information_type}_html", link:)
+    end
+
+    links.compact.join(" ").html_safe
+  end
+
+  def further_information_types
+    %w[
+      publication_scheme
+      welsh_language_scheme
+      personal_information_charter
+      social_media_use
+      about_our_services
+    ]
+  end
 end
--- a/app/presenters/corporate_information_page_presenter.rb
+++ b/app/presenters/corporate_information_page_presenter.rb
@@ -1,12 +1,23 @@
 class CorporateInformationPagePresenter < ContentItemPresenter
   include LinkHelper
 
-  def corporate_information_heading
-    I18n.t("formats.corporate_information_page.corporate_information")
+  def headers_for_contents_list_component
+    @headers = contents_list_headings
+
+    return [] unless show_contents_list?
+
+    ContentsOutlinePresenter.new(@headers).for_contents_list_component
   end
 
-  def corporate_information_heading_id
-    corporate_information_heading.tr(" ", "-").downcase
+  def corporate_information_heading
+    heading_text = I18n.t("formats.corporate_information_page.corporate_information")
+    heading_id = heading_text.tr(" ", "-").downcase
+
+    {
+      "text" => heading_text,
+      "level" => 2,
+      "id" => heading_id,
+    }
   end
 
   def further_information
@@ -16,6 +27,24 @@ class CorporateInformationPagePresenter < ContentItemPresenter
   end
 
 private
+
+  def show_contents_list?
+    @headers.present? && @headers.level_two_headers?
+  end
+
+  def contents_list_headings
+    content_item.headers << corporate_information_heading if content_item.corporate_information?
+
+    exclude_nested_headings
+
+    ContentsOutline.new(content_item.headers) if content_item.headers.present?
+  end
+
+  def exclude_nested_headings
+    content_item.headers.each do |header|
+      header.delete("headers") unless header["headers"].nil?
+    end
+  end
 
   def corporate_information_pages?
     content_item.corporate_information_pages.any?
--- a/app/views/corporate_information_page/show.html.erb
+++ b/app/views/corporate_information_page/show.html.erb
@@ -1,3 +1,5 @@
+<% content_for :title, "#{@presenter.page_title} - #{content_item.default_organisation.title} - GOV.UK" %>
+
 <% content_for :extra_headers do %>
   <meta name="description" content="<%= content_item.description %>">
   <%= render "govuk_publishing_components/components/machine_readable_metadata", { content_item: content_item.to_h, schema: :article } %>
--- a/app/views/corporate_information_page/show.html.erb
+++ b/app/views/corporate_information_page/show.html.erb
@@ -7,21 +7,22 @@
 
 <% @additional_body = capture do %>
   <% if content_item.corporate_information? %>
-    <%= content_item.corporate_information_heading_tag %>
+    <%= content_tag(:h2, @presenter.corporate_information_heading, id: @presenter.corporate_information_heading_id) %>
+
     <% content_item.corporate_information.each do |group| %>
-      <%= group[:heading] %>
+      <%= content_tag(:h3, group[:title], id: group[:id]) %>
       <ul>
         <% group[:links].each do |link| %>
           <li>
-            <%= link %>
+            <%= govuk_styled_link(link[:title], path: link[:path]) %>
           </li>
         <% end %>
       </ul>
     <% end %>
 
-    <% if content_item.further_information.present? %>
+    <% if @presenter.further_information.present? %>
       <p>
-        <%= content_item.further_information %>
+        <%= @presenter.further_information %>
       </p>
     <% end %>
   <% end %>
--- a/app/views/corporate_information_page/show.html.erb
+++ b/app/views/corporate_information_page/show.html.erb
@@ -41,7 +41,14 @@
     <div class="govuk-grid-column-two-thirds">
       <%= render "govuk_publishing_components/components/lead_paragraph", text: content_item.description %>
       <% if content_item.withdrawn? %>
-        <%= render "govuk_publishing_components/components/notice", content_item.withdrawal_notice_component %>
+        <% withdrawn_time_tag = tag.time(display_date(content_item.withdrawn_at), datetime: content_item.withdrawn_at) %>
+
+        <%= render "govuk_publishing_components/components/notice", {
+          title: I18n.t("withdrawn_notice.title", schema_name: I18n.t("formats.#{content_item.schema_name}", count: 1, locale: :en).downcase, withdrawn_time: withdrawn_time_tag).html_safe,
+          description_govspeak: content_item.withdrawn_explanation&.html_safe,
+          time: withdrawn_time_tag,
+          lang: "en",
+        } %>
       <% end %>
     </div>
   </div>
--- a/app/views/corporate_information_page/show.html.erb
+++ b/app/views/corporate_information_page/show.html.erb
@@ -25,7 +25,7 @@
   <% end %>
 <% end %>
 
-<div class="<%= content_item.organisation_brand_class %>">
+<div>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <%= render "govuk_publishing_components/components/heading",
@@ -47,7 +47,15 @@
     <div class="govuk-grid-column-two-thirds">
       <hr class="govuk-section-break govuk-section-break--visible">
       <div class="govuk-!-width-one-half govuk-!-margin-top-3 responsive-bottom-margin">
-        <%= render "govuk_publishing_components/components/organisation_logo", content_item.organisation_logo %>
+        <%= render "govuk_publishing_components/components/organisation_logo", {
+          organisation: {
+            name: content_item.default_organisation.logo.formatted_title.html_safe,
+            url: content_item.default_organisation.base_path,
+            brand: content_item.default_organisation.brand,
+            crest: content_item.default_organisation.logo.crest,
+            image: content_item.default_organisation.logo.image,
+          },
+        } %>
       </div>
 
       <%= render "govuk_publishing_components/components/contents_list_with_body", contents: content_item.contents do %>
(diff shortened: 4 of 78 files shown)


spec/components/contents_list_with_body_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    ]
+  end
+
+  def block
+    "<p>Foo</p>".html_safe
+  end
+
+  it "renders nothing without a block" do
+    render_component(contents: contents_list)
+    expect(rendered).to be_empty
```
KludgeKML, line 19: Could you add a space before the first expectation here? (a sort of non-lintable convention we have)

spec/components/contents_list_with_body_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  it "yields the block without contents data" do
+    render_component({}) { block }
+    expect(rendered).to include(block)
```
KludgeKML, line 24: Same here

spec/components/contents_list_with_body_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    expect(rendered).to have_css("#contents.app-c-contents-list-with-body")
+    expect(rendered).to have_css("#contents.app-c-contents-list-with-body")
+    expect(rendered).to have_css("#contents[data-module='sticky-element-container']")
+  end
+
+  it "does not apply the sticky-element-container data-module without contents data" do
+    render_component(contents: contents_list) { block }
+
+    expect(rendered).to have_css("#contents[data-module='sticky-element-container']")
```
KludgeKML, line 38: Is this test right? Test text says "does not apply the data module", but the expectation seems to thing it would still be there.

app/models/concerns/contents_list.rb
```diff
@@ -0,0 +1,104 @@
...
+      @contents ||=
+        if show_contents_list?
+          contents_items.each { |item| item[:href] = "##{item[:id]}" }
+        else
+          []
+        end
+    end
+
+    def contents_items
+      extract_headings_with_ids
```
KludgeKML, line 20: This line could do with a test - unless it's never expected to be called (always overwritten by the class the module is included in) - in which case we could maybe remove it entirely?

app/models/concerns/contents_list.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    def first_item_has_image?
+      element = first_item
+
+      until element.name == "h2"
+        return true if element.name == "div" && element["class"] == "img"
+
+        element = element.next_element
+        return false if element.nil?
```
KludgeKML, line 92: This line could do with another test (looks like in tests it's always triggering). We're also not explicitly returning a value if we trigger the until (which _probably_ means nil, which is probably okay, but it might be better to be strict in our ? methods and explicitly return a false.) You can also see more clearly that the false actually got returned.

app/presenters/corporate_information_page_presenter.rb
```diff
@@ -0,0 +1,157 @@
...
+  end
+
+  def page_title
+    page_title = super
+    page_title += " - #{default_organisation['title']}" if default_organisation
+    page_title
+  end
+
+  def organisation_logo(organisation = default_organisation)
+    return nil unless organisation && organisation.dig("details", "logo")
```
KludgeKML, line 14: I think we don't have a test for if this triggers and returns the nil?

app/presenters/corporate_information_page_presenter.rb
```diff
@@ -0,0 +1,157 @@
...
+  end
+
+private
+
+  def default_organisation
+    organisation_content_id = content_item.content_store_hash.dig("details", "organisation")
+    organisation = nil
+    if organisations.present?
+      organisation = organisations.detect { |org| org["content_id"] == organisation_content_id }
+      raise "No organisation in links that matches the one specified in details: #{organisation_content_id}" unless organisation
```
KludgeKML, line 80: We don't test this raise. Is it possible to get here at all with our current content?

leenagupte (commented):
WIP Review. I got as far as: Add ContentsList concern and tests

spec/components/contents_list_with_body_spec.rb
deborahchua: Ah good spot thanks both! I copied the test over from government-frontend incorrectly - https://github.com/alphagov/government-frontend/blob/842a887dc0c83311a9ead2daa08bd430ad0ac347/test/components/contents_list_with_body_test.rb#L34-L37

app/models/concerns/contents_list.rb
deborahchua: I had a look at the test coverage and [the method itself](https://github.com/alphagov/frontend/pull/4608/commits/ef5ef221c521c6b84d0bf8ec4e809aab5a898ea3#diff-73e131976ad489a72d889fad2320885cced2b1c302b0c89373ea30d874f3cd4eR44-R50) is covered but not this specific line as `contents_items` is always overwritten for this document.

Some of the other document types don't overwrite this method, so this line should hopefully be covered by their tests when they are moved across.

app/presenters/corporate_information_page_presenter.rb
```diff
@@ -65,4 +103,55 @@ def executive_order_crest?(organisation)
...
+    else
+      @view_context.link_to(group_item["title"], group_item["path"] || group_item["url"])
+    end
+  end
+
+  def corporate_information_page_links
+    expanded_links_from_content_item("corporate_information_pages")
+  end
+
+  def expanded_links_from_content_item(type)
```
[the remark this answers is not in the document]
deborahchua: I've done some further refactoring and moved as much as possible into the model.
The remaining methods in the presenter help generate a paragraph which includes links to further information pages if they are available, and need `@view_context` to generate the links. I haven't been able to find a nice way of directly displaying them in view instead.

spec/models/corporate_information_page_spec.rb
```diff
@@ -107,4 +112,12 @@
       expect(presented_groups.first[:id]).to eq("access-our-information")
     end
   end
+
+  describe "#contents_items" do
+    it "includes the corporate information heading in the contents list if corporate informations groups are present" do
+      expect(corporate_information_page.content_store_response["details"]["corporate_information_groups"].present?).to be(true)
```
leenagupte, line 118: Should this be checking for the presence of `corporate_information_page.corporate_information_groups` instead?

leenagupte (commented):
We've already chatted about modelling Organisations in person, but as promised I've reviewed the `ContentsList` concern. 
It's very confusing! So I can understand why you've kept most of the code as it was in government-frontend.

Part of me wonders whether it's worth trying to deconstruct the ContentsList a bit in to it's model and presentational parts.

For example, the decision to `show_contents_list?` is quite presentational, not something the model cares about. Could the model return a structured object of headers? The code in the ContentsList is confusing because as I understand it, much of it won't be reached unless `contents_items` is overwritten.

I'm in two minds about whether it's worth the effort to simplify it now, or leave it for a firebreak. I think perhaps if the system tests are thorough, then the refactoring could be left for "later".

spec/models/corporate_information_page_spec.rb
```diff
@@ -30,4 +30,43 @@
...
+    it "does not present corporate information groups on about pages if information isn't available" do
+      content_store_response["details"]["corporate_information_groups"] = nil
+
+      expect(described_class.new(content_store_response).corporate_information?).to be(false)
+    end
+  end
+
+  describe "#corporate_information" do
+    it "includes group links that are guids" do
+      presented_groups = described_class.new(content_store_response).corporate_information
```
leenagupte, line 53: Perhaps set `subject` at the beginning of the describe block so we don't get presenters and models confused in the tests. I had to double check what type of test I was looking at!

e.g.

```ruby
subject(:corporate_information) { described_class.new(content_store_response).corporate_information }
```

then the expectation can use `expect(corporate_information.groups.first` etc

spec/requests/corporate_information_page_spec.rb
```diff
@@ -0,0 +1,24 @@
+RSpec.describe "Corporate Information Page" do
+  before do
+    content_store_has_example_item("/government/organisations/government-digital-service/about", schema: :corporate_information_page)
+    content_store_has_example_item("/government/organisations/government-digital-service/about/our-governance", schema: :corporate_information_page)
```
leenagupte, line 4: Why does the "our-governance" page need to be stubbed?

spec/system/corporate_information_page_spec.rb
```diff
@@ -0,0 +1,127 @@
+RSpec.describe "CorporateInformationPage" do
+  context "when visiting a Corporate Information page" do
+    before do
+      @content_store_response = GovukSchemas::Example.find("corporate_information_page", example_name: "corporate_information_page")
```
leenagupte, line 4: I'm surprised rubocop has moaned about this, but it normally recommends using a `let` rather than instance variables in tests.

leenagupte (commented):
This is making good progress. I've added some recommendations in line.

I think that the next thing we need to think about is the Contents List and how it's modelled and presented. It's something that's going to be used by a lot of document types, so it'll be good to get together as a group and dissect the problem.

spec/requests/corporate_information_page_spec.rb
deborahchua: I had to configure both `/about` and `/about/:slug` to render the show template in routes.rb so I though I should test both routes here.

app/views/corporate_information_page/show.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :title, "#{@presenter.page_title} - #{content_item.default_organisation.title} - GOV.UK" %>
```
KludgeKML, line 1: Again, in the original code they're testing if a default organisation is present. If it's possible one won't be, we should be testing too. If it's guaranteed to be there, we don't have to test but we should remove guard code in the previous commit.

app/presenters/corporate_information_page_presenter.rb
```diff
@@ -1,2 +1,49 @@
 class CorporateInformationPagePresenter < ContentItemPresenter
+  def initialize(content_item, view_context)
```
KludgeKML, line 2: We can probably get rid of the view context by including the link_helper, a la https://github.com/alphagov/frontend/pull/4687/commits/d94fa1399967c20c524e18fbdb8ae1352ff5c095

KludgeKML: (which in turn means we don't need this initializer at all)

app/presenters/corporate_information_page_presenter.rb
```diff
@@ -1,2 +1,49 @@
...
+
+  def further_information_types
+    %w[
+      publication_scheme
+      welsh_language_scheme
+      personal_information_charter
+      social_media_use
+      about_our_services
+    ]
+  end
```
KludgeKML, line 55: Not a blocking comment, but _maybe_ this isn't entirely presentational and it might make sense for it to be in the model. Maybe we should discuss

app/models/concerns/contents_list.rb
```diff
@@ -0,0 +1,32 @@
+module ContentsList
+  extend ActiveSupport::Concern
+
+  included do
```
KludgeKML, line 4: I don't think this actually needs to be in an included block, and we don't need to extend ActiveSupport::Concern - default module behaviour allows object method definitions (which this is)

KludgeKML: Actually, I wonder if this should just be a model - it appears to rely on the class it's included in having a body method/attribute, and that's the only thing that gets passed in. Perhaps one to discuss on Monday morning.

KludgeKML (changes requested):
Very nice, almost complete I think, but I have added a few not particularly serious notes - a couple of model-related ones that it might be best to talk about briefly.

app/models/corporate_information_page.rb
```diff
@@ -1,2 +1,9 @@
 class CorporateInformationPage < ContentItem
-end
\ No newline at end of file
+  def default_organisation
+    return if organisations.blank?
```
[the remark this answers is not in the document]
deborahchua: I've just checked the schema rules and the [default organisation ID is required](https://github.com/alphagov/publishing-api/blob/main/content_schemas/formats/corporate_information_page.jsonnet#L9) so I don't think we need a check in the view.

I don't think it's possible for organisations to be blank (there should be at least one) as you need an existing organisation to be able to create a corporate information page in Whitehall. I've removed the guard as you've suggested 👍

app/models/logo.rb
```diff
@@ -4,7 +4,7 @@ class Logo
   def initialize(logo)
     @image = get_image(logo["image"])
     @crest = logo["crest"]
-    @formatted_title = logo["formatted_title"]
+    @formatted_title = logo["formatted_title"].html_safe
```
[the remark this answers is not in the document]
deborahchua: Thanks! I've updated in commit [Update title in Logo model](https://github.com/alphagov/frontend/pull/4608/commits/efb5aaac3572a9acbe8b64f9c542ca9a75a0387e)

app/controllers/corporate_information_page_controller.rb
```diff
@@ -1,3 +1,5 @@
 class CorporateInformationPageController < ContentItemsController
-  def show; end
+  def show
+    @presenter = CorporateInformationPagePresenter.new(@content_item, view_context)
```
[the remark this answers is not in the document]
deborahchua: I've included the LinkHelper in the presenter so we no longer need the `view_context` - I've updated this in commit
[Add basic model, controller and presenter](https://github.com/alphagov/frontend/pull/4608/commits/598f04014359faa54f88ef0ea56c9710a585546e)

app/presenters/corporate_information_page_presenter.rb
deborahchua: I've updated this now in commit [Add corporate information group methods to CorporateInformationPage presenter](https://github.com/alphagov/frontend/pull/4608/commits/a8448d31cbed05af936a21b2c323be72be4bf802)

spec/models/corporate_information_page_spec.rb
```diff
@@ -17,4 +17,64 @@
...
+        expect(corporate_information.first[:links].last).to eq({ title: "Statistics at DHSC", path: "/government/organisations/department-of-health-and-social-care/about/statistics" })
+      end
+
+      it "excludes group links that do not have matching guids" do
+        groups = content_store_response.dig("details", "corporate_information_groups")
+        groups.first["contents"] << "123"
+
+        corporate_information = described_class.new(content_store_response).corporate_information
+
+        expect(corporate_information.first[:links].count).not_to eq(groups.count)
```
KludgeKML, line 61: Non-blocking note: You could make this `.to eq(groups.count - 1)`, I think? And it would be a bit of a tighter test, because it wouldn't pass if all the links were gone, or there were a thousand additional links? (I'm assuming there's more than one link, of course, so that setting the bogus GUID doesn't take this from 1 link to 0)

app/controllers/corporate_information_page_controller.rb
```diff
@@ -1,5 +1,6 @@
 class CorporateInformationPageController < ContentItemsController
   def show
     @presenter = CorporateInformationPagePresenter.new(@content_item)
+    @contents_outline_presenter = ContentsOutlinePresenter.new(@presenter.contents_list_headings)
```
KludgeKML, line 4: I think the preferred way of getting at a ContentsOutlinePresenter is to include it in the main presenter behind a method (as here: https://github.com/alphagov/frontend/blob/main/app/presenters/specialist_document_presenter.rb#L5-L9 ) - I appreciate I originally did have separate presenters when I first altered specialist documents, so this is on me.

spec/requests/corporate_information_page_spec.rb
```diff
@@ -0,0 +1,24 @@
+RSpec.describe "Corporate Information Page" do
```
KludgeKML, line 1: We might need an audit trail message for the system/request tests?

KludgeKML (changes requested):
Looking pretty good - I have a few comments, but most of them are about commit messages and one is just something I fear I led you astray on myself.

app/controllers/corporate_information_page_controller.rb
```diff
@@ -0,0 +1,3 @@
+class CorporateInformationPageController < ContentItemsController
```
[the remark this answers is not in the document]
deborahchua: Ah thank you! Yeah I wish it was possible to comment on just a commit, my workaround is to quote the commit title.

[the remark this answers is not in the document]
deborahchua: That's now updated in https://github.com/alphagov/frontend/pull/4608/commits/4fc126182f28b607f64352062dd304ae0ccdee36

app/controllers/corporate_information_page_controller.rb
deborahchua: No worries @KludgeKML , that makes sense to include in the main presenter! I've updated in https://github.com/alphagov/frontend/pull/4608/commits/c246acece6b021b63a0f5511ca5fa0ddccb989a3 - `contents_list_headings` is now a private method which is called from a new method`headers_for_contents_list_component` similar to that in specialist documents.

spec/requests/corporate_information_page_spec.rb
deborahchua: Good spot thanks @KludgeKML , I've updated in https://github.com/alphagov/frontend/pull/4608/commits/c21d7fa549aa9856b2c15ebe89dfd7c481c9587a

KludgeKML (approved):
LGTM!

app/models/corporate_information_page.rb
```diff
@@ -0,0 +1,61 @@
+class CorporateInformationPage < ContentItem
+  attr_reader :headers
+
+  def initialize(content_store_response)
+    super(content_store_response)
+
+    @headers = content_store_response.dig("details", "headers")
```
leenagupte, line 7: Hmmm, so `headers` aren't being modelled with ContentsOutline?

app/presenters/corporate_information_page_presenter.rb
```diff
@@ -0,0 +1,69 @@
...
+
+private
+
+  def contents_list_headings
+    content_item.headers << corporate_information_heading if content_item.corporate_information?
+
+    exclude_nested_headings
+
+    ContentsOutline.new(content_item.headers)
```
leenagupte, line 34: Oh, it's modelled here. Perhaps the modelling should be moved to the model for [consistency](https://github.com/alphagov/frontend/blob/main/app/models/specialist_document.rb#L15)?

app/presenters/corporate_information_page_presenter.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def exclude_nested_headings
```
leenagupte, line 37: Is there any restriction on the level? For example, if there are no level 2 headings, and level 3 headings are the top level in the headers hash should the contents list still be displayed?

leenagupte (commented):
Following the conversation in Slack I've added a couple of inline thoughts / questions.

app/presenters/corporate_information_page_presenter.rb
deborahchua: Ah yes, I did consider this initially, but both the model and presenter uses the method [corporate_information_heading](https://github.com/alphagov/frontend/pull/4608/commits/d45d542fd523fe4af3e71c878232454d9c75c88b#diff-59cebc1a6c28f24f1ed9f80839a0e0a226b94925414b84bfc32fd143d200f356R10) to show the header in the view, and to add it to the `content_item.headers` array.

app/presenters/corporate_information_page_presenter.rb
deborahchua: That is a great point, I hadn't thought of that!

I think I'll have to use [this additional check](https://github.com/alphagov/frontend/blob/4af199c3df5f52b86e37b50664c12e847046c0f1/app/models/contents_outline.rb#L10), similar to that in specialist documents - https://github.com/alphagov/frontend/blob/4af199c3df5f52b86e37b50664c12e847046c0f1/app/presenters/specialist_document_presenter.rb#L30

app/presenters/corporate_information_page_presenter.rb
leenagupte: Do you mean the view and the presenter? rather than model and presenter?

It's not a blocker. Just a question.

deborahchua: I've moved this to the model now in commit [Model headers for contents list using ContentsOutline model,](https://github.com/alphagov/frontend/pull/4608/commits/db22e38f0cb0c2689901b6016eb91d526cc6a561) as it does make more sense 👍

app/presenters/corporate_information_page_presenter.rb
deborahchua: I've updated this in commit [Update presenter with contents list headers](https://github.com/alphagov/frontend/pull/4608/commits/e830fee696735aad587a87ca1f08394eaec68a85)

KludgeKML (approved):
LGTM

AshGDS (dismissed):
EDIT: Ignore this, apologies, was looking at an older commit.

AshGDS (approved):
LGTM 👍

PR #4615, opened 2025-01-28
Upgrade to Heroku-24

hannako:
https://devcenter.heroku.com/articles/upgrading-to-the-latest-stack#updating-app-json


jon-kirwan (approved):
It looks good to me

PR #4632, opened 2025-02-11
Tidy styles

KludgeKML:

## What

Tidy up application.scss by moving styles that are only used by specific views into view-specific stylesheets.

## Why

Random styles in application.scss are harder to maintain (if a view is removed, it's not always obvious that you have to remove a system-wide style), and lead to a bigger download on unrelated pages.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -11,12 +11,6 @@ $govuk-include-default-font-face: false;
 // frontend mixins
 @import "mixins/margins";
 
-.gem-c-govspeak {
-  address {
-    font-style: normal;
-  }
-}
-
 .article-container {
   margin-bottom: govuk-spacing(6);
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -40,7 +40,3 @@ $govuk-include-default-font-face: false;
     @include responsive-bottom-margin;
   }
 }
-
-.bank-hols .govuk-panel {
-  margin-bottom: govuk-spacing(6);
-}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -41,27 +41,6 @@ $govuk-include-default-font-face: false;
   }
 }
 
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/views/_calendars.scss
```diff
@@ -52,3 +52,7 @@
     background-size: 231px 52px;
   }
 }
+
+.bank-hols .govuk-panel {
+  margin-bottom: govuk-spacing(6);
```
andysellick, line 57: This is interesting - is it me or is this only used in an AB test variant? The GOV.UK browser extension isn't showing a variant on the bank holidays page right now, so is this needed?

If it is needed, it's worth pointing out that the panel component now has the component wrapper helper, which includes an option for margin bottom. So I _think_ this CSS could be removed and the code in `bank_holidays.html.erb` could be updated to be something like the below.

```
<% if division.upcoming_event %>
  <%
    margin_bottom = 6 if variant_b_page?
  %>
  <% rendered_panel = render "govuk_publishing_components/components/panel", {
    prepend: t("common.next_holiday_in_is", :in_nation => t("#{division.title}_in")),
    title: division.upcoming_event.date == Time.zone.today ? t("common.today") : l(division.upcoming_event.date, format: "%e %B"),
    append: division.upcoming_event.title,
    margin_bottom: margin_bottom
  } %>
  <%= rendered_panel %>
<% end %>

```

andysellick: Also if you decide to leave this as-is, does this CSS need including in the view, as per the previous commit?

KludgeKML: The view stylesheet was already included in the view, but looks like the component version of this is better, so I'll do that.

KludgeKML: Okay, have updated how this works, and confirmed that it still results in a small visible difference on the bank holidays page when the B variant is selected.

andysellick: Nice 👍

PR #4633, opened 2025-02-11
Pass in example name to meta tags shared examples

leenagupte:

## What

Pass in example name to meta tags shared examples

## Why

Not all content schema examples have an example that matches the document type. Rather the shared examples should take the example name as a parameter and then use that to calculate the path value.

## Screenshots?

N/A


KludgeKML (approved):
Nice.

PR #4644, opened 2025-02-17
Fix Sass mixin error

jon-kirwan:
## What

Fix Sass mixin error

## Why

Produces an invalid CSS propery value


MartinJJones (approved):
Good spot, thanks for fixing 👍

PR #4650, opened 2025-02-19
Remove pfg scaffolding

KludgeKML:

## What

Remove PfG scaffolding code

## Why

PfG is now live, so we don't need the scaffolding that was necessary to test it before go-live. The ability to load content items from the local path has been 

Relies on / Cannot be merged before: https://github.com/alphagov/publishing-api/pull/3153


spec/support/content_item_helpers.rb
leenagupte, line 1: I think this commit message isn't quite right. It makes sure that t

leenagupte (commented):
I think it's a great that the scaffolding is being removed. 🎉 

I have a couple of inline comments / suggestions.

KludgeKML:
Closing - better idea here: https://github.com/alphagov/frontend/pull/4656

PR #4656, opened 2025-02-24
Remove PfG scaffolding and simplify content items.

KludgeKML:

## Why

[Trello card?](url)


spec/models/landing_page_spec.rb
```diff
@@ -1,103 +1,64 @@
...
-          {
-            "type" => "exact",
-            "path" => "/landing-page",
-          },
-        ],
-      }.to_json,
-    )
-    GdsApi::Response.new(http_response)
-  end
+  subject(:landing_page) { ContentItemFactory.build(landing_page_example) }
```
leenagupte, line 2: Why is `ContentItemFactory.build` being called here rather than `described_class.new`?

leenagupte (commented):
It's looking good, but I do have an inline question about a test setup.

spec/models/landing_page_spec.rb
KludgeKML: Oh yeah, this is just a failure to follow all the way down (I was fixated on removing the helper, forgot this example could go even further). Thanks! fixed now.

leenagupte (approved):
👍

PR #4660, opened 2025-02-25
Add inverse flag to govuk_styled_links_list helper

leenagupte:

### DIFF
--- a/app/helpers/link_helper.rb
+++ b/app/helpers/link_helper.rb
@@ -16,7 +16,7 @@ module LinkHelper
     "<a href='#{path}' class='#{classes}'>".html_safe + text + "</a>".html_safe
   end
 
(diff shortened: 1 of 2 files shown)


app/helpers/link_helper.rb
```diff
@@ -17,6 +17,6 @@ def govuk_styled_link(text, path: nil, inverse: false)
   end
 
   def govuk_styled_links_list(links)
```
KludgeKML, line 19: Ah, I think I was assuming that this would be an all or nothing thing - that you'd have:

`def govuk_styled_links_list(links, inverse: false)`

...is there a market for having some links in a single list inverse, and some not?

leenagupte: Hmm, that's a good shout. Will update.

PR #4661, opened 2025-02-26
Add check for available translations

leenagupte:

## What

Reinstate the original logic from government-frontend to check for the presence of translations.

## Why

The available translations method looks through the `available_translations` and replaces the English translation with `""`.

That means that if there's only an English translation, then the `available_translations` will return `[""]`.

`any?` was being used in the view to determine whether or not to display the translations partial, however, `any?` returns true for `[""]`.

```
frontend(dev)> a = [""]
=> [""]
frontend(dev)> a.any?
=> true
frontend(dev)> a.length > 1
=> false
```

So rather than not rendering anything when there is only an English translation, an empty `div` was being rendered:

`<div class="govuk-grid-column-one-third"></div>`

Instead of `any?` the original logic from government-frontend is being reinstated.

This logic has been made into a method in the model so that it can be easily tested.

[government-frontend](https://github.com/alphagov/government-frontend/blob/85f4681feed9ed3b57bb7f2ea0dce10bf6e0338a/app/views/shared/_translations.html.erb#L1)

See: https://github.com/alphagov/frontend/pull/4600#pullrequestreview-2644341439


deborahchua (approved):

KludgeKML (commented):
Do we know _why_ the available translations is an array with an empty string? I feel like it should either be an empty array (and this is unnecessary), or an array with "en" in it (and then this method gets modified slightly)

PR #4666, opened 2025-02-28
Remove user survey banners references

jon-kirwan:
## What

Remove user survey banners references

https://trello.com/c/4tJWLAuf/3333-remove-custom-user-survey-banners-from-static, [Jira issue NAV-15329](https://gov-uk.atlassian.net/browse/NAV-15329)

## Why

Code simplification.
> The survey was removed nearly 3 years ago, but the code is still included on production. https://github.com/alphagov/static/pull/2772
Removing the code would provide a small performance improvement to our end users, it would also help avoid copying code to other applications when moving away from Slimmer / Static.

### DIFF
--- a/app/views/static_error_pages/_error_page.html.erb
+++ b/app/views/static_error_pages/_error_page.html.erb
@@ -6,7 +6,6 @@
 
 <%= render "govuk_publishing_components/components/layout_for_public", {
   full_width: false,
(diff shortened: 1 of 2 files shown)


KludgeKML (approved):
LGTM! Thanks, I was wondering what we were going to do about this.

MartinJJones (approved):
The changes look good to me 👍

PR #4667, opened 2025-02-28
Change logo alignment on csv previews

beccapearce:
## What

On CSV preview pages sometimes multiple organisation logos are present, this can happen when there’s machinery of government (MOG) changes. We leave the closed orgs tagged to the content for context of which org first published or has been involved in the content as well as adding the new and current organisation.

Since we want to keep the content tagged to both organisations, we wanted to change the logos to line up horizontally across the page such as on [html publication pages](https://www.gov.uk/government/publications/digital-inclusion-action-plan-first-steps/digital-inclusion-action-plan-first-steps) instead of vertically down the page.

## Why

[Trello card](https://trello.com/c/7LxcBuOw/3342-review-duplicate-organisation-logos-on-csv-preview-pages), [Jira issue NAV-15582](https://gov-uk.atlassian.net/browse/NAV-15582)

## How

We've added a new css class to add this styling to the logos on CSV preview pages.

## Screenshots

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/app/assets/stylesheets/views/_csv_preview.scss
+++ b/app/assets/stylesheets/views/_csv_preview.scss
@@ -31,3 +31,25 @@
     outline: 3px solid govuk-colour("yellow");
   }
 }
+
+.organisation-logos {
+  list-style-type: none;
+  padding: 0;
+}
+
+.organisation-logos__logo {
+  display: inline-block;
+  vertical-align: top;
+  padding-bottom: govuk-spacing(3);
+  margin-right: govuk-spacing(6);
+
+  & img {
+    @include govuk-media-query($until: desktop) {
+      width: 140px;
+    }
+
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
Nice work!, I like the approach here of keeping the logo layout consistent with the approach used in `government-frontend` - https://github.com/alphagov/government-frontend/blob/d946e851a1fd4eae4c292c37022c451421eaf036/app/assets/stylesheets/helpers/_organisation-logos.scss

It does make me wonder if we should add a new layout feature to the [organisation logo component](https://components.publishing.service.gov.uk/component-guide/organisation_logo) in the future in the govuk_publishing_components gem, I'll have a further look at this and create an issue for it if needed

PR #4672, opened 2025-03-03
Update contributors handling

KludgeKML:

## What

Update `ContentItem#contributors` so that it returns a list of contributor content item objects rather than an array of hashed link values.

## Why

The hashes are presentational (they're to be passed into the govuk_styled_links_list helper method), so better that they're in presenter code. This lets us simplify how things are passed around a little, and is extensible to allow individual presenters to handle unorthodox additions to contributor lists (eg speaker_without_profile in speech items)

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -5,6 +5,12 @@ class ContentItemPresenter
     @content_item = content_item
   end
 
+  def contributor_links
+    content_item.contributors.map do |content_item|
+      { text: content_item.title, path: content_item.base_path }
+    end
+  end
+
   def page_title
     content_item.withdrawn? ? "[Withdrawn] #{content_item.title}" : content_item.title
   end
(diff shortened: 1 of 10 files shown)


app/presenters/content_item_presenter.rb
```diff
@@ -5,6 +5,12 @@ def initialize(content_item)
     @content_item = content_item
   end
 
+  def contributor_links
+    content_item.contributors.map do |content_item|
+      { "title" => content_item.title, "base_path" => content_item.base_path }
```
leenagupte, line 10: If you're moving this to the presenter, rather remodelling a cut-down content item, I'd change the string keys to symbols, and change them and the params for `govuk_styled_link_list` to be `text` and `path` so it matches `govuk_styled_link`.

KludgeKML: Done!

deborahchua (approved):
LGTM

leenagupte (approved):

PR #4675, opened 2025-03-05
Update documentation around calendars

KludgeKML:
- this was missed when the last set of calendar rendering changes were made.

### DIFF
--- a/docs/calendars.md
+++ b/docs/calendars.md
@@ -1,32 +1,37 @@
 # Calendars
 
-Calendar JSON data files are stored in `lib/data/<type>.json`, with a `divisions` hash for separate data per region (`united-kingdom`, `england-and-wales`, `scotland` or `northern-ireland`).
+## Bank Holidays
+
+The calendar data is in `lib/data/bank-holidays.json`, with a `divisions` hash for separate data per region (`united-kingdom`, `england-and-wales`, `scotland` or `northern-ireland`).
 
 Each type's data file contains a list of divisions, containing a list of years, each with a list of events:
 
 ```json
 {
-  "title": "UK bank holidays",
-  "description": "UK bank holidays calendar - see UK bank holidays and public holidays for 2012 and 2013",
+  "title": "bank_holidays.calendar.title",
+  "description": "bank_holidays.calendar.description",
   "divisions": {
     "england-and-wales": {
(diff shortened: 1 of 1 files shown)


docs/calendars.md
```diff
@@ -1,32 +1,42 @@
...
 ```
 
-The division `title` attribute is optional.  If this is not present the slug will be humanized and used instead.
+The strings are all localisation paths, which allows the calendar to be available in English at /bank-holidays and
+in Welsh at /gwyliau-banc
+
+## When do the clocks change
+
+The calendar data is in `lib/data/when-do-the-clocks-change.json`. Since the clocks change in all four regions at
+the same time, there is only one common division. As with the bank holiday data, all strings are localisation paths (although at the moment this data exists only for english)
```
yndajas, line 32: **nitpick:** missing full stops on these paragraphs

docs/calendars.md
```diff
@@ -1,32 +1,42 @@
...
 
 ## Publishing Calendars
 
-You'll need to run the `publishing_api:publish_calendars` rake task against the `frontend` app on the `frontend` machine.
+These two calendars are special routes, published using
+
+`rails special_routes:publish`
 
-This will update <https://www.gov.uk/when-do-the-clocks-change> and <https://www.gov.uk/bank-holidays>.
+in `publishing-api`.
```
yndajas, line 40: **question:** should this be a single paragraph or use ```?

docs/calendars.md
```diff
@@ -1,32 +1,42 @@
...
 
-To test locally, you may need to `govuk-docker-up` on `publishing-api` in a separate shell, before you run the rake task. You may also need to run the rake task a couple of times if you encounter timeouts.
+Note that the routes do not need to be republished when data in the json file is changed - only if the routes are missing from content-store, or if the rendering app changes.
```
yndajas, line 42: **nitpick:** JSON 🙈

yndajas (approved):
Looks good! Just a few really unimportant nits

docs/calendars.md
KludgeKML: Done (the one ending in a path I generally hate putting full stops on, because it makes it harder to copy/paste, but I've fixed by making this a proper link)

PR #4657, opened 2025-02-24
Move service manual homepage

andysellick:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

To do:

- [x] Rename everything to `service_manual`

## Visual changes

Trello card: https://trello.com/c/ANPt33UZ/397-move-route-service-manual-from-government-frontend-to-frontend, [Jira issue PNP-7337](https://gov-uk.atlassian.net/browse/PNP-7337)

PR #4683, opened 2025-03-10
Refactor breadcrumb logic

andysellick:

## What
Clean up the logic for whether breadcrumbs should be shown or not.

- clean up the layout view logic for whether to show breadcrumbs by moving to a application helper method
- refactor to be shorter and cleaner
- remove specific root path logic as can instead rely on homepage content item schema name
- update tests

Thanks to @KludgeKML for mostly writing this 👍 

## Why
Came up in https://github.com/alphagov/frontend/pull/4657 - we'll need to include the service manual homepage in the list of pages that don't show the breadcrumb, but there wasn't a clean way of doing that.

## Visual changes
None.

Trello card: https://trello.com/c/ANPt33UZ/508-move-route-service-manual-from-government-frontend-to-frontend, [Jira issue PNP-7337](https://gov-uk.atlassian.net/browse/PNP-7337)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -21,13 +21,13 @@ module ApplicationHelper
     request.original_fullpath.split("?", 2).first
   end
 
-  def remove_breadcrumbs(content_item)
-    remove_breadcrumbs = false
+  def show_breadcrumbs?(content_item)
+    return false if content_item.nil?
 
-    if content_item.respond_to?(:is_a_landing_page?) && content_item.is_a_landing_page?
-      remove_breadcrumbs = true
-    end
+    no_breadcrumbs_for = %w[homepage landing_page]
 
-    remove_breadcrumbs
+    return false if no_breadcrumbs_for.include?(content_item.schema_name)
+
+    true
   end
 end
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -41,28 +41,28 @@ RSpec.describe ApplicationHelper do
     end
   end
 
-  describe "#remove_breadcrumbs" do
-    describe "when is_landing_page? is true" do
+  describe "it doesn't show breadcrumbs" do
+    describe "when the content item is a homepage" do
       let(:content_item) { ContentItem.new({ "schema_name" => "landing_page" }) }
 
-      it "removes breadcrumbs" do
-        expect(remove_breadcrumbs(content_item)).to be(true)
+      it "does not show breadcrumbs" do
+        expect(show_breadcrumbs?(content_item)).to be(false)
       end
     end
 
-    describe "when is_landing_page? is false" do
+    describe "when the content item is a landing page" do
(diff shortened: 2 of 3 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -41,28 +41,28 @@
     end
   end
 
-  describe "#remove_breadcrumbs" do
+  describe "#show_breadcrumbs?" do
```
leenagupte, line 44: I think perhaps these tests need to be amended to when the content item is a homepage, landing_page or is nil rather than referring to "is_landing_page?"

spec/helpers/application_helper_spec.rb
```diff
@@ -41,28 +41,28 @@
...
     describe "when is_landing_page? is true" do
       let(:content_item) { ContentItem.new({ "schema_name" => "landing_page" }) }
 
       it "removes breadcrumbs" do
```
leenagupte, line 48: I'd rename this test to "it doesn't show breadcrumbs"

spec/helpers/application_helper_spec.rb
```diff
@@ -41,28 +41,28 @@
...
-        expect(remove_breadcrumbs(content_item)).to be(true)
+        expect(show_breadcrumbs?(content_item)).to be(false)
       end
     end
 
     describe "when is_landing_page? is false" do
       let(:content_item) { ContentItem.new({ "schema_name" => "a_different_page" }) }
 
       it "does not remove breadcrumbs" do
```
leenagupte, line 56: And rename this test to "it shows breadcrumbs" etc

app/helpers/application_helper.rb
```diff
@@ -21,13 +21,13 @@ def current_path_without_query_string
     request.original_fullpath.split("?", 2).first
   end
 
-  def remove_breadcrumbs(content_item)
-    remove_breadcrumbs = false
+  def show_breadcrumbs?(content_item)
+    return false if content_item.nil?
```
leenagupte, line 25: Is content_item ever `nil`?

leenagupte (commented):
This refactor is so much nicer! Getting rid of that `unless` in the view is lovely.

I have a couple of minor inline comments.

spec/helpers/application_helper_spec.rb
andysellick: Done 👍

spec/helpers/application_helper_spec.rb
andysellick: Done 👍

app/helpers/application_helper.rb
andysellick: I don't know, but the original logic included this check so I wasn't confident in removing it.

KludgeKML: Yes, it can be nil because not all controllers are derived from ContentItemsController (I think there's an argument for a further refactoring where _everything_ has a content_item - some would just be fake - but that's outside the scope of this PR.

KludgeKML (approved):
LGTM (thanks @leenagupte for the improvements to the test suite)

PR #4674, opened 2025-03-04
Move news articles from government frontend

KludgeKML:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

Add rendering support for content items with schema_name "news_article" into Frontend, mirroring existing code in Government Frontend.

Note that this PR doesn't include the GraphQL trial code for news_article, that will be added in a subsequent PR. (https://github.com/alphagov/frontend/pull/4696)

## Why

https://trello.com/c/6qhFX0hH/329-move-newsarticle-from-government-frontend-to-frontend, [Jira issue PNP-7335](https://gov-uk.atlassian.net/browse/PNP-7335)

## Example pages on preview apps
- https://govuk-frontend-app-pr-4674.herokuapp.com/government/news/the-personal-independence-payment-amendment-regulations-2017-statement-by-paul-gray (news story)
- https://govuk-frontend-app-pr-4674.herokuapp.com/government/news/christmas-2016-prime-ministers-message (with history mode and translations)
- https://govuk-frontend-app-pr-4674.herokuapp.com/government/news/industrial-strategy-launch-to-hardwire-stability-for-investors (press release)

### News story

image

### News story with history mode (in English)

image

### News story with history mode (in Urdu)

image

### Press release

image

PR #4689, opened 2025-03-11
Migrate service manual homepage

andysellick:

## What
Migrates the [service manual homepage](https://www.gov.uk/service-manual) from `government-frontend` to here. Supersedes https://github.com/alphagov/frontend/pull/4657

Todo:

- [x] add a system test (see https://github.com/alphagov/frontend/pull/4600/commits/cbe6fe59b08e92b852b2ff438b04475adfadee29 for example)
- [x] test that the search box is scoped to the service manual, add to system test
- [x] add a request test (see https://github.com/alphagov/frontend/pull/4600/commits/9e5876bea354ee8fbcc0f7d827827233f0a75c0a)

## Why
Part of the app consolidation work.

Related PRs:

- https://github.com/alphagov/service-manual-publisher/pull/1711

## Visual changes
Includes a number of visual changes including several requested by the Service Manual team.

- use standard GOV.UK header (this has now been done on the current Service Manual homepage)
- removal of the phase banner
- use of the [cards component](https://components.publishing.service.gov.uk/component-guide/cards) for the list of Service Manual topics
- minor spacing changes around the two sections at the bottom of the page, plus removing the horizontal lines above those headings as unnecessary (plus now the cards have the same style lines, seemed like too many lines)

Trello card: https://trello.com/c/ANPt33UZ/508-move-route-service-manual-from-government-frontend-to-frontend, [Jira issue PNP-7337](https://gov-uk.atlassian.net/browse/PNP-7337)

### DIFF
--- a/app/models/service_manual_homepage.rb
+++ b/app/models/service_manual_homepage.rb
@@ -1,0 +1,5 @@
+class ServiceManualHomepage < ContentItem
+  def topics
+    linked("children")
+  end
+end
--- a/app/views/service_manual/index.html.erb
+++ b/app/views/service_manual/index.html.erb
@@ -1,0 +1,1 @@
+<p>placeholder</p>
--- a/app/views/service_manual/index.html.erb
+++ b/app/views/service_manual/index.html.erb
@@ -1,47 +1,55 @@
-<% content_for :title, "Service Manual" %>
-
-<header class="app-hero">
-  <div class="app-hero__content govuk-width-container">
-    <div class="govuk-main-wrapper">
-      <div class="govuk-grid-row">
-        <div class="govuk-grid-column-two-thirds">
-          <%= render "govuk_publishing_components/components/heading", {
-            text: "Service Manual",
-            font_size: "xl",
-            heading_level: 1,
-            inverse: true,
-            margin_bottom: 6,
+<%
+  content_for :title, "Service Manual"
+  content_for :main_classes, "govuk-!-padding-top-0"
+%>
+<%= render "govuk_publishing_components/components/inverse_header", {
+  full_width: true,
+  padding_top: 8,
+  padding_bottom: 6,
+} do %>
+  <div class="govuk-width-container">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <%= render "govuk_publishing_components/components/heading", {
+          text: "Service Manual",
+          font_size: "xl",
+          heading_level: 1,
+          inverse: true,
+          margin_bottom: 6,
+        } %>
+        <% lead_para = capture do %>
+          Helping teams to create and run great public services that meet the <%= link_to "Service Standard", "/service-manual/service-standard", class: "govuk-link govuk-link--inverse" %>.
+        <% end %>
+        <%= render "govuk_publishing_components/components/lead_paragraph", {
+          text: lead_para,
+          inverse: true,
+        } %>
+        <form
+          action="/search"
+          method="get"
+          role="search"
+          data-module="ga4-form-tracker"
+          data-ga4-form-include-text
+          data-ga4-form-no-answer-undefined
+          data-ga4-form='{"event_name": "search", "type": "content", "url": "/search/all", "section": "Service Manual", "action": "search"}'>
+          <input type="hidden" name="filter_manual" value="/service-manual">
+          <%= render "govuk_publishing_components/components/search", {
+            label_text: "Search the service manual",
+            name: "q",
+            on_govuk_blue: true,
+            type: "search",
           } %>
-          <p class="govuk-body-lead app-hero-lead app-hero__body--inverse govuk-!-padding-bottom-1">
-            Helping teams to create and run great public services that meet the <%= link_to "Service Standard", "/service-manual/service-standard", class: "govuk-link govuk-link--inverse" %>.
-          </p>
-          <form
-            action="/search"
-            method="get"
-            role="search"
-            data-module="ga4-form-tracker"
-            data-ga4-form-include-text
-            data-ga4-form-no-answer-undefined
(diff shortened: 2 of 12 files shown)


app/views/service_manual/index.html.erb
```diff
@@ -1,47 +1,55 @@
...
-      <div class="govuk-grid-row">
-        <div class="govuk-grid-column-two-thirds">
-          <%= render "govuk_publishing_components/components/heading", {
-            text: "Service Manual",
-            font_size: "xl",
-            heading_level: 1,
-            inverse: true,
-            margin_bottom: 6,
+<%
+  content_for :title, "Service Manual"
```
leenagupte, line 2: I think this is supposed to be "Service Manual - GOV.UK"

In other formats it's been rendered as `<% content_for :title, "#{content_item.title} - GOV.UK" %>`

app/views/service_manual/index.html.erb
```diff
@@ -1,47 +1,55 @@
...
+  content_for :main_classes, "govuk-!-padding-top-0"
+%>
```
leenagupte, line 4: The meta description needs to be added separately too.

It populates this field in the page source:

```html
<meta name="description" content="Helping government teams create and run great digital services that meet the Service Standard.">
```

```ruby
<% content_for :extra_headers do %>
  <meta name="description" content=content="<%= strip_tags(@content_item.description) %>">
<% end %>
```

leenagupte: The existing Service Manual page doesn't have open graph meta tags, so it's probably to leave them off here too.

leenagupte (commented):
This is looking great! 🎉 

I've added a couple of inline comments, but they are really minor.

The changes you've made to use components rather than bespoke styling is awesome. The only question I have about that are the missing lines above the "The Service Standard" and "Communities of practice" headings. Was that a conscious decision?

e.g.
Screenshot 2025-03-14 at 14 39 21

app/models/service_manual_homepage.rb
```diff
@@ -0,0 +1,5 @@
+class ServiceManualHomepage < ContentItem
+  def topics
+    linked("children")
```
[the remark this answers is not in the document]
andysellick: Added, thanks.

app/views/service_manual/index.html.erb
andysellick: Thanks, have updated in a separate commit.

andysellick:
@leenagupte thanks for the review, have made changes. Regarding the missing lines, yes, this was a conscious decision - they don't add anything to the page (the two sections are differentiated in appearance already) and lines like this now appear between the cards already. Adding more dividing lines seemed like overkill for a relatively short page like this. Have updated the PR description to mention this.

spec/system/service_manual_homepage_spec.rb
```diff
@@ -0,0 +1,69 @@
+RSpec.describe "Service Manual homepage" do
```
KludgeKML, line 1: Only note I have is that in this file a lot of the expectation values are written out explicitly rather than as paths into the example response (eg: https://github.com/alphagov/frontend/pull/4674/commits/5e41596180cacef25003076ed141eafad8e3fcf3). I'm in two minds about this, because I personally prefer things written out explicitly, but I have nothing to back up that preference rather than "I like it", and the idiom of the changes we've made recently has been to have the paths into the response, so this should probably be following that.

andysellick: Thanks @KludgeKML. I think the only test where the text comes from the content item is line 48, so I've updated that one, but let me know if there are others I can change.

KludgeKML (approved):
LGTM!

AshGDS:
@andysellick Just wanted to check, was it a conscious decision to use `gem_layout_full_width` over `gem_layout_full_width_no_footer_navigation` for the migrated page?

andysellick:
> @andysellick Just wanted to check, was it a conscious decision to use `gem_layout_full_width` over `gem_layout_full_width_no_footer_navigation` for the migrated page?

It wasn't, no. What's the difference between the two? Is the second one more appropriate in this situation?

AshGDS:
@andysellick The live version of this page has a smaller footer, which I believe is achieved using `gem_layout_full_width_no_footer_navigation`: https://www.gov.uk/service-manual

andysellick:
Thanks @AshGDS that's good to know, I didn't realise there was a difference. I think we should probably stick with the current option, given our general push towards simplifying things and also that the team was happy to switch to the common site header, would be inconsistent to have a different footer now.

hannako:
@andysellick and @AshGDS 

Moving away from the custom footer on these pages (ie using `gem_layout_full_width` instead of `gem_layout_full_width_no_footer_navigation`) has been agreed and is the expectation of Ben Carpenter the product owner, as part of the simplification and standardisation of these pages. Daniel Susser was on the email chain, but I'll forward that to you now Andy for visibility.

app/views/service_manual/index.html.erb
```diff
@@ -52,21 +52,11 @@
...
-            heading_level: 2,
-            font_size: "s",
-            margin_bottom: 1,
-          } %>
-          <p class="govuk-body-s"><%= topic.description %></p>
-        </div>
-      <% end %>
-    </div>
-  <% end %>
+  <%= render "govuk_publishing_components/components/cards", {
```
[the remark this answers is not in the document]
hannako: Is it fine that we're not sanitising the topic title set in service manual publisher anymore? https://service-manual-publisher.integration.publishing.service.gov.uk/topics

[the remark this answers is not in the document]
KludgeKML: Hmm. There's no explanation in the commit that introduced the sanitisation as to why it was necessary: https://github.com/alphagov/service-manual-frontend/commit/cd6b754dafa7b3935f56ab086e3509c27e3282d6 

We don't generally sanitise titles, do we?

PR #4696, opened 2025-03-12
(Re-)Enable GraphQL Trial support for news_article documents

KludgeKML:

## What

Adds the ability for content items to be loaded from the GraphQL endpoint on publishing-api. For news_articles only, frontend will default to loading from the content store as usual unless the param `graphql=true` is appended to the path. If the env var `GRAPHQL_FEATURE_FLAG=true`, if will default to loading from GraphQL unless the param `graphql=false` is appended to the path. 

## Why

The publishing platform team are testing GraphQL as an option for querying content at the moment (as per [RFC-172](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-172-graphql-for-govuk.md)). The only schema they currently support is news_article, and since that has been moved over to frontend, 

https://trello.com/c/rKkLJkhX/336-add-graphql-support-to-frontend

## How

Replicate the edition query from government-frontend, then update the ContentItemLoader to allow it 1) access to the request that it was created for (which means it can read the `graphql` param), and 2) allow it to load from GraphQL if the content item is in the allowed list of schemas (in this case just news_article)

## Screenshots?

No visual changes expected (except that GraphQL news articles don't return related links, so some pages will not have them where they would if served from the content store.)

### DIFF
--- a/app/queries/graphql/edition_query.rb
+++ b/app/queries/graphql/edition_query.rb
@@ -1,0 +1,90 @@
+class Graphql::EditionQuery
+  def initialize(base_path)
+    @base_path = base_path
+  end
+
+  def query
+    <<-QUERY
+        {
+          edition(
+            base_path: "#{@base_path}",
+            content_store: "live",
+          ) {
+            ... on Edition {
+              base_path
+              description
+              details {
+              body
+                change_history
+                default_news_image {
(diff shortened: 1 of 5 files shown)


deborahchua (approved):
Looks good to me! 👍

PR #4700, opened 2025-03-14
Render non filterable text facets in a list

leenagupte:

## What

There's a case where a facet could have an array of allowed values but
not be filterable, so a link to a finder shouldn't be rendered.

This was not being picked up so rather than rendering:

```
Reason for protection: Application to UK scheme from 2021
```

A hash was rendered instead:

```
Reason for protection:
{:label=>"Application to UK scheme from 2021", :value=>"uk-gi-after-2021"}
```

## Why

This case was missed when the specialist documents were moved over.

## How

A new type is added to account for this case "preset_text". This is
"text" that the publisher selected from a redefined set of options, but
is not filterable so shouldn't be a link.

### On GOV.UK (government-frontend)
Screenshot 2025-03-14 at 10 48 37

### Before (frontend)
Screenshot 2025-03-14 at 10 19 22

### After
Screenshot 2025-03-14 at 10 43 53

### DIFF
--- a/app/models/specialist_document.rb
+++ b/app/models/specialist_document.rb
@@ -31,6 +31,8 @@ class SpecialistDocument < ContentItem
 
       f[:type] = if link?(selected_facet, f[:value])
                    "link"
+                 elsif preset_text?(selected_facet, f[:value])
+                   "preset_text"
                  else
                    selected_facet["type"]
                  end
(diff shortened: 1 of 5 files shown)


deborahchua (approved):
LGTM!

KludgeKML (approved):
LGTM2!

PR #4720, opened 2025-03-26
[DO NOT MERGE] Local authority api with Fake results (leave in place for the moment)

KludgeKML:

## What

This is a version of this actual PR:
https://github.com/alphagov/frontend/pull/4770

## Why

to add dummy data so that this endpoint can be called from a preview app (which otherwise wouldn't work since they can't access locations-api or Local Links Manager)


KludgeKML:
Not needed any more (actual endpoint is live)

PR #4733, opened 2025-04-02
Redirect old csv preview urls

catalinailie:

Trello card https://trello.com/c/kwSXPt3Y/3392-redirect-old-asset-csv-preview-urls, [Jira issue NAV-1867](https://gov-uk.atlassian.net/browse/NAV-1867)

### DIFF
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -9,7 +9,7 @@ class CsvPreviewController < ApplicationController
 
     return error_410 if @asset["deleted"] || @asset["redirect_url"].present?
     if draft_asset? && served_from_live_host?
-      redirect_to(Plek.find("draft-origin", external: true) + request.path, allow_other_host: true) and return
+      redirect_to(URI.parse(Plek.find("draft-origin", external: true) + request.path), allow_other_host: true) and return
     end
 
     parent_document_uri = @asset["parent_document_url"]
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -8,8 +8,8 @@ class CsvPreviewController < ApplicationController
     @asset = GdsApi.asset_manager.asset(params[:id]).to_hash
 
     return error_410 if @asset["deleted"] || @asset["redirect_url"].present?
-    if draft_asset? && served_from_asset_host?
-      redirect_to(Plek.find("draft-assets") + request.path, allow_other_host: true) and return
+    if draft_asset? && served_from_live_host?
+      redirect_to(Plek.find("draft-origin", external: true) + request.path, allow_other_host: true) and return
     end
 
     parent_document_uri = @asset["parent_document_url"]
@@ -52,7 +52,7 @@ private
     @asset["draft"] == true
   end
 
-  def served_from_asset_host?
-    request.hostname == URI.parse(Plek.find("assets")).hostname
+  def served_from_live_host?
+    request.hostname == URI.parse(Plek.website_root).hostname
   end
 end
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -23,7 +23,7 @@ class CsvPreviewController < ApplicationController
       redirect_to(parent_document_uri, status: :see_other, allow_other_host: true) and return
     end
 
-    return cacheable_404 if @attachment_metadata["content_type"] != "text/csv"
+    return cacheable_404 unless csv_content_type.include?(@attachment_metadata["content_type"])
 
     @csv_rows, @truncated = CsvPreviewService
       .new(GdsApi.asset_manager.media(params[:id], params[:filename]).body)
@@ -54,5 +54,9 @@ private
 
   def served_from_live_host?
     request.hostname == URI.parse(Plek.website_root).hostname
+  end
+
+  def csv_content_type
+    ["text/csv", "application/csv"]
   end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/csv_preview_redirect_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  before_action { expires_in(1.day, public: true) }
+
+  def redirect
+    host = if served_from_draft_asset_host?
+             # PLEK_HOSTNAME_PREFIX is set on draft environments,
+             # which means a 'draft-' prefix will be added to this host
+             Plek.find("origin", external: true)
+           else
+             Plek.find("www", external: true)
+           end
```
hannako, line 11: In the live stack:
- Integration environment `Plek.find("www", external: true)` returns: https://www.integration.publishing.service.gov.uk 👍 
- Staging environment: `Plek.find("www", external: true)` returns: https://www.staging.publishing.service.gov.uk 👍
- But in production it returns https://www.publishing.service.gov.uk/ and I think we'd want it to return https://www.gov.uk

I think we might want to just fetch the [website_root](https://github.com/alphagov/plek/blob/3933b6a4d999984b58d2bd4b971e03b33e6588f0/lib/plek.rb#L115) instead: `Plek.website_root`

app/controllers/csv_preview_controller.rb
```diff
@@ -8,8 +8,8 @@ def show
     @asset = GdsApi.asset_manager.asset(params[:id]).to_hash
 
     return error_410 if @asset["deleted"] || @asset["redirect_url"].present?
-    if draft_asset? && served_from_asset_host?
-      redirect_to(Plek.find("draft-assets") + request.path, allow_other_host: true) and return
+    if draft_asset? && served_from_origin_host?
```
hannako, line 11: In the first commit, anything [routed](https://github.com/alphagov/govuk-helm-charts/blob/a787602eb4734d74babaf943eaa9a9dbb805eb8d/charts/app-config/values-production.yaml#L1141-L1154) to this application from the assets domain is being [redirected](https://github.com/alphagov/frontend/blob/3fda530eff0f2fe1f1860548b532fb7001dd8842/config/routes.rb#L81) to your new `/csv-preview` route via the CsvPreviewRedirectController.

Anything being routed to the show action of the CsvPreviewController will be coming from the gov.uk domain as there's nothing in helm charts that routes the /csvpreview path with the assets host to this application.

TLDR: I don't think you need to check if the draft asset is coming is `served_from_origin_host?`

I'm also a bit confused by in the previous commit you have:

```
host = if served_from_draft_asset_host?
             # PLEK_HOSTNAME_PREFIX is set on draft environments,
             # which means a 'draft-' prefix will be added to this host
             Plek.find("origin", external: true)
```

But here we've got `redirect_to(Plek.find("draft-origin", external: true)`

Won't this lead to the `draft-draft-origin` problem?


So I don't think there's any value in checking that the draft asset has come from www.gov.uk

hannako (changes requested):
Thanks Catalina - this is a really tricky one!! I have a feeling we need to make some small changes before we ship this, but I might be wrong so please have a read and let me know if I've got the wrong end of the stick (which is a definite possibility 😂 )

config/routes.rb
```diff
@@ -78,7 +78,7 @@
   get "/find-licences/:slug/:authority_slug/:interaction", to: "licence_transaction#authority_interaction", as: "licence_transaction_authority_interaction"
 
   # Old style media previews (compatible with preview_url)
-  get "/media/:id/:filename/preview", to: "csv_preview#show", filename: /[^\/]+/
+  get "/media/:id/:filename/preview", to: "csv_preview_redirect#redirect", filename: /[^\/]+/
```
hannako, line 81: nit pick! Please could we amend the comment, to lose `(compatible with preview_url)` as it wasn't and isn't!

app/controllers/csv_preview_controller.rb
catalinailie: Hmm but saying this, I'm not sure we need this redirect at all. With the old `/media` route, it made sense because the preview url was created using the asset url. But now, the `/csv-preview` link using the existing page domain, so it will always match

catalinailie: And about the `redirect_to(Plek.find("draft-origin", external: true)` response, it won't be `draft-draft-origin` because `PLEK_HOSTNAME_PREFIX` is set only on [draft environments](https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/values-production.yaml#L1417:L1418), and this case will happen only on origin hosts.

app/controllers/csv_preview_redirect_controller.rb
catalinailie: Ohh I totally missed this one. Thank you 🙌

app/controllers/csv_preview_controller.rb
catalinailie: Hmm haven't thought about that edge case when I implemented this. More like when somebody types the url manually and misses the draft-origin domain while typing it. See my comment above.
But it will also cover that case yes, although not making things work for the user since the regular users don't have access to our draft- pages.

hannako: I think we should leave it in. Is this easier to follow:

```ruby
if draft_asset? && served_from_live_host?
  redirect_to (draft_host + request.path, allow_other_host: true) and return
end

def served_from_live_host?
  request.hostname != URI.parse(draft_host).hostname
end

def draft_host
  @draft_host ||= Plek.find("draft-origin", external: true)
end
```

hannako (approved):
Thank you for all your sterling work on this - it's looking really good. Just one request please to amend the specs: 

In the first commit, could you update the the system specs so that the main url being tested in every spec is your new url,  rather than the now legacy one? Hopefully it will be as easy as updating [this line](https://github.com/alphagov/frontend/blob/main/spec/system/csv_preview_spec.rb#L6)

PR #4735, opened 2025-04-02
Add conditional check for metadata block component

GDSNewt:
This change will make rendering the blue metadata box at the top of specialist documents dependent on a value defined in the schema of each finder.

This is a replication of a recently [implemented government frontend feature](https://github.com/alphagov/government-frontend/pull/3639), in anticipation of the specialist document migration to frontend.


hannako (approved):
LGTM

app/presenters/specialist_document_presenter.rb
```diff
@@ -22,6 +22,10 @@ def important_metadata
     end
   end
 
+  def show_metadata_block?
+    content_item.to_h.dig("links", "finder", 0, "details", "show_metadata_block")
```
leenagupte, line 26: I think this code should be in the `Finder` model.

So create an attribute in the that model called `show_metadata_block` that does this `dig`

Then this method could be changed to call

`finder.show_metadata_block`

leenagupte: We're keeping model and presenter code separate where possible, so anything that does a `dig` on `content_store_response` should be in the model.

leenagupte (commented):
I've added a couple of recommendations for how the `show_metadata_block` should be modelled.

app/presenters/specialist_document_presenter.rb
GDSNewt: I've made the suggested changes 🙂

app/models/finder.rb
```diff
@@ -1,9 +1,10 @@
 class Finder < ContentItem
-  attr_reader :facets
+  attr_reader :facets, :show_metadata_block
 
   def initialize(content_store_response)
     super(content_store_response)
 
     @facets = content_store_response.dig("details", "facets")
+    @show_metadata_block = to_h.dig("details", "show_metadata_block")
```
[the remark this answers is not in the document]
leenagupte: Could you add a [test](https://github.com/alphagov/frontend/blob/main/spec/models/finder_spec.rb) for this attribute as well?

leenagupte (commented):
Thanks for making the requested change. 

There's just two small inline comments. Apart from that it's looking good.

app/models/finder.rb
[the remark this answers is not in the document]
GDSNewt: Amended 😃

leenagupte (approved):
Looks good 👍

PR #4736, opened 2025-04-03
Add A/B testing for GraphQL on supported document types

brucebolt:
, [Jira issue PP-2556](https://gov-uk.atlassian.net/browse/PP-2556)This will allow us to serve a defined amount of traffic from GraphQL, to measure response times in a real environment.

Note: all requests where the GraphQL variant is in use will involve a query to Publishing API's GraphQL endpoint, since we need to determine whether the document is a news article before deciding how to render it.  This should give us a good indication of how our implementation would cope with all documents types being migrated to use GraphQL.

[Trello card](https://trello.com/c/stxnESVI)

### DIFF
--- a/lib/content_item_loader.rb
+++ b/lib/content_item_loader.rb
@@ -43,7 +43,26 @@ private
   end
 
   def use_graphql?
-    request&.params&.[]("graphql") == "true"
+    return false unless request
+
+    if request.params["graphql"] == "true"
+      return true
+    elsif request.params["graphql"] == "false"
+      return false
+    end
+
+    if request.headers
+      ab_test = GovukAbTesting::AbTest.new(
+        "GraphQLNewsArticles",
+        allowed_variants: %w[A B Z],
+        control_variant: "Z",
+      )
+      @requested_variant = ab_test.requested_variant(request.headers)
+
+      return true if @requested_variant.variant?("B")
+    end
+
+    false
   end
 
   def use_local_file?
--- a/spec/unit/content_item_loader_spec.rb
+++ b/spec/unit/content_item_loader_spec.rb
@@ -3,11 +3,18 @@ require "gds_api/test_helpers/publishing_api"
 RSpec.describe ContentItemLoader do
   include GdsApi::TestHelpers::PublishingApi
   include ContentStoreHelpers
+  include GovukAbTesting::RspecHelpers
 
   subject(:content_item_loader) { described_class.new }
 
   let!(:item_request) { stub_content_store_has_item("/my-random-item") }
   let(:graphql_query) { Graphql::EditionQuery.new("/my-random-item").query }
+
+  before do
+    GovukAbTesting.configure do |config|
+      config.acceptance_test_framework = :active_support
+    end
+  end
 
   describe ".for_request" do
     it "returns a new object per request" do
@@ -39,8 +46,8 @@ RSpec.describe ContentItemLoader do
     end
 
     it "restricts cache to the specific instance of the class, so does not cache across requests" do
-      request_1 = instance_double(ActionDispatch::Request, path: "/my-random-item", env: {}, params: {})
-      request_2 = instance_double(ActionDispatch::Request, path: "/my-random-item", env: {}, params: {})
+      request_1 = instance_double(ActionDispatch::Request, path: "/my-random-item", env: {}, headers: {}, params: {})
+      request_2 = instance_double(ActionDispatch::Request, path: "/my-random-item", env: {}, headers: {}, params: {})
 
       loader_1 = described_class.for_request(request_1)
       loader_2 = described_class.for_request(request_2)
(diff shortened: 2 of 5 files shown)


lib/content_item_loader.rb
```diff
@@ -43,7 +43,22 @@ def load_from_sources(base_path)
...
 
   def use_graphql?
-    request&.params&.[]("graphql") == "true"
+    if request && request.params["graphql"] == "true"
+      return true
+    elsif request && request.params["graphql"] == "false"
+      return false
+    end
+
+    if request && request.headers
```
mike3985, line 52: We're checking that `request` isn't nil for every conditional in this method. It might be nice to just check that once at the beginning of the method?

lib/content_item_loader.rb
```diff
@@ -43,7 +43,22 @@ def load_from_sources(base_path)
...
+      ab_test = GovukAbTesting::AbTest.new(
+        "GraphQLRoles",
+        allowed_variants: %w[A B Z],
+        control_variant: "Z",
+      )
+      @requested_variant = ab_test.requested_variant(request.headers)
+
+      true if @requested_variant.variant?("B")
+    end
```
mike3985, line 61: With these changes, there are scenarios where this method can now return nil instead of just true or false

spec/unit/content_item_loader_spec.rb
```diff
@@ -80,16 +203,90 @@
...
+        end
+      end
+    end
+
+    context "with ALLOW_LOCAL_CONTENT_ITEM_OVERRIDE=true" do
+      before do
+        ENV["ALLOW_LOCAL_CONTENT_ITEM_OVERRIDE"] = "true"
+        stub_const("ContentItemLoader::LOCAL_ITEMS_PATH", "spec/fixtures/local-content-items")
+      end
+
```
mike3985, line 271: It looks like the additions below this line are accidental duplications. Unless I'm failing a spot the difference test 😵‍💫

lib/content_item_loader.rb
```diff
@@ -43,7 +43,22 @@ def load_from_sources(base_path)
...
-    request&.params&.[]("graphql") == "true"
+    if request && request.params["graphql"] == "true"
+      return true
+    elsif request && request.params["graphql"] == "false"
+      return false
+    end
+
+    if request && request.headers
+      ab_test = GovukAbTesting::AbTest.new(
+        "GraphQLRoles",
```
mike3985, line 54: I was expecting Frontend to be handling the News Articles A/B test and Collections to be handling Roles?

mike3985: I just double-checked for my own sanity... Collections is already handling the Roles test. I guess this is just a copy-paste error 😄

lib/content_item_loader.rb
brucebolt: I hadn't spotted that.  I'll add a guard clause at the beginning of the method.

lib/content_item_loader.rb
brucebolt: I think that would've been the case before it there's no `request`, since the safe nav operator will return nil.

lib/content_item_loader.rb
brucebolt: Oops.  I'll update this to `GraphQLNewsArticles`.

lib/content_item_loader.rb
mike3985: I still think the previous version was doing the right thing because it would always ultimately compare the result of that safe-navigated-to value with `"true"`

mike3985: I think it's good form for a Ruby `?` predicate method to produce booleans in general?

brucebolt: I've updated to always return a boolean value, including an explicit `return false` for when none of the conditions have been met.

mike3985 (approved):

lib/content_item_loader.rb
```diff
@@ -43,7 +43,26 @@ def load_from_sources(base_path)
...
-    request&.params&.[]("graphql") == "true"
+    return false unless request
+
+    if request.params["graphql"] == "true"
+      return true
+    elsif request.params["graphql"] == "false"
+      return false
+    end
+
+    if request.headers
```
KludgeKML, line 54: There isn't a test for if request.headers is nil (we also appear to have lost coverage of line 36 above, testing if there's a graphql response, but it's not in the allowed schemas?

brucebolt: I've added a test for headers being `nil` and also ensured we're testing on the B variant for schemas that aren't allowed.

KludgeKML: Hmm - this line is still showing as only being covered for the case when request.headers is not nil. (I can see that line 36 is uncovered in the main branch though, so I'll take care of that, but you might have to do a small rebase)

brucebolt: Looks like I never pushed the branch after adding the test for the header being nil 🤦‍♂️.

KludgeKML (approved):
LGTM!

lib/content_item_loader.rb
KludgeKML: Thanks!

PR #4737, opened 2025-04-03
Remove GraphQL feature flag

brucebolt:
, [Jira issue PP-2556](https://gov-uk.atlassian.net/browse/PP-2556)This is not being used for anything at the moment, since we are manually selecting GraphQL by adding a query parameter to requests.

Removing in preparation for adding a A/B test for GraphQL instead.

[Trello card](https://trello.com/c/stxnESVI)

### DIFF
--- a/lib/content_item_loader.rb
+++ b/lib/content_item_loader.rb
@@ -43,10 +43,7 @@ private
   end
 
   def use_graphql?
-    return true if request && request.params["graphql"] == "true"
-    return false if request && request.params["graphql"] == "false"
-
-    ENV["GRAPHQL_FEATURE_FLAG"] == "true"
+    request&.params&.[]("graphql") == "true"
   end
 
   def use_local_file?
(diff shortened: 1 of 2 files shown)


lib/content_item_loader.rb
```diff
@@ -43,10 +43,11 @@ def load_from_sources(base_path)
...
   def use_graphql?
-    return true if request && request.params["graphql"] == "true"
-    return false if request && request.params["graphql"] == "false"
-
-    ENV["GRAPHQL_FEATURE_FLAG"] == "true"
+    if request && request.params["graphql"] == "true"
+      true
+    elsif request && request.params["graphql"] == "false"
+      false
+    end
```
yndajas, line 50: **suggestion:** could this be simplified?

```rb
return request&.params["graphql"] == "true"
```

yndajas: Ah yeah, it doesn't like you making any assumptions after a safe nav, which I guess makes sense, given all the subsequent methods in the chain will be called on `nil`

yndajas (approved):

PR #4740, opened 2025-04-04
Add conditional check for contents list

GDSNewt:
This change will make rendering the table of contents component in specialist documents dependent on a value defined in the schema of each finder. We [recently implemented a similar feature for the metadata block ](https://github.com/alphagov/frontend/pull/4735). 

Tests in this PR depend on example schemas from the Publishing API repository, which may cause CI failures until the related changes are merged.

New HRMC document:
> Screenshot 2025-04-25 at 13 07 36

Preexisting document type (CMA case):
- **all documents will require a republish to set the field to true as default**
> Screenshot 2025-04-25 at 13 12 45

**Associated PRs**

[Publishing API](https://github.com/alphagov/publishing-api/pull/3265)
[Government-Frontend](https://github.com/alphagov/government-frontend/pull/3645)
[Specialist Publisher](https://github.com/alphagov/specialist-publisher/pull/3083)

[Trello](https://trello.com/c/2eakEGOo/3570-make-it-possible-to-hide-the-table-of-contents-on-specialist-documents)

### DIFF
--- a/app/presenters/specialist_document_presenter.rb
+++ b/app/presenters/specialist_document_presenter.rb
@@ -3,7 +3,7 @@ class SpecialistDocumentPresenter < ContentItemPresenter
   include LinkHelper
 
   def headers_for_contents_list_component
-    return [] unless content_item.headers.level_two_headers?
+    return [] unless show_table_of_contents?
 
     ContentsOutlinePresenter.new(content_item.headers).for_contents_list_component
   end
@@ -24,6 +24,10 @@ class SpecialistDocumentPresenter < ContentItemPresenter
 
   def show_metadata_block?
     content_item.finder.show_metadata_block
+  end
+
+  def show_table_of_contents?
+    content_item.headers.level_two_headers? && content_item.finder.show_table_of_contents
   end
 
   def protection_image_path
--- a/app/models/finder.rb
+++ b/app/models/finder.rb
@@ -1,10 +1,11 @@
 class Finder < ContentItem
-  attr_reader :facets, :show_metadata_block
+  attr_reader :facets, :show_metadata_block, :show_table_of_contents
 
   def initialize(content_store_response)
     super(content_store_response)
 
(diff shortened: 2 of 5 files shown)


app/presenters/specialist_document_presenter.rb
```diff
@@ -26,6 +26,10 @@ def show_metadata_block?
     content_item.finder.show_metadata_block
   end
 
+  def hide_contents_list?
+    content_item.finder.hide_contents_list
+  end
```
leenagupte, line 31: I've mentioned this in the other PRs, but this method should be renamed to `show_contents_list?` with the appropriate change in logic.

app/views/specialist_document/show.html.erb
```diff
@@ -43,7 +43,9 @@
       <% end %>
     <% end %>
 
-    <%= render "govuk_publishing_components/components/contents_list_with_body", contents: @contents_outline_presenter.for_contents_list_component do %>
+      <% if @presenter.show_table_of_contents_list? %>
```
leenagupte, line 46: The guard clause can't be added here as it means the body won't display at all if there aren't any headers in the content.

What we've done in the past is add the guard clause to the for_contents_list_component.
e.g. https://github.com/alphagov/frontend/blob/1dae521cfd339f1187220e0fa68aeb5982e7963c/app/presenters/specialist_document_presenter.rb#L12
       
That way `[]` will be passed as `contents` to the component, but the body content in the block will still render.

As it stands here nothing will be rendered.

lauraghiorghisor-tw: Just got around to this. 

All updated now, plus an additional context added in the presenter spec. 

Struggled to name the flag, we were between "show_table_of_contents" and "show_headers_for_contents_list" but decided on the first as the second is too close to the implementation and it might become obsolete when/if the components change.

leenagupte (approved):

PR #4741, opened 2025-04-07
Request multiple levels of taxons in GraphQL queries

brucebolt:
, [Jira issue PP-2876](https://gov-uk.atlassian.net/browse/PP-2876)At the moment, we only retrieve the parent taxon when requesting content from GraphQL.
    
Taxonomy exists up to five levels, so we need to request all five levels on every request to allow us to properly form the breadcrumbs needed to render a news article.

[Trello card](https://trello.com/c/8JajAb6r)

### DIFF
--- a/spec/requests/news_article_spec.rb
+++ b/spec/requests/news_article_spec.rb
@@ -1,6 +1,7 @@
 require "gds_api/test_helpers/publishing_api"
 
 RSpec.describe "News Article" do
+  include Capybara::RSpecMatchers
   include GdsApi::TestHelpers::PublishingApi
 
   describe "GET show" do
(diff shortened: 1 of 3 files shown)


spec/requests/news_article_spec.rb
```diff
@@ -1,28 +1,64 @@
...
-      stub_content_store_has_item(base_path, content_item)
-    end
+      before do
+        stub_content_store_has_item(base_path, content_item)
+      end
 
-    it "succeeds" do
-      get base_path
+      it "succeeds" do
+        request
```
yndajas, line 18: **thought:** would it be clearer what's happening if `get base_path` was moved to the `before` block, rather than using RSpec's lazy let/variable assignment in each example?

yndajas (approved):

spec/requests/news_article_spec.rb
brucebolt: That's a good idea.  Looking back, I'm not really sure why I did it like this in the first place 😕.

PR #4742, opened 2025-04-07
Field of operation migration

AshGDS:
, [Jira issue PNP-6443](https://gov-uk.atlassian.net/browse/PNP-6443)## What / Why
- Migrates these routes to `frontend`:
  - https://www.gov.uk/government/fields-of-operation/iraq
  - https://www.gov.uk/government/fields-of-operation/united-kingdom
  - https://www.gov.uk/government/fields-of-operation/afghanistan
  - https://www.gov.uk/government/fields-of-operation/northern-ireland
  - https://www.gov.uk/government/fields-of-operation/other-locations
- PR version of these pages can be accessed through here: https://govuk-frontend-app-pr-4742.herokuapp.com/government/fields-of-operation
- Part of the app consolidation work. Trello card: https://trello.com/c/qbIWfPqV

## Visual Changes
I've done some work to hide the `Field of operation` heading when no `description` exists. The code was meant to be doing this already, but it was checking if the `description` was blank which wasn't working, as an empty HTML element was being returned as the description instead of an empty string. Therefore I've used `strip_tags` to ensure it also doesn't render when it's given an empty HTML element.

See https://govuk-frontend-app-pr-4742.herokuapp.com/government/fields-of-operation/united-kingdom versus https://www.gov.uk/government/fields-of-operation/united-kingdom - the live example has a "Field of operation" heading with nothing underneath it

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ some hard-coded routes.
 |Case studies           |[case_study](https://docs.publishing.service.gov.uk/content-schemas/case_study.html)|https://www.gov.uk/government/case-studies/aiding-capability-decision-making-for-the-royal-navy|
 |Cookies                |hardcoded|https://www.gov.uk/help/cookies|
 |Fatality notice        |[fatality_notice](https://docs.publishing.service.gov.uk/content-schemas/fatality_notice.html)|https://www.gov.uk/government/fatalities/corporal-lee-churcher-dies-in-iraq|
+|Field of operation     |[field_of_operation](https://docs.publishing.service.gov.uk/document-types/field_of_operation.html)|https://www.gov.uk/government/fields-of-operation/united-kingdom|
 |Fields of operation    |[fields_of_operation](https://docs.publishing.service.gov.uk/content-schemas/fields_of_operation.html)|https://www.gov.uk/government/fields-of-operation|
 |Find electoral office  |hardcoded|https://www.gov.uk/contact-electoral-registration-office|
 |Find local council     |hardcoded|https://www.gov.uk/find-local-council|
--- a/app/models/field_of_operation.rb
+++ b/app/models/field_of_operation.rb
@@ -1,0 +1,5 @@
+class FieldOfOperation < ContentItem
+  def fatality_notices
+    linked("fatality_notices")
+  end
+end
--- a/app/presenters/field_of_operation_presenter.rb
+++ b/app/presenters/field_of_operation_presenter.rb
@@ -17,7 +17,7 @@ class FieldOfOperationPresenter
   end
 
   def description
-    description = @content_item["description"]
+    description = @content_item.description
 
     description.html_safe if description.present?
   end
@@ -25,8 +25,16 @@ class FieldOfOperationPresenter
   def contents
     contents = []
     contents << { href: "#field-of-operation", text: "Field of operation" } if description.present?
-    contents << { href: "#fatalities", text: "Fatalities" } if fatality_notices.present?
+    contents << { href: "#fatalities", text: "Fatalities" } if @content_item.fatality_notices.present?
 
     contents
+  end
+
+  def roll_call_introduction(fatality_notice)
+    fatality_notice.content_store_response.dig("details", "roll_call_introduction")
+  end
+
+  def casualties(fatality_notice)
+    fatality_notice.content_store_response.dig("details", "casualties")
   end
 end
--- a/app/presenters/field_of_operation_presenter.rb
+++ b/app/presenters/field_of_operation_presenter.rb
@@ -1,12 +1,9 @@
-class FieldOfOperationPresenter < ContentItemPresenter
-  include ContentItem::HeadingAndContext
+class FieldOfOperationPresenter
 
-  def heading_and_context
-    super.tap do |t|
-      t[:context] = I18n.t("field_of_operation.context")
-      t[:text] = "#{I18n.t('field_of_operation.title')} #{@content_item['title']}"
-      t[:font_size] = "xl"
-    end
+  attr_reader :content_item
+
+  def initialize(content_item)
+    @content_item = content_item
   end
 
   def organisation
--- a/app/presenters/field_of_operation_presenter.rb
+++ b/app/presenters/field_of_operation_presenter.rb
@@ -1,5 +1,4 @@
 class FieldOfOperationPresenter
-
   attr_reader :content_item
 
   def initialize(content_item)
@@ -7,9 +6,8 @@ class FieldOfOperationPresenter
   end
 
   def organisation
-    org = @content_item.dig("links", "primary_publishing_organisation", 0)
+    org = @content_item.content_store_response.dig("links", "primary_publishing_organisation", 0)
     logo = org.dig("details", "logo")
-
     {
       name: logo["formatted_title"].html_safe,
       url: org["base_path"],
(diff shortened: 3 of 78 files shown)


README.md
```diff
@@ -26,6 +26,7 @@ some hard-coded routes.
 |Homepage               |[homepage](https://docs.publishing.service.gov.uk/content-schemas/homepage.html)|https://www.gov.uk/|
 |Get involved           |[get_involved](https://docs.publishing.service.gov.uk/content-schemas/get_involved.html)|https://www.gov.uk/government/get-involved|
 |Licence finder         |schema: [specialist_document](https://docs.publishing.service.gov.uk/content-schemas/specialist_document.html)|https://www.gov.uk/find-licences/premises-licence|
+|Field of operation     |[field_of_operation](https://docs.publishing.service.gov.uk/document-types/field_of_operation.html)|https://www.gov.uk/government/fields-of-operation/united-kingdom|
```
KludgeKML, line 29: Can we stick this in alphabetical order, please?

app/presenters/field_of_operation_presenter.rb
```diff
@@ -29,20 +27,6 @@ def description
     description.html_safe if description.present?
   end
 
-  def fatality_notices
```
deborahchua: I think as part of this commit we can include the presenter tests that were updated/added in commits [Migrate field_of_operation presenter test from government-frontend](https://github.com/alphagov/frontend/pull/4742/commits/5bdda531974a6981163f8a5e8060693019c7f0f1) and [Make existing presenter tests pass](https://github.com/alphagov/frontend/pull/4742/commits/13b190151b4d2f473b359cebafd76ef8b2a8bf1d)

I also think the commit [Add FieldOfOperation model](https://github.com/alphagov/frontend/pull/4742/commits/cfdb2aa91d60cce6cc63f734affd10c57db88da1) should either directly follow this commit, or is included as part of this commit, which helps follow the flow of changes.

AshGDS: @deborahchua Thanks - I've made this change 👍

README.md
AshGDS: Thanks @KludgeKML - I've done that now. Not sure how it ended up there!

app/models/field_of_operation.rb
```diff
@@ -0,0 +1,5 @@
+class FieldOfOperation < ContentItem
+  def fatality_notices
```
deborahchua, line 2: I think we can combine the commit [Add field_of_operation model test](https://github.com/alphagov/frontend/pull/4742/commits/633bb78c3cfe6ee129e82c1d552c78ed2bedc23d) into this commit [Add FieldOfOperation model](https://github.com/alphagov/frontend/pull/4742/commits/c5bea5832af58c38e8e2ffb277602bd77a7bc866)

spec/models/field_of_operation_spec.rb
```diff
@@ -0,0 +1,15 @@
+RSpec.describe FieldOfOperation do
+  describe "#field_of_operation" do
```
deborahchua, line 2: I think we can update this to be `fatality_notices` to match the name of the method being tested.

app/presenters/field_of_operation_presenter.rb
```diff
@@ -17,16 +17,24 @@ def organisation
...
   def contents
     contents = []
     contents << { href: "#field-of-operation", text: "Field of operation" } if description.present?
-    contents << { href: "#fatalities", text: "Fatalities" } if fatality_notices.present?
+    contents << { href: "#fatalities", text: "Fatalities" } if @content_item.fatality_notices.present?
 
     contents
   end
+
+  def roll_call_introduction(fatality_notice)
```
deborahchua, line 38: I think we can merge commit [Add new tests for roll_call_introduction and casualties presenter methods](https://github.com/alphagov/frontend/pull/4742/commits/b7349acae639c9df578c0fc019d53a378c2b3f88) into this commit [Modify presenter and view to render content](https://github.com/alphagov/frontend/pull/4742/commits/54e14349e9bf78123c34d7a5cfe0cd5cece82f06)

app/views/field_of_operation/show.html.erb
```diff
@@ -1,3 +1,6 @@
+<% content_for :extra_headers do %>
+  <meta name="description" content="<%= strip_tags(@content_item.description) %>">
```
deborahchua, line 2: I think we still need to include `<%= render "govuk_publishing_components/components/machine_readable_metadata", { content_item: content_item.to_h, schema: :article } %>` in this block, similar to https://github.com/alphagov/frontend/blob/main/app/views/fatality_notice/show.html.erb#L6

spec/system/field_of_operation_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+      within(".gem-c-heading__context") do
+        expect(page).to have_text("British fatalities")
+      end
+
+      locale = page.find("main div:first-of-type > .gem-c-heading")["lang"]
+      expect(locale).to eq content_store_response["locale"]
+    end
+
+    it "has the correct subheadings and text" do
```
deborahchua: I think we can add some additional tests where the fatality notice heading and list aren't shown if there are no notices (when [`@content_item.fatality_notices.any?`](https://github.com/alphagov/government-frontend/blob/8d2872c29486c4806cdd134cf1eb0a931f2a146c/app/views/content_items/field_of_operation.html.erb#L34) returns `false`)

You could wrap the existing tests in a `context "when there are fatality notices present" do` and have a separate context for "when there are no fatality notices present".

deborahchua (commented):
This is looking good! I was mid reviewing this end of last week and have a few more minor inline comments (mainly about squashing some commits and tests).

AshGDS:
Thanks @deborahchua - I've made all the requested changes 👍

app/views/field_of_operation/show.html.erb
AshGDS: Thanks, I've added this, though I can't see any difference in the HTML.

deborahchua: Oh! Maybe we don't need it anymore then? 🤔 I guess we can revisit and remove it from all the other migrated routes later?

deborahchua (approved):
Looks great! 👍

app/views/field_of_operation/show.html.erb
KludgeKML: Just to double-check: The machine readable metadata is right at the end of the HTML, is that right? So is it possible something else is adding it in this case (and either adding or not adding it in others)?

KludgeKML: This is the current metadata in the live page:
```
<script src="[/assets/frontend/application-633967b9ddd3282900f0cc893c3bf16fa13b1caa003db4d3cd022d3f02819108.js](https://www.gov.uk/assets/frontend/application-633967b9ddd3282900f0cc893c3bf16fa13b1caa003db4d3cd022d3f02819108.js)" type="module"></script><script type="application/ld+json">
  {
  "@context": "http://schema.org",
  "@type": "BreadcrumbList",
  "itemListElement": [
    {
      "@type": "ListItem",
      "position": 1,
      "item": {
        "name": "Home",
        "@id": "https://www.gov.uk/"
      }
    }
  ]
}
</script>
```

AshGDS: @KludgeKML @deborahchua Oh yeah, ignore me as Keith is right - I was wrongly looking in the `<head>` of the page for the difference in HTML.

PR #4750, opened 2025-04-09
Contents outline

KludgeKML:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

Add a ContentsOutline model and a presenter that will convert it into a format useful for the ContenstList component.

## Why

Specialist Documents, Corporate information pages, and the upcoming Flexible Page trial all have a concept of a list of links within the page generated from the content in some way (currently Specialist Documents have a nested list of headers passed to it, CI pages have to extract it directly from the content body , and Flexible Pages will define it in Flexible Section). This adds a model and presenter for the concept to simplify working with them.

https://trello.com/c/H2TSoDLE/

PR #4753, opened 2025-04-10
Contents outline refactor

KludgeKML:

## What

Refactor contents outline work added here (https://github.com/alphagov/frontend/pull/4750)

## Why

Refactor improvement opportunities noted after PR had been merged.

### DIFF
--- a/app/presenters/specialist_document_presenter.rb
+++ b/app/presenters/specialist_document_presenter.rb
@@ -3,6 +3,8 @@ class SpecialistDocumentPresenter < ContentItemPresenter
   include LinkHelper
 
   def headers_for_contents_list_component
+    return [] unless content_item.headers.level_two_headers?
+
     ContentsOutlinePresenter.new(content_item.headers).for_contents_list_component
   end
 
--- a/app/presenters/specialist_document_presenter.rb
+++ b/app/presenters/specialist_document_presenter.rb
@@ -2,6 +2,10 @@ class SpecialistDocumentPresenter < ContentItemPresenter
   include DateHelper
   include LinkHelper
 
+  def headers_for_contents_list_component
+    ContentsOutlinePresenter.new(content_item.headers).for_contents_list_component
+  end
+
   def show_protection_type_image?
     protected_food_drink_name? && content_item.protection_type_image.present?
   end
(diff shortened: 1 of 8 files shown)


app/presenters/specialist_document_presenter.rb
```diff
@@ -2,6 +2,10 @@ class SpecialistDocumentPresenter < ContentItemPresenter
   include DateHelper
   include LinkHelper
 
+  def headers_for_contents_list_component
```
leenagupte, line 5: Yes, adding the "show table of contents" guard clause here in the future will mimic the [previous behaviour](https://github.com/alphagov/frontend/blob/1dae521cfd339f1187220e0fa68aeb5982e7963c/app/presenters/specialist_document_presenter.rb#L12).

leenagupte: The issue about not showing the contents list if there aren't any `h2` headers hasn't been addressed. The "body" field is free text, so I think it's possible for `h3` and `h4` headers to be added in the body without an `h2` so I think we should put the check back in: https://github.com/alphagov/frontend/blob/1dae521cfd339f1187220e0fa68aeb5982e7963c/app/presenters/specialist_document_presenter.rb#L49

leenagupte (commented):
Something is happening with the tests. Apart from that it's looking good. I have an inline question about the `h2` header check.

app/presenters/specialist_document_presenter.rb
KludgeKML: Have added another commit to handle this.

app/presenters/specialist_document_presenter.rb
```diff
@@ -3,6 +3,8 @@ class SpecialistDocumentPresenter < ContentItemPresenter
   include LinkHelper
 
   def headers_for_contents_list_component
+    return [] unless content_item.headers.level_two_headers?
```
leenagupte, line 6: I think the test for an empty array being returned here needs to be added back in.

leenagupte (commented):
Apart from one missing test and one flaky test, I think this is looking good.

app/presenters/specialist_document_presenter.rb
KludgeKML: Done!

leenagupte (approved):
Looks good 👍

PR #4762, opened 2025-04-16
Run basic News Article tests against GraphQL, too

mike3985:
## What
1. Create a test fixture based on the vanilla News Article content schema example.
2. Run the basic News Article system tests against the fixture.

## Why

We're working towards having GraphQL producing content as close to Content Store's as is possible and reasonable. A bit more test coverage is going to help.

[Trello card](https://trello.com/c/HZRt4r50/1666-make-meta-tags-etc-of-graphql-version-of-news-articles-the-same-as-regular-version), [Jira issue PP-2906](https://gov-uk.atlassian.net/browse/PP-2906)

### DIFF
--- a/spec/fixtures/graphql/news_article_christmas_2016.json
+++ b/spec/fixtures/graphql/news_article_christmas_2016.json
@@ -1,0 +1,145 @@
+{
+  "data": {
+    "edition": {
+      "base_path": "/government/news/christmas-2016-prime-ministers-message",
+      "description": "Theresa May sends her best wishes to everyone celebrating Christmas in the UK and around the world.",
+      "details": {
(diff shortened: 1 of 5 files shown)


spec/fixtures/graphql/news_article_christmas_2016.json
```diff
@@ -0,0 +1,145 @@
...
+        "change_history": [
+           {
+              "note": "Added translation",
+              "public_timestamp": "2016-12-25T00:15:02.000+00:00"
+           }
+        ],
```
yndajas, line 13: **question:** did you get the data from the live version rather than the content schemas example? Some of this is missing from the latter

spec/system/news_article_spec.rb
```diff
@@ -54,4 +54,18 @@
...
+  context "when content item is from Content Store" do
+    let(:content_item) { content_store_has_example_item(path, schema: :news_article) }
+    let(:path) { "/government/news/christmas-2016-prime-ministers-message" }
+
+    it_behaves_like "a news article page"
+  end
+
+  context "when content item is from Publishing API's GraphQL" do
+    let(:content_item) { graphql_has_example_item("news_article_christmas_2016") }
+    let(:path) { "#{content_item.fetch('base_path')}?graphql=true" }
```
yndajas, line 67: **question:** is there a reason for only hard coding it here (unlike the Content Store example)?

yndajas (approved):

spec/system/news_article_spec.rb
mike3985: for _not_ hard coding it here? that's the result of a chaotic development process

spec/fixtures/graphql/news_article_christmas_2016.json
mike3985: I started with the GraphQL response in integration and then updated some values where they differed enough to upset the tests

mike3985: updated some values -> copied some values from the content schema example

spec/system/news_article_spec.rb
mike3985: they're now both hard coded

spec/requests/news_article_spec.rb
```diff
@@ -29,11 +28,8 @@
     end
 
     context "when loaded from GraphQL" do
-      let(:graphql_fixture) { fetch_graphql_fixture("news_article") }
-      let(:base_path) { graphql_fixture.dig("data", "edition", "base_path") }
-
       before do
-        stub_publishing_api_graphql_content_item(Graphql::EditionQuery.new(base_path).query, graphql_fixture)
+        base_path = graphql_has_example_item("news_article").fetch("base_path")
```
leenagupte, line 32: This should be in a `let` block. At the moment I think the test is picking up the value of `base_path` from line 9. If the test is passing, I wonder it's actually testing anything from the fixture?

spec/system/news_article_spec.rb
```diff
@@ -3,8 +3,8 @@
 
   it_behaves_like "it has meta tags", "news_article", "news_article"
 
-  context "when visiting a page" do
-    before { visit "/government/news/christmas-2016-prime-ministers-message" }
+  shared_examples "a news article page" do
```
leenagupte, line 6: I think these examples should be moved to the bottom of the test file as it's a little jarring to have the history mode tests between the shared example definitions and the tests that run them.

leenagupte (commented):
This looks good to me. I have a couple of minor inline comments.

spec/system/news_article_spec.rb
mike3985: I've taken your point but resolved it the opposite way: I've moved the contexts that actually use the shared examples above the history mode contexts (based on the understanding that the more basic/generic a test, the nearer the top of the test file it should be)

spec/requests/news_article_spec.rb
mike3985: I've made an attempt at making this better to read. None of the variables from the `let`s were actually being used in the example blocks, so I've gone ahead and moved all of those setup variables into their respective `before`s

spec/system/news_article_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe "Fatality Notice" do
+RSpec.describe "News Article" do
```
leenagupte, line 1: Thanks for fixing this. I've just realised that there is a typo in the commit message 🍝

leenagupte (approved):
👍 

There's a typo in the commit message that should be fixed, but doesn't need a re-review.

spec/system/news_article_spec.rb
mike3985: It's quite rare for my commit messages to not have typos, so you're probably right only I can't see it. But the pasta part was intentional, https://en.wikipedia.org/wiki/Copypasta#Technology 😋

leenagupte: In the future I wouldn't put intentional typos in as it just looks odd.

mike3985: Ha yeah, it wasn't the typo that was intentional, just the reference to copy-pastaing. I suppose I made some poor punctuation choices that resulted in what looked like a typo!

leenagupte: Next time please fix the typos (intentional or not) before merging.

PR #4764, opened 2025-04-17
Test image caption

yndajas:
In #4757 support was added for rendering image captions when GraphQL is the data source. This wasn't tested for pages rendered by Content Store or GraphQL, so this fixes that

[Trello](https://trello.com/c/AfhRJ4EP/1672-make-graphql-version-of-news-articles-include-figcaption), [Jira issue PP-2940](https://gov-uk.atlassian.net/browse/PP-2940)

### DIFF
--- a/spec/system/news_article_spec.rb
+++ b/spec/system/news_article_spec.rb
@@ -35,11 +35,26 @@ RSpec.describe "News Article" do
     end
   end
 
+  shared_examples "a news article page with an image caption" do
+    before { visit path }
+
+    it "renders a caption for the lead image" do
+      expect(page).to have_css("figcaption p", text: "British High Commissioner, Jane Marriott CMG OBE with Chief Guest, Ahsan Iqbal at the Islamabad KBP.")
+    end
+  end
+
   context "when content item is from Content Store" do
     let(:content_item) { content_store_has_example_item(path, schema: :news_article) }
     let(:path) { "/government/news/christmas-2016-prime-ministers-message" }
 
     it_behaves_like "a news article page"
+
+    context "when content item has an image caption" do
+      let(:path) { "/government/news/british-high-commission-marks-his-majesty-king-charles-iiis-birthday-with-brilliantly-british-celebrations" }
+      let(:content_item) { content_store_has_example_item(path, schema: :news_article, example: :news_article_with_image_caption) }
+
+      it_behaves_like "a news article page with an image caption"
+    end
   end
 
   context "when content item is from Publishing API's GraphQL" do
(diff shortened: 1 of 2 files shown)


spec/system/news_article_spec.rb
```diff
@@ -1,5 +1,8 @@
 RSpec.describe "News Article" do
-  let!(:content_item) { content_store_has_example_item("/government/news/christmas-2016-prime-ministers-message", schema: :news_article) }
+  let(:news_article_christmas_2016_path) { "/government/news/christmas-2016-prime-ministers-message" }
+  let(:path) { news_article_christmas_2016_path }
+  let(:example) { :news_article }
```
mike3985, line 5: I'm not convinced that this is simplification. Is it doing something other than deduplication?

mike3985: Fair dos. The approach with the duplicated `let!(:content_item) ...` was a bit more followable from where I'm looking and this simplified version has a bit less charm, but that preference is not strong enough for me to argue about 😆

mike3985 (approved):

hannako (approved):
LGTM

spec/system/news_article_spec.rb
leenagupte, line 5: I think it's more confusing to have these values here. For readability I'd assign them within the context they're used in.

spec/system/news_article_spec.rb
```diff
@@ -1,5 +1,8 @@
 RSpec.describe "News Article" do
-  let!(:content_item) { content_store_has_example_item("/government/news/christmas-2016-prime-ministers-message", schema: :news_article) }
+  let(:news_article_christmas_2016_path) { "/government/news/christmas-2016-prime-ministers-message" }
```
leenagupte, line 2: I'm not sure what the value of having this here is? I'd prefer duplication if it increases readability.

I think that trying to reduce the duplication has made the tests much harder to follow.

leenagupte (changes requested):
It's great that the extra tests have been added, however I'd remove the "simplication" commit and prioritise readble tests over DRY-ed up tests.

I've set this to "request changes" so that it can't be accidentally merged.

yndajas:
@leenagupte (cc: @mike3985) I think the 'simplification' felt a bit different before rebasing in Mike's more recent changes. Agree that the current approach is easier to follow now, so I've cut that out and left this as just adding the new shared example to two contexts

leenagupte (commented):
This is much easier to follow! 🎉  
Thanks for adding the tests to the "from content store" context as well.

leenagupte (approved):

PR #4769, opened 2025-04-22
Add single page notification button partial

sairamya93:

## What

Add partial for `single page notification button`

## Why

As part of app consolidation, `single_page_notification_button` partial is required for the document type `detailed_guide`. The partial would also be needed for other document types - `call_for_evidence`, `consultation`, `document_collection` and `publication`.

Trello card https://trello.com/c/L3F8O1R5/571-single-page-notification-button-partial-for-app-consolidation, [Jira issue PNP-6397](https://gov-uk.atlassian.net/browse/PNP-6397)

## How

Based on the code in government-frontend where the pages are currently rendered

### DIFF
--- a/app/models/concerns/single_page_notification_button.rb
+++ b/app/models/concerns/single_page_notification_button.rb
@@ -1,0 +1,19 @@
+module SinglePageNotificationButton
+  extend ActiveSupport::Concern
+
+  # Add the content id of the publication, detailed_guide or consultation that should be exempt from having the single page notification button
+  EXEMPTION_LIST = %w[
+    c5c8d3cd-0dc2-4ca3-8672-8ca0a6e92165
+    70bd3a76-6606-45dd-9fb5-2b95f8667b4d
+    a457220c-915c-4cb1-8e41-9191fba42540
+    5f9c6c15-7631-11e4-a3cb-005056011aef
+  ].freeze
+
+  def page_is_on_exemption_list?
+    EXEMPTION_LIST.include? content_id
+  end
+
+  def display_single_page_notification_button?
+    !page_is_on_exemption_list? && locale == "en"
+  end
+end
--- a/spec/support/concerns/single_page_notification_button.rb
+++ b/spec/support/concerns/single_page_notification_button.rb
@@ -1,0 +1,17 @@
+RSpec.shared_examples "it can have single page notifications" do |document_type, example_name|
+  let(:content_store_response) { GovukSchemas::Example.find(document_type, example_name:) }
+
+  it "knows it has single page notification for English pages and page is not in exemption list" do
+    expect(described_class.new(content_store_response).display_single_page_notification_button?).to be(true)
+  end
+
+  it "knows it does not have single page notification when page is on exemption list" do
+    content_store_response["content_id"] = "c5c8d3cd-0dc2-4ca3-8672-8ca0a6e92165"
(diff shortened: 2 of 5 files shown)


leenagupte (commented):
This is looking good.

I've added a few inline recommendations. 

I've assumed the concern will be added to the specific models for that render the single page notification button, and that's why it hasn't been included in anything yet.

app/models/concerns/single_page_notification_button.rb
```diff
@@ -0,0 +1,19 @@
...
+  # Add the content id of the publication, detailed_guide or consultation that should be exempt from having the single page notification button
+  EXEMPTION_LIST = %w[
+    c5c8d3cd-0dc2-4ca3-8672-8ca0a6e92165
+    70bd3a76-6606-45dd-9fb5-2b95f8667b4d
+    a457220c-915c-4cb1-8e41-9191fba42540
+    5f9c6c15-7631-11e4-a3cb-005056011aef
+  ].freeze
+
+  def page_is_on_exemption_list?
+    EXEMPTION_LIST.include? content_store_response["content_id"]
```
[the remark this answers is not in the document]
sairamya93: Hmm yes we could access the `content_id` directly when we include that concern in the model

sairamya93:
Yes. The tests will be included in the spec for the models that include this concern.

app/models/concerns/single_page_notification_button.rb
```diff
@@ -0,0 +1,19 @@
...
+    a457220c-915c-4cb1-8e41-9191fba42540
+    5f9c6c15-7631-11e4-a3cb-005056011aef
+  ].freeze
+
+  def page_is_on_exemption_list?
+    EXEMPTION_LIST.include? content_id
+  end
+
+  def display_single_page_notification_button?
+    !page_is_on_exemption_list? && locale == :en
```
leenagupte, line 17: I think the locale should be a string, so `"en"` rather than `:en`. Otherwise I'll think you'll find that the button will never be displayed.

spec/support/concerns/single_page_notification_button.rb
```diff
@@ -0,0 +1,17 @@
...
+    expect(described_class.new(content_store_response).display_single_page_notification_button?).to be(true)
+  end
+
+  it "knows it does not have single page notification when page is on exemption list" do
+    content_store_response["content_id"] = "c5c8d3cd-0dc2-4ca3-8672-8ca0a6e92165"
+    expect(described_class.new(content_store_response).display_single_page_notification_button?).to be(false)
+  end
+
+  it "knows it does not have single page notification for foreign language pages" do
+    content_store_response["locale"] = :cy
```
leenagupte, line 14: Same here regarding using a string rather than a symbol.

leenagupte (commented):
Two easy to fix inline comments.

It does make me think that being able to call the first shared example test would have been beneficial as it should have caught this. Perhaps it is worth creating a basic model for one of the new document types that just includes the concern and has one line in the tests to call the single page notification shared examples.

sairamya93:
I have added `DetailedGuide` model just to test the SinglePageNotification concern tests. The entire route will be added in the subsequent PR.

leenagupte (approved):
Thanks for making the updates and including a basic Detailed Guide model and tests ⭐

PR #4774, opened 2025-04-24
Add published_dates_with_notification_button partial

sairamya93:
, [Jira issue PNP-6369](https://gov-uk.atlassian.net/browse/PNP-6369)⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

Add `published_dates_with_notification_button` partial

## Why

As part of app consolidation, `published_dates_with_notification button` partial is required for the document type `detailed_guide`. The partial would also be needed for other document types - `call_for_evidence`, `consultation` and `publication`.

Trello card: https://trello.com/c/XJekbr1s

## How

Based on the code in government-frontend where the pages are rendered from currently.

### DIFF
--- a/app/views/shared/_published_dates_with_notification_button.html.erb
+++ b/app/views/shared/_published_dates_with_notification_button.html.erb
@@ -1,0 +1,41 @@
+<% add_view_stylesheet("published-dates-button-group") %>
+<% skip_account = skip_account || "false" %>
+<% user_account_present = local_assigns.fetch(:user_account, false) %>
+
+<%= render "govuk_publishing_components/components/published_dates", {
+    published: display_date(content_item.initial_publication_date),
+    last_updated: display_date(content_item.updated),
+    history: formatted_history(content_item.history),
+    margin_bottom: 5,
+  } %>
+
+<div class="published-dates-button-group">
+  <h2 class="govuk-visually-hidden">
+    <% if content_item.display_single_page_notification_button? %>
+      <%= I18n.t("common.email_and_print_link") %>
+    <% else %>
+      <%= I18n.t("common.print_link") %>
+    <% end %>
+  </h2>
+
+  <%= render "govuk_publishing_components/components/single_page_notification_button", {
+    base_path: content_item.base_path,
+    js_enhancement: user_account_present,
+    button_location: "bottom",
+    margin_bottom: 3,
+    skip_account: skip_account,
+    ga4_data_attributes: {
+      module: "ga4-link-tracker",
+      ga4_link: {
+        event_name: "navigation",
+        type: "subscribe",
+        index_link: 2,
+        index_total: 2,
+        section: "Footer",
+      },
+    },
+  } if content_item.display_single_page_notification_button? %>
+  <%= render "govuk_publishing_components/components/print_link", {
+    margin_bottom: 8,
+  } %>
+</div>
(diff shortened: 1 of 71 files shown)


app/views/shared/_published_dates_with_notification_button.html.erb
```diff
@@ -0,0 +1,40 @@
...
+  <h2 class="govuk-visually-hidden">
+    <% if @content_item.display_single_page_notification_button? %>
+      <%= I18n.t("common.email_and_print_link") %>
+    <% else %>
+      <%= I18n.t("common.print_link") %>
+    <% end %>
+  </h2>
+
+  <%= render "govuk_publishing_components/components/single_page_notification_button", {
+    base_path: @content_item.base_path,
```
KludgeKML, line 21: Could lose the @ from content_item? (here and below)

app/views/shared/_published_dates_with_notification_button.html.erb
```diff
@@ -0,0 +1,40 @@
...
+    js_enhancement: @has_govuk_account,
```
KludgeKML, line 22: Do we know what this does?

app/views/shared/_published_dates_with_notification_button.html.erb
deborahchua, line 1: Can we update the audit trail links in the commit description? I think this file has changed slightly since I made the commit you've cherrypicked.

spec/views/shared/_published_dates_with_notification_button.html.erb_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      allow(content_item).to receive(:display_single_page_notification_button?).and_return(false)
+
+      render partial: "shared/published_dates_with_notification_button", locals: { content_item: content_item }
+
+      expect(rendered).to include(I18n.t("common.print_link"))
+    end
+  end
+end
+
+# To remove this once DetailedGuide PR is made
```
deborahchua, line 31: Is this meant to be included as part of the PR?

sairamya93: Not exactly. I had added in a dummy class here to handle the tests for the partial as we don't yet have the specific document type views like `detailed_guide`  that actually render this partial. Once we have the detailed_guide view, we could remove the dummy class.

app/views/shared/_published_dates_with_notification_button.html.erb
KludgeKML: Okay for the moment, but we probably need to look into this variable (the only other place it's used in frontend is in another shared template, and it's never set. So I think we should ultimately make this use and the other locals to the partials which can be set when called)

config/locales/ar.yml
```diff
@@ -96,10 +97,12 @@ ar:
       scotland_slug:
     next_holiday_in_is:
     past_bank_holidays:
+    print_link:
     substitute_day:
     today:
     united-kingdom_slug:
     upcoming_bank_holidays:
+    visit: 'تفضّل بزيارة:'
```
KludgeKML, line 105: It's possible this command was a bit too broad - it's brought in this common.visit, which isn't used in this partial. Perhaps run the command again twice, once for common.print_link, once for common.email_and_print_link, so it doesn't bring in other things?

app/views/shared/_published_dates_with_notification_button.html.erb
sairamya93: I have updated the partial to set the value for the `account` from the view that renders this partial else defaults to false.

KludgeKML (approved):
LGTM

PR #4779, opened 2025-04-28
Include more metadata fields in GraphQL query

mike3985:
This brings the GraphQL version of rendered news articles pages a lot closer to the regular, Content Store version.

I've explained the changes in detail in the commit message. At a high level they're:

* add more fields to query
* update test fixtures to include the new fields

Question: would this benefit from test coverage? I'm happy to add some. I wrote some in the process of making these changes but a few things about them didn't feel quite right so I wanted to check what the general opinion was before proceeding.

https://trello.com/c/HZRt4r50/1666-make-meta-tags-etc-of-graphql-version-of-news-articles-the-same-as-regular-version, [Jira issue PP-2906](https://gov-uk.atlassian.net/browse/PP-2906)


app/queries/graphql/edition_query.rb
```diff
@@ -101,21 +124,35 @@ def query
                 }
 
                 world_locations {
+                  analytics_identifier
                   base_path
                   content_id
                   title
                 }
+                worldwide_organisations {
```
yndajas, line 132: **nitpick:** per your previous commit, this could do with a preceding line break

spec/fixtures/graphql/news_article.json
```diff
@@ -1,13 +1,18 @@
 {
-  "data":
```
yndajas, line 2: **question:** so are you of the opinion that we don't test within the news article system test spec that all of this is present when rendered using both GraphQL and Content Store? Do we have an alternative, other than relying on our manual diffing checks?

app/queries/graphql/edition_query.rb
```diff
@@ -12,12 +12,17 @@ def query
           ) {
```
yndajas, line 12: **suggestion:** struggling to parse this from the commit message because of the multiple "for"s and list(s). Is there a way to reword it for clarity?

> `analytics_identifier` for `organisations`, `world_locations` and `worldwide_organisations` for populating the "govuk:world-locations" and "govuk:organisations" meta tags.

app/queries/graphql/edition_query.rb
```diff
@@ -13,58 +13,74 @@ def query
             ... on Edition {
               base_path
               description
+              document_type
```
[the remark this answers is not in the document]
mike3985: Yeah, that theory does make sense. I'll revert this change if there're no votes in favour.

spec/fixtures/graphql/news_article.json
mike3985: Nope, that's the only alternative I had in mind.

I can add some tests if it makes sense.

app/queries/graphql/edition_query.rb
mike3985: Oof, that is a particularly bad one. I'll give it a go 📝

spec/fixtures/graphql/news_article.json
mike3985: I've added GraphQL versions of the News Article model specs. It could be nice to have the same for the system specs, too, but I've left my work-in-progress on that on a separate branch for now https://github.com/alphagov/frontend/compare/graphql-news-articles-meta-tags-system-test?expand=1

app/queries/graphql/edition_query.rb
yndajas: Looks like the spaces were kicked along with the other formatting changes, so I'll mark this as resolved

yndajas (dismissed):
This is a lot of work and - in my opinion - a lot of good work!

Re: "a lot", it might be helpful to split this into two or three PRs covering:
- the substantial refactors
- the improved test coverage (if differentiable)
- the introduction of more fields in the query

... but maybe it's not worth the rewrite

spec/fixtures/graphql/news_article.json
KludgeKML: I'm going to reiterate what I said in Slack, which is: I don't think you need request or system tests for GraphQL.

The GraphQL loading is segregated inside the ContentItemLoader, which ultimately resolves (through the ContentItemFactory) into a ContentItem (or subclass) object. Once you get to that point, as long as the model values are identical between ContentStore and GraphQL, anything that's going wrong will go wrong for either source. So if you've proved that the model generated by GraphQL is the same as the one from ContentStore, all the existing system and request tests provide enough coverage already.

spec/support/concerns/emphasised_organisations.rb
```diff
@@ -1,12 +1,13 @@
-RSpec.shared_examples "it can have emphasised organisations" do |document_type, example_name|
+RSpec.shared_examples "it can have emphasised organisations" do |document_type, example_name, data_source: :content_store|
   let(:content_store_response) { GovukSchemas::Example.find(document_type, example_name: example_name) }
+  let(:api_response) { fetch_content_item(document_type, example_name, data_source:) }
```
leenagupte, line 3: I'm not sure about the naming here. Is it always an `api_response`? Also it feels a bit generic.

spec/system/news_article_spec.rb
```diff
@@ -126,5 +126,23 @@
 
       it_behaves_like "a news article page with a high resolution image"
     end
+
+    context "when visiting a page in history mode" do
+      let(:path) { "/government/news/final-care-act-guidance-published?graphql=true" }
+      # rubocop:disable RSpec/LetSetup
```
leenagupte, line 132: Can you add something to the commit message about why this Cop has been disabled.

KludgeKML: (Particularly, can we double-check this isn't just a before block in disguise)

leenagupte: Yes, agreed. I did wonder why a `let!` is being used in these tests. I've noticed that they've been littered about the tests with no explanation.

spec/support/concerns/emphasised_organisations.rb
yndajas: I've changed `api_response` to `content_item` throughout

spec/system/news_article_spec.rb
leenagupte: Can you make sure that when the tests are added in [graphql-news-articles-meta-tags-etc-remainder](https://github.com/alphagov/frontend/compare/main...graphql-news-articles-meta-tags-etc-remainder) that you add something to the appropriate commit message about why `let!` is being used therefore why `RSpec/LetSetup` is being disabled. I had a quick look and at time of writing that hadn't been done.

yndajas: Yeah I just moved those onto another branch for reference. The intention isn't to open a PR with those changes at the moment given Keith's provocations on our broader approach to testing, but if we do, we'll be sure to address this

spec/models/news_article_spec.rb
```diff
@@ -1,25 +1,29 @@
...
+      it_behaves_like "it can be withdrawn", "news_article", "news_article_withdrawn", data_source: data_source
+    end
 
-  describe "#contributors" do
-    subject(:news_article) { described_class.new(content_store_response) }
+    describe "#contributors" do
+      subject(:news_article) { described_class.new(api_response) }
 
-    let(:content_store_response) { GovukSchemas::Example.find("news_article", example_name: "best-practice-government-response") }
+      let(:api_response) { fetch_content_item("news_article", "best-practice-government-response", data_source: data_source) }
```
leenagupte, line 17: A rogue `api_response` has been left in.

leenagupte (commented):
Thanks for renaming the test variables. One minor inline comment.

KludgeKML (approved):
LGTM! Thanks for the changes.

KludgeKML:
Oh, if you can fix that one variable name before merging (see Leena's comment).

PR #4805, opened 2025-05-20
Fix an error in the govspeak block

andysellick:

## What
Fixes an error discovered in the govspeak block, in the `landing_page` item.

- when used inside a columns block, the UI was erroring because no options object is passed from the columns to the child govspeaks
- this works in other layout blocks because they pass that object
- since there's no options to pass in this context, seemed best to create a default object in case none existed inside the govspeak block

## Why
Discovered during related work.

## Visual changes
None.

Trello card: https://trello.com/c/4g67KQHy/629-build-map-block, [Jira issue PNP-6391](https://gov-uk.atlassian.net/browse/PNP-6391)

### DIFF
--- a/app/views/landing_page/blocks/_columns_layout.html.erb
+++ b/app/views/landing_page/blocks/_columns_layout.html.erb
@@ -21,7 +21,5 @@
 %>
 
 <%= tag.div class: "columns-layout", data: data_attributes do %>
-  <% block.blocks.each do |child| %>
-    <%= render "landing_page/blocks/#{child.type}", block: child %>
-  <% end %>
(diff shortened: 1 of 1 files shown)


KludgeKML:
I think it might be better to update the columns view to use the render_block helper method (which supplies a default options).

- You could replace these lines: https://github.com/alphagov/frontend/blob/main/app/views/landing_page/blocks/_columns_layout.html.erb#L24-L26

...with:
`<% block.blocks.each { |child| %><%= render_block(child) %><% } %>`

andysellick:
@KludgeKML thanks, have updated 👍

KludgeKML (approved):
LGTM

PR #4824, opened 2025-05-30
Add additional GraphQL links for related navigation component

brucebolt:
, [Jira issue PP-2924](https://gov-uk.atlassian.net/browse/PP-2924)News articles include the related navigation component, which is populated from [the links](https://github.com/alphagov/govuk_publishing_components/blob/7e37b0a92d66860fe2dcb0c1827ffc3f2db0b1ae/lib/govuk_publishing_components/presenters/related_navigation_helper.rb).

Adding the links we are currently missing, to ensure this component is able to display all the existing links that are currently included on the content store version of the page.

I haven't added any tests, as there's nothing to test directly here, since it's just a hardcoded string.  In previous PRs, we've indirectly tested `govuk_publishing_components`, but this doesn't feel like the right approach and we'll need to think about the long term testing of GraphQL in frontend applications.

[Trello card](https://trello.com/c/bBztfL8K)

### DIFF
--- a/app/queries/graphql/edition_query.rb
+++ b/app/queries/graphql/edition_query.rb
@@ -39,11 +39,17 @@ class Graphql::EditionQuery
                   base_path
                   locale
                 }
+                document_collections {
+                  ...RelatedItem
+                }
                 government {
                   details {
                     current
                   }
                   title
+                }
+                mainstream_browse_pages {
+                  ...RelatedItem
                 }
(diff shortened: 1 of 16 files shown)


yndajas (approved):

brucebolt:
I forgot to commit the fixture files 🤦‍♂️.  I've added them now.

catalinailie (approved):

PR #4839, opened 2025-06-12
Migrate document collection

KludgeKML:

## What

Migrate the document_collection schema rendering from government-frontend to Frontend. 

Note that there's no locale copying in this PR, because document collections only need the names of the various document types, and those were copied into the right place in https://github.com/alphagov/frontend/pull/4848

## Why

https://trello.com/c/gCX6e1Kb/674-move-documentcollection-from-government-frontend-to-frontend, [Jira issue PNP-6446](https://gov-uk.atlassian.net/browse/PNP-6446)

### Document collection with single notification button (https://www.gov.uk/government/collections/statutory-guidance-schools)

Before | After
------- | -------
image | image

### Document collection with topic notification link (https://www.gov.uk/government/collections/self-employment-detailed-information)

Before | After
------ | -------
image |  image

### DIFF
--- a/spec/system/document_collection_spec.rb
+++ b/spec/system/document_collection_spec.rb
@@ -67,62 +67,57 @@ RSpec.describe "Document Collection" do
         expect(page).not_to have_selector("nav a", text: "Empty Group")
       end
     end
+
+    context "when all document collection groups are empty" do
+      context "when no headers are supplied" do
+        let(:content_item) do
+          GovukSchemas::Example.find(:document_collection, example_name: "document_collection").tap do |item|
+            item["details"]["headers"] = []
+
+            item["details"]["collection_groups"] = [
+              {
+                "body" => "<div class=\"empty group\">\n</div>",
+                "documents" => [],
+                "title" => "Empty Group",
+              },
+            ]
+          end
+        end
+
+        it "does not render a contents list" do
+          visit base_path
+
+          expect(page).not_to have_selector(".gem-c-contents-list")
+        end
+      end
+
+      context "when h2 headers are supplied" do
+        let(:content_item) do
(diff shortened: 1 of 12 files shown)


spec/system/document_collection_spec.rb
```diff
@@ -67,62 +67,57 @@
...
-  #   assert page.has_css?(".gem-c-contents-list")
-  # end
+        it "does not render a contents list" do
+          visit base_path
+
+          expect(page).not_to have_selector(".gem-c-contents-list")
+        end
+      end
+
+      context "when h2 headers are supplied" do
```
deborahchua, line 94: I think we should also [test that nested headings aren't included in the contents list](https://github.com/alphagov/frontend/blob/main/spec/presenter/corporate_information_page_presenter_spec.rb#L15-L38), as Whitehall will automatically extract all headings from the body and will be shown unless [specifically removed](https://github.com/alphagov/frontend/blob/main/app/presenters/corporate_information_page_presenter.rb#L43-L47).

deborahchua (commented):
This is looking great! 👏 I think you're now in the lead for the most number of commits in a consolidation PR 🥇 The atomic commits really helped me review easily, and find my place again when there were changes still being pushed up.

I've left a couple of comments mainly about the contents list headers.

AshGDS (commented):
Frontend looks OK to me 👍

spec/system/document_collection_spec.rb
KludgeKML: It was tested implicitly, in fact, because there are L3 headers in the document_collection_with_body example, but I've made it more explicit https://github.com/alphagov/frontend/pull/4839/files#diff-50919a8b21f824666a72e806f02b0ff691f7050bc5064142d7aa5aba1f01c931R106-R119 (as

deborahchua (approved):
LGTM!

PR #4841, opened 2025-06-12
Fix retrieval of default news image for GraphQL news articles

brucebolt:
, [Jira issue PP-3294](https://gov-uk.atlassian.net/browse/PP-3294)We were previously looking for the `default_news_image` on the news article itself. However this is not what the presenter is looking for.

We actually [needed to retrieve](https://github.com/alphagov/frontend/blob/e428d61951eed2fee50572d965a3083eb6727d48/app/models/concerns/news_image.rb#L10-L13) this from the `primary_publishing_organisation` instead, so updating the GraphQL to use that.

[Trello card](https://trello.com/c/QK8BsDKn)

### DIFF
--- a/app/queries/graphql/edition_query.rb
+++ b/app/queries/graphql/edition_query.rb
@@ -17,10 +17,6 @@ class Graphql::EditionQuery
               details {
                 body
                 change_history
-                default_news_image {
-                  alt_text
-                  url
-                }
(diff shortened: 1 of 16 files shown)


callumknights (approved):

JonathanHallam (approved):
LGTM!

PR #4848, opened 2025-06-17
Arrange formats translations

KludgeKML:

## What

Rearrange some translations keys that were causing confusion (both to developers and to the i18n tasks that keep
the locale files tidy)

## Why

It wasn't possible to safely call `rails translation:add_missing`, which is necessary to keep the locale files tidy, without
causing the rake tests for translation file tidiness to fail. 

## How

Move all pluralisation keys under "formats.*" to "formats.*.name" to keep them separate from other key types, move
the csv_preview.document_type.* keys to "formats.*.name" to organise format names more logically, and remove a key leftover from other work that only existed in the Kyrgyz translation file.


deborahchua (approved):
LGTM 🚀

PR #4851, opened 2025-06-18
PP-636 Add prometheus labels for graphql errors and error status codes

callumknights:

## What

Adding reporting to prometheus for graphql query errors, and general error codes related to graphql requests.

## Why

So our team (publishing-platform) can pro-actively monitor error rates for graphql rendered pages vs normal content store rendered pages. This code is needed to differentiate those in our reporting. The `contains_errors` label here also allows us to separate graphql queries with errors in by frontend app (another for PR for collections is forthcoming).

https://gov-uk.atlassian.net/browse/PP-636

## How

Adding prometheus labels based on actions similar to how it's done in publishing-api.

[Prior art](https://github.com/alphagov/publishing-api/commit/3e08dfbc25e79bbc165edea3e0e3f0a7c0ef47c8)

### DIFF
--- a/lib/content_item_loader.rb
+++ b/lib/content_item_loader.rb
@@ -40,16 +40,17 @@ private
   def load_from_graphql(base_path)
     graphql_response = GdsApi.publishing_api.graphql_content_item(Graphql::EditionQuery.new(base_path).query)
     if graphql_response.to_hash.blank?
-      set_prometheus_labels("graphql_contains_errors" => true)
+      set_prometheus_labels(graphql_contains_errors: true)
       nil
     elsif GRAPHQL_ALLOWED_SCHEMAS.include?(graphql_response["schema_name"])
+      set_prometheus_labels
       graphql_response
     end
   rescue GdsApi::HTTPErrorResponse => e
-    set_prometheus_labels("graphql_status_code" => e.code)
+    set_prometheus_labels(graphql_status_code: e.code)
     nil
   rescue GdsApi::TimedOutException
-    set_prometheus_labels("graphql_api_timeout" => true)
+    set_prometheus_labels(graphql_api_timeout: true)
     nil
   end
 
@@ -100,8 +101,14 @@ private
     { cache_control: "max-age=0, public", expires: "" }
   end
 
-  def set_prometheus_labels(hash)
+  def set_prometheus_labels(graphql_status_code: 200, graphql_contains_errors: false, graphql_api_timeout: false)
     prometheus_labels = request.env.fetch("govuk.prometheus_labels", {})
+
+    hash = {
+      "graphql_status_code" => graphql_status_code,
+      "graphql_contains_errors" => graphql_contains_errors,
+      "graphql_api_timeout" => graphql_api_timeout,
+    }
 
     request.env["govuk.prometheus_labels"] = prometheus_labels.merge(hash)
   end
--- a/spec/unit/content_item_loader_spec.rb
+++ b/spec/unit/content_item_loader_spec.rb
@@ -218,6 +218,16 @@ RSpec.describe ContentItemLoader do
           expect(graphql_request).to have_been_made
           expect(item_request).not_to have_been_made
         end
+
+        it "sets the appropriate prometheus labels" do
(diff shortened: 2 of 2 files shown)


lib/content_item_loader.rb
```diff
@@ -29,19 +29,29 @@ def load_from_sources(base_path)
...
       GdsApi.content_store.content_item(base_path)
     end
   rescue GdsApi::HTTPErrorResponse, GdsApi::InvalidUrl => e
     e
   end
 
+  def load_from_graphql(base_path)
+    graphql_response = GdsApi.publishing_api.graphql_content_item(Graphql::EditionQuery.new(base_path).query)
+    if graphql_response.to_hash.key?("errors")
+      set_prometheus_labels("contains_errors" => true)
```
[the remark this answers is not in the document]
callumknights: Fixed with a mildly ugly explicit `nil` under the call to `set_prometheus_labels`

KludgeKML:
Looks like we're still missing some coverage (one of the branches on line 104 and all of line 53)

spec/unit/content_item_loader_spec.rb
```diff
@@ -219,6 +219,35 @@
...
+        subject(:content_item_loader) { described_class.for_request(request) }
+
+        let!(:graphql_request) { stub_publishing_api_graphql_query(graphql_query, { data: { edition: { schema_name: "news_article", "errors" => [{ "message" => "some_error" }] } } }) }
+        let(:request) { instance_double(ActionDispatch::Request, path: "/my-random-item", env: {}, params: { "graphql" => "true" }) }
+
+        it "reports the presence of errors as a prometheus label" do
+          content_item_loader.load("/my-random-item")
+
+          expect(graphql_request).to have_been_made
+          expect(item_request).to have_been_made
```
mike3985, line 233: I'd like to see separate `it` examples analogous to a lot of the others in this spec file... like, `it "falls back to calling the Content Store"` including these ^ two expectations.

Same for the `context` below, too. Especially if the outcome is actually different.

Both things seem worthwhile documenting and this seems like a reasonable time to do it?

JonathanHallam (approved):
Looks good to me, just waiting on review from the team that owns the app :)

leenagupte (commented):
I have a couple of inline comments, otherwise it looks good.

lib/content_item_loader.rb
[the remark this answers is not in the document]
leenagupte: Perhaps you could link to both? I think it's good to see where the code is copied from.

mike3985:
I've made those changes that you asked for, @leenagupte. Are you OK for me to merge this now? Thanks

lib/content_item_loader.rb
```diff
@@ -38,6 +38,7 @@ def load_from_sources(base_path)
   end
 
   def load_from_graphql(base_path)
+    set_prometheus_labels("graphql_status_code" => 200, "graphql_contains_errors" => false, "graphql_api_timeout" => false)
```
leenagupte, line 41: Call this here feels a bit like false reporting. I think this should be set after the call to publishing-api so it can accurately report the status codes and the lack of errors.

leenagupte:
> I've made those changes that you asked for, @leenagupte. Are you OK for me to merge this now? Thanks

The existing changes are good to go. I've added comment about the new commit.

lib/content_item_loader.rb
JonathanHallam: So it doesn't actually send anything to prometheus until further down the line when the prometheus exporter is invoked. [Here it's just building up an object to send later](https://github.com/alphagov/frontend/pull/4851/files#diff-4c2ac1f535a1643c98d54d305323a278558c08f9309b72e18c224179575d02ddR104-R108). Setting it here and waiting for it to be overwritten by the calls to publishing API failing in some way seemed like the cleanest way to do it, but either way we won't be sending anything false over to prometheus

leenagupte: Perhaps the logic can be re-thought?

leenagupte (approved):
Looks good. Thanks for updating the last commit.

PR #4888, opened 2025-07-08
Simplify single page notification button calls in partials

KludgeKML:

## What

Simplify the partials that render a single page notification button component.

## Why

In the original partials copied from government-frontend, we used various different ways to determine if a one-login or non-one-login link should be the action of the button, and whether or not the js_enhancement flag should be set., including local_assigns and class-level variables passed into the partials. In fact, in government-frontend none of the views that use these partials actually require customisation, so it's possible to just check the logged in status and set js_enhancement to match it, and skip_account to an inverted string version of it (due to a current issue in the component gem which requires this parameter to be a string).

## How

Use the [logged_in?](https://github.com/alphagov/govuk_personalisation/blob/main/lib/govuk_personalisation/controller_concern.rb#L65-L67) method provided by the govuk_personalisation controller concern. This is an encapsulated way of checking if a user account header is present.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -6,6 +6,8 @@ class ContentItemsController < ApplicationController
 
   attr_reader :content_item
 
+  helper_method :logged_in?
+
 private
 
   def set_content_item_and_cache_control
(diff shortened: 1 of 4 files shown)


sairamya93 (approved):
LGTM!
I have put in a tiny suggestion if that might be fine.

PR #4889, opened 2025-07-09
Fix subscription partials

KludgeKML:

I introduced a mistake in logic in https://github.com/alphagov/frontend/pull/4888 - this fixes that.

The js_enhancement changes were solid, but the skip_account change was wrong. The skip_account flag should almost always default to "false", regardless of whether someone is logged in or not. It should only be set to "true" as an exception (in Document Collections).

https://docs.publishing.service.gov.uk/manual/email-signup-journeys.html#single-page-notification-button-component

This change adds an optional local assign `:skip_account` which can be left out for correct behaviour in most callers, but set explicitly to "true" in the soon to be added document collection view.

### DIFF
--- a/app/assets/javascripts/dependencies.js
+++ b/app/assets/javascripts/dependencies.js
@@ -8,6 +8,7 @@
 //= require govuk_publishing_components/components/metadata
 //= require govuk_publishing_components/components/print-link
 //= require govuk_publishing_components/components/radio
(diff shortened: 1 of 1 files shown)


sairamya93 (approved):
Tested the changes locally for `detailed_guide` which does not set `skip_account` explicitly and it is working fine.

AshGDS (approved):

PR #4897, opened 2025-07-14
[PP-7253] Restore GraphQL A/B test for news articles

brucebolt:
In https://github.com/alphagov/frontend/pull/4895, we switched off all traffic to GraphQL for news articles as we were getting an error in the draft deployment of Frontend.

This switches the A/B test back on and forces all draft requests to Content Store.


yndajas (approved):
I'd probably move the fix commit before the revert commit here, but otherwise this looks good

PR #4899, opened 2025-07-15
Update action link component options and styles

jon-kirwan:
## What

Update action link component options and styles

https://trello.com/c/iiekbtY8/3561-updates-to-the-action-link-component, [Jira issue NAV-2462](https://gov-uk.atlassian.net/browse/NAV-2462)

## Why

Due to the changes outlined here https://github.com/alphagov/govuk_publishing_components/pull/4915, the `light_icon` has been removed, as it is now the default option. Also, custom override styles need to be updated to target the default element styles.


MartinJJones (approved):
Nice work!

PR #4915, opened 2025-07-24
Add support for extra fields in local transactions

KludgeKML:
## What

Adds support for new fields in local_transaction content items, inititally to support https://gov.uk/find-registered-childminder

- `cta_text` will (if present) replace the text in the cta button on the search page
- `before_results` will (if present) be rendered as a govspeak box above the council result
- `after_results` will (if present) be rendered as a govspeak box below the "search for another postcode" link.
- a small additional border has been added below the "search for another postcode" link to give some room before `after_results`
- we also remove the hacks added to support https://gov.uk/apply-foster-child-council, since they can now be replicated in the UI.

## Why

1. There's an existing hack around the foster through your council page that this would allow us to remove
2. Content would like to make some updates to the [find-a-registered-childminder page](https://www.gov.uk/find-registered-childminder) which require these new fields.

https://trello.com/c/3ZomQTGq/753-update-find-a-registered-childminder-postcode-finder, [Jira issue PNP-5827](https://gov-uk.atlassian.net/browse/PNP-5827)

### https://www.gov.uk/find-registered-childminder

Before|After
------|------
image|image
image|image

### https://gov.uk/apply-foster-child-council

(No visible changes on search page, on results page the `before_results1 moves above the "We've matched..." paragraph.

Before|After
------|------
image|image

### DIFF
--- a/app/models/local_transaction.rb
+++ b/app/models/local_transaction.rb
@@ -1,11 +1,15 @@
 class LocalTransaction < ContentItem
-  attr_reader :country_name, :introduction,
+  attr_reader :after_results, :before_results, :country_name,
+              :cta_text, :introduction,
               :lgil_code, :lgil_override, :lgsl_code,
               :more_information, :need_to_know
 
   def initialize(content_store_response)
     super(content_store_response)
 
+    @after_results = content_store_response.dig("details", "after_results")
+    @before_results = content_store_response.dig("details", "before_results")
+    @cta_text = content_store_response.dig("details", "cta_text")
     @introduction = content_store_response.dig("details", "introduction")
     @lgil_code = content_store_response.dig("details", "lgil_code")
     @lgil_override = content_store_response.dig("details", "lgil_override")
--- a/app/models/local_transaction.rb
+++ b/app/models/local_transaction.rb
@@ -40,12 +40,6 @@ class LocalTransaction < ContentItem
     URI.parse(base_path).path.sub(%r{\A/}, "")
   end
 
-  # the apply to foster page has a few hard-coded exceptions to its
-  # layout, so add a method to detect it.
-  def apply_foster_child_council?
-    slug == "apply-foster-child-council"
-  end
-
 private
 
   def country_name_availability_for(key)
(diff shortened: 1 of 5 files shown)


app/models/local_transaction.rb
```diff
@@ -1,7 +1,7 @@
 class LocalTransaction < ContentItem
   attr_reader :country_name, :introduction,
-              :lgil_code, :lgil_override, :lgsl_code,
-              :more_information, :need_to_know
+              :lgil_code, :lgil_override, :lgsl_code, :cta_text,
```
deborahchua, line 3: I think we can reorder these to be in alphabetical order.

app/models/local_transaction.rb
```diff
@@ -10,8 +10,11 @@ def initialize(content_store_response)
     @lgil_code = content_store_response.dig("details", "lgil_code")
     @lgil_override = content_store_response.dig("details", "lgil_override")
     @lgsl_code = content_store_response.dig("details", "lgsl_code")
+    @cta_text = content_store_response.dig("details", "cta_text")
```
deborahchua, line 13: Similar for these, we can rearrange to be alphabetical.

app/views/local_transaction/index.html.erb
```diff
@@ -29,7 +29,7 @@
              postcode: current_page?(electoral_services_path) ? @postcode.sanitized_postcode : @postcode,
              margin_top: @location_error ? 5 : 0,
              publication_title: content_item.title,
-             go_button_text: content_item.apply_foster_child_council? ? t("formats.local_transaction.find_adoption_hub") : nil,
+             go_button_text: content_item.cta_text.presence,
```
deborahchua, line 32: Should this text default to something else if `cta_text` is not present?

deborahchua (commented):
Looks good! I've left a couple of minor comments.

app/views/local_transaction/index.html.erb
sairamya93: I realised while running the tests that `.presence` is equivalent to `cta_text.present? ?  content_item.cta_text : nil`.It is not as explicit but more concise.

deborahchua (approved):

PR #4936, opened 2025-08-05
fix: remove double base_path in graphql fixture

KludgeKML:
- Causing a deprecation warning from the json gem


spec/fixtures/graphql/world_news_story_news_article.json
```diff
@@ -1,7 +1,6 @@
 {
   "data": {
     "edition": {
-      "base_path": "/government/news/christmas-2016-prime-ministers-message-from-azerbaijan",
       "base_path": "/government/news/christmas-2016-prime-ministers-message",
```
mike3985, line 4: I reckon the Azerbaijan one is the one to keep because it matches the base_path in the real content schema (https://github.com/alphagov/publishing-api/blob/main/content_schemas/examples/news_article/frontend/world_news_story_news_article.json#L2)

KludgeKML: Updated

mike3985 (approved):
Double trouble

PR #4940, opened 2025-08-07
Remove historic bank holidays

sihugh:
## What

Delete the old bank holidays from the API from 2019 - 2023 inclusive.

## Why

[This API](https://www.gov.uk/bank-holidays.json) is responsible for about 6% of GOV.UK traffic. Reducing the payload, even by a small amount, has a significant effect on our bandwidth, and processing by the consumers.

I don't think we have a documented commitment to retain old dates indefinitely. Retaining last year in full feels fair to me as it allows for records keeping for tax purposes, etc.

This [API is preserved by the National Archives](https://webarchive.nationalarchives.gov.uk/ukgwa/*/https://www.gov.uk/bank-holidays.json), which allows spelunking if people wish, plus the config is readily available in Github.

An improvement in future might be to have the default API response to show current + future dates, with historic dates available separately.

[The bank holidays page](https://www.gov.uk/bank-holidays) only shows today's and future events, so is unaffected by this.

(Scotland's GitHub diff is a little confusing, so possibly worth checking in the review app. The others look sensible)


andysellick (approved):
I've checked the changed file particularly the Scotland section and it looks sensible.

PR #4942, opened 2025-08-07
[PP-7284] Switch GraphQL method for retrieving content

brucebolt:
We have moved the query for GraphQL from frontend applications into Publishing API, and instead exposed an API that behaves exactly like Content Store.

Therefore switching to use this new API method and removing most of the GraphQL-specific tests.

Depends on https://github.com/alphagov/gds-api-adapters/pull/1360.


spec/unit/content_item_loader_spec.rb
```diff
@@ -199,7 +203,12 @@
       context "with graphql param=true" do
         subject(:content_item_loader) { described_class.for_request(request) }
 
-        let(:graphql_request) { stub_publishing_api_graphql_query(graphql_query, { data: { edition: { schema_name: "news_article" } } }) }
+        let(:graphql_request) do
+          stub_publishing_api_graphql_has_item(
```
mike3985, line 207: I missed the introduction of these API test helpers... and so I'm just taking the opportunity here to think out loud: I wonder if we can think of a catchy name for this endpoint and if doing so would make it clearer what methods like this are for 💭

brucebolt: I tried to keep this name as similar to the Content Store equivalents as possible, since they both do the same thing (albeit stubbing a different URL) and accept the same arguments.

spec/support/content_item_helpers.rb
```diff
@@ -1,14 +1,5 @@
-require "support/publishing_api_graphql_helpers"
-
 module ContentItemHelpers
```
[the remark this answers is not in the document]
brucebolt: Ooh, hadn't spotted this was introduced by the GraphQL work.  I'll swap this back to how it was before.

PR #4943, opened 2025-08-07
(CM-294) Add styling for content blocks

pezholio:
## What

As part of the work to introduce Contact blocks in Content Block Manager, we've had to introduce some [minimal styling](https://github.com/alphagov/govuk_content_block_tools/pull/76) to our contact blocks. After some discussion it was decided that the best place to distribute these styles was the `content_block_tools` gem, as it was closer to the layout of the blocks themselves, and didn't quite fit in with the `govuk_publishing_components` gem, as they weren't components as such.

## Why

[Jira ticket](https://gov-uk.atlassian.net/jira/software/c/projects/CM/boards/134?selectedIssue=CM-294)

## How

This adds the `content_block_tools` gem as a dependency and loads in the `content_block_tools` styles.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -4,6 +4,7 @@
 $govuk-include-default-font-face: false;
 
 @import "govuk_publishing_components/govuk_frontend_support";
+@import "content_block_tools";
 
 // Helper stylesheets (things on more than one page layout)
 @import "helpers/content-bottom-margin";


KludgeKML:
Question - this pulls in the govspeak gem which (amazingly, tbf) isn't currently in Frontend (the situation at the moment is that publishing apps tend to do govspeak processing and then pass us blocks of preprocessed HTML with the correct classes to style from). Can I double-check that the intent of including the gem is _solely_ for including the styling?

pezholio:
No, so it's for styling and the processing of blocks too (which sometimes contain Govspeak)

PR #4968, opened 2025-08-19
Build history pages from content item

KludgeKML:

## What

Allows pages under /government/history to be built from the new history documents that will come out of Whitehall (based on their standard edition).

- adds support for /government/history/:slug to render the location history pages currently on government-frontend (eg: https://www.gov.uk/government/history/10-downing-street).
- adds support for rendering /government/history via a little trick (we respond to that route, but intercept it at content item loader time and load the content item from /government/history/history-of-the-uk-government), since Whitehall doesn't support arbitrary paths or heirarchical paths at the moment. We can remove this hack when Whitehall supports those paths fully. 

## Why

https://trello.com/c/pMEJlBME/796-build-history-pages-from-content-item, [Jira issue PNP-5622](https://gov-uk.atlassian.net/browse/PNP-5622)

## Screenshots

Location history pages change very little (just a change in the spacing at the top), but the main history homepage changes a lot (from the zig-zag before to a content list/content column pair). This has been agreed with our designer.

Before|After
------|------
image|image
image|image

### DIFF
--- a/app/models/history.rb
+++ b/app/models/history.rb
@@ -1,4 +1,11 @@
 class History < FlexiblePage
+  def breadcrumbs
+    super << {
+      title: "History of the UK Government",
+      url: "/government/history",
+    }
+  end
+
 private
 
   def default_flexible_sections
--- a/app/models/history.rb
+++ b/app/models/history.rb
@@ -1,0 +1,30 @@
+class History < FlexiblePage
+private
+
+  def default_flexible_sections
+    [
+      {
+        type: "page_title",
+        context: "History",
+        heading_text: title,
+        lead_paragraph: content_store_response.dig("details", "lead_paragraph"),
+      },
+      {
+        type: "rich_content",
+        contents_list: (content_store_response.dig("details", "headers") || []).map { |header| header.except("headers") },
+        govspeak: body,
+        image: image_hash,
+      },
+    ]
+  end
+
+  def image_hash
+    sidebar_image = content_store_response.dig("details", "sidebar_image")
+    return nil unless sidebar_image
+
+    {
+      alt: sidebar_image["caption"],
+      src: sidebar_image["url"],
+    }
+  end
+end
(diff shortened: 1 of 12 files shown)


app/models/history.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def image_hash
+    return nil unless content_store_response.dig("details", "sidebar_image")
+
+    {
+      alt: content_store_response.dig("details", "sidebar_image", "caption"),
+      src: content_store_response.dig("details", "sidebar_image", "url"),
+    }
+  end
```
AshGDS, line 35: Is it more efficient to store  the `content_store_response.dig("details", "sidebar_image")` part in a variable here so that it's only called once?

AshGDS (commented):
Was just wondering where/when it was agreed for the History of the UK Government to have its layout changed? I'm guessing it's a compromise to make the pages editable now?

And not sure if I'm doing something wrong, but the history pages do not load properly for me locally - `/government/history` 404s, and the No 10 page returns this:

image

(Code looks good though 👍 )

KludgeKML:
> Was just wondering where/when it was agreed for the History of the UK Government to have its layout changed? I'm guessing it's a compromise to make the pages editable now?

Yeah, we discussed it with Nikin. The zig-zag styling wasn't awful, but it isn't something we've done anywhere else, and this lets us standardise and clear all the old hardcoded history pages off government-frontend for low effort.

> And not sure if I'm doing something wrong, but the history pages do not load properly for me locally - `/government/history` 404s, and the No 10 page returns this:

Ah, you need to be pointing your local machine at integration, not prod.

AshGDS (approved):
LGTM 👍  Two things to note:

- On the Number 10 page, the `first-published-at` etc. dates are now 2025, do we need to preserve the original dates?
- The GA4 indexes shift slightly on the Number 10 page, so we probably need to let the performance analysts know when we publish this page.

deborahchua (approved):
LGTM!

PR #4973, opened 2025-08-21
refactor: extract functionality from ContentItemLoader to make it easier to understand

KludgeKML:
ContentItemLoader was getting a bit too complex for its base case (loading a content item from the content store), so split out the four loading cases (loading from a json file, a yaml file, graphql or content store) into submodules.

The ContentItemLoaders behaviours are: They have a #can_load? method which returns true if the loader thinks it can/should try to load the item. They have a #load method which actually tries to load and return the item. This method should throw an `ContentItemLoaders::UnableToLoad` error if it fails to load when it said it would. The main loader catches this and returns the simple content_store response.

ContentStoreLoader gets to be a bit of a special snowflake - it has no #can_load? method because that's assumed to always be true. It can't refuse to try to load, and it will always be called as the last resort.

JsonFileLoader and YamlFileLoader are simple and similar enough that they can be based on the same parent class (LocalFileLoader) which handles initialisation, checking for the environment variable, base paths, etc (this also affects the test, making this not quite a 100% pure refactor commit, but it seems better to keep this change in the commit). We don't use the error to fallback to the content_store version because it's okay (indeed, preferable) for this to fail more spectacularly if the env var is set.

GraphqlLoader is obviously the most complicated of the three, because it has to use the content_store_loader and request variables passed in during initialisation to load the content item from the content_store to check its schema and to write request information to prometheus labels.

Additionally we add a `safe_path?` check to the loader which fails fast with an InvalidUrl if the path passed in contains path-traversal tricks (ie `/government/../../test.js`) I don't believe this is currently an exploitable security flaw, since it's impossible to find out anything this way that you can't find out easier just by looking at the source code, but it does prevent nuisance calls to content-store.

### DIFF
--- a/lib/content_item_loader.rb
+++ b/lib/content_item_loader.rb
@@ -1,109 +1,45 @@
 require "ostruct"
 
 class ContentItemLoader
-  include DraftHelper
-
-  LOCAL_ITEMS_PATH = "lib/data/local-content-items".freeze
-  GRAPHQL_ALLOWED_SCHEMAS = %w[news_article].freeze
-  GRAPHQL_TRAFFIC_RATE = 0.5 # This is a decimal version of a percentage, so can be between 0 and 1
-
   def self.for_request(request)
     request.env[:loader] ||= ContentItemLoader.new(request:)
   end
 
-  attr_reader :cache, :request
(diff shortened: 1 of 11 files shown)


KludgeKML:
Yes, sorry about that.

PR #4974, opened 2025-08-22
Fix skip-link styling on error pages

MartinJJones:
## What

Import `skip-link` component styles

## Why

Fix skip-link styling on error pages


kevindew (approved):
Nice one, thanks for fixing so quickly.

PR #5003, opened 2025-09-16
PP-7352 Allow travel_advice via GraphQL with query param

mike3985:
This change results in Frontend serving travel_advice pages using Publishing API's "/graphql/content" endpoint as its backend data source, but this will only happen if the `?graphql=true` query param is included with the request. If the query param isn't set or if it's not set to a recognised value or if it's set to "false", then Content Store data will be used as normal.

https://gov-uk.atlassian.net/browse/PP-7352


KludgeKML (approved):
LGTM

PR #5027, opened 2025-09-26
[PNP-9586] Add manual title concern

sairamya93:

## What

Move `manual_section` content_item from `government_frontend` to `frontend`.
Note: `published` and `other_metadata` methods in the content_item would be handled in the corresponding [shared manual header views](https://github.com/alphagov/government-frontend/blob/3012ee93eeba7f9fbc58dba09a0f9e04d5691cf6/app/views/content_items/manuals/_header.html.erb#L25) as they contain the `public_updated_date` metadata.  

## Why

`manual_section` content_item is shared with `manual_section` and `hmrc_manual_section` document_types. We migrate the shared component first to avoid conflicts and duplication when moving over the dependent routes.

[JIRA Issue](https://gov-uk.atlassian.net/jira/software/c/projects/PNP/boards/1356?selectedIssue=PNP-9586)

### DIFF
--- a/spec/support/concerns/manual_sections.rb
+++ b/spec/support/concerns/manual_sections.rb
@@ -1,3 +1,5 @@
+require "gds_api/test_helpers/content_store"
+
 RSpec.shared_examples "it can have manual title" do |document_type, example_name|
   let(:content_item) { GovukSchemas::Example.find(document_type, example_name:) }
 
@@ -94,5 +96,36 @@ RSpec.shared_examples "it can have manual base path" do |document_type, example_
 
   it "returns the manual base path" do
     expect(described_class.new(content_item).manual_base_path).to eq(content_item.dig("details", "manual", "base_path"))
+  end
+end
+
+RSpec.shared_examples "it can have manual content item" do |_document_type, _example_name|
+  include GdsApi::TestHelpers::ContentStore
+
+  let(:base_path) { "/test-manual" }
+
+  let(:content_item) do
+    {
+      "base_path" => base_path,
+      "details" => {
+        "body" => "Test manual",
+        "headers" => [],
+        "manual" => {
+          "base_path" => "/test-manual",
+        },
+      },
+      "title" => "Manual Content Item",
+      "manual" => { "title" => "Test manual" },
+      "document_type" => "manual_section",
+    }
+  end
+
+  before do
+    stub_content_store_has_item(base_path, content_item)
+  end
+
+  it "returns the manual content item" do
+    manual = described_class.new(content_item).manual_content_item
+    expect(manual.content_store_response.parsed_content).to include(content_item)
   end
 end
--- a/app/models/concerns/manual_sections.rb
+++ b/app/models/concerns/manual_sections.rb
@@ -30,6 +30,13 @@ module ManualSections
     def manual_base_path
       content_store_response&.dig("details", "manual", "base_path")
     end
+
+    def manual_content_item
+      @manual_content_item ||= begin
+        response = ContentItemLoaders::ContentStoreLoader.new.load(base_path: manual_base_path)
+        ContentItemFactory.build(response)
+      end
+    end
   end
 
 private


app/models/concerns/manual_parts.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    if hmrc?
+      I18n.t("formats.manuals.hmrc_title", title:)
+    else
+      I18n.t("formats.manuals.title", title:)
+    end
+  end
+
+  def hmrc?
+    %w[hmrc_manual hmrc_manual_section].include?(schema_name)
```
KludgeKML: Is this manual parts concern going to be included in `hmrc_manual` items, or just `hmrc_manual_section` items?

app/models/concerns/manual_parts.rb
```diff
@@ -26,6 +26,14 @@ def breadcrumbs
         },
       ]
     end
+
+    def manual_content_item
+      @manual_content_item ||= Services.content_store.content_item(base_path)
```
KludgeKML: I think we shouldn't have this - is this supposed to be pointing to the manual that this manual section is linked to? If so, we should be handling it via `linked("manual")` - or some function built into that if the linked item doesn't have enough information.

spec/support/concerns/manual_parts.rb
```diff
@@ -0,0 +1,29 @@
+RSpec.shared_examples "it can have manual title" do |document_type, example_name|
```
deborahchua: Could we include a link to the tests in the commit description?

app/models/concerns/manual_parts.rb
```diff
@@ -9,6 +9,13 @@ def title
     def page_title
       "#{breadcrumb} - #{manual_page_title}"
     end
+
+    def document_heading
```
deborahchua: Can we include a link to the audit trail from government-frontend to the description in commit https://github.com/alphagov/frontend/pull/5027/commits/d26a7929a8a434fcc8455a7c361313e9052fd3f8

app/models/concerns/manual_parts.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def breadcrumb
+    content_store_response["details"]["section_id"] || title
+  end
+
+  def manual_page_title
+    title = content_store_response["title"] || ""
+    title += " - " if title.present?
+
+    if hmrc?
```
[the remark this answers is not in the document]
sairamya93: Hmm the particular concern - `manual_parts` consists of `page_title` which makes a call to `manual_page_title` alias method. As that method is in `manual` content_item in [government_frontend](https://github.com/alphagov/government-frontend/blob/main/app/presenters/content_item/manual.rb#L16),I have added it here along with the dependent method i.e `hmrc?`
If I remember correctly, we had decided on not having methods in one concern calling methods in another concern.
As both `hmrc_manual` and `hmrc_manual_section` call this method through `manual` content_item, I have included that in this shared concern.

app/models/concerns/manual_parts.rb
sairamya93: I suppose we would have to refactor [manual](https://github.com/alphagov/government-frontend/blob/c035e4ba316b90a485b4cf34cacb8f6fa4a64c00/app/presenters/content_item/manual.rb) content_item too to include the relevant methods required in one concern. We aim to to avoid dependency of methods between concerns. In that case `hmrc_manual` would need to include the shared `manual_parts` concern.

app/models/concerns/manual_parts.rb
sairamya93: We actually use the method `parsed_content_item` [here](https://github.com/alphagov/government-frontend/blob/c035e4ba316b90a485b4cf34cacb8f6fa4a64c00/app/views/layouts/application.html.erb#L13). I am not sure if we really require here as we have not migrated it yet but looks like we can possibly avoid it.

spec/support/concerns/manual_sections.rb
```diff
@@ -0,0 +1,29 @@
...
+      expect(described_class.new(content_item).page_title).to eq("Content design: planning, writing and managing content - What is content design? - HMRC internal manual")
+    end
+  end
+
+  context "when schema is not hmrc manual" do
+    it "knows it has page title" do
+      expect(described_class.new(content_item).page_title).to eq("Content design: planning, writing and managing content - What is content design? - Guidance")
+    end
+
+    it "does not add separator when title is blank" do
```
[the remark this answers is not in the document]
sairamya93: Yes right hyphen is the separator. 
Page title comprises of `title`, `manual_page_title` and one of `guidance` or `hmrc internal manual`. So if there is no title, we just don't add the hyphen and render `manual_page_title` + `guidance`.

PR #5030, opened 2025-09-30
NAV-18497 Ensure h1 elements have the same font-size on travel advice pages

MartinJJones:
## What

Ensure h1 elements have the same font-size on travel advice pages

## Why

There are two h1 elements on travel advices pages that have different font sizes. When different font sizes are used for each h1 element, it implies a structure and relationship that may not be reliably perceived by some users of assistive technology.

This change ensures that both h1 elements use the same font-size to fix the accessibility issue, Level A failure of WCAG 2.2 Success Criteria 1.3.1 Info and Relationships.

Jira: https://gov-uk.atlassian.net/browse/NAV-18497

### Mobile

| Before | After |
| --- | --- |
| | |

### Desktop

| Before | After |
| --- | --- |
| | |


jon-kirwan (approved):
Thanks @MartinJJones, looks good to me!

PR #5031, opened 2025-09-30
Fix `travel_advice` page styles

MartinJJones:
## What

Ensure the correct selector is used on travel advice pages, updated from `.travel-advice` to `.travel_advice`.

## Why

Fix travel_advice page styles, noticed the issue when working on https://github.com/alphagov/frontend/pull/5030

## Visual Changes

Preview link: https://govuk-frontend-app-pr-5031.herokuapp.com/foreign-travel-advice/afghanistan

| Before | After |
| --- | --- |
| mobile-before | mobile-after |

### DIFF
--- a/app/assets/stylesheets/views/_travel-advice.scss
+++ b/app/assets/stylesheets/views/_travel-advice.scss
@@ -70,7 +70,7 @@
 
 // TODO: remove input margin and border reset once the global input[type="text"]
 // styles are gone from static.
-.travel-advice {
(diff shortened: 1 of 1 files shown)


jon-kirwan (approved):
Good find! The change looks good to me.

I'm wondering if we should also create a backlog ticket to align the image and download link with the left margin of the text content on this page? Right now, it seems to have roughly 40px of extra spacing from the browser.

PR #5037, opened 2025-10-03
Introduce complete fallback behaviour for lead image

lauraghiorghisor-tw:
## What
Introduce additional logic for defaulting to news image. Fallback to primary organisation if available (as it was), then to the first organisation, then to worldwide organisation.

## Why
This behaviour did not actually get triggered (at least from whitehall) because we have the same duplicate behaviour in the publisher. Whitehall used this fallback logic and produced a lead image based on the organisation default news image, meaning it got to this bit code with a lead image already set.

We are keen to remove that logic from whitehall, since it already exists here. Nonetheless, whitehall has some additional fallback logic, so it defaults to primary org, then to organisation if primary is not set, and eventually to worldwide organisation if it's available. The latter is only the case for WorldNewsStory.

The context in which this work is being done is that we are reimplementing some WH functionality, and reevaluating business logic. There is some UR taking place that might mean we can remove this behaviour altogether. We are also exploring what the similitudes are between the default news image and crest usage, since they seem to largely serve the same purpose of departmental "branding".

Moving all of the defaulting behaviour to FE makes sense because it’s already opinionated about it, and it might need that logic for other publishing apps. It also has opinions about crests, and it feels like having both lead images and crests in the same repo is sensible.

## Screenshots
TBC (will test)

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2549)


lauraghiorghisor-tw:
Closed as keeping this functionality WH side for now.

PR #5077, opened 2025-10-16
Remove phase banner support from Frontend

KludgeKML:

## What

Remove phase banner support from Frontend.

## Why

There are no document types in phases alpha or beta that are rendered by frontend, and there aren't likely to be any any time soon (there are none in government-frontend, arguably no valid ones in collections or finder-frontend either), so we can just remove support for it entirely and reduce style downloads slightly.

## Visual changes

None, because nothing on frontend uses the banner at the moment (or maybe ever)


andysellick (approved):
Looks good 👏 

The translation files contain entries in the `help/index` area for `beta` and `beta_description` - are these related to the phase banner at all?

PR #5092, opened 2025-10-23
[PNP-9666] simplify layouts in frontend

KludgeKML:

## What

The current layouts in frontend are inverted (body/head/call to layout) to support the individual component loading. Since we're not doing that at the moment, put the elements back in order.

This change is expected to make no visual difference to any page.

## Why

https://gov-uk.atlassian.net/browse/PNP-9666


app/views/layouts/full_width.html.erb
andysellick: Is it possible to have full width extend from the application layout? It seems like we end up duplicating a fair bit of stuff here.

PR #5102, opened 2025-10-28
Add 'feed' flexible_section

AshGDS:

## What / Why
- Adds a Flexible section called `Feed` that includes four things - the `document_list` component, `share_links` component, a `see all` link and our `subscription_links` component
- Also adds GA4 tracking to the HTML where appropriate
- What I've done should hopefully be a fairly inoffensive way of implementing it - though I did use `data-flexible-section` attributes to make the HTML a bit easier to target in tests, especially considering the flexibility of these pages could make it harder to target them by their position in the DOM when new things are added. But happy to remove that if needed.

Test page: https://govuk-frontend-app-pr-5102.herokuapp.com/test/flexible-page

Jira card: https://gov-uk.atlassian.net/browse/9667

## Visual changes

image

### DIFF
--- a/app/models/flexible_page/flexible_section/feed.rb
+++ b/app/models/flexible_page/flexible_section/feed.rb
@@ -1,11 +1,10 @@
 module FlexiblePage::FlexibleSection
   class Feed < Base
-    attr_reader :context, :email_signup_link, :feed_heading_text, :see_all_items_link, :share_links_heading_text
+    attr_reader :email_signup_link, :feed_heading_text, :see_all_items_link, :share_links, :share_links_heading_text
 
     def initialize(flexible_section_hash, content_item)
       super
 
-      @context = flexible_section_hash["context"]
       @feed_heading_text = flexible_section_hash["feed_heading_text"]
 
       @share_links_heading_text = flexible_section_hash["share_links_heading_text"]
@@ -27,17 +26,6 @@ module FlexiblePage::FlexibleSection
             public_updated_at: Time.zone.parse(i["metadata"]["public_updated_at"] || ""),
             document_type: i["metadata"]["document_type"],
           },
-        }
-      end
-    end
-
-    def share_links
-      flexible_section_hash["share_links"].map do |i|
-        {
-          href: i["href"],
-          text: i["text"],
-          icon: i["icon"],
-          data_attributes: i["data_attributes"],
         }
       end
     end
--- a/app/views/flexible_page/flexible_sections/_feed.html.erb
+++ b/app/views/flexible_page/flexible_sections/_feed.html.erb
@@ -14,14 +14,13 @@
     },
   } %>
 
-  <%= tag.p(class: "govuk-body", data: { flexible_section: "feed_see_all", module: "ga4-link-tracker", ga4_link: { event_name: "navigation", type: "see all", section: flexible_section.feed_heading_text }.to_json }) do %>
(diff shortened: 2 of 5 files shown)


app/models/flexible_page/flexible_section/feed.rb
andysellick: Do we have presenters for flexible content? Fine if not, but it feels like we're doing some presenter-y things here in the model, wondering if it would be better to get all the individual bits of data here, then have a presenter that brings them all together to spit out the individual bits needed for each component to then pass to the view?

app/views/flexible_page/flexible_sections/_feed.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  <%= render "govuk_publishing_components/components/heading", {
+    text: flexible_section.share_links_heading_text,
+    margin_bottom: 1,
+    padding: true,
+    border_top: 5,
+  } %>
+
+  <%= render "govuk_publishing_components/components/share_links", {
+    square_icons: true,
+    flexbox: true,
```
andysellick: What does the flexbox option here do?

app/views/flexible_page/flexible_sections/_feed.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="govuk-grid-column-two-thirds" data-flexible-section="feed_document_list">
```
andysellick: About the data attributes - since we know what's going to be in this particular bit of the page (so far) it might be enough to put one data attribute wrapping the whole section, and base the tests off that. For example the test for the 'see all' link - we know there should be a link with that href and text somewhere in here, I think it's enough that it's within this section rather than more granular.

app/models/flexible_page/flexible_section/feed.rb
```diff
@@ -0,0 +1,45 @@
...
+    def share_links
+      flexible_section_hash["share_links"].map do |i|
+        {
+          href: i["href"],
+          text: i["text"],
+          icon: i["icon"],
+          data_attributes: i["data_attributes"],
+        }
+      end
+    end
```
KludgeKML: You could simplify this quite a lot - down to basically this:

`flexible_section_hash["share_links"].map(&:symbolize_keys)`

...which is probably simple enough that you could put it directly in the view.

app/models/flexible_page/flexible_section/feed.rb
```diff
@@ -0,0 +1,45 @@
+module FlexiblePage::FlexibleSection
+  class Feed < Base
+    attr_reader :context, :email_signup_link, :feed_heading_text, :see_all_items_link, :share_links_heading_text
+
+    def initialize(flexible_section_hash, content_item)
+      super
+
+      @context = flexible_section_hash["context"]
```
KludgeKML: We don't actually use a context in this view, do we?

app/views/flexible_page/flexible_sections/_feed.html.erb
andysellick: No, it's good, I just couldn't see the difference with/without 👍

app/views/flexible_page/flexible_sections/_feed.html.erb
AshGDS: Thanks, I've simplified it now 👍

AshGDS:
@andysellick @KludgeKML  thanks both - should be ready for another review now (have put the changes in a separate commit then will squash)

PR #5426, opened 2026-03-24
[PNP-9822] expose place as geojson

KludgeKML:

## What

Exposes a places-manager service data set as GeoJSON, suitable for consumption by a JS map component.

In order to avoid exposing the endpoint in production / limiting it to
  certain datasets, at the moment this is controlled by an ALLOWED_GEOJSON_SLUGS
  variable which has to contain the service slug from places-manager. Any
  unpermitted slugs will return a 404.
- Called by (eg) /api/places/driving-test-centres.geojson
- Require /api/places added as a prefix special_route
  (see https://github.com/alphagov/publishing-api/pull/3971)
- Currently routes are not cached, before production release we will either
  add caching or consider whether to write these to (eg) an S3 bucket.

## Why

Jira card: https://gov-uk.atlassian.net/browse/PNP-9822

### DIFF
--- a/app/controllers/api/places_controller.rb
+++ b/app/controllers/api/places_controller.rb
@@ -1,0 +1,50 @@
+# rubocop:disable Rails/ApplicationController
+module Api
+  class PlacesController < ActionController::Base
+    def show
+      return render json: {}, status: :not_found unless allowed_slugs.include? params[:service_slug]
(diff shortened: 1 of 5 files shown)


deborahchua (approved):
LGTM 👍

PR #5425, opened 2026-03-24
[WIP] [DO NOT MERGE] Extend prototype map component

andysellick:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What
Prototype a basic map component.

## Why

https://gov-uk.atlassian.net/jira/software/c/projects/PNP/boards/1356?quickFilter=1456&selectedIssue=PNP-9892&useStoredSettings=true

PR #5455, opened 2026-04-07
Add map component

andysellick:

## What
Add a basic map component. This PR is based on the first commit from https://github.com/alphagov/frontend/pull/5425, which implements a basic map component using leaflet. That other PR will continue, in order to explore using [DEFRA's more accessible map](https://defra.github.io/interactive-map/), however that work may depend on features not yet implemented. In the meantime, we'll work on this PR.

Todo:

- [x] output full popup text in external list
- [x] sort external list alphabetically
- [x] hide external list if JS disabled
- [x] fix CORS issue
- [ ] sort out OS maps API key
- [x] redo the yarn.lock, far too big for one change?

The map depends upon an OS maps API key, which is stored as an environment variable in production/staging/etc. so it's possible it might not work locally for anyone else. If that's the case I can supply it for reviewing purposes.

## Why

https://gov-uk.atlassian.net/jira/software/c/projects/PNP/boards/1356?quickFilter=1456&selectedIssue=PNP-9892&useStoredSettings=true


andysellick:
@MartinJJones I've attempted to address your comments, including:

- `that` has been removed thanks to the arrow function suggestion
- `GOVUK.extendObject` has been replaced with `Object.assign`
- all markers have a popup by default
- popup close icon increased size

app/assets/javascripts/components/map.js
```diff
@@ -0,0 +1,141 @@
...
+        maxBounds: [
+          [49.5, -10.5],
+          [62, 6]
+        ],
+        attributionControl: false
+      }
+      const passedConfig = JSON.parse(this.$module.getAttribute('data-config')) || {}
+      this.config = Object.assign(allMapOptions, passedConfig)
+      this.markers = JSON.parse(this.$module.getAttribute('data-markers'))
+      this.geoJsonUrl = this.$module.getAttribute('data-geojson')
```
AshGDS, line 44: I'm guessing our content security policy will block any geojson download that isn't hosted on a GOV.UK domain (which is a good thing) but should we double check that this is the case, and if not put a safeguard in the code too?

app/views/components/_map.html.erb
```diff
@@ -0,0 +1,53 @@
...
+    <% end %>
+    <div id="<%= skip_link_id %>"></div>
+    <% if markers and !hide_markers_beneath %>
+      <div class="app-c-map__markers-list">
+        <%= render "govuk_publishing_components/components/details", {
+          title: "Locations on this map",
+          classes: "js-list-markers",
+        } do %>
+          <ul class="govuk-list govuk-list--bullet">
+          </ul>
```
AshGDS, line 62: Might need a comment to explain why we're rendering an empty `<ul>`?

app/assets/config/manifest.js
```diff
@@ -3,13 +3,16 @@
...
+//= link components/map/leaflet.js
 //= link views/travel-advice.js
 //= link views/landing_page/map/leaflet.js
 //= link views/landing_page/map/data/lookup.js
 //= link views/landing_page/map/data/cdc.geojson.js
 //= link views/landing_page/map/data/hub.geojson.js
 //= link views/landing_page/map/data/icb.geojson.js
 
+//= link components/map.js
+
```
AshGDS: Nitpick: Should both component statements be together? I'm guessing it was done like this because the landing page needs `leaflet.js` too.

AshGDS (dismissed):
LGTM! 👍 I think in the first commit it may be worth removing the line about content security policy issues.

andysellick:
@AshGDS thanks for reviewing but in addition to your comments there are a few more things I think this PR needs, so can you withdraw your approval?

app/assets/javascripts/components/map.js
andysellick: The CORS policy will block any request to an unrecognised domain, but I've added a check to not attempt it unless the URL is relative anyway.

app/views/components/_map.html.erb
andysellick: Added 👍

app/assets/config/manifest.js
andysellick: Absolutely, have fixed.

AshGDS:
@andysellick No worries, have dismissed my approval 👍

andysellick:
@AshGDS I haven't finished implementing all the features I think this needs, but I think this has been in review long enough and it works as an MVP plus, so ready for another review when you are.

AshGDS (approved):
LGTM 👍

PR #5474, opened 2026-04-14
Assume No. 10 theme on landing pages

KludgeKML:
- The current landing pages aren't going to be expanded on, so the PfC pages are the only ones that will exist. To get round a change in publishing where the theme has gone missing, we simply default to the No. 10 theme and remove the default theme option.

Before: (https://www.gov.uk/missions/nhs/progress) 
image

After: (https://govuk-frontend-app-pr-5474.herokuapp.com/missions/nhs/progress)
image


AshGDS (approved):

maxgds (approved):
Not thrilled about hacking it but it's the path of least resistance at this point.

KludgeKML:
> Not thrilled about hacking it but it's the path of least resistance at this point.

@maxgds To be honest, I think this is actually a clean PR (otherwise I'd add a tech debt issue for it at the same time). I don't think any new landing page work is going to build on the PfC landing page system, so anything we can remove from it would be fine.

(I'd be delighted if we could replicate PfC in the new flexible pages system and tear this all out).

PR #5481, opened 2026-04-16
Fix impact header captions when empty strings

andysellick:

## What
- it looks like the content item can sometimes contain an empty string for the image caption, which previously was getting through and causing the details component around the caption to render, with nothing inside it
- adds some extra checks for this kind of situation

## Why
Found during testing of content on integration.

## Visual changes
Fixes this situation:

Screenshot 2026-04-16 at 10 59 38

### DIFF
--- a/app/models/flexible_page/flexible_section/impact_header.rb
+++ b/app/models/flexible_page/flexible_section/impact_header.rb
@@ -43,7 +43,7 @@ module FlexiblePage::FlexibleSection
     end
 
     def build_caption
-      return unless image && image[:caption] && flexible_section_hash["image_type"] != "logo"
+      return unless image && image[:caption].present? && flexible_section_hash["image_type"] != "logo"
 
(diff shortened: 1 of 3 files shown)


app/models/flexible_page/flexible_section/impact_header.rb
```diff
@@ -43,7 +43,7 @@ def build_modifier_classes
     end
 
     def build_caption
-      return unless image && image[:caption] && flexible_section_hash["image_type"] != "logo"
+      return unless image && !image[:caption].to_s.strip.empty? && flexible_section_hash["image_type"] != "logo"
```
KludgeKML: You can replace `!image[:caption].to_s.strip.empty?` with `image[:caption].present?`, which is only true if the value is not nil, not an empty string, and not a string of purely whitespace.

andysellick: Brilliant thanks, have changed.

KludgeKML (approved):
LGTM!

PR #5484, opened 2026-04-16
Hide feed latest section if no documents

andysellick:

## What / why
- if there's nothing attached to a topical event page at the moment, the feed section appears with no content in it
- we've been advised that this shouldn't appear in this situation, not even to show the 'get emails' link
- adjusting the code to do this means that if there are no documents but there is a share links, it'll mean the share links appears by itself in a one third column with some whitespace to the right

Might need some design thought around this as having a one third column by itself on a row leaves a lot of space to the right. Having said that, it's probably unlikely that this situation would occur often.

## Visual changes
Feed section without any documents:

Screenshot 2026-04-16 at 16 52 33

### DIFF
--- a/app/models/flexible_page/flexible_section/feed.rb
+++ b/app/models/flexible_page/flexible_section/feed.rb
@@ -21,6 +21,8 @@ module FlexiblePage::FlexibleSection
     end
 
     def items_no_description
+      return unless items
+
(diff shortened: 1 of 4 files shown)


KludgeKML (approved):
LGTM

PR #5496, opened 2026-04-21
Use DEFRA map with map component

andysellick:

## What
Modifies the map component to use the [DEFRA Interactive map](https://defra.github.io/interactive-map/).

Improvements:

- new map is navigable with a keyboard, focus on keyboard shows a shortcut for all key bindings
- new feature: custom marker shapes and colours
- heading levels throughout component are now calculated automatically based on top level heading level option
- new feature: can set an optional manual bounding box for the map, instead of letting it be automatically determined by marker positions
- new map will allow for further accessibility improvements (not yet implemented)

Replaces https://github.com/alphagov/frontend/pull/5488

## Why
We're trying it out as it offers a lot of accessibility features to make maps more usable.

## Visual changes
New map, showing keyboard shortcut menu on focus:

Screenshot 2026-05-01 at 11 52 02


andysellick:
Hey @AshGDS I think this is ready for review now. The obvious outstanding issue is the automated security alert above, but ignore that for now, still under discussion.

I haven't implemented as many accessibility fixes from our testing session last week as I can't get any screenreaders to access the map markers anymore, which is a bit strange. Still planning to hand over to Calum for some proper testing, that'll be the next step.

app/assets/javascripts/components/map.js
```diff
@@ -123,43 +102,82 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-        }
+      if (popupsListWrapper && popupsListEl) {
+        popupsListWrapper.classList.add('app-c-map__markers-list--visible')
+        var popupsList = []
+        this.markers.forEach(marker => {
+          popupsList.push(this.createPopupContent(marker))
+        })
+        popupsList.forEach(popup => {
+          const listItem = document.createElement('li')
+          // listItem.appendChild(popup)
```
AshGDS: Code comment left in?

AshGDS (commented):
Code generally looks good to me, quite a lot of new code though so there's a chance I missed something 😄 

I experienced some bugs while testing, not sure if they're caused by you though:

1. Not sure if intentional but there is no back button on the first two map previews on mobile?

Image

(Also, why does the back button on mobile have a disabled "Continue" button next to it as well?)

2. Sometimes I managed to get in a state where clicking "Map view" would cause the button to do nothing and disappear. Then when a Map view button did work, pressing the back button in the browser would cycle through all the maps that never displayed. I'm guessing the code is getting in a weird state somehow, maybe when multiple maps are on the page? Seems like you can replicate it by opening `/component-guide/map/preview` in  mobile view in Incognito, clicking Map view on the first map, hitting the browser back button, then clicking Map view on another map?

Image

3. In VoiceOver it looks like this text needs a space after the K key, otherwise it says "Press Alt + Kto view keyboard shortcuts"

Image

4. Sometimes pressing Alt + K doesn't seem to do anything? It seems to work intermittently 

5. I think there needs to be a message displayed when JavaScript is disabled

6. Looks like the map doesn't work at all in Microsoft Edge?

Image

7. Also worth keeping an eye on the performance of a page while using this component - looks like the component preview alone uses about 5x the RAM of the GOVUK homepage depending on how the map is interacted with (Homepage is 60MB vs 250MB-350MB for the map) - probably fine for most modern devices but might want to keep an eye on budget devices like Chromebooks with 4GB RAM total

andysellick:
@AshGDS thanks for the review.

1. and 2. As you noted, it's sometimes impossible to exit the map on mobile when it is triggered fullscreen from the 'map' button, so I have reconfigured the map to always be inline, to sidestep these issues.

3. and 4. The keyboard shortcuts are internal to the map, so I will raise with the developer, thanks.

5. Have added some text for JS/no JS.

6. and 7. Will raise with the developer.

I'll raise some of the issues above with the map developer, as they are internal to that code.

app/assets/javascripts/components/map.js
andysellick: Whoops, fixed, thanks.

AshGDS (approved):
LGTM 👍 

Not sure if the performance thing will actually be an issue - I've seen some long guidance pages on GOVUK using the same amount of RAM.

andysellick:
@KludgeKML I think this is pretty much ready to go - only thing is the security alert on appending content into the DOM.

app/assets/stylesheets/components/_map.scss
```diff
@@ -1,3 +1,4 @@
+@import "@defra/interactive-map/dist/css/index";
```
KludgeKML, line 1: Does this mean that we end up serving Defra CSS to the PfC Map?

KludgeKML: Oh, I see you remove the leaflet bit in a later commit.

app/assets/javascripts/components/map.js
```diff
@@ -123,43 +102,53 @@
...
+      if (popupsListWrapper && popupsListEl) {
+        popupsListWrapper.classList.add('app-c-map__markers-list--visible')
+        var popupsList = []
+        this.markers.forEach(marker => {
+          popupsList.push(this.createPopupContent(marker))
+        })
+        popupsList.sort()
+        popupsList.forEach(popup => {
+          const listItem = document.createElement('li')
+          listItem.innerHTML = popup
```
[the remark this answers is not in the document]
KludgeKML: In this case we are controlling the HTML that comes into the DOM - it's loaded from gov.uk frontend, and covered by the CSP. So we believe this to be a false alarm.

KludgeKML (approved):
LGTM

PR #5506, opened 2026-04-23
Update `Gone` to handle `Archived` pages [WHIT-3282]

patrickpatrickpatrick:
## What

Update `Gone` to handle new archived unpublishing reason.

## Why

`Archived` is a new unpublishing reason that has been added to Whitehall. The previous implementation `Gone` assumes the page was removed in error, so it needed to be changed to take into account this new reason.

Related to https://github.com/alphagov/whitehall/pull/11355

## Visual changes

For a page that has a URL that is of the National Archives:

Screenshot 2026-04-23 at 15 13 30

No visual changes for `Gone` pages that were published in error.


ChrisBAshton (approved):
Nice one 🎉 

I'd probably move the locale change into its own (first) commit, then a commit where you use the locale field 👍 That way CI passes each time. But code looks great!

PR #5507, opened 2026-04-24
Change impact header image alignment

andysellick:

## What / why
- change the alignment of the impact header image on desktop from centered to left aligned
- this means that on smaller desktop screens only the right hand side of the image will be cropped, rather than both sides
- this is being changed to accommodate an image on a topical event that is less of a photo and more of a logo, where cropping both sides makes it look wrong but the 'logo' slot is already occupied with an image, so we can't treat it as a logo
- cropping only the right hand side of the image on smaller desktop screens means that it looks much better, even though the right hand side is still cropped

This is a surprisingly complex change due to two key problems:

- The image caption didn't line up with the image on small desktop screens because it was wrapped in a `govuk-width-container`, which has margin either side, whereas the elements above don't. I've recreated the layout of the text/image in the caption element by using a pseudo element to serve as a blank copy of the text box, and adjusted the styles to match
- The impact header without a background uses a max width on the picture element to line up the right hand edge of the image with any content beneath it (when there's a background it extends out by an extra 30px). In order to make this behave I've had to rewrite the layout without a background considerably - there's now no gap in the flex layout, the gap is instead achieved with padding on the text details element.

## Visual changes

Before | After
----- | ----
Screenshot 2026-04-24 at 10 24 35 | Screenshot 2026-04-24 at 10 24 14

### DIFF
--- a/app/assets/stylesheets/views/_flexible-content.scss
+++ b/app/assets/stylesheets/views/_flexible-content.scss
@@ -11,13 +11,36 @@ $gap: 30px;
   margin-bottom: 30px;
 }
 
-.impact-header__caption {
-  padding-top: 15px;
-  padding-bottom: 15px;
+.impact-header__caption-wrapper {
+  max-width: 1020px;
+  margin: 0 auto;
 
   @include govuk-media-query($from: desktop) {
-    margin-left: 50%;
-    padding-left: $gap / 2;
+    display: flex;
+    gap: $gap;
+
+    &::before {
+      content: "";
+      display: block;
+      height: 1px;
+      width: 50%;
+      flex-shrink: 1;
+      max-width: 495px;
+    }
+  }
+}
+
+.impact-header__caption {
+  padding: 15px;
+
+  @include govuk-media-query($from: tablet) {
+    padding: 30px;
+  }
+
+  @include govuk-media-query($from: desktop) {
+    box-sizing: border-box;
+    width: 50%;
+    padding-left: 0;
   }
 }
 
@@ -49,7 +72,6 @@ $gap: 30px;
   @include govuk-media-query($from: desktop) {
     box-sizing: border-box;
     width: 50%;
-    max-width: $box-size;
     padding: 30px 0 30px 30px;
   }
 }
@@ -147,9 +169,20 @@ $gap: 30px;
 }
 
 .impact-header--plain {
-  .impact-header__image {
-    @include govuk-media-query($from: desktop) {
-      transform: translateX(-$gap);
+  @include govuk-media-query($from: desktop) {
+    .impact-header__container {
+      gap: 0;
+    }
+
+    .impact-header__detail {
+      padding-left: $gap;
+      padding-right: $gap;
+      max-width: calc(50% + 15px);
+      width: auto;
+    }
+
+    .impact-header__image-wrapper {
+      max-width: 465px;
     }
   }
 }
--- a/app/views/flexible_page/flexible_sections/_impact_header.html.erb
+++ b/app/views/flexible_page/flexible_sections/_impact_header.html.erb
@@ -28,7 +28,7 @@
         alt_text: "" %>
     </div>
     <% if flexible_section.caption %>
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/views/_flexible-content.scss
```diff
@@ -11,13 +11,40 @@ $gap: 30px;
...
   @include govuk-media-query($from: desktop) {
-    margin-left: 50%;
-    padding-left: $gap / 2;
+    box-sizing: border-box;
+    width: 50%;
+    flex-shrink: 1;
+    max-width: 495px;
+    padding-left: 0;
+    padding-right: $gap;
   }
```
AshGDS, line 44: Do we need the `flex-shrink`, `max-width` and `padding-right` here? (I can't really see a difference when toggling them)

app/assets/stylesheets/views/_flexible-content.scss
```diff
@@ -69,6 +96,8 @@ $gap: 30px;
     aspect-ratio: auto;
     max-width: $box-size;
     align-self: flex-end;
+    display: flex;
+    justify-content: flex-start;
```
AshGDS: Similarly, do we need `max-width` and `display: flex` on this anymore?

app/assets/stylesheets/views/_flexible-content.scss
```diff
@@ -11,13 +11,40 @@ $gap: 30px;
   margin-bottom: 30px;
 }
 
+.impact-header__caption-wrapper {
+  max-width: 1020px;
+  margin: 0 auto;
+
+  @include govuk-media-query($from: desktop) {
+    display: flex;
+    justify-content: flex-end;
```
AshGDS: Do we need `justify-content: flex-end`?

app/assets/stylesheets/views/_flexible-content.scss
AshGDS: We may only need one of `width` and `max-width` here?

AshGDS (commented):
Looks good to me  👍  was just wondering if we can remove some CSS rules now - I was toggling things on and off to understand how everything worked and found some rules that didn't seem to be doing anything, though I may be wrong.

app/assets/stylesheets/views/_flexible-content.scss
andysellick: `width: auto` overrides the mobile/tablet setting of 50%, without which the box isn't wide enough (this is the one I had to restructure for the non-background variant).

`max-width` similarly overrides the mobile styles, allowing this box width to equal half the space plus the gap, which has been removed for the non-background variant.

I found when toggling either of them the box size would collapse to less than what it should be.

app/assets/stylesheets/views/_flexible-content.scss
andysellick: No we don't - good shout, thanks.

app/assets/stylesheets/views/_flexible-content.scss
andysellick: No we don't - good shout, thanks.

app/assets/stylesheets/views/_flexible-content.scss
andysellick: No we don't - good shout, thanks.

app/assets/stylesheets/views/_flexible-content.scss
AshGDS: Thanks - no worries if it's not needed. Just double checking, you mentioned `width: auto` here but I was referring to `width: 50%`/`max-width: 495px` on `.impact-header__detail` on variants that aren't plain

AshGDS (approved):
LGTM 👍

app/assets/stylesheets/views/_flexible-content.scss
andysellick: Ah sorry for the confusion. I've checked again and I think you're right - we don't need the max-width. But please double check for me 😁 (have pushed)

AshGDS (approved):
Nice, thanks 👍 LGTM

PR #5508, opened 2026-04-24
Realign impact header images

andysellick:

## What
Actually fixes the problem https://github.com/alphagov/frontend/pull/5507 was raised for.

## Why
Somehow in the midst of the complexity of the previous PR I managed to not solve the original point of it. So this is the fix.

## Visual changes
Impact header image on desktop is now aligned to its left edge, rather than being centered. This means that when the screen is a narrow desktop, only the right hand edge of the image is lost, rather than both left and right edges. This allows images to look more naturally cropped, particularly in the situation below.

Before | After
---- | -----
Screenshot 2026-04-24 at 10 24 35 | Screenshot 2026-04-24 at 10 24 14

### DIFF
--- a/app/assets/stylesheets/views/_flexible-content.scss
+++ b/app/assets/stylesheets/views/_flexible-content.scss
@@ -82,7 +82,7 @@ $gap: 30px;
   overflow: hidden;
   display: flex;
   align-items: center;
-  justify-content: center;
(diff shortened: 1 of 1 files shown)


AshGDS (approved):
LGTM 👍

PR #5516, opened 2026-04-29
[PNP-9961] Pin SHAs for third-party actions

KludgeKML:
Pin non-alphagov/non-github actions

-As per policy: https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#pin-untrusted-github-actions-to-a-commit-sha

JIRA Card: https://gov-uk.atlassian.net/browse/PNP-9961


deborahchua (approved):
SHA matches https://github.com/ruby/setup-ruby/releases/tag/v1.305.0

PR #5527, opened 2026-05-08
Add GA4 indexes to image cards

AshGDS:

## What/Why
- Adds position indexes to image cards on flexible sections when clicked for GA4 tracking purposes
- Also removes some Universal Analytics code I spotted
- As requested by our PA in https://gov-uk.atlassian.net/browse/PNP-9952
- Test page: https://govuk-frontend-app-pr-5527.herokuapp.com/government/topical-events/budget-2025

## Visual changes

None

### DIFF
--- a/app/models/flexible_page/flexible_section/featured.rb
+++ b/app/models/flexible_page/flexible_section/featured.rb
@@ -26,10 +26,18 @@ module FlexiblePage::FlexibleSection
       }
     end
 
-    def ga4_tracking
+    def ga4_tracking(index_link = nil, index_total = nil)
       return nil unless flexible_section_hash["ga4_image_card_json"]
 
-      flexible_section_hash["ga4_image_card_json"].to_json
+      if index_link && index_total
+        index_obj = { index: { index_link: index_link }, index_total: index_total }
+        flexible_section_hash["ga4_image_card_json"].merge!(index_obj)
+      end
+
+      {
+        module: "ga4-link-tracker",
+        ga4_link: flexible_section_hash["ga4_image_card_json"].to_json,
+      }
     end
 
   private
--- a/app/views/flexible_page/flexible_sections/_featured.html.erb
+++ b/app/views/flexible_page/flexible_sections/_featured.html.erb
@@ -9,14 +9,7 @@
 
   <div class="govuk-grid-row">
     <% flexible_section.items.each.with_index(1) do |item, index| %>
-      <%
-        ga4_tracking = {
-          module: "ga4-link-tracker",
-          ga4_track_links_only: true,
-          ga4_link: flexible_section.ga4_tracking,
-        } if flexible_section.ga4_tracking
-      %>
-      <%= tag.div(class: flexible_section.grid_layout_options[:class], data: ga4_tracking) do %>
+      <%= tag.div(class: flexible_section.grid_layout_options[:class], data: flexible_section.ga4_tracking(index, flexible_section.items.count)) do %>
         <%= render "govuk_publishing_components/components/image_card", {
           href: item[:href],
           image_src: item[:image_src],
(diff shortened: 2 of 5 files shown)


app/models/flexible_page/flexible_section/featured.rb
```diff
@@ -26,9 +26,12 @@ def grid_layout_options
       }
     end
 
-    def ga4_tracking
-      return nil unless flexible_section_hash["ga4_image_card_json"]
+    def has_ga4_tracking?
+      flexible_section_hash["ga4_image_card_json"] ? true : false
```
andysellick: I think you might be able to do something like `flexible_section_hash.has_key?('ga4_image_card_json')` here?

app/views/flexible_page/flexible_sections/_featured.html.erb
```diff
@@ -12,9 +12,8 @@
       <%
         ga4_tracking = {
           module: "ga4-link-tracker",
-          ga4_track_links_only: true,
-          ga4_link: flexible_section.ga4_tracking,
-        } if flexible_section.ga4_tracking
+          ga4_link: flexible_section.ga4_tracking({ index: { index_link: index }, index_total: flexible_section.items.count  }),
+        } if flexible_section.has_ga4_tracking?
```
andysellick: This feels a bit like the code is spread across two locations when it could just be in the model? Have you tried something like:

```Ruby
# featured.rb
def ga4_tracking(index_link = nil, index_total = nil)
  return nil unless flexible_section_hash.has_key?('ga4_image_card_json')

  if index_link
    index_obj = { index: { index_link: index_link }, index_total: index_total  }
    flexible_section_hash["ga4_image_card_json"].merge!(index_obj)
  end
  flexible_section_hash["ga4_image_card_json"].to_json
}),
end

# _featured.html.erb
ga4_tracking = {
  module: "ga4-link-tracker",
  ga4_link: flexible_section.ga4_tracking(index, flexible_section.items.count)
}
```

Also not sure from the code but is the call to `ga4_tracking` on line 16 of this file the only call to that method? Because it looks like `index` is always passed (as a hash, although sometimes containing unfilled values) so the `if` on line 34 of `featured.rb` always happens?

Happy to chat if you want some clarification.

andysellick (commented):
index values look like they're being generated correctly on the preview page, but I think the code could be optimised a little.

app/views/flexible_page/flexible_sections/_featured.html.erb
AshGDS: Good idea, thanks - I've refactored it 👍

app/models/flexible_page/flexible_section/featured.rb
AshGDS: I've opted to use `flexible_section_hash["ga4_image_card_json"]` because `has_key?` will return `true` if the key exists but is `nil`, so I think evaluating the value rather than the key is safer.

AshGDS:
Thanks @andysellick - should be ready for another review

app/views/flexible_page/flexible_sections/_featured.html.erb
andysellick: Sorry, thought of a further optimisation. Does this work?

```Ruby
# featured.rb
def ga4_tracking(index_link = nil, index_total = nil)
  return nil unless flexible_section_hash["ga4_image_card_json"]

  if index_link && index_total
    index_obj = { index: { index_link: index_link }, index_total: index_total }
    flexible_section_hash["ga4_image_card_json"].merge!(index_obj)
  end

  {
    module: "ga4-link-tracker",
    ga4_link: flexible_section_hash["ga4_image_card_json"].to_json
  }  
end

# _featured.html.erb
ga4_tracking = flexible_section.ga4_tracking(index, flexible_section.items.count)

```

app/models/flexible_page/flexible_section/featured.rb
andysellick: Okay, fair enough.

app/views/flexible_page/flexible_sections/_featured.html.erb
AshGDS: @andysellick Thanks, have used this instead 👍

andysellick (approved):
👍

PR #5532, opened 2026-05-12
Add cards flexible section

andysellick:

## What
- using flexbox to create a flexible section that will display grid column content at a consistent width regardless of the space it is put into
- our current grid when nested will only make things smaller
- uses a trick to ensure that any elements on a row by themselves don't expand to fill the space, using the before/after pseudo elements (only works where the content is no more than 3 items per row)
- styled to look like our existing cards component

## Why
There seems to be a frequent requirement on campaign sites for something like this - a series of horizontal boxes with simple content that link to other pages.

Also experimenting to see if we can create a flexible section that adapts to its surroundings, rather than having to be told e.g. you're in a two-thirds column.

Future iterations of this flexible section might include allowing image card content to be shown in the boxes.

## Visual changes

Screenshot 2026-05-15 at 10 46 48

https://gov-uk.atlassian.net/jira/software/c/projects/PNP/boards/1356?quickFilter=1456&selectedIssue=PNP-9973

### DIFF
--- a/app/models/flexible_page/flexible_section/cards.rb
+++ b/app/models/flexible_page/flexible_section/cards.rb
@@ -1,0 +1,11 @@
+module FlexiblePage::FlexibleSection
+  class Cards < Base
+    attr_reader :items
+
+    def initialize(items:)
+      super
+
+      @items = items
+    end
+  end
+end
--- a/app/models/flexible_page/flexible_section/cards.rb
+++ b/app/models/flexible_page/flexible_section/cards.rb
@@ -2,10 +2,15 @@ module FlexiblePage::FlexibleSection
   class Cards < Base
     attr_reader :items
 
-    def initialize(items:)
+    def initialize(items:, heading_level: nil)
       super
 
       @items = items
+      @heading_level = [2, 3, 4, 5, 6].include?(heading_level) ? heading_level : 2
+    end
+
+    def heading
+      "h#{@heading_level}"
     end
   end
 end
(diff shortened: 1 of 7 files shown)


KludgeKML (approved):
LGTM

AshGDS (approved):
LGTM 👍 

Should the heading level be customisable? (I'm guessing that was probably out of scope for this PR?)

andysellick:
> Should the heading level be customisable? (I'm guessing that was probably out of scope for this PR?)

Ooh, good feature. Yes, I think I'll add that in, thanks.

andysellick:
@AshGDS have added a custom heading level option, see the new commit 👍

app/models/flexible_page/flexible_section/cards.rb
```diff
@@ -2,10 +2,15 @@ module FlexiblePage::FlexibleSection
   class Cards < Base
     attr_reader :items
 
-    def initialize(items:)
+    def initialize(items:, heading_level: nil)
```
AshGDS, line 5: I guess you could default it to 2 here instead of having the if statement on line 9?

AshGDS (approved):
Nice, thanks - LGTM 👍

app/models/flexible_page/flexible_section/cards.rb
andysellick: You'd think that would work, wouldn't you? But for some reason it doesn't. I've had a similar problem with https://github.com/alphagov/frontend/pull/5584 and I think the whole way we're doing this needs to change slightly, particularly when passing options to components within flexible sections. Is it okay if I address that in a separate PR? It feels like a bigger piece of work.

AshGDS: Ah OK, interesting, no worries 👍

PR #5533, opened 2026-05-12
Rework flexible sections width layout

andysellick:

## What
Change how flexible content is laid out, specifically with regard to full width flexible sections.

Instead of using a full width page template and wrapping most of the flexible sections in a `govuk-width-container`, use the normal layout and apply a special class to those flexible sections that need full width, which adds a negative margin either side of it to make it break out of the container element.

Relies upon https://github.com/alphagov/govuk_publishing_components/pull/5455, because we need to append the class for the full width container query element onto the body, so any width calculation takes account of the width of the scrollbar (now included).

Tested in a variety of browsers in Windows and MacOs, plus various mobile devices.

## Why
It felt a bit weird for every flexible section to be wrapped in its own `govuk-width-container` given that we only have one that doesn't need it (impact header) - better to wrap the page and apply a fix for the exceptions.

## Visual changes
None.


KludgeKML (approved):
Looks good to me!

AshGDS (approved):
LGTM 👍

KludgeKML (approved):
LGTM

PR #5534, opened 2026-05-12
[PNP-10000] Refactor flexible pages

KludgeKML:

## What

Refactor how we create FlexiblePage subclasses (History, TopicalEvent, TopicalEventAboutPage) to make them simpler to use.

## Why

Originally we expected FlexiblePage to be a class that would encompass a lot of different types, and which would receive arbitrary lists of flexible sections from content items, so it would need to handle hashes of data. In fact that's not the case - we're defining flexible pages as subclasses ourselves. That allows us to simplify a little by adding container classes (which we don't want to be used by editors, but can be useful for making simple sets of content/sidebar composite types), and remove a bunch of test fixtures that are hard to maintain.

It also makes FlexibleSections a little more self-contained, and more in line with components, which ties in nicely with the plan to render sections in the component guide.

Jira card: https://gov-uk.atlassian.net/browse/PNP-10000

## Visual changes

None expected.


app/models/flexible_page/flexible_section/breadcrumbs.rb
```diff
@@ -2,10 +2,10 @@ module FlexiblePage::FlexibleSection
   class Breadcrumbs < Base
     attr_reader :breadcrumbs
 
-    def initialize(flexible_section_hash, content_item)
+    def initialize(breadcrumbs:)
```
andysellick, line 5: Just curious to learn what `breadcrumbs:` does in this context, how is it different to `breadcrumbs`?

app/models/flexible_page/flexible_section/impact_header.rb
```diff
@@ -1,52 +1,54 @@
...
 
-  private
-
-    def build_modifier_classes
+    def build_modifier_classes(image_type)
       base_class = "impact-header"
-      logo = flexible_section_hash["image_type"] == "logo"
-      return "#{base_class}--plain" unless logo || @variant
+      logo = image_type == :logo
+      return "#{base_class}--plain" unless logo || variant
```
andysellick, line 38: Gap in my knowledge: do these references to `variant` need to be `@variant` or is this enough? (presumably, since the tests pass)

app/views/flexible_page/flexible_sections/_impact_header.html.erb
```diff
@@ -1,5 +1,5 @@
 <% text = capture do %>
-  <% inverse = true if flexible_section.variant == "notable-death" %>
+  <% inverse = true if flexible_section.variant == :notable_death %>
```
andysellick, line 2: Not part of this PR: could we use `flexible_section.caption[:inverse]` here instead of this?

spec/models/flexible_page/flexible_section/impact_header_spec.rb
```diff
@@ -70,8 +39,9 @@
     end
 
     it "returns the correct caption information" do
-      allow(SecureRandom).to receive(:hex).and_return("1234")
-      expect(impact.caption).to eq({
+      # allow(SecureRandom).to receive(:hex).and_return("1234")
```
andysellick: Commented line?

andysellick (commented):
I've only reviewed the impact header commit, as that's the bit of the code I'm most familiar with. BE changes seem to be sensible and much cleaner, only some minor comments/questions.

app/models/flexible_page/flexible_section/breadcrumbs.rb
KludgeKML: `breadcrumbs:` makes it a named parameter rather than a positional parameter.

app/models/flexible_page/flexible_section/impact_header.rb
KludgeKML: variant is enough because we've specified variant as a reader attribute on line 3 (ie we want externally to be able to say `my_impact_header.variant`, and internally that also means we can say `variant` if we're only reading the value, we only have to say `@variant` if we're modifying it).

app/views/flexible_page/flexible_sections/_impact_header.html.erb
KludgeKML: Uh, probably! That would make more sense, to not have knowledge of specific variant types getting into the view, but just making it about the appearance.

spec/models/flexible_page/flexible_section/impact_header_spec.rb
KludgeKML: Oh yeah, I should investigate when this was added and why.

app/models/flexible_page/flexible_section/breadcrumbs.rb
andysellick: So you can call the method without worrying about the order? Like `breadcrumbs.new(something: 1, breadcrumbs: []`?

app/views/flexible_page/flexible_sections/_impact_header.html.erb
andysellick: I'll make a note and think about how to make this better 👍

spec/models/flexible_page/flexible_section/page_title_spec.rb
```diff
@@ -1,14 +1,12 @@
...
-      "lead_paragraph" => "Welcome to this page",
-    }, FlexiblePage.new({}))
-  end
+  subject(:page_title) { described_class.new(context:, heading_text:, lead_paragraph:) }
+
+  let(:context) { "My Context" }
+  let(:heading_text) { "My Heading" }
+  let(:lead_paragraph) { "Welcome to this page" }
 
   describe "#initialize" do
```
deborahchua, line 8: Should we also test the default values (of `nil`) when only `heading_text` is set?

app/models/flexible_page/flexible_section/breadcrumbs.rb
KludgeKML: Yeah, which is kind of worse here (because there's only one parameter), but it's more useful on the other flexible sections, and keeps this one consistent. In practise it's not too bad because of named parameter auto-use (ie if you've got a variable / method named `breadcrumbs` you can use that automatically by saying `Breadcrumbs.new(breadcrumbs:)`)

spec/models/flexible_page/flexible_section/impact_header_spec.rb
KludgeKML: Removed in https://github.com/alphagov/frontend/pull/5545 - I'll rebase to fix the conflict when everyone's okay with this PR.

docs/adr/0003-creating-flexible-pages.md
```diff
@@ -0,0 +1,55 @@
...
++ Keeps consistency with older content types
++ Simple for devs
+
+#### Cons
+- Requires changing the existing pages
+- Risk of drift between how the pages handle various things as new types added - as is a known problem with existing page types.
+
+### Option 3
+Create flexible sections directly in initializer of the various flexible pages,
+simplify data passed to
```
deborahchua: Is this meant to read:

> simplify data passed to it

deborahchua (approved):
Great refactor! This approach looks much cleaner 🎉 

Just a couple of small comments:

docs/adr/0003-creating-flexible-pages.md
KludgeKML: Oh yeah, there's actually more to the end of that sentence! Thanks, fixed!

spec/models/flexible_page/flexible_section/breadcrumbs_spec.rb
```diff
@@ -1,20 +1,18 @@
...
+    [
+      {
+        title: "History of the UK Government",
+        url: "/government/history",
+      },
+    ]
+  end
 
   describe "#initialize" do
     it "sets the attributes from the flexible section hash" do
```
sairamya93: I think we can update the test description too as we directly set the breadcrumb values and no longer from flexible section hash?

spec/models/flexible_page/flexible_section/page_title_spec.rb
KludgeKML: Added!

sairamya93 (approved):
Looks nice! Helpful commit descriptions too 👍 
Just added a tiny note to update test description.
Should we add `tech debt` label too? (to track the couple of commits where it was introduced)

spec/models/flexible_page/flexible_section/breadcrumbs_spec.rb
KludgeKML: Good spot!

KludgeKML:
> Looks nice! Helpful commit descriptions too 👍 Just added a tiny note to update test description. Should we add `tech debt` label too? (to track the couple of commits where it was introduced)

@sairamya93 Yes, I'll add that label. And now that it's been approved, I'll add a tech debt issue as well that points to this commit. Thanks for the reminder!

KludgeKML:
Tech debt: 
https://github.com/alphagov/frontend/issues/5552
https://github.com/alphagov/frontend/issues/5554

PR #5535, opened 2026-05-12
[NAV-18643] Update layout for Get Involved page

luke-booker:

## What

Using [this Universal Credit page](https://www.gov.uk/guidance/how-to-verify-your-identity-for-universal-credit) as an example to update the 'Get Involved' page. 

**Heading changes** 

Updated all headings on 'Get Involved' page in alignment with [Design System here](https://design-system.service.gov.uk/styles/headings/#:~:text=%3Ch1%20class%3D%22govuk%2Dheading%2Dl%22%3Egovuk%2Dheading%2Dl%3C/h1%3E%0A%3Ch2%20class%3D%22govuk%2Dheading%2Dm%22%3Egovuk%2Dheading%2Dm%3C/h2%3E%0A%3Ch3%20class%3D%22govuk%2Dheading%2Ds%22%3Egovuk%2Dheading%2Ds%3C/h3%3E)

`<h1 class="govuk-heading-l">govuk-heading-l</h1>`
`<h2 class="govuk-heading-m">govuk-heading-m</h2>`
`<h3 class="govuk-heading-s">govuk-heading-s</h3>`

**Border changes**

- Added a section break below the lead paragraph. 
- Removed all current solid black borders. 

## Why

Update the 'Get Involved' page for consistence and maintainability. 

Jira card: https://gov-uk.atlassian.net/browse/NAV-18643

## Visual changes

| Live  | Updated Layout |
| ------------- | ------------- |
| Screenshot 2026-05-13 at 15 55 15 | Screenshot 2026-05-18 at 16 27 20 |

### DIFF
--- a/app/views/get_involved/show.html.erb
+++ b/app/views/get_involved/show.html.erb
@@ -4,20 +4,17 @@
 <% end %>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-3">
     <%= render "govuk_publishing_components/components/heading", {
       context: "Government activity",
       text: t("formats.get_involved.page_heading"),
       heading_level: 1,
-      font_size: "xl",
+      font_size: "l",
       margin_bottom: 8,
     } %>
 
-    <%= render "govuk_publishing_components/components/lead_paragraph", {
-      text: t("formats.get_involved.intro_paragraph.body_html",
-      ),
-      margin_bottom: 9,
-    } %>
+    <%= render "govuk_publishing_components/components/lead_paragraph", { text: t("formats.get_involved.intro_paragraph.body_html") } %>
+    <hr class="govuk-section-break govuk-section-break--visible">
   </div>
 </div>
 <div class="govuk-grid-row">
@@ -26,7 +23,7 @@
     <%= render "govuk_publishing_components/components/heading", {
       text: t("formats.get_involved.engage_with_gov"),
       heading_level: 2,
-      font_size: "l",
+      font_size: "m",
       padding: true,
       id: "engage-with-government",
     } %>
@@ -37,8 +34,7 @@
     <%= render "govuk_publishing_components/components/heading", {
       text: t("formats.get_involved.respond_to_consultations"),
       heading_level: 2,
-      font_size: "l",
-      margin_bottom: 3,
+      font_size: "m",
     } %>
 
     <%# Big numbers %>
@@ -66,6 +62,7 @@
         <%= render "govuk_publishing_components/components/heading", {
           text: @presenter.time_until_next_closure,
           heading_level: 3,
+          font_size: "s",
           margin_bottom: 4,
         } %>
         <%= link_to @content_item.next_closing_consultation["title"], @content_item.next_closing_consultation["link"], class: "govuk-link" %>
@@ -81,7 +78,7 @@
       text: t("formats.get_involved.recently_opened"),
       padding: true,
       heading_level: 2,
-      font_size: "l",
+      font_size: "m",
       margin_bottom: 4,
     } %>
     <%= render "govuk_publishing_components/components/document_list", {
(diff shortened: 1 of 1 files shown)


MartinJJones:
The change looks good to me overall, I have spotted one potential issue with the h1 heading and caption size that we may want to look at further.

When looking at the [design system guidance for headings with captions](https://design-system.service.gov.uk/styles/headings/#headings-with-captions-typography-example-open), all of the examples show the headings with the same `caption-[size]` and `heading-[size]` t-shirt size, for example if the heading is using XL, the caption is also set to XL.

With this change, the caption size will be XL, but the heading size will be L.

This is because the [heading component currently hardcodes the caption size to XL](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_heading.html.erb#L19) and the govuk_publishing_components guide recommends using a larger font-size - https://components.publishing.service.gov.uk/component-guide/heading/with_context to ensure the heading text is always larger than the caption text.

I've checked and in this case in both mobile and desktop, the heading text is still larger than the caption text, but wonder if we should be following the design system examples to the letter and if this would improve the balance.

@deberny @jon-kirwan just wanted to get your thoughts on this as well, I wonder if this should be a non-blocking issue and we can look to review this in the future, possibly as part of https://gov-uk.atlassian.net/browse/NAV-18930? 🤔

MartinJJones (commented):
Nice work, the changes look good to me overall, I just have one question around the heading and caption size, which I've tagged Agz and Jon in on.

jon-kirwan:
> Nice work, the changes look good to me overall, I just have one question around the heading and caption size, which I've tagged Agz and Jon in on.

Interesting. I wonder if this is related to merging the 'context' option from the title component into the heading component, where `xl` was previously the only option - which perhaps made more sense for a dedicated title component but a heading component might need more flexibility.

[https://github.com/alphagov/govuk_publishing_components/pull/4510](https://github.com/alphagov/govuk_publishing_components/pull/4510)
[https://github.com/alphagov/govuk_publishing_components/commit/3dfb20e1f9e6df0604c6f19c58ab82448c555859#diff-3218a674596e8214625153780930423aa3a2ee00210af108066c7980343f2caf](https://github.com/alphagov/govuk_publishing_components/commit/3dfb20e1f9e6df0604c6f19c58ab82448c555859#diff-3218a674596e8214625153780930423aa3a2ee00210af108066c7980343f2caf)

I’d say this is probably something we should add to the backlog, or raise in the components repo, for further investigation/discussion.

deberny:
It's a good point @MartinJJones on the caption size. I agree that we should follow the Design System, and if a heading is "l" the caption should be be matching it. Not sure why we have these hardcoded, but is there a way to make it smarter?

Also I found the [Heading component guide](https://components.publishing.service.gov.uk/component-guide/heading) confusing, as it uses different terminology to DS, so I wonder if we could standardise it.

Is there a reason some of the component options for Headings come with the top border and where are they used? (as it's not explained)

MartinJJones:
> I’d say this is probably something we should add to the backlog, or raise in the components repo, for further investigation/discussion.

@jon-kirwan thanks that makes sense to me, I've created a new card relating to this and added it to the backlog - https://gov-uk.atlassian.net/browse/NAV-18941

> Also I found the [Heading component guide](https://components.publishing.service.gov.uk/component-guide/heading) confusing, as it uses different terminology to DS, so I wonder if we could standardise it.
>
> Is there a reason some of the component options for Headings come with the top border and where are they used? (as it's not explained)

@deberny good point, I'm not sure of this history of where the top border came from and where it should be used, but I've also added this as something we should look at further as part of the new card added to the backlog - [Align heading component styles with design system](https://gov-uk.atlassian.net/browse/NAV-18941)

I've approved the PR as the code changes look good to me and we can make further improvements as part of the card created and added to the backlog.

MartinJJones (approved):
Nice work the changes look good to me 👍

luke-booker:
Thank you @MartinJJones for spotting the caption and overall heading improvements we can make. Along with heading sizes, borders and spacings, it would be great to audit and standardise as templates are brought into frontend.

PR #5540, opened 2026-05-14
Remove YAML support from content loader

KludgeKML:

## What

Remove the ability for the content loader to load YAML-formatted content items from the local path (it retains the ability to load JSON content items), and reformat the one development file and two test fixtures from YAML to JSON

## Why

All content items in actual use are JSON or JSON-structured anyway, so having two file options doesn't really give us much useful flexibility (it could be argued that YAML is a bit easier to work with when editing by hand, but json<->yaml conversion is easy enough). Ultimately we want to remove the three existing YAML files anyway (the two fixtures by refactoring FlexiblePage tests, and the remaining one by making the component guide responsible for displaying flexible sections so that a developer can work with them easily), and it makes no difference in the meantime what format those files are in (we can just as easily delete three JSON files as three YAML files 😉  )


andysellick (commented):
Seems sensible, can't see anything amiss but I will let a BE dev approve.

AshGDS (commented):
Looks sensible to me as well 👍

sairamya93 (approved):
LGTM 👍

CodeSonia (approved):
Looks good to me  👍 - the only tiny thing was just a spelling error on this commit:
`Update ContentItemLoader to usr LocalFileLoader` should be use. But no biggie!

PR #5561, opened 2026-05-21
[PNP-10020] Handle probe attempts more gracefully

KludgeKML:
## What

Default to responding `:not_acceptable` to any request with a format that's not HTML (except for those routes that have explicit format handling, such as for atom/json/ics responses).

## Why

We get a lot of Sentry errors like this:
https://govuk.sentry.io/issues/5157654714/?project=202225&referrer=project-issue-stream

...which are not really actionable (because they're just attempts to probe the system which don't break into anything and can't be stopped), but:
1) we should be responding to them with an appropriate code, and
2) we shouldn't be seeing them in Sentry

(This is equivalent to how we shouldn't see NotFounds in sentry)

Jira card: https://gov-uk.atlassian.net/browse/PNP-10020

## Visual changes

None

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,7 +1,6 @@
 class ApplicationController < ActionController::Base
-  before_action do
-    I18n.locale = I18n.default_locale
-  end
+  before_action { I18n.locale = I18n.default_locale }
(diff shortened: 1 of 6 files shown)


deborahchua (approved):
🎉

PR #5578, opened 2026-06-01
[PNP-10036] Catch additional bad requests before passing them to ContentStore

KludgeKML:

## What

Improve safe_path? to add an additional check similar to the one used by content-store. We also add additional tests to make sure the backing content-store isn't called (because it might return the same InvalidURL value otherwise, making the test for that return value not enough)

## Why

At the moment, there are some classes of bad requests that we're passing straight through to Content Store. This is "safe" in that Content Store handles it safely by returning a 400 before doing any database looking up, since: https://github.com/alphagov/content-store/pull/946

However, it's not ideal because 1) we should be able to catch this closer to the perimeter, saving us a call to content-store, and 2) because we're not currently catching the error from content-store we see these as Sentry errors. We could just stop the errors appearing in Sentry by adding GdsApi::HTTPBadRequest to the list of errors to catch/return from content loader, but that still means that some additional bad traffic is going to content store that we could have avoided.

Jira card: https://gov-uk.atlassian.net/browse/PNP-10036

### DIFF
--- a/spec/unit/content_item_loader_spec.rb
+++ b/spec/unit/content_item_loader_spec.rb
@@ -71,6 +71,24 @@ RSpec.describe ContentItemLoader do
       it "returns (but does not raise) an InvalidUrl exception" do
         expect(content_item_loader.load("/../my-missing-item")).to be_a(GdsApi::InvalidUrl)
       end
+
+      it "does not call the default loader to try to load the content item" do
+        expect(content_item_loader.default_loader).not_to receive(:load)
+
+        content_item_loader.load("/../my-missing-item")
+      end
+    end
+
+    context "when the path includes header repeat tricks" do
+      it "returns (but does not raise) an InvalidUrl exception" do
+        expect(content_item_loader.load("/my-missing-item HTTP/1.1\r\nOld-Header: bb-8")).to be_a(GdsApi::InvalidUrl)
+      end
+
+      it "does not call the default loader to try to load the content item" do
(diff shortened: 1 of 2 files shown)


deborahchua (approved):

spec/unit/content_item_loader_spec.rb
```diff
@@ -71,6 +71,24 @@
       it "returns (but does not raise) an InvalidUrl exception" do
         expect(content_item_loader.load("/../my-missing-item")).to be_a(GdsApi::InvalidUrl)
       end
+
+      it "does not call the default loader to try to load the content item" do
+        expect(content_item_loader.default_loader).not_to receive(:load)
+
+        content_item_loader.load("/../my-missing-item")
```
sairamya93, line 78: Are we setting the expectation first and then loading the invalid url here?

sairamya93 (approved):
LGTM 👍 
Just had a clarification on how we order the test expectation.

spec/unit/content_item_loader_spec.rb
KludgeKML: Yeah, with these sorts of expectations RSpec needs to have set something up on the class so that it can tell whether the method got called, so the expect is first, then the trigger.

KludgeKML:
Closing this in favour of a better solution currently under discussion in https://github.com/alphagov/govuk-rfcs/pull/192

PR #5584, opened 2026-06-02
Add breadcrumbs flexible section options

andysellick:

## What
Adds options to the breadcrumbs flexible section:

- `margin_bottom` gets passed to the breadcrumbs component (see below)
- `full_width` makes the wrapper for the breadcrumb (not the breadcrumb itself) reach the edge of the screen
- `background` sets a black background on the section, and inverts the breadcrumb component so it can still be read (we might want to add more nuanced background options in the future, but right now we only need this one)

## Future work
Passing individual options to the components inside a flexible section (like we're doing here with `margin_bottom`, and likely in other places) and then documenting and testing that feels like duplication of effort. I think there's a better solution, but given that we don't have that right now and we need these options I'm submitting this as-is, and will circle around and rework this when I have a better solution.

## Why
Looking at using flexible sections to render the missions pages.

## Visual changes

Screenshot 2026-06-03 at 09 09 44

### DIFF
--- a/app/assets/stylesheets/views/_flexible-content.scss
+++ b/app/assets/stylesheets/views/_flexible-content.scss
@@ -1,5 +1,6 @@
 @import "govuk_publishing_components/base";
 @import "flexible_sections/body-with-image";
+@import "flexible_sections/breadcrumbs";
 @import "flexible_sections/impact-header";
 
 .flexible-section {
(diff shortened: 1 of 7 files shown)


KludgeKML (approved):
LGTM

PR #5588, opened 2026-06-03
Switch Impact header to use strings, not syms

andysellick:

## What / why
Change options passed to the impact header flexible section that were symbols to be strings.

- the impact header was working correctly, except in the component guide, where some of the variants with backgrounds weren't being styled properly
- the problem was that the code expects to be passed a symbol e.g. `:govuk` for the variant, but the value coming from the YML for the component guide was a string, so it was failing to match and not adding the right background
- modify all variables used by the impact header that were syms to be strings, including variant, header, logo
- also allows us to consistently use `notable-death` instead of a mix of that and `notable_death`

## Visual changes
None.


KludgeKML:
Have confirmed bridges taxon still works with this branch.

KludgeKML (approved):
LGTM

PR #5602, opened 2026-06-09
[PNP-10047] Tidy layouts

KludgeKML:

## What

We have two layouts with a lot of overlapping content. Simplify them, and add a third layout for manuals and HMRC manuals, recognising that they have enough in common and are different enough from other pages to warrant their own layout.

## Why

We'd like to simplify the number of different ways we render pages on gov.uk. A reasonable solution is to lean into layouts a bit more - recognising that we have certain pages that have a similar essential structure and moving that structure into a layout file that we can use. But our current layout files are a bit daunting, even simplified as they've been post-static. So we need to simplify them to the point where adding additional layouts is relatively simple.

Jira card: https://gov-uk.atlassian.net/browse/PNP-10047

## Visual changes

None

### DIFF
--- a/app/views/layouts/manual.html.erb
+++ b/app/views/layouts/manual.html.erb
@@ -19,6 +19,8 @@
 
     <%= render "govuk_web_banners/recruitment_banner" %>
 
+    <%= yield :before_main %>
+
     <main id="content" class="govuk-main-wrapper <%= rtl_attribute %> <%= yield :main_classes %>" <%= lang_attribute %>>
       <span id="Top"></span>
       <%= yield %>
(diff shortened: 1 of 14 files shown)


app/views/layouts/manual.html.erb
andysellick, line 1: Is there anything to be deleted now that this layout is in place? There's no equivalent template somewhere?

andysellick (approved):
Very nice change, much cleaner and simpler 👏

CodeSonia (approved):
LGTM 👌

app/views/layouts/manual.html.erb
KludgeKML: To do (in a later PR) - we can move handling of manual headers and back links into this layout.

Issue #5603, opened 2026-06-09
Postcode lookup makes two identical calls to locations api if an address fallback is required


KludgeKML:
https://github.com/alphagov/frontend/blob/7dcde1843cc756398090e529491517d878746b6a/app/models/postcode_lookup.rb#L4

A postcode lookup always makes a call to `locations_api.local_custodian_code_for_postcode` (except when there's an invalid postcode format passed), because it's in the initializer. It may also make a call to `locations_api.results_for_postcode` if it needs to provide a list of addresses to disambiguate a postcode. But behind the scenes, these are the same call to Locations API, the adapter just throwing away different code. PostcodeLookup should handle this itself, making the initial call, filtering out the 7655 records, and then returning one of two views on the data it stores. 

This might also solve: https://govuk.sentry.io/issues/7224384142/?project=202225&query=is%3Aunresolved&referrer=issue-stream, which appears to occur when a split postcode includes 7655 records

(7655 records are where the local custodian code is 7655 - this is the Ordnance Survey themselves, and cannot be translated into a local authority).

KludgeKML:
Fixed by https://github.com/alphagov/frontend/pull/5607

PR #5607, opened 2026-06-10
Update postcode lookup to do a single lookup

KludgeKML:

## What

Update PostcodeLookup model to do a single lookup from Locations API and filter it for 7655 codes at that point, then use those saved values for both the `local_custodian_code` and `addresses` methods.

## Why

- Previously this model was doing two effectively identical lookups (both calls made the same call to locations api under the hood), which was not only inefficient in the address-picker path, but also meant that the address-picker path didn't include 7655 custodian code filtering, which led to bugs when trying to pick those addresses.
- Fixes issue https://github.com/alphagov/frontend/issues/5603]

## Visual changes

None


deborahchua (approved):
LGTM!

PR #5610, opened 2026-06-10
Add header sidebar content layout

KludgeKML:

## What

Adds a Header/Sidebar/Content layout, and converts `field_of_operation` to use it.

## Why

There are a handful of pages that use this layout (a header block, a sidebar with usually a contents list in it, then a govspeak content). We can standardise these a bit by moving the gross layout into its own actual layout file, then in the views providing content for the three available "slots" (header, sidebar, content). This is a proof of concept, if accepted we would initially use it also for the TopicalEventAboutPage update, and post that we'd find other pages with the same rough layout and convert them.

This allows us to start narrowing down the number of subtly different layouts we have on gov.uk. 

We also add some visual fixes (removing nested govuk-width-containers and making the bullet points behave correctly, improving margins, removing an unnecessary section).

## Visual changes

|before|after|
|-----|-----|
|image|image|

### DIFF
--- a/app/views/field_of_operation/show.html.erb
+++ b/app/views/field_of_operation/show.html.erb
@@ -15,61 +15,63 @@
     } %>
   </div>
 
-  <div class="govuk-grid-column-one-third govuk-!-margin-bottom-8">
-    <%= render "govuk_publishing_components/components/organisation_logo", organisation: @presenter.organisation %>
+  <div class="govuk-grid-column-one-third">
+    <%= render "govuk_publishing_components/components/organisation_logo",
+      organisation: @presenter.organisation,
+      margin_bottom: 8 %>
   </div>
 <% end %>
 
 <% content_for :sidebar do %>
   <% if @presenter.contents.present? %>
-    <%= render "govuk_publishing_components/components/contents_list", {
+    <%= render "govuk_publishing_components/components/contents_list",
           contents: @presenter.contents,
-        } %>
+          margin_bottom: 8 %>
   <% end %>
 <% end %>
 
 <% content_for :content do %>
-  <section id="field-of-operation">
-    <% unless @presenter.description.blank? %>
-      <%= render "govuk_publishing_components/components/heading", {
-        text: t("formats.fatality_notice.field_of_operation"),
-        margin_bottom: 4,
-      } %>
-      <%= render "govuk_publishing_components/components/govspeak", {
-          margin_bottom: 8,
-        } do %>
-          <%= @presenter.description %>
-      <% end %>
+  <% unless @presenter.description.blank? %>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: t("formats.fatality_notice.field_of_operation"),
+      id: "field-of-operation",
+      margin_bottom: 4,
+    } %>
+    <%= render "govuk_publishing_components/components/govspeak", {
+        margin_bottom: 8,
+      } do %>
+        <%= @presenter.description %>
     <% end %>
-    <% if @content_item.fatality_notices.any? %>
-      <%= render "govuk_publishing_components/components/heading", {
-        text: "Fatalities",
-        id: "fatalities",
-        margin_bottom: 4,
-      } %>
-      <ul class="govuk-list">
-        <% @content_item.fatality_notices.each do |fatality_notice| %>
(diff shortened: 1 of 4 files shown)


app/views/layouts/header_sidebar_content.html.erb
```diff
@@ -0,0 +1,37 @@
+<%
```
deborahchua, line 1: Can we include a link to the application layout in the commit description?

KludgeKML: Done!

app/views/field_of_operation/show.html.erb
```diff
@@ -2,75 +2,74 @@
...
+      margin_bottom: 8,
+      lang: content_item.locale,
+    } %>
   </div>
 
-  <div class="govuk-grid-row">
-    <% if @presenter.contents.present? %>
-      <div class="govuk-grid-column-one-third">
-        <%= render "govuk_publishing_components/components/contents_list", {
+  <div class="govuk-grid-column-one-third govuk-!-margin-bottom-8">
```
andysellick: This margin class is bothering me (probably from the original page). I think it would be better if it was removed and both the heading component (above) and the organisation logo (below) had a `margin_bottom` option of 8 (the heading already has it).

app/views/field_of_operation/show.html.erb
```diff
@@ -2,75 +2,74 @@
...
+    <%= render "govuk_publishing_components/components/organisation_logo", organisation: @presenter.organisation %>
+  </div>
+<% end %>
+
+<% content_for :sidebar do %>
+  <% if @presenter.contents.present? %>
```
andysellick, line 26: Quick question - is the outcome of this `if` ever `false`? In that situation the whole of this grid column would collapse, and the `content:` part of the page would move to the left to fill the gap. We can fix this fairly easily with CSS but wanted to check.

app/views/field_of_operation/show.html.erb
```diff
@@ -2,75 +2,74 @@
...
+    <%= render "govuk_publishing_components/components/contents_list", {
           contents: @presenter.contents,
         } %>
-      </div>
+  <% end %>
+<% end %>
+
+<% content_for :content do %>
+  <section id="field-of-operation">
```
andysellick: Possibly out of scope for this PR. This ID is hard coded into the template, and appears to be used by the contents list (first item, jump link), but that comes from the content item. Are there ever any more items in that contents list? If not, what's the value of it? Seems a bit strange.

andysellick (commented):
Nice change, some particularly good frontend improvements as well 😁 

Just a few thoughts/questions.

app/views/field_of_operation/show.html.erb
KludgeKML: Theoretically it's possible. There are a maximum of two content items:
- Field of Operation (which is the top item if there's a description, which there isn't always)
- Fatalities (which would not be present if no fatalities)

At the moment, every page has fatalities, some don't have a description. If we added a new one with no description where no-one had died yet, it would be false. But in practice, it's always true at the moment and it feels unlikely it would ever be false.

app/views/field_of_operation/show.html.erb
KludgeKML: As above - there are two potential items - I think we could move this inside the following if, so that this section didn't appear unless the contents list link also appeared. I'll do that.

app/views/field_of_operation/show.html.erb
KludgeKML: Done.

app/views/field_of_operation/show.html.erb
KludgeKML: I've removed the section entirely and moved the id onto the heading element for Field of Operation, so it matches how the Fatalities heading/contents are handled.

app/views/field_of_operation/show.html.erb
andysellick: Thanks. Also just realised we could make it consistent (and give the page a little more breathing space on mobile) by putting the same `margin_bottom` of 8 on the contents list (it's currently 4).

KludgeKML: Done!

deborahchua (approved):
LGTM!

PR #5611, opened 2026-06-10
Force nil return from error path

KludgeKML:
- When we are about to follow too many redirects, we write a warning into Sentry (via GovukError.notify), and we're supposed to return a nil so that the calling function can detect a content item isn't present and return a 404. This works fine in tests, because in tests Sentry sending is turned off, so the notify call short circuits just before the point where it would try to send the message, and instead returns a nil. In production, it returns the Sentry::ErrorEvent that comes back from the send. To test this we make sure that GovukError.notify actually returns something non-nil, then check afterwards to confirm that it definitely was called.

https://govuk.sentry.io/issues/6215960168/?alert_type=issue&notification_uuid=b8be73be-41d5-47f8-8e2a-df08394b1563&project=202225&referrer=slack&workflow_id=3547982

### DIFF
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -89,6 +89,7 @@ private
     end
 
     Rails.logger.debug("CSV attachment details missing for '#{attachment_id}' at '#{current_path}'")
+
     GovukError.notify(
       "CSV preview - couldn't resolve attachment details",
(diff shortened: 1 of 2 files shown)


deborahchua (approved):
Nice!

CodeSonia (approved):
LGTM!

PR #5612, opened 2026-06-11
[PNP-10042] Support rendering embedded "about" sections from Topical Events

KludgeKML:
## What

Adds support for topical event about pages embedded in the topical event content item. At this point we still support separate content items from the `topical_event_about_page` schema, but once these are all migrated in Whitehall we can remove the old support.

## Why

It's been decided that separate about pages aren't really a good way to handle these pages, so to migrate the old pages (which still have them), we will embed the about page info into the topical event itself. During the republish we'll want to handle both separate and integrated content items, after that's done we can remove the old model.

Jira card: https://gov-uk.atlassian.net/browse/PNP-10042

## To Do

- [ ] If merged, add tech debt issue around search being called in topical event even when not strictly needed for rendering.

## Notes for Reviewers

There aren't any embedded pages yet, but we want to make sure that old (separate) about pages aren't affected, so compare:
- https://govuk-frontend-app-pr-5612.herokuapp.com/government/topical-events/western-balkans-summit-london-2018/about
- https://www.gov.uk/government/topical-events/western-balkans-summit-london-2018/about

But we do have a local-content-item with embedded about page you can see here:

- https://govuk-frontend-app-pr-5612.herokuapp.com/government/topical-events/sample/
- https://govuk-frontend-app-pr-5612.herokuapp.com/government/topical-events/sample/about

(Note: PR is in draft for the moment because this will be removed before merging)

### DIFF
--- a/app/views/shared/headers/_page_title.html.erb
+++ b/app/views/shared/headers/_page_title.html.erb
@@ -1,0 +1,12 @@
+<div class="govuk-grid-column-two-thirds">
+  <%= render "govuk_publishing_components/components/heading", {
+    text: options[:heading_text],
+    context: options[:context],
+    heading_level: 1,
+    font_size: "xl",
+    margin_bottom: 8,
+  } %>
+  <%= render "govuk_publishing_components/components/lead_paragraph", {
+    text: options[:lead_paragraph],
+  } %>
+</div>
--- a/app/views/topical_event/about.html.erb
+++ b/app/views/topical_event/about.html.erb
@@ -1,5 +1,7 @@
+<% content_for :title, build_page_title([content_item.about.title]) %>
+
 <% content_for :head do %>
-  <meta name="description" content="<%= content_item.description %>">
+  <meta name="description" content="<%= content_item.about.summary %>">
 
   <%= render "govuk_publishing_components/components/machine_readable_metadata",
     schema: :article,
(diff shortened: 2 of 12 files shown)


app/views/topical_event/about.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :head do %>
+  <meta name="description" content="<%= content_item.description %>">
```
AshGDS: Off topic, but I'm surprised we haven't standardised setting this meta tag across views.

AshGDS (commented):
LGTM 👍

app/views/topical_event/about.html.erb
KludgeKML: Yeah, it's on my to-do list.

app/views/topical_event/about.html.erb
```diff
@@ -1,16 +1,28 @@
...
     schema: :article,
     content_item: content_item.to_h %>
 <% end %>
 
 <% content_for :header do %>
+  <%= render "shared/headers/page_title", options: content_presenter.page_title_options %>
 <% end %>
 
 <% content_for :sidebar do %>
+  <nav>
```
andysellick: The contents list is already a `nav` element, so this extra wrapping one can be removed.

app/views/shared/headers/_page_title.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="govuk-grid-column-two-thirds">
```
andysellick, line 1: Where is the wrapping `govuk-grid-row`? I think it would be easier to follow the logic if this grid column element was in the same place.

KludgeKML: Updated as per our discussion

KludgeKML: But grid-row and grid-column remain separate for now.

AshGDS (approved):

sairamya93 (approved):
Nice 👍 
Looks good to me too though I couldn't view the sample pages in the PR description.

PR #5629, opened 2026-06-24
Update references to Sass variables to use the govuk-functional-colour function

MartinJJones:
## What

Update references to Sass variables to use the govuk-functional-colour function

## Why

The Sass variables for accessing functional colours are deprecated, and will be removed in a future breaking release of govuk-frontend

Jira card: https://gov-uk.atlassian.net/browse/NAV-18923


luke-booker (approved):
Checked out and looking good. Nice one 👍

PR #5648, opened 2026-07-07
Header sidebar content improvements

KludgeKML:

## What

Improve handling of breadcrumbs in header/sidebar/content layout.

## Why

By focusing on the two things currently rendered by this layout, we can bring in incremental improvements. At the moment both of the routes handle their breadcrumbs independently. By moving breadcrumbs into the presenter and making a standard presenter variable that layouts will always have access to, we can simplify breadcrumb handling in the individual models (they just provide a breadcrumbs method, or not if they want the default Home breadcrumb), and in the layout (it can rely on a presenter with a breadcrumb method being there to determine whether or not it shows a breadcrumb, and the layout options of bottom_margin / collapse_on_mobile are defined in the layout, close to the component, rather than in a presenter somewhere), ensuring that we don't allow a proliferation of styles on this layout.

## Visual changes

None, although we will have to review how about pages look when topical events with embedded about pages are tested coming out of Whitehall.

## Reviewer notes:

Check:
- https://govuk-frontend-app-pr-5648.herokuapp.com/government/fields-of-operation/iraq
... against its [production version](https://www.gov.uk/government/fields-of-operation/iraq).

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -3,7 +3,9 @@ class ContentItemsController < ApplicationController
   before_action :reroute_to_gone, if: -> { content_item.schema_name == "gone" }
   before_action :set_locale, if: -> { request.format.html? }
 
-  attr_reader :content_item
+  attr_reader :content_item, :content_item_presenter
+
+  helper_method :content_item_presenter
 
 private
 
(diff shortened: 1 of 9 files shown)


andysellick (commented):
All looks sensible to me, but I'm not an expert. I like the idea of moving component options closer to the component in particular.

AshGDS (commented):
LGTM as well!

sairamya93 (approved):
Looks good to me too!

deborahchua (approved):
This is a nice solution!

PR #5655, opened 2026-07-09
Standardise History Pages by not deriving them from FlexiblePages any more

KludgeKML:

## What

Replaces FlexiblePage version of History pages with standardized version (with its own controller, a presenter to contain presentational methods previously in the model, and its own view based on the header_sidebar_content layout).

## Why

Flexible Pages don't follow the model we set when migrating paths from government frontend that they should have their own controller, model, presenter with presentational methods, and view. Since we're no longer looking for the free layout flexibility we originally designed these pages for, it's better to have them work like other things.

## Visual Changes

Small improvement (increase) in spacing between sidebar image and contents list.

## Reviewer Notes:

Compare:
- https://govuk-frontend-app-pr-5655.herokuapp.com/government/history with https://www.gov.uk/government/history
- https://govuk-frontend-app-pr-5655.herokuapp.com/government/history/10-downing-street with https://www.gov.uk/government/history/10-downing-street

### DIFF
--- a/app/models/concerns/has_images.rb
+++ b/app/models/concerns/has_images.rb
@@ -1,0 +1,16 @@
+module HasImages
+  def images
+    @images ||= extract_images
+  end
+
+  def image_for(type)
+    images.find { it.type == type }
+  end
+
+private
+
+  def extract_images
+    images_info = content_store_response.dig("details", "images") || []
+    images_info.map { |info| Image.new(info) }
+  end
+end
--- a/app/presenters/history_presenter.rb
+++ b/app/presenters/history_presenter.rb
@@ -1,0 +1,10 @@
+class HistoryPresenter < ContentItemPresenter
+  def breadcrumbs
+    default_breadcrumbs + [
+      {
+        title: "History of the UK Government",
+        url: "/government/history",
+      },
+    ]
+  end
+end
--- a/app/presenters/history_presenter.rb
+++ b/app/presenters/history_presenter.rb
@@ -7,4 +7,12 @@ class HistoryPresenter < ContentItemPresenter
       },
     ]
   end
+
+  def page_title_options
+    {
+      context: "History",
+      heading_text: content_item.title,
+      lead_paragraph: content_item.lead_paragraph,
+    }
+  end
 end
--- a/app/presenters/history_presenter.rb
+++ b/app/presenters/history_presenter.rb
@@ -8,6 +8,10 @@ class HistoryPresenter < ContentItemPresenter
     ]
   end
 
+  def contents
+    ContentsOutlinePresenter.new(content_item.contents_outline).for_contents_list_component
+  end
+
   def page_title_options
     {
       context: "History",
--- a/app/views/history/show.html.erb
+++ b/app/views/history/show.html.erb
@@ -1,0 +1,21 @@
+<% content_for :title, build_page_title([content_item.title]) %>
+
+<% content_for :head do %>
+  <meta name="description" content="<%= content_item.description %>">
+
+  <%= render "govuk_publishing_components/components/machine_readable_metadata",
+    schema: :article,
+    content_item: content_item.to_h %>
+<% end %>
+
+<% content_for :header do %>
+<% end %>
+
+<% content_for :sidebar do %>
+<% end %>
+
+<% content_for :content do %>
+  <%= render "govuk_publishing_components/components/govspeak", { margin_bottom: 8 } do %>
+    <%= content_item.body.html_safe %>
+  <% end %>
+<% end %>
--- a/app/views/history/show.html.erb
+++ b/app/views/history/show.html.erb
@@ -19,6 +19,10 @@
         alt: content_item.image_for("sidebar").caption,
       } %>
     <% end %>
+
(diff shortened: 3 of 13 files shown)


app/presenters/history_presenter.rb
```diff
@@ -0,0 +1,11 @@
+class HistoryPresenter < ContentItemPresenter
+  def breadcrumbs
+    default_breadcrumbs + [
+      {
+        title: "History of the UK Government",
+        url: "/government/history",
+      },
+
```
andysellick: Nitpick: empty line.

app/models/concerns/has_images.rb
```diff
@@ -0,0 +1,14 @@
+module HasImages
+  def image_for(type)
+    images.find { it.type == type }
```
andysellick, line 7: Out of interest, what is `it` in this line?

app/models/concerns/has_images.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  def images
```
andysellick: Non-blocking: should this method come before `image_for`, since that method uses it?

andysellick: Also should any of these methods be private?

app/views/history/show.html.erb
```diff
@@ -13,6 +13,9 @@
 <% end %>
 
 <% content_for :sidebar do %>
+    <% if content_item.image_for("sidebar") %>
+      <%= image_tag(content_item.image_for("sidebar").src(key: "s960"), alt: content_item.image_for("sidebar").caption, class: "card__image govuk-!-width-full", loading: "lazy") %>
```
andysellick: Just wondering about the use of the `card__image` class here - it looks like that's a class from the `landing_page` type, which we're trying to retire. Do you need a different class here, rather than relying on that? (happy to consult)

app/views/history/show.html.erb
```diff
@@ -16,6 +16,11 @@
     <% if content_item.image_for("sidebar") %>
       <%= image_tag(content_item.image_for("sidebar").src(key: "s960"), alt: content_item.image_for("sidebar").caption, class: "card__image govuk-!-width-full", loading: "lazy") %>
     <% end %>
+
+    <%= render "govuk_publishing_components/components/contents_list", {
+      contents: content_item_presenter.contents,
+      id: "contents",
```
andysellick: I was going to ask what the `id` here is for (possibly a jump link from the bottom of the page although I don't think there is one) but I can't actually see this attribute being rendered by the component in the preview, which is a bit strange.

andysellick (commented):
Mostly minor comments, one I'm happy to discuss further about styling and one question about an option on the contents list component.

andysellick:
Also to note that visually the example pages appear exactly as they do on live 👍

app/models/concerns/has_images.rb
KludgeKML: It's like the default "thing", the variable being passed to the block. It used to be _1, which I think looked a bit gross and people didn't use much. It's equivalent to:

`images.find { |it| it.type == type }`

(or, say)

`images.find { |image| image.type == type }`

(I'm not perfectly in love with it, if you prefer I'll change it to the older more explicit style)

app/views/history/show.html.erb
KludgeKML: Oh, I can see it - it's on the nav element which is (I think) the root of the component?

KludgeKML: We can remove it if it's not being used.

app/models/concerns/has_images.rb
KludgeKML: Order doesn't really matter, we don't need things to be defined before they're used because the ruby interpreter loads things before using them. Typically I'd default to this order: initializer (if there is one), public methods in alphabetical order, private methods in alphabetical order. I think here extract_images should be private. I'll tinker. Probably images is alphabetically before image_for, so they could swap for that reason.

app/presenters/history_presenter.rb
KludgeKML: Thanks! Fixed.

app/models/concerns/has_images.rb
KludgeKML: Have rearranged and made extract_images private (I'm not 100% sure what private means to modules, but the including classes shouldn't be calling it directly, and this at least signals to developers even if it doesn't enforce it on the code)

app/views/history/show.html.erb
KludgeKML: (As discussed, replaced with figure component)

app/models/concerns/has_images.rb
andysellick: Thanks for the explanation 👍

app/views/history/show.html.erb
andysellick: Updated to use figure component.

app/views/history/show.html.erb
KludgeKML: Removed!

andysellick (approved):
🚀

Issue #5659, opened 2026-07-13
Local transactions don't filter out the 7655 custodian code


KludgeKML:
The local_transaction lookup doesn't filter 7655 custodian codes, so it incorrectly triggers the multiple authorities path, which then fails (for postcodes like LS1 5AT and S11 8NX, for instance).

KludgeKML:
Fixed by PR 5660

PR #5660, opened 2026-07-13
[PNP-10081] Filter out 7655 from local custodian codes in local transactions

KludgeKML:
## What

Filter out 7655 from local custodian codes in local transactions

## Why 

- These have just started appearing and causing problems with the single_authority method (and possibly location_not_found but that hasn't shown up yet).
- ideally we want to replace this whole file with the saner PostcodeLookup model, but that requires a complete rewrite of local_transaction controller so for now just a fix in place.
- local_custodian_codes is sometimes nil, so we need the safe navigation operator (&) in the added line.

Addresses issue: https://github.com/alphagov/frontend/issues/5659

Jira card: https://gov-uk.atlassian.net/browse/PNP-10081

## Visual changes

None


AshGDS (approved):
LGTM 👍

PR #5676, opened 2026-07-16
Use layouts for call to evidence pages

andysellick:

## What
Convert the call to evidence pages to use one of the new layouts.

This has raised some interesting things:

- the layout has a content and sidebar, but this page also has a kind of footer as well, which contains the 'Explore this topic' section. This is relevant from a layout perspective as we may need to add a `footer` section to the layout, as the current implementation in this PR changes the order of some elements on mobile (this footer section now exists, and this PR has been updated to use it)
- need an option on the metadata component to put a border along the top, as this seems to be a thing in a few places (this is done as well and included here, see https://github.com/alphagov/govuk_publishing_components/pull/5592)

Examples of call for evidence pages:

- [open call for evidence](https://www.gov.uk/government/calls-for-evidence/smart-data-multi-sector-call-for-evidence)
- [closed call for evidence](https://www.gov.uk/government/calls-for-evidence/review-of-civil-legal-aid-call-for-evidence)
- [call for evidence outcome](https://www.gov.uk/government/calls-for-evidence/building-a-future-tech-sector-that-works-for-everyone)

## Why

## Visual changes
This change (I think) brings some visual improvements to layout of these pages. See this call for evidence outcome:

Before | After
------ | -----
Screenshot 2026-08-19 at 14 41 20 | Screenshot 2026-08-19 at 14 41 30

Other than that there's some minor spacing changes around the footer, but I think within acceptable limits.

### DIFF
--- a/app/views/call_for_evidence/show.html.erb
+++ b/app/views/call_for_evidence/show.html.erb
@@ -5,86 +5,82 @@
 
 <%= render "shared/email_subscribe_unsubscribe_flash", { title: content_item.title } %>
 
-<%= render "shared/headers/page_title", options: content_item_presenter.page_title_options %>
-
-<%= render "shared/publisher_metadata", locals: {
-  from: govuk_styled_links_list(content_item_presenter.contributor_links),
-  first_published: display_date(content_item.initial_publication_date),
-  last_updated: display_date(content_item.updated),
-  see_updates_link: true,
-} %>
-
-<%= render "shared/single_page_notification_button", content_item: content_item %>
-<%= render "shared/history_notice", content_item: content_item %>
-
-<% if content_item.withdrawn? %>
-  <% withdrawn_time_tag = tag.time(display_date(content_item.withdrawn_at), datetime: content_item.withdrawn_at) %>
-
-  <%= render "govuk_publishing_components/components/notice", {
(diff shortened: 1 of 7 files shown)


app/views/call_for_evidence/show.html.erb
```diff
@@ -5,86 +5,84 @@
...
 
-      <%= render "shared/published_dates_with_notification_button" %>
+  <%= render "shared/published_dates_with_notification_button" %>
+<% end %>
+
+<% content_for :sidebar do %>
+  <%= render "govuk_publishing_components/components/contextual_sidebar", content_item: @content_item.to_h %>
+<% end %>
+
+<% content_for :footer do %>
```
AshGDS, line 82: Should the layout be determining our footer grid?

AshGDS (dismissed):
LGTM 👍

app/views/call_for_evidence/show.html.erb
andysellick: Yeah, I was going to do that in a followup but this is probably the right moment. I'll push another commit 👍

andysellick: Done.

AshGDS (approved):
LGTM 👍

PR #5683, opened 2026-07-20
Add topical_event type to list of pages

AshGDS:
## What/Why
- Add `topical_event` to the list of frontend routes in our README
- We should keep this up to date as it makes it easier to audit our website.


KludgeKML:
Should we also add to: https://github.com/alphagov/frontend/blob/main/config/govuk_examples.yml

AshGDS:
@KludgeKML ah yeah good point, will do!

deborahchua (dismissed):

AshGDS:
Thanks @KludgeKML @deborahchua - just pushed again so will need a re-review

KludgeKML (approved):
LGTM

PR #5673, opened 2026-07-16
Change service manual topic pages to use layouts

andysellick:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What
Change service manual topic pages to use standard layout and standard page header.

## Why
Continuing the work from https://github.com/alphagov/frontend/pull/5665

## Visual changes
[Service manual topic](https://www.gov.uk/service-manual/agile-delivery)

Before | After
------ | -----
Screenshot 2026-07-16 at 14 02 06 | Screenshot 2026-07-16 at 14 02 15

PR #5691, opened 2026-07-23
[PNP-10090] Use shared layout for fields_of_operation page

AshGDS:

## What / Why

- Use the reusable layout for `fields_of_operation` removing the need to define columns in the view
- The code diff is a lot easier to read if you hide whitespace in GitHub's settings
- Jira card: https://gov-uk.atlassian.net/browse/PNP-10090

## Visual changes
Should be none, compare: https://www.gov.uk/government/fields-of-operation and https://govuk-frontend-app-pr-5691.herokuapp.com/government/fields-of-operation

### DIFF
--- a/app/views/fields_of_operation/index.html.erb
+++ b/app/views/fields_of_operation/index.html.erb
@@ -1,27 +1,16 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <%= render "govuk_publishing_components/components/heading", {
-      text: content_item.title,
-      context: I18n.t("formats.fields_of_operation.context"),
-      font_size: "xl",
-      margin_bottom: 8,
-      heading_level: 1,
-      lang: content_item.locale,
-    } %>
+<% content_for :header do %>
+  <%= render "shared/headers/page_title", options: content_item_presenter.page_title_options %>
+<% end %>
 
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <%
-          list_items = content_item.fields_of_operation.map do |content_item|
-            sanitize("<a href='#{content_item.base_path}' class='govuk-link'>#{content_item.title}</a>")
-          end
-        %>
+<% content_for :content do %>
+  <%
+    list_items = content_item.fields_of_operation.map do |content_item|
+      sanitize("<a href='#{content_item.base_path}' class='govuk-link'>#{content_item.title}</a>")
+    end
+  %>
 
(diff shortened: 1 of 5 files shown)


app/views/fields_of_operation/index.html.erb
```diff
@@ -1,27 +1,21 @@
...
-    <%= render "govuk_publishing_components/components/heading", {
-      text: content_item.title,
-      context: I18n.t("formats.fields_of_operation.context"),
-      font_size: "xl",
-      margin_bottom: 8,
-      heading_level: 1,
-      lang: content_item.locale,
-    } %>
+<% content_for :content do %>
+  <%= render "govuk_publishing_components/components/heading", {
```
andysellick: I think you could use the new shared `page-title` for this heading?

https://github.com/alphagov/frontend/blob/main/app/views/service_manual/service_manual_guide.html.erb#L9

andysellick: See also: https://github.com/alphagov/frontend/pull/5673/changes/a84620fb223e604d5e65a41add9d783946328d0d

AshGDS: Good spot, will update - thanks!

AshGDS: Done, but I had to merge the context into the render as it isn't accessed via the content item. @deborahchua @KludgeKML is there a better way to handle that?
https://github.com/alphagov/frontend/blob/01d716c03a08a1511483d98fa129473bb60b277d/app/views/fields_of_operation/index.html.erb#L1-L4

AshGDS: @andysellick A similar pattern to the above was merged in https://github.com/alphagov/frontend/pull/5698 so we can probably go ahead with it here

KludgeKML: The canonical way to do this is to make a FieldsOfOperationPresenter that inherits from ContentItemPresenter and overrides the page_title_options method (as per https://github.com/alphagov/frontend/blob/main/app/presenters/history_presenter.rb#L15-L21, say - although you could also do super.merge! in that method if all the other options are default).

AshGDS: Thanks Keith - i'll do that instead

andysellick: Looks good @AshGDS but will wait for your next change before approving 👍

AshGDS:
Thanks @andysellick - should be ready for another review now

andysellick (approved):
Nice, looks good.

PR #5694, opened 2026-07-24
Remove `px` values in font-size styles

MartinJJones:
## What

Remove px values used in font-size styles

## Why

The font-size rules are duplicated to create a CSS fallback for older ([Grade X](https://frontend.design-system.service.gov.uk/browser-support/#grade-x)) browsers that do not support rem values, IE6-10 for example.

The CSS callback approach can safely be removed as rem values are now widely supported - https://caniuse.com/rem

Jira card: https://gov-uk.atlassian.net/browse/NAV-18930

## Visual Changes

No visual changes


jon-kirwan (approved):
Nice. All looks good to me! 👍

PR #5698, opened 2026-07-27
Move help_page document type to shared layout

AshGDS:

## What/Why
- Make `help_page` document types use the shared layout
- There was a partial that was only ever used for this document type, so I've removed that as well.
- This page type doesn't render its `lead_paragraph`, so I've had to override that when rendering `page_title` - maybe this shouldn't be done using the view?
- I also noticed an empty column being rendered where the translations would be, so I've tried to fix that. (Is there a place I can add a test for this?)
- Jira card: https://gov-uk.atlassian.net/browse/PNP-10093

## Visual changes

Here are some sample pages:

- [GOV.UK](https://www.gov.uk/help/accessibility-statement) vs. [Heroku](https://govuk-frontend-app-pr-5698.herokuapp.com/help/accessibility-statement)
- [GOV.UK](https://www.gov.uk/help/about-govuk) vs. [Heroku](https://govuk-frontend-app-pr-5698.herokuapp.com/help/about-govuk)
- [GOV.UK](https://www.gov.uk/help/terms-conditions) vs. [Heroku](https://govuk-frontend-app-pr-5698.herokuapp.com/help/terms-conditions)

There's some minor spacing changes, the new pages are a bit shorter now.


app/views/help_page/show.html.erb
```diff
@@ -4,6 +4,32 @@
...
 <% end %>
 
-<%= render "shared/body_with_related_links" %>
+<% content_for :content do %>
+
+  <%= render "govuk_publishing_components/components/govspeak", {
+    direction: page_text_direction,
+    disable_youtube_expansions: true,
+  } do %>
+    <%= raw(@content_item.body) %>
```
KludgeKML: I don't think this is what brakeman is complaining about, but you don't need the @ in front of content_item (here or lower down)

AshGDS: Thanks @KludgeKML - done 👍

KludgeKML (approved):
LGTM

PR #5708, opened 2026-07-30
[PNP-10078] Allow `utm_campaign` parameter passthrough in service start buttons

KludgeKML:

## What

Adds passthrough of `utm_campaign` parameter to `transaction` content item rendering. These pages are start pages, so calling a start page with (eg) `https://www.gov.uk/my-service-start-page?utm_campaign=test` will result in `utm_campaign=test` being safely appended as a query parameter to the service start button link.

## Why

Allows aggregate tracking of service referrals from other services (currently impossible because all service start pages have to be on gov.uk) (for more details see [attached ADR](https://github.com/alphagov/frontend/blob/74a4fa474cc7ff11ab2cd41ab8e7daa5b257af42/docs/adr/0005-utm_campaign-passthrough.md))

Jira card: https://gov-uk.atlassian.net/browse/PNP-10078

## Visual changes

None

### DIFF
--- a/app/views/transaction/show.html.erb
+++ b/app/views/transaction/show.html.erb
@@ -34,7 +34,7 @@
       <%= render "govuk_publishing_components/components/button",
                   text: @transaction_presenter.start_button_text.html_safe,
                   rel: "external",
-                  href: content_item.transaction_start_link,
+                  href: add_parameter_if_present(content_item.transaction_start_link, "utm_campaign"),
                   start: true,
                   info_text: info_text,
                   margin_bottom: true %>
--- a/app/helpers/url_helper.rb
+++ b/app/helpers/url_helper.rb
@@ -1,4 +1,13 @@
 module UrlHelper
+  def add_parameter_if_present(url, parameter)
+    return url if params[parameter].nil?
+
+    parsed_url = URI.parse(url)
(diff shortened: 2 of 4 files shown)


deborahchua (approved):
Looks good! 👍

app/views/transaction/show.html.erb
```diff
@@ -34,7 +34,7 @@
       <%= render "govuk_publishing_components/components/button",
                   text: @transaction_presenter.start_button_text.html_safe,
                   rel: "external",
-                  href: content_item.transaction_start_link,
+                  href: add_parameter_if_present(content_item.transaction_start_link, "utm_campaign"),
```
AgaDufrat, line 37: Do we need to restrict this to users who consented to tracking?

KludgeKML: No and Yes. Basically, this isn't for gov.uk to use (it's for the receiving service), so we don't need to restrict it because we're not actually using it. The receiving service has to ignore it if the person hasn't consented to aggregate-level tracking on that service (but that's out of scope of our application)

KludgeKML:
Problem: We filter out utm_* parameters at the fastly level, so the backend doesn't get to see them. We will have to look into a frontend solution.

KludgeKML:
Closed in favour of https://github.com/alphagov/frontend/pull/5728

PR #5713, opened 2026-07-31
Use shared layout for help index page

AshGDS:

## What/Why
- Moves the `help_index` to use our shared layout
- The diff is easier to read when you hide whitespace in GitHub's settings
- Compare [GOV.UK](https://www.gov.uk/help) and [Heroku](https://govuk-frontend-app-pr-5713.herokuapp.com/help)
- Jira card: https://gov-uk.atlassian.net/browse/PNP-10096

## Visual changes

The shared layout modifies the spacing slightly.

### DIFF
--- a/app/views/help/index.html.erb
+++ b/app/views/help/index.html.erb
@@ -3,103 +3,97 @@
   <meta name="description" content="<%= t("help.index.find_out") %>">
 <% end %>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <%= render "govuk_publishing_components/components/heading", {
-      text: "Help using GOV.UK",
-      heading_level: 1,
-      font_size: "xl",
-      margin_bottom: 8,
-    } %>
-    <p class="govuk-body govuk-!-margin-bottom-8"><%= t("help.index.find_out") %></p>
-    <%= render "govuk_publishing_components/components/document_list", {
-      equal_item_spacing: true,
-      items: [
-        {
-          link: {
-            text: t("help.index.about"),
-            path: "/help/about-govuk",
-            description: t("help.index.about_description"),
-            full_size_description: true,
-          },
-        },
-        {
-          link: {
-            text: t("help.index.accessibility"),
-            path: "/help/accessibility",
-            description: t("help.index.accessibility_description"),
-            full_size_description: true,
-          },
-        },
-        {
-          link: {
-            text: t("help.index.beta"),
-            path: "/help/beta",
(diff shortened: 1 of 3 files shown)


app/views/help/index.html.erb
```diff
@@ -3,103 +3,99 @@
...
-            path: "/help/accessibility",
-            description: t("help.index.accessibility_description"),
-            full_size_description: true,
-          },
+<% content_for :header do %>
+  <%= render "shared/headers/page_title", options: content_item_presenter.page_title_options.merge({ lead_paragraph: nil }) %>
+<% end %>
+
+<% content_for :content do %>
+  <p class="govuk-body govuk-!-margin-bottom-8"><%= t("help.index.find_out") %></p>
```
andysellick, line 11: Would it be sensible to pass this to the `page_title` partial as the `lead_paragraph` option? It's a visual change, but I think an improvement.

spec/system/help_spec.rb
```diff
@@ -6,6 +6,7 @@
         format: "special_route",
         title: "Help using GOV.UK",
         description: "",
+        links: {},
```
andysellick, line 9: Just interested why the addition of `links` is needed now?

app/views/help/index.html.erb
AshGDS: @andysellick Good idea - it does make the Related Content border slightly misaligned with the content's borders. This is due to the document_list having `equal_item_spacing` which gives every item a 15px margin on the top and bottom. Without equal item spacing the border aligns. Should we remove the equal item spacing?

## With equal item spacing

image

## Without equal item spacing

image

spec/system/help_spec.rb
AshGDS: It's because the `content_item.page_title_options` calls `content_item.available_translations`, which expects `links` to exist on a `content_item`. The fake content item in the test didn't have `links` so it would crash. https://github.com/alphagov/frontend/blob/995d6d588813ab7d313e552c5832c2ab0ef0f0bd/app/models/content_item.rb#L57-L61

app/views/help/index.html.erb
andysellick: Yep, sounds good 👍

AshGDS: Done!

AshGDS:
Thanks @andysellick - should be ready for another review 👍 

https://govuk-frontend-app-pr-5713.herokuapp.com/help

andysellick (approved):
New layout looks much better 👍

PR #5724, opened 2026-08-05
Use layouts for fatality notice pages

andysellick:

## What
Convert the fatality notice pages ([example](https://www.gov.uk/government/fatalities/corporal-lee-churcher-dies-in-iraq)) to use one of the new layouts.

## Why

## Visual changes
Couple of things:

- the metadata component now has its own border top, and things have moved around fractionally, but not enough to be concerning
- the published date at the bottom of the page has moved up a little bit on mobile, but without changing the position of the elements around it. We're planning to remove this in future anyway

### DIFF
--- a/app/views/fatality_notice/show.html.erb
+++ b/app/views/fatality_notice/show.html.erb
@@ -5,58 +5,61 @@
   <meta name="description" content="<%= strip_tags(content_item.description) %>">
 <% end %>
 
-<%= render "shared/headers/page_title", options: content_item_presenter.page_title_options %>
-
-<%= render "shared/publisher_metadata", locals: {
-  from: govuk_styled_links_list(content_item_presenter.contributor_links),
-  first_published: display_date(content_item.initial_publication_date),
-  last_updated: display_date(content_item.updated),
-  } %>
-
-<% if content_item.withdrawn? %>
-  <% withdrawn_time_tag = tag.time(display_date(content_item.withdrawn_at), datetime: content_item.withdrawn_at) %>
-  <%= render "govuk_publishing_components/components/notice", {
-    title: t("withdrawn_notice.title", schema_name: t("formats.#{content_item.schema_name}.name", count: 1, locale: :en).downcase, withdrawn_time: withdrawn_time_tag).html_safe,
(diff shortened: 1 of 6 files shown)


app/views/fatality_notice/show.html.erb
```diff
@@ -5,58 +5,55 @@
...
-        history: formatted_history(content_item.history),
-      } %>
-    </div>
-  </div>
+<% content_for :sidebar do %>
   <%= render "shared/sidebar_navigation" %>
-</div>
-
-<%= render "shared/footer_navigation" %>
+<% end %>
```
AshGDS: I don't think you need to render the sidebar, it results in the page rendering empty div elements - if you remove these lines it should render OK still.

AshGDS (commented):
Looks good to me, just one small change needed

app/views/fatality_notice/show.html.erb
andysellick: Are you certain the sidebar never renders anything on this page type? I'll happily take it out, but I deliberately left it in as it was already there, and I didn't want to break anything.

AshGDS: I think Rails is smart enough to just not render anything, I've done what I've suggested here:

https://github.com/alphagov/frontend/blob/main/app/views/fields_of_operation/index.html.erb#L1-L16

https://www.gov.uk/government/fields-of-operation

AshGDS: Ah sorry I misinterpreted what you meant, I can check what you asked.

AshGDS: @andysellick I think you should be OK as that partial renders the contextual sidebar component, which in turn  only renders Related content (unless it's a step by step/ukraine cta page).

There's no "Related content" appearing in Google search for these pages https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fwww.gov.uk%2Fgovernment%2Ffatalities+%22Related+content%22

andysellick: Thanks @AshGDS have updated. I've also added another commit to restore the border on the metadata component, which I'd overlooked.

AshGDS (approved):
LGTM 👍

PR #5739, opened 2026-08-18
Use layout for case study pages

andysellick:

## What
Convert the case study pages to use layouts.

Examples:

- [normal](https://www.gov.uk/government/case-studies/aiding-capability-decision-making-for-the-royal-navy)
- [with sidebar](https://www.gov.uk/government/case-studies/the-rana-plaza-disaster)
- [with footer and published dates](https://www.gov.uk/government/case-studies/tax-avoidance-dont-get-caught-out) (including because I spotted a lack of spacing late on with this component, fixed now)

## Why

## Visual changes
There are some very minor spacing changes around the metadata at the top of the page, and the border along the top of the metadata now extends to only 2/3rds, not full width. Otherwise should be broadly the same.

Before | After
------- | ------
Screenshot 2026-09-04 at 10 28 51 | Screenshot 2026-09-04 at 10 28 59

### DIFF
--- a/app/controllers/case_study_controller.rb
+++ b/app/controllers/case_study_controller.rb
@@ -1,5 +1,7 @@
 class CaseStudyController < ContentItemsController
+  layout "header_content_sidebar"
+
   def show
-    @content_item_presenter = ContentItemPresenter.new(content_item)
+    @content_item_presenter = CaseStudyPresenter.new(content_item)
   end
 end
--- a/app/presenters/case_study_presenter.rb
+++ b/app/presenters/case_study_presenter.rb
@@ -1,0 +1,19 @@
+class CaseStudyPresenter < ContentItemPresenter
+  include LinkHelper
+  include DateHelper
+
+  def use_contextual_components?
+    true
+  end
+
+  def page_title_options
+    super.merge({
+      metadata: {
+        from: govuk_styled_links_list(contributor_links),
+        first_published: display_date(content_item.initial_publication_date),
+        last_updated: display_date(content_item.updated),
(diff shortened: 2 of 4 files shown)


AshGDS (approved):
LGTM 👍 Maybe a bit of a nitpick so non-blocking but I think the spacing of the metadata on https://www.gov.uk/government/case-studies/tax-avoidance-dont-get-caught-out looks a bit better on the original GOVUK page, is it possible to adjust the spacing on the updated page? Could it also be part of the `content` block so that the sidebar sits on the same level as it? I'm guessing that's not possible as it's in the shared header now?

andysellick:
@AshGDS thanks for the review. Can you be more specific about the spacing issue? You might be referring to the space between the top of the first metadata item and the border along the top, but I'm not certain (if it is that, I agree, but that border is now internal to the component so the change would be separate from this).

Here's the two pages I'm seeing for comparison. The layout before/after is broadly the same:

Live | PR
---- | -----
Screenshot 2026-09-07 at 08 26 57 | Screenshot 2026-09-07 at 08 27 04

And yes, the metadata is now part of the `page-title` partial, so we can't move the Related content section up. We could look at adding that in a future iteration, but it would mean it would come before the page content on mobile, which I don't think would be desirable.

AshGDS:
@andysellick No worries - yeah I was referring to the spacing between `From:` and the grey border, but looking at it again it's not that bad so don't worry about it 👍

andysellick:
@AshGDS thanks for the clarification. I think I agree with you though - I might iterate it in the component. Thanks.

PR #5745, opened 2026-08-20
[PNP-10131] Add extra routes to be excluded from search engines

AshGDS:

## What/Why

- Adds extra routes to be excluded from search engines, as requested in https://gov-uk.atlassian.net/browse/PNP-10131
- It will automatically exclude the child html publication pages from search engines too - fortunately that was accounted for when this code was initially introduced into the repo.
- I had to modify the code so that it worked with translated pages. Translated pages have a different base path, so the simple base path check doesn't work. For translated pages, I removed the `.cy` part from the base path before checking if it's in the array. Let me know if you think of a more efficient way to do this.
  - I think ideally we could change this code to use the `content_id` instead, as that stays the same across translations - but we're currently adding in pages that aren't yet published, so we can't be sure the `content_id` will stay the same.
- Test pages are in the JIRA ticket - look for the `noindex` meta tag. The last test page isn't on integration currently as it was created today so hasn't been synced to integration. 
  - If you can't see the meta tag anymore you may need to redeploy this branch to Integration.

## Visual changes

None.

### DIFF
--- a/app/presenters/publication_presenter.rb
+++ b/app/presenters/publication_presenter.rb
@@ -5,6 +5,10 @@ class PublicationPresenter < ContentItemPresenter
     /government/publications/govuk-app-testing-privacy-notice-how-we-use-your-data
     /government/publications/govuk-test-app-privacy-notice
     /government/publications/pension-credit-claim-form--2
+    /government/publications/hpv-self-testing-kit-instructions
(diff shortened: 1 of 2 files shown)


andysellick (commented):
Looks sensible to me (and I haven't thought of a better alternative yet). Best to wait for @deborahchua to confirm all okay though.

deborahchua (approved):
This looks good to me 👍

PR #5746, opened 2026-08-20
Use layouts for worldwide organisation pages

andysellick:

## What

Example worldwide organisation pages:

- [translation nav and news](https://www.gov.uk/world/organisations/british-embassy-dakar)
- [news but no translation nav](https://www.gov.uk/world/organisations/high-commission-victoria)
- [right to left](https://www.gov.uk/world/organisations/british-embassy-tehran.fa)

This change means these pages aren't using the `header` partial in `worldwide_organisation` anymore, but this is relied upon by other pages. We might be able to remove it after this and https://github.com/alphagov/frontend/pull/5749 are merged.

There's also probably some CSS for the thing that the metadata component is being used for that can be removed after these two PRs are merged.

## Why

## Visual changes
Possibly the most important part of this PR is the final commit, where I've refactored the header partial to use CSS grid instead of the GOVUK grid, in an attempt to fix the problem of varying height content for different things (lead para, metadata, translation nav) and the layout of elements on desktop compared with the ordering of the same elements on mobile. I think this is a good solution, but it probably needs further testing and interation.

This PR also fixes some obvious layout and spacing problems in the existing pages, particularly where some components are vertically too close together.

Embassy page on desktop. Note the increased spacing between elements, increased H1 size, and better alignment of elements overall.

Before | After
------ | -----
Screenshot 2026-09-01 at 14 29 21 | Screenshot 2026-09-01 at 14 29 31

Same page on mobile. Note the increased spacing. I've also switched the news/part of section to use the metadata component, which looks very similar but has bold links. I think that's okay as it's a bit more consistent (with the other links in the sidebar on desktop).

Before | After
------ | -----
Screenshot 2026-09-01 at 14 33 51 | Screenshot 2026-09-01 at 14 33 59

Bonus improvement on https://www.gov.uk/government/publications/breast-screening-helping-women-decide

Before | After
------ | -----
Screenshot 2026-09-01 at 15 15 27 | Screenshot 2026-09-01 at 15 15 34

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -112,3 +112,40 @@
     text-align: right;
   }
 }
+
+@include govuk-media-query($from: tablet) {
+  .page-title {
+    display: grid;
+    grid-template-columns: [content] 66% [sidebar] auto;
+    column-gap: 30px;
+  }
+
+  .page-title__org,
+  .page-title__heading,
+  .page-title__lead {
+    grid-column: content;
+  }
+
+  .page-title__org {
+    grid-row: 1;
+  }
+
+  .page-title__heading {
+    grid-row: 2;
+  }
+
+  .page-title__lead {
+    grid-row: 3;
+  }
+
+  .page-title__translations {
+    grid-column: sidebar;
+    grid-row: 2/4;
+  }
+
+  .direction-rtl {
+   .page-title {
+     grid-template-columns: [sidebar] auto [content] 66%;
+   }
+  }
+}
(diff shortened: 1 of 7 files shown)


AshGDS (dismissed):
Nice, LGTM 👍

andysellick:
Thanks @AshGDS. I've just pushed another commit as I spotted a problem with the new grid layout. Basically grid can't exactly replicate the GOVUK grid (see commit message) but I've found a middle ground option that seems to be pretty close. Works fine on desktop, just a few pixels out just before it drops to mobile. Take a look and let me know what you think.

Some screenshots (annotated to compared the CSS grid with the GOVUK grid beneath it):

Desktop | Just before mobile
------- | --------
Screenshot 2026-09-01 at 15 37 47 | Screenshot 2026-09-01 at 15 37 58

AshGDS (approved):
Good spot 👍

PR #5749, opened 2026-08-21
Use shared layout for worldwide_corporate_information_about_page views

AshGDS:

## What/ Why
- Use the shared layout for `worldwide_corporate_information_about_page` such as https://www.gov.uk/world/organisations/british-embassy-dakar/about/recruitment and https://www.gov.uk/world/organisations/british-embassy-dakar/about/complaints-procedure
- PR links: https://govuk-frontend-app-pr-5749.herokuapp.com/world/organisations/british-embassy-dakar/about/recruitment and https://govuk-frontend-app-pr-5749.herokuapp.com/world/organisations/british-embassy-dakar/about/complaints-procedure 
- I've moved the Contents List above the content. Without this, the page looked a bit weird as the Contents and Header would be left aligned, but the content would be further to the right. Therefore stacking it looked a bit nicer in this instance.
- Jira card: https://gov-uk.atlassian.net/browse/PNP-10129

## Visual changes
| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/app/controllers/worldwide_corporate_information_page_controller.rb
+++ b/app/controllers/worldwide_corporate_information_page_controller.rb
@@ -3,5 +3,6 @@ class WorldwideCorporateInformationPageController < ContentItemsController
 
   def show
     @presenter = WorldwideCorporateInformationPagePresenter.new(content_item)
+    render layout: "header_content_sidebar"
   end
 end
--- a/app/controllers/worldwide_corporate_information_page_controller.rb
+++ b/app/controllers/worldwide_corporate_information_page_controller.rb
@@ -2,6 +2,7 @@ class WorldwideCorporateInformationPageController < ContentItemsController
   include Cacheable
 
   def show
-    @presenter = WorldwideCorporateInformationPagePresenter.new(content_item)
+    @content_item_presenter = WorldwideCorporateInformationPagePresenter.new(content_item)
+    render layout: "header_content_sidebar"
   end
 end
--- a/app/views/worldwide_corporate_information_page/show.html.erb
+++ b/app/views/worldwide_corporate_information_page/show.html.erb
@@ -1,27 +1,24 @@
-<%= render partial: "worldwide_organisation/header", locals: {
-  worldwide_organisation: @presenter.worldwide_organisation,
-  show_header_title: false,
-} %>
 
-<article class="govuk-grid-row">
-  <div class="govuk-grid-column-one-third">
+
+<% content_for :header do %>
+  <%= render partial: "worldwide_organisation/header", locals: {
+    worldwide_organisation: @presenter.worldwide_organisation,
+    show_header_title: false,
+  } %>
+  <%= render "shared/headers/page_title", options: @presenter.page_title_options %>
+<% end %>
+
+<% content_for :content do %>
+  <aside>
     <%= render "govuk_publishing_components/components/contents_list",
-               contents: @presenter.headers_for_contents_list_component,
-               underline_links: true %>
-  </div>
-
-  <div class="govuk-grid-column-two-thirds">
-    <%= render "govuk_publishing_components/components/heading", {
-      text: @content_item.title,
-      heading_level: 2,
-      font_size: "xl",
-      margin_bottom: 4,
-    } %>
-
+            contents: @presenter.headers_for_contents_list_component,
+            underline_links: true %>
+  </aside>
+  <article>
     <p class="govuk-body-l"><%= @content_item.description %></p>
 
     <%= render "govuk_publishing_components/components/govspeak", {} do
       raw(@content_item.body)
     end %>
-  </div>
-</article>
+  </article>
+<% end %>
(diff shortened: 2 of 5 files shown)


app/views/worldwide_corporate_information_page/show.html.erb
andysellick, line 1: The lead paragraph is duplicated for some reason?

Image

andysellick: I think it's because the template is rendering both the worldwide organisation header partial and the page title partial, which probably both have the lead paragraph in them.

If you can wait for https://github.com/alphagov/frontend/pull/5746 I've added the organisation logo into page title there (or you can copy what I've done if you like).

AshGDS: Good spot, yeah I waited for your review as I suspected you'd be doing some good work to standardise these headers - Happy to wait 👍

AshGDS: Thanks, I've removed the `lead_paragraph` from the page title options, since it looks like it's rendered below in the content for this page (unless you think it's better above the contents list?)

andysellick: That's a good question. I think we have to think carefully about the order that things appear on the page. I think there are two options:

```
Heading 1
Contents list
Lead paragraph
```
- Pros: users find the contents first without having to read the lead paragraph.
- Cons: not consistent with other pages that use the `page-title` partial (although they're not necessarily right)

```
Heading 1
Lead paragraph
Contents list
```
- Pros: consistent with other pages
- Cons: contents list is less prominent

I'm genuinely uncertain of the best option here, although I'm leaning towards the second only because it makes it consistent with other pages, and allows us to iterate everything all at once in the future. It might be that contents list also belongs in the `page-title` partial, but we don't have to decide right now.

What do you think?

app/controllers/worldwide_corporate_information_page_controller.rb
```diff
@@ -3,5 +3,6 @@ class WorldwideCorporateInformationPageController < ContentItemsController
 
   def show
     @presenter = WorldwideCorporateInformationPagePresenter.new(content_item)
```
andysellick: I think we were going to use `@content_item_presenter` for consistency across page types?

app/views/worldwide_corporate_information_page/show.html.erb
```diff
@@ -1,27 +1,18 @@
...
+  <article>
+    <p class="govuk-body-l"><%= content_item.description %></p>
 
     <%= render "govuk_publishing_components/components/govspeak", {} do
-      raw(@content_item.body)
+      raw(content_item.body)
     end %>
-  </div>
-</article>
+  </article>
```
andysellick, line 13: Non-blocking: I note that this page already used the `article` tag. I think this is the first time I've noticed it in a page during this piece of work, so we should maybe have a broader conversation about the use of `article` in page types, as if it's appropriate we should use it everywhere to be consistent.

app/views/worldwide_corporate_information_page/show.html.erb
```diff
@@ -1,27 +1,18 @@
...
-  show_header_title: false,
-} %>
+<% content_for :header do %>
+  <%= render "shared/headers/page_title", options: @presenter.page_title_options %>
+<% end %>
 
-<article class="govuk-grid-row">
-  <div class="govuk-grid-column-one-third">
+<% content_for :content do %>
+  <aside>
```
andysellick: I see you've wrapped the contents list in an `aside`. It's worth noting that the contents list is itself a `nav` element, so this is now a `nav` inside an `aside`. I'm not sure `aside` is appropriate here, given that the spec says it should be used for content that `represents a portion of a document whose content is only indirectly related to the document's main content` - and the contents list does relate to the main content?

app/views/worldwide_organisation/_header.html.erb
```diff
@@ -4,6 +4,7 @@
         organisation: worldwide_organisation.organisation_logo,
         inline: true,
         heading_level: 1,
+        margin_bottom: 4,
```
andysellick: Why has margin been added here - it looks like this partial isn't now being used by this page type?

app/controllers/worldwide_corporate_information_page_controller.rb
AshGDS: Thanks, have changed this 👍

app/views/worldwide_corporate_information_page/show.html.erb
AshGDS: Good spot, not sure why I added that. Removed now!

app/views/worldwide_organisation/_header.html.erb
AshGDS: Thanks, was likely from when I was trying to get the old page to look correct with the layout.

app/views/worldwide_corporate_information_page/show.html.erb
AshGDS: @andysellick Thanks, I've moved it to the second option so it's consistent.

AshGDS:
@andysellick Should be ready for another review 👍

andysellick (approved):
Looks good. I think this is the last use of some of the worldwide partials - want to pair on cleaning them up?

AshGDS:
Thanks @andysellick, sounds good 👍

PR #5760, opened 2026-09-01
Use layout for worldwide office pages

andysellick:

## What
Convert worldwide office pages to use layouts, and the shared `page-title` partial for the header.

Example pages:

- [with a long contents list](https://www.gov.uk/world/organisations/british-embassy-washington/office/british-embassy-washington)
- [with no contents list](https://www.gov.uk/world/organisations/british-embassy-amman/office/visa-services) (left hand column collapses)
- [with almost no detail](https://www.gov.uk/world/organisations/british-embassy-stockholm/office/visa-enquiries)

Interestingly on these pages the main text title is not the page H1 element. The H1 is actually the organisation logo, and what looks like the H1 is a H2. This has required some changes to the `page-title` partial. The reason for this seems to be to avoid duplication, particularly for screen reader users. The organisation logo is read as (e.g.) `British High Commission Victoria` and the H2 is read as `About British High Commission Victoria` (where `About` is visually hidden text inside the H2).

## Why
Part of the work to move pages to use layouts and a shared header.

## Visual changes
The major visual change is to remove the left hand one third column (containing only the contents list) and move it into the main 2/3rds column. This is for several reasons:

- the contents list is (for most of these pages) only a few items, which seems like a waste of a whole column
- its better for users who depend on screen magnifiers, so they don't have to move across the screen so much to find/read the content
- there are several examples of these pages (see list above) where there's no contents list, so the content collapses across to the left of the page anyway

From https://www.gov.uk/world/organisations/high-commission-victoria/office/british-high-commission-victoria on desktop:

Before | After
------ | -----
Screenshot 2026-09-03 at 10 02 05 | Screenshot 2026-09-03 at 10 02 18

Same page on mobile. This fixes some spacing issues and reorders the contents list to come after the page title.

Before | After
------ | -----
Screenshot 2026-09-03 at 10 07 45 | Screenshot 2026-09-03 at 10 07 55

### DIFF
--- a/app/controllers/worldwide_office_controller.rb
+++ b/app/controllers/worldwide_office_controller.rb
@@ -1,7 +1,9 @@
 class WorldwideOfficeController < ContentItemsController
   include Cacheable
 
+  layout "header_content_sidebar"
+
   def show
-    @presenter = WorldwideOfficePresenter.new(content_item)
+    @content_item_presenter = WorldwideOfficePresenter.new(content_item)
   end
 end
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -32,7 +32,9 @@ class ContentItemPresenter
 
   def page_title_options
     {
+      organisation_logo_heading_level: nil,
       heading_text: content_item.title,
+      heading_level: 1,
       context: content_item.context,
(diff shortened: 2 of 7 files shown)


AshGDS (approved):
LGTM 👍

KludgeKML (approved):
LGTM

PR #5767, opened 2026-09-03
[PNP-10142] replace standardx with standard for ES linting

KludgeKML:

## What

Replaces standardx linter with standard, and reinstates no-var and fixes those lints detected by it.

## Why

StandardX seems to be abandonware relative to standard (standard's last update was a year ago compared to 3 years for StandardX), and it's increasingly the source of hard-to-fix dependency alerts. 

Jira card: https://gov-uk.atlassian.net/browse/PNP-10142

## Visual changes

None

## Manual Checklist

- [x] Maps: when a selection changes
- [x] Maps: When popups are created
- [x] HideOtherLinks: Creation of elements and show links
- [x] Sticky elements: when they move or change visibility (do they hide? TIL)
- [x] Webchat: Advisor state get
- [x] Travel advice: filter in general.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -379,21 +379,70 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@eslint/eslintrc@npm:^0.3.0":
-  version: 0.3.0
-  resolution: "@eslint/eslintrc@npm:0.3.0"
+"@eslint-community/eslint-utils@npm:^4.2.0":
+  version: 4.10.1
+  resolution: "@eslint-community/eslint-utils@npm:4.10.1"
+  dependencies:
+    eslint-visitor-keys: ^3.4.3
+  peerDependencies:
+    eslint: ^6.0.0 || ^7.0.0 || >=8.0.0
+  checksum: ac9da8475207591237f79462fef7f201611b22481df1972e6b81bd2a05dc546e0b69b0094771d23b7a3075d7d69d413c955989473abc0e227f9473108124ded4
+  languageName: node
+  linkType: hard
+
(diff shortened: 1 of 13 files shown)


yarn.lock
andysellick, line 1: There's a lot of changes in this file. Is that expected? What was the command you ran to do this update?

andysellick (commented):
Just want to check on the first commit, but otherwise looks good 👍

yarn.lock
KludgeKML: I just changed package.json and then ran `yarn install` - I assume this is all the differences between standard 16 and standard 17, plus the removal of standardx. Does any of it look suspicious?

andysellick: Not suspicious, just incredibly long. I've tried it locally and also get massive changes, so I don't think it's wrong, but I'm still surprised by the extent of the changes.

andysellick: Might be worth getting a second opinion - @AshGDS ?

AshGDS: Might be worth doing `yarn remove standardx` and then `yarn add standard` instead. That still results in ` 415 insertions(+), 384 deletions(-)` for me though.

KludgeKML: With that I get:

```
❯ git diff --stat HEAD~4 HEAD~5
 package.json |    2 +-
 yarn.lock    | 1289 +++++++++++++++++++++++++++-------------------------------
 2 files changed, 594 insertions(+), 697 deletions(-)
```

But seems like a more repeatable way of doing it.

andysellick (dismissed):

andysellick (approved):

PR #5769, opened 2026-09-04
Fix page title columns layout

andysellick:

## What / why
Fix the CSS grid styles in the page title partial so they correctly line up with the GOVUK grid. Previously they were close but not exact, due to what I thought was a quirk of CSS grid compared to floating elements. Turns out the solution was a little easier - just copy the GOVUK grid a bit more.

Here's a few test pages, including a right to left:

- [translation nav and news](https://www.gov.uk/world/organisations/british-embassy-dakar)
- [news but no translation nav](https://www.gov.uk/world/organisations/high-commission-victoria)
- [right to left](https://www.gov.uk/world/organisations/british-embassy-tehran.fa)

## Visual changes
Lines added for reference. Works at full desktop at all sizes right down to just before mobile, where the columns collapse.

Screenshot 2026-09-04 at 08 39 03
Screenshot 2026-09-04 at 08 39 07
Screenshot 2026-09-04 at 08 39 22

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -114,10 +114,17 @@
 }
 
 @include govuk-media-query($from: tablet) {
+  $content-width: 66.6667%;
+  $sidebar-width: auto;
+
   .page-title {
     display: grid;
-    grid-template-columns: [content] 65.666% [sidebar] auto;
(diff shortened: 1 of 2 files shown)


AshGDS (approved):
Nice 👍

PR #5771, opened 2026-09-07
Use layouts for document collection pages

andysellick:

## What
Convert the document collection page type to use page layouts.

Examples:

- [document collection](https://www.gov.uk/government/collections/visa-processing-times)
- [long enough to have the floating 'contents' at the bottom](https://www.gov.uk/government/collections/intellectual-property-trade-marks)
- [with 'get emails about this topic' link](https://www.gov.uk/government/collections/paying-hmrc-detailed-information) (as opposed to 'get emails about this page' single page notification button)
- [without change history](https://www.gov.uk/government/collections/resources-for-asylum-seekers-in-the-uk)

Some stuff about the 'Get emails..' button. In summary - it was too complicated, so I've not merged it into the `page-title` partial for now, just left it as it is in the view.

- previously added the button to `page-title` in https://github.com/alphagov/frontend/pull/5676 for call for evidence pages
- logic for CFE is a lot simpler, just call `shared/single_page_notification_button` with the content_item
- more complex for document collection and this seems to be unique to this page type
- if `show_email_signup_link` method returns true, shows the signup link component
- else if `display_single_page_notification_button` then include the `shared/single_page_notification_button` partial, but pass an additional param for `skip_account`, `!logged_in?`
- `logged_in?` comes from https://github.com/alphagov/govuk_personalisation/blob/ce735240c0cfc572345e2da68ceea33a91253ad8/lib/govuk_personalisation/controller_concern.rb#L65 and again only seems to be used by document collection
- when I tried to merge this into `page-title` and included the `logged_in` param it changed how call for evidence pages displayed the button, specifically in the target for the form
- we might circle back to this later

## Why

## Visual changes
Included in this PR is the removal of the published dates component, which was at the foot of the content. The information this contained (page history) has been moved into the metadata component at the top of the page.

Other than that there should be very little visual difference, although the vertical alignment of some of the elements in the page header is improved.

Before | After
------ | -----
Screenshot 2026-09-10 at 10 00 01 | Screenshot 2026-09-10 at 10 00 11

### DIFF
--- a/app/controllers/document_collection_controller.rb
+++ b/app/controllers/document_collection_controller.rb
@@ -1,6 +1,7 @@
 class DocumentCollectionController < ContentItemsController
   include Cacheable
   include Personalisable
+  layout "header_content_sidebar"
 
   def show
     @content_item_presenter = DocumentCollectionPresenter.new(content_item)
--- a/app/presenters/document_collection_presenter.rb
+++ b/app/presenters/document_collection_presenter.rb
@@ -1,5 +1,22 @@
 class DocumentCollectionPresenter < ContentItemPresenter
   include ContentsList
+  include DateHelper
+  include LinkHelper
+
+  def use_contextual_components?
+    true
+  end
+
+  def page_title_options
+    super.merge({
+      metadata: {
+        from: govuk_styled_links_list(contributor_links),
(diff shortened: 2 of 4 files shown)


KludgeKML (approved):
LGTM

PR #5777, opened 2026-09-08
Upgrade standard and stylelint dependencies

MartinJJones:
## What

Upgrade standard and stylelint dependencies

## Why

Ensure we are using the latest dependencies


MartinJJones:
Closed in favour of https://github.com/alphagov/frontend/pull/5780

PR #5778, opened 2026-09-08
Update brakeman to ignore false positives

KludgeKML:

## What

Update brakeman ignore file to silence some false positives.

## Why

These are showing up as code scanning alerts, but I believe them to be false positives based on the fact that brakeman doesn't appear to look inside the method. Most of these values aren't user-supplied (they come from a content item, so they're controlled on our side).

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,7 +1,277 @@
 {
   "ignored_warnings": [
-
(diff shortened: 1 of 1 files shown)


deborahchua (approved):
Looks okay to me!

PR #5782, opened 2026-09-09
[PNP-10145] Allow will_continue_on to be blank

KludgeKML:

## What

- In specialist publisher, metadata/continuation_link is required, but metadata/will_continue_on (the explanatory text) is optional. We haven't had one of these without a will_continue_on before, but we do now, so we need to handle it.
- We also remove the strip here, as values should be stripped when they get into the content item.

## Why

Jira card: https://gov-uk.atlassian.net/browse/PNP-10145

## Visual changes

None


sairamya93 (approved):
LGTM!

CodeSonia (approved):
Nice 👌

PR #5783, opened 2026-09-09
Redirect draft HTML requests to Asset Manager preflight route [WHIT-3983]

ChrisBAshton:
## What

This PR introduces a new AssetManagerPreflightController, which forces an Asset Manager session to be set up.

The AssetManagerPreflightController loads a single, permanent, harmless placeholder image from Asset Manager's draft assets host. That's enough to complete one Signon handshake and establish an Asset Manager session cookie in the browser - the user should already have an active Signon session by the time they reach a draft preview, so this is just completing a handshake, not asking them to sign in again.

Once that's done - or once we've given it a reasonable amount of time to finish - we bounce the user back to the page they originally asked for. From then on, every image request on that page reuses the now-warm Asset Manager session instead of starting its own handshake.

## Why

Last week, we [shipped a change in Whitehall](https://github.com/alphagov/whitehall/pull/11667), changing how images work and bringing them in line with attachments. Images are now uploaded to the draft stack by default, when uploaded to a draft document - and only become 'live' when the document is published. Access-limiting metadata on the document (individual/organisation access-limiting, auth-bypass tokens) are also proliferated through to the assets in the same way as they are for attachments.

Unlike attachments, which are exclusively viewed in isolation in browser (e.g. a PDF in its own tab) or downloaded at their own endpoint, images are **embedded** in the page. Whereas attachments, by virtue of being visited/downloaded in isolation, funnel publishers through Signon and then allow the visit/download, images are not able to do so because they live within the page.

If only one image is embedded in the page, it works: the publisher is already signed into Signon at the point of getting through to draft-frontend, and so under the hood when the browser renders the DOM, the draft image is parsed, several round trip network requests are made to Asset Manager and Signon, and if the user is authorised, the image downloads and renders.

The problem occurs when we're dealing with more than one draft asset in the page. A draft document can contain several draft images, and each one independently triggers its own Asset Manager/Signon authentication handshake when there's no existing Asset Manager session. Because those handshakes can complete out of order, they invalidate each other's session state and the images fail to load.

To avoid this, we need to make sure a single Asset Manager session has already been established before rendering any draft page that might contain draft images.

An equivalent fix has been applied to Whitehall, where the same issue occurs when viewing the Images tab of a draft document: https://github.com/alphagov/whitehall/pull/11775

**This is seen as a suboptimal fix**. We will raise a ticket on the Publishing Tech Debt board to reconsider a better engineered solution to the problem, but any alternatives we've considered so far require significantly more work and come with additional risk. Given the draft assets issue is affecting publishers today, we want to prioritise what we think is an acceptable patch here and pay down the technical debt later.

Jira card: https://gov-uk.atlassian.net/browse/WHIT-3983

## Visual changes

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -21,6 +21,8 @@ Rails.application.routes.draw do
   # Favicon redirect
   get "/favicon.ico", to: "favicon#redirect_to_asset"
 
+  get "/draft-asset-preflight", to: "asset_manager_preflight#show", as: :asset_manager_preflight
+
   unless Rails.env.production?
     get "/development", to: "development#index"
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,9 @@
 class ApplicationController < ActionController::Base
+  include DraftHelper
+
   before_action { I18n.locale = I18n.default_locale }
   before_action :allow_only_html_requests
+  before_action :redirect_to_asset_manager_preflight_if_required, if: -> { request.format.html? }
 
   rescue_from GdsApi::TimedOutException, with: :error_503
   rescue_from GdsApi::EndpointNotFound, with: :error_503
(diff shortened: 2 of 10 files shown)


config/routes.rb
```diff
@@ -21,6 +21,8 @@
   # Favicon redirect
   get "/favicon.ico", to: "favicon#redirect_to_asset"
 
+  get "/draft-asset-preflight", to: "asset_manager_preflight#show", as: :asset_manager_preflight
```
KludgeKML, line 24: Could you move this under the /api section? (we're trying to make this file as alphabetical as possible given the constraints).

ChrisBAshton:
Closing in favour of #5788.

This approach wouldn't work because, as per [our docs](https://docs.publishing.service.gov.uk/manual/cookies-and-sessions-in-frontend-apps.html):

> GOV.UK frontend applications, with the exception of Licensing and Email Alert Frontend, do not currently require cookies.
> We configure Varnish to strip cookies for [inbound](https://github.com/alphagov/govuk-puppet/blob/a6c51d887a6501f02766b7279127b60f02037a7f/modules/varnish/templates/default.vcl.erb#L83) and [outbound](https://github.com/alphagov/govuk-puppet/blob/a6c51d887a6501f02766b7279127b60f02037a7f/modules/varnish/templates/default.vcl.erb#L119) requests.
> This means any features in frontend application relying on sessions/cookies will not work.

The docs point to Puppet, which was replaced by Kubernetes a couple of years ago, but the logic still stands. There is VCL in govuk-helm-charts that prevents `Set-Cookie`, apart from for an allow-listed few. And even if we add our new endpoint to that allowlist, the nginx config clobbers the cookie on read, to an empty string, so we can't make use of the value. See:

PR #5786, opened 2026-09-09
Add more accurate BadRequest handling

KludgeKML:

## What

Add explicit capture/report of BadRequests from GdsApi::HTTPBadRequest

## Why

Currently bad requests are getting turned into 503s, which means they turn up in Sentry. We should ideally not be alerting on these, so return a correct 400 (this will be important when we include gds-api-adapters 103.5.0, which returns BadRequest immediately when we ask content store for a path that doesn't conform to RFC-192


CodeSonia (approved):
LGTM!

PR #5789, opened 2026-09-10
Force asset manager preflight login

eYinka:
## What
Load a placeholder image from the draft-assets host on every draft page. Once that request has settled, retry every draft `` already on the page with a cache-busting query param, so they get a fresh attempt against what should now be a warm Asset Manager session cookie.

## Why

As of https://github.com/alphagov/whitehall/pull/11667, images can now be uploaded to the draft stack, meaning Frontend needs to be able to render the draft images in the page when on draft preview.

Unlike live images, draft images require the user to be authenticated (via Signon) and for an Asset Manager session to exist. When rendering a draft asset, the request for the asset from Asset Manager is redirected through a chain of calls including Authenticating Proxy and Signon, and if 1) the user is logged into Signon, and 2) they have permission to view the asset, the asset is then downloaded. Subsequent attempts to render any asset avoids the OAuth round-trip, as the user already has a warm Asset Manager session cookie by then.

The problem arises when more than one draft asset is included in a page, and a user is coming to the page cold. When rendering the page, every draft asset begins its own authentication journey, and the various callbacks and tokens will often arrive out of order, meaning no session takes hold. The images fail to load, and subsequent refreshes of the page rarely help (unless the order of requests sent and received happens in a very particular order). Visiting a draft asset in isolation, e.g. on a new tab, forces the session cookie and then subsequent attempts to render a page full of draft assets will succeed.

In #5783 we explored adding a preflight mechanism to Frontend whereby 'cold' users are redirected to a page containing a single asset, and then redirected back to their original page. Unfortunately our nginx config would require that the single-asset-page be added to an allowlist of pages that are allowed to set cookies, and even then, subsequent pages would not be allowed to read the cookie, which is then clobbered to an empty string:

The blast radius is small - we only inject this request on the draft stack, so only publishers will see it, and the real world ramifications are perhaps a very slight delay in the rendering of the page.

The asset chosen is the default lead image placeholder referenced throughout Frontend. It is the same asset used in a similar solution in Whitehall: https://github.com/alphagov/whitehall/pull/11775

Jira: https://gov-uk.atlassian.net/browse/WHIT-3983

### DIFF
--- a/app/views/shared/_asset_manager_session.html.erb
+++ b/app/views/shared/_asset_manager_session.html.erb
@@ -1,0 +1,18 @@
+<% if draft_host? %>
+  <%
+    host = if GovukEnvironment.current == "staging"
+             "staging.publishing.service.gov.uk"
+           elsif GovukEnvironment.current == "integration"
+             "integration.publishing.service.gov.uk"
+           else
+             "publishing.service.gov.uk"
+           end
+    placeholder_asset_url = "https://draft-assets.#{host}/media/5e59279b86650c53b2cefbfe/placeholder.jpg"
+  %>
+  <%= javascript_include_tag "lib/asset-manager-session.js", integrity: false %>
+  <%= javascript_tag nonce: true do %>
+    window.addEventListener("load", function () {
+      GOVUK.warmUpAssetManagerSession('<%= placeholder_asset_url %>')
+    })
+  <% end %>
+<% end %>
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -9,6 +9,7 @@
 //= link components/map/map-test.geojson
 //= link components/map/maplibre-gl-csp-worker.js
 //= link views/travel-advice.js
+//= link lib/asset-manager-session.js
 
 //= link static-error-pages.js
 
--- a/app/assets/javascripts/lib/asset-manager-session.js
+++ b/app/assets/javascripts/lib/asset-manager-session.js
@@ -1,0 +1,33 @@
+/* Warms up the Asset Manager draft-assets session cookie via a single
+ * placeholder image request, then reloads any draft images already on
+ * the page once that request has settled (success or failure).
+ */
+/* istanbul ignore next */
+window.GOVUK = window.GOVUK || {};
+
+(function (root) {
+  function draftAssetImages () {
+    return [...document.images].filter((image) =>
+      image.src.includes('assets.') // currently, asset urls in draft preview point to their live link: 'assets.xyz' instead of 'draft-assets.xyz'. Created a backlog item for this: https://gov-uk.atlassian.net/browse/WHIT-4008.
+    )
+  }
+
+  function reloadDraftImages (draftAssets) {
+    draftAssets.forEach((image) => {
+      // re-setting src tries to fetch the image again, with the hope that a prior success is just served from cache
+      image.setAttribute('src', image.getAttribute('src'))
+    })
+  }
+
+  root.GOVUK.warmUpAssetManagerSession = function (placeholderAssetUrl) {
+    const draftAssets = draftAssetImages()
+    // fewer than 2 images means there's no concurrent-request race to fix, so nothing to warm up
+    if (draftAssets.length < 2) return
+
+    const reload = () => reloadDraftImages(draftAssets)
+    const placeholder = new Image()
+    placeholder.onload = reload
+    placeholder.onerror = reload
+    placeholder.src = placeholderAssetUrl
+  }
+}(window))
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -24,4 +24,5 @@
       <%= yield %>
     </main>
   </div>
+  <%= render "shared/asset_manager_session" %>
 <% end %>
--- a/app/views/layouts/full_width.html.erb
+++ b/app/views/layouts/full_width.html.erb
@@ -14,4 +14,5 @@
     <span id="Top"></span>
     <%= yield %>
   </main>
+  <%= render "shared/asset_manager_session" %>
 <% end %>
--- a/app/views/layouts/header_content_sidebar.html.erb
+++ b/app/views/layouts/header_content_sidebar.html.erb
@@ -60,5 +60,6 @@
         </div>
       <% end %>
     </main>
+    <%= render "shared/asset_manager_session" %>
   </div>
 <% end %>
--- a/app/views/layouts/header_sidebar_content.html.erb
+++ b/app/views/layouts/header_sidebar_content.html.erb
@@ -34,5 +34,6 @@
         </div>
       </div>
     </main>
+    <%= render "shared/asset_manager_session" %>
   </div>
 <% end %>
--- a/app/views/layouts/manual.html.erb
+++ b/app/views/layouts/manual.html.erb
@@ -17,5 +17,6 @@
       <span id="Top"></span>
       <%= yield %>
     </main>
+    <%= render "shared/asset_manager_session" %>
   </div>
 <% end %>
--- a/spec/javascripts/lib/asset-manager-session.spec.js
+++ b/spec/javascripts/lib/asset-manager-session.spec.js
@@ -1,0 +1,90 @@
+describe('GOVUK.warmUpAssetManagerSession', function () {
+  let originalImage
+
+  const addImage = (src) => {
+    const img = document.createElement('img')
+    img.src = src
+    document.body.appendChild(img)
+    return img
+  }
+
+  const trackReloads = (image) => {
+    const reloadedUrls = []
+    const originalSetAttribute = image.setAttribute.bind(image)
+    image.setAttribute = (name, value) => {
+      if (name === 'src') reloadedUrls.push(value)
+      originalSetAttribute(name, value)
+    }
+    return reloadedUrls
+  }
+
+  const stubImage = ({ succeeds }) => {
+    class FakeImage {
+      get src () {
+        return this._src
+      }
+
+      set src (value) {
+        this._src = value
+        succeeds
+          ? this.onload && this.onload()
+          : this.onerror && this.onerror()
+      }
+    }
+    window.Image = FakeImage
+  }
+
+  beforeEach(function () {
+    originalImage = window.Image
+  })
+
+  afterEach(function () {
+    window.Image = originalImage
+    document.querySelectorAll('img').forEach((img) => img.remove())
+  })
+
+  it('does nothing when fewer than two draft images are on the page', function () {
+    addImage('https://draft-assets.test.gov.uk/media/1/one.jpg')
+    stubImage({ succeeds: true })
+
+    GOVUK.warmUpAssetManagerSession('https://draft-assets.test.gov.uk/media/placeholder/placeholder.jpg')
+
+    expect(document.images[0].src).toEqual(
+      'https://draft-assets.test.gov.uk/media/1/one.jpg'
+    )
+  })
+
+  it('retries all draft images once the placeholder request succeeds', function () {
+    const first = addImage('https://draft-assets.test.gov.uk/media/1/one.jpg')
+    const second = addImage('https://draft-assets.test.gov.uk/media/2/two.jpg?foo=bar')
+    const firstReloads = trackReloads(first)
+    const secondReloads = trackReloads(second)
+    stubImage({ succeeds: true })
+
+    GOVUK.warmUpAssetManagerSession('https://draft-assets.test.gov.uk/media/placeholder/placeholder.jpg')
+
+    expect(firstReloads).toEqual(['https://draft-assets.test.gov.uk/media/1/one.jpg'])
+    expect(secondReloads).toEqual(['https://draft-assets.test.gov.uk/media/2/two.jpg?foo=bar'])
+  })
+
+  it('also retries all draft images when the placeholder request errors', function () {
+    const first = addImage('https://draft-assets.test.gov.uk/media/1/one.jpg')
+    addImage('https://draft-assets.test.gov.uk/media/2/two.jpg')
+    const firstReloads = trackReloads(first)
+    stubImage({ succeeds: false })
+
+    GOVUK.warmUpAssetManagerSession('https://draft-assets.test.gov.uk/media/placeholder/placeholder.jpg')
+
+    expect(firstReloads).toEqual(['https://draft-assets.test.gov.uk/media/1/one.jpg'])
+  })
+
+  it('ignores non-asset-manager images when counting/retrying', function () {
+    const other = addImage('https://static.test.gov.uk/media/1/one.jpg')
+    addImage('https://draft-assets.test.gov.uk/media/2/two.jpg')
+    stubImage({ succeeds: true })
+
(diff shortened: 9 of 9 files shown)


app/assets/javascripts/asset-manager-session.js
```diff
@@ -0,0 +1,48 @@
...
+    this.element = element
+  }
+
+  AssetManagerSession.prototype.init = function () {
+    window.addEventListener('load', this.warmUpSession.bind(this))
+  }
+
+  AssetManagerSession.prototype.warmUpSession = function () {
+    const draftAssets = this.draftAssetImages()
+    if (draftAssets.length < 2) return
```
ChrisBAshton: I'd add an inline comment here explaining why 1 or 0 images is OK to early return.

app/assets/javascripts/modules/asset-manager-session.js
```diff
@@ -0,0 +1,48 @@
...
+    const retry = this.retryDraftAssets.bind(this, draftAssets)
+    const placeholder = new Image()
+    placeholder.onload = retry
+    placeholder.onerror = retry
+    placeholder.src = this.element.getAttribute('data-placeholder-asset-url')
+  }
+
+  AssetManagerSession.prototype.draftAssetImages = function () {
+    return [...document.images].filter((image) =>
+      image.src.includes('assets.') // currently, asset urls in draft preview point to their live link: 'assets.xyz' instead of 'draft-assets.xyz'
```
ChrisBAshton: We should create a backlog Whitehall item to fix that 🤔 and link to it in the comment here

app/views/shared/_footer_navigation.html.erb
```diff
@@ -8,3 +8,21 @@
...
+             "integration.publishing.service.gov.uk"
+           else
+             "publishing.service.gov.uk"
+           end
+    placeholder_asset_url = "https://draft-assets.#{host}/media/5e59279b86650c53b2cefbfe/placeholder.jpg"
+  %>
+  <div
+    data-module="AssetManagerSession"
+    data-placeholder-asset-url="<%= placeholder_asset_url %>"
+    hidden>
```
ChrisBAshton: Is this tag necessary? We don't render anything in the div, so there's nothing to hide?

app/assets/javascripts/modules/asset-manager-session.js
```diff
@@ -0,0 +1,48 @@
...
+
+  AssetManagerSession.prototype.init = function () {
+    window.addEventListener('load', this.warmUpSession.bind(this))
+  }
+
+  AssetManagerSession.prototype.warmUpSession = function () {
+    const draftAssets = this.draftAssetImages()
+    if (draftAssets.length < 2) return
+
+    const retry = this.retryDraftAssets.bind(this, draftAssets)
```
ChrisBAshton: I'd rename `retry` as it implies on lines 29-30 that we want to retry downloading the placeholder asset. I'd call it `reloadDraftImages` (and do the same for the method name).

app/assets/javascripts/modules/asset-manager-session.js
```diff
@@ -0,0 +1,48 @@
...
+
+  AssetManagerSession.prototype.draftAssetImages = function () {
+    return [...document.images].filter((image) =>
+      image.src.includes('assets.') // currently, asset urls in draft preview point to their live link: 'assets.xyz' instead of 'draft-assets.xyz'
+    )
+  }
+
+  AssetManagerSession.prototype.retryDraftAssets = function (draftAssets) {
+    draftAssets.forEach((image) => {
+      const separator = image.src.includes('?') ? '&' : '?'
```
ChrisBAshton: Might be worth adding a unit test for this separator logic too

ChrisBAshton (commented):
LGTM 👍 Some small comments below, which I can help with later if you don't get a chance to in the meantime.

app/assets/javascripts/modules/asset-manager-session.js
ChrisBAshton: Actually, second thoughts - we could also test dropping the cache-busting entirely. I imagine the 'failed' image would not be re-served from cache, and any 'successful' image _would_ be re-served from cache, so might help keep the page fast.

app/views/shared/_footer_navigation.html.erb
eYinka: What I intended to do was to replicate `govuk-visually-hidden` just like in [place.html.erb:17](https://github.com/alphagov/frontend/blob/main/app/views/place/_place.html.erb#L17). I pasted the wrong thing in  🤦 
The idea behind that was to ensure the page has zero layout shifts; just as a safety net in case that block magically adds 1px of height.

app/assets/javascripts/modules/asset-manager-session.js
eYinka: Created https://gov-uk.atlassian.net/browse/WHIT-4008

ChrisBAshton (dismissed):
Looks good to me - nice one for keeping it scoped only to the draft stack, to minimise blast radius, whilst still keeping it modular and testable 🎉 

Will get a 2i from a frontend dev

app/views/shared/_footer_navigation.html.erb
```diff
@@ -8,3 +8,25 @@
...
+    placeholder_asset_url = "https://draft-assets.#{host}/media/5e59279b86650c53b2cefbfe/placeholder.jpg"
+  %>
+  <div
+    id="asset-manager-session"
+    data-placeholder-asset-url="<%= placeholder_asset_url %>"
+    class="govuk-visually-hidden">
+  </div>
+  <%= javascript_include_tag "asset-manager-session.js", integrity: false %>
+  <%= javascript_tag nonce: true do %>
+    new GOVUK.Modules.AssetManagerSession(document.getElementById("asset-manager-session")).init()
```
andysellick: I don't think you need to manually initialise this code here (unless I've misunderstood how it works). All of our Javascript modules are started automatically when an element is given a `data-module` attribute e.g. in this case `data-module="asset-manager-session"`. This would mean you can also remove the id from the element.

app/views/shared/_footer_navigation.html.erb
```diff
@@ -8,3 +8,25 @@
...
+             "publishing.service.gov.uk"
+           end
+    placeholder_asset_url = "https://draft-assets.#{host}/media/5e59279b86650c53b2cefbfe/placeholder.jpg"
+  %>
+  <div
+    id="asset-manager-session"
+    data-placeholder-asset-url="<%= placeholder_asset_url %>"
+    class="govuk-visually-hidden">
+  </div>
+  <%= javascript_include_tag "asset-manager-session.js", integrity: false %>
```
andysellick: This could probably be included in the main `application.js` rather than manually calling it here.

app/views/shared/_footer_navigation.html.erb
andysellick, line 1: Can I check that this is best place to put this new code? I think only 11 content types use this partial. It might be better in the main layouts (although you would have to duplicate it, so it would need to go in its own partial).

app/views/shared/_footer_navigation.html.erb
```diff
@@ -8,3 +8,25 @@
     </div>
   </div>
 <% end %>
+
+<% if draft_host? %>
```
andysellick: Where is `draft_host` defined?

app/views/shared/_footer_navigation.html.erb
```diff
@@ -8,3 +8,25 @@
...
+  <%
+    host = if GovukEnvironment.current == "staging"
+             "staging.publishing.service.gov.uk"
+           elsif GovukEnvironment.current == "integration"
+             "integration.publishing.service.gov.uk"
+           else
+             "publishing.service.gov.uk"
+           end
+    placeholder_asset_url = "https://draft-assets.#{host}/media/5e59279b86650c53b2cefbfe/placeholder.jpg"
```
andysellick: This URL looks fingerprinted - is that fingerprint ever likely to change?

app/assets/javascripts/asset-manager-session.js
```diff
@@ -0,0 +1,49 @@
...
+window.GOVUK = window.GOVUK || {}
+/* istanbul ignore next */
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function AssetManagerSession (element) {
+    this.element = element
+  }
+
+  AssetManagerSession.prototype.init = function () {
```
andysellick: Non-blocking - ideally any new JS modules would be written using a newer syntax. Example:

```Javascript
(function (Modules) {
  class AssetManagerSession {
    constructor ($module) {
    }

    init () {
    }
  }
  Modules.AssetManagerSession = AssetManagerSession
})(window.GOVUK.Modules)
```

See https://github.com/alphagov/frontend/blob/main/app/assets/javascripts/components/map.js for example.

app/assets/javascripts/asset-manager-session.js
```diff
@@ -0,0 +1,49 @@
+/* Warms up the Asset Manager draft-assets session cookie via a single
+ * placeholder image request, then retries any draft images already on
+ * the page once that request has settled (success or failure).
+ *
+ * Usage: add `id="..."` and `data-placeholder-asset-url="..."` to an element,
+ * then call `new GOVUK.Modules.AssetManagerSession(element).init()`.
+ */
+
+/* istanbul ignore next */
```
andysellick: Were these ignore lines necessary for the code coverage?

app/views/shared/_footer_navigation.html.erb
eYinka: Well, that was the initial implementation but it didn't solve our race-condition problem. The auto-init is scanned on DOMContentLoaded, which meant the images already fired and triggered a csrf problem before the module kicks in. This was the only other way I could think of.

app/views/shared/_footer_navigation.html.erb
eYinka: Yes ideally, but we didn't want it bundled in production's compiled JS, just to reduce the blast radius if anything ever goes wrong with it i.e it's scoped and mounted on the draft stack only.

ChrisBAshton: It would then be included on the live frontend too, which we probably don't want - in case there are any existing (or future) bugs with the module we've added. We deliberately want it scoped to only happen on the draft stack.

app/views/shared/_footer_navigation.html.erb
ChrisBAshton: Ooh, good shout - worth us double-checking this.

app/views/shared/_footer_navigation.html.erb
eYinka: That's in https://github.com/alphagov/frontend/blob/main/app/helpers/draft_helper.rb#L2

ChrisBAshton: https://github.com/alphagov/frontend/blob/2056b0ee3aba1b3de62c805b5ed7084fe326a29e/app/helpers/draft_helper.rb#L2-L3

app/views/shared/_footer_navigation.html.erb
ChrisBAshton: Good question. I think 'no' - it's already referenced in [several places in Frontend and Whitehall](https://github.com/search?q=org%3Aalphagov+5e59279b86650c53b2cefbfe&type=code) and was (I guess?) uploaded manually to Asset Manager, never to be touched again. It's the default 'lead image' for news articles if no image has been selected.

eYinka: Not likely anytime soon. We use it in Whitehall to set the [default placeholder](https://github.com/alphagov/whitehall/blob/24e2c442a7e2d5ce2eb9e5a4a6c52881a76481d2/app/models/concerns/standard_edition/lead_image.rb#L4) as well. Also, this is meant to be a temporary fix for the ongoing incident. A proper solution is needed, which will likely start from signon and cut across several other apps. I think this url should be fine for now.

app/assets/javascripts/asset-manager-session.js
eYinka: I think so. I just copied over the same convention from  https://github.com/alphagov/frontend/blob/main/app/assets/javascripts/modules/sticky-element-container.js

app/views/shared/_footer_navigation.html.erb
andysellick: In that case I wouldn't write it as a module. I'd write it as a self executing block of code and put it in `app/javascripts/lib` or something. See https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js for an example.

app/views/shared/_footer_navigation.html.erb
andysellick: Ah okay that makes sense, thanks.

app/views/shared/_footer_navigation.html.erb
andysellick: Thanks for clarifying.

app/assets/javascripts/asset-manager-session.js
andysellick: This comment can now be ignored given that I think this shouldn't be written as a module - see my earlier comment.

app/assets/javascripts/asset-manager-session.js
andysellick: Non-blocking, but I'm not sure they're necessary. Might be worth removing them and seeing if the coverage still passes.

eYinka: Just got tripped by its absence in ci. See [coverage error](https://github.com/alphagov/frontend/actions/runs/34612437477/job/103305995850?pr=5789)

app/views/shared/_footer_navigation.html.erb
eYinka: I've now extracted it to its own partial and used across the 5 layouts in `app/views/layouts`

app/views/shared/_footer_navigation.html.erb
eYinka: Thanks for the tip. That's all done now!

app/views/shared/_asset_manager_session.html.erb
```diff
@@ -0,0 +1,18 @@
...
+             "integration.publishing.service.gov.uk"
+           else
+             "publishing.service.gov.uk"
+           end
+    placeholder_asset_url = "https://draft-assets.#{host}/media/5e59279b86650c53b2cefbfe/placeholder.jpg"
+  %>
+  <%= javascript_include_tag "lib/asset-manager-session.js", integrity: false %>
+  <%= javascript_tag nonce: true do %>
+    window.addEventListener("load", function () {
+      GOVUK.warmUpAssetManagerSession('<%= placeholder_asset_url %>')
```
ChrisBAshton, line 15: Very nice.

ChrisBAshton (approved):
Thanks for working through the comments! ⭐

app/views/shared/_footer_navigation.html.erb
andysellick: Nice.

app/assets/javascripts/asset-manager-session.js
andysellick: Okay, fair enough.

Issue #5716, opened 2026-08-03
Maplibre-gl no longer supports UMD


andysellick:
We use the DEFRA map in `frontend`s map component. In order to have it function correctly within our CSP we rely upon `maplibre-gl`s CSP worker, which is used in place of the default.

Unfortunately, as of v6 of maplibre-gl the UMD approach has been removed, and can only be used with ES6 modules. Our Rails setup does not yet support this, so we've had to [pin to an older version of maplibre-gl](https://github.com/alphagov/frontend/pull/5715).

## Context
Recording in case this is helpful.

We currently use the [UMD approach for the DEFRA map](https://defra.github.io/interactive-map/getting-started). This works as follows:

- we [import the csp worker from the node_modules](https://github.com/alphagov/frontend/blob/main/app/assets/javascripts/components/map/maplibre-gl-csp-worker.js) via a JS file specifically created for this
- since we don't know what the fingerprinted filename for this file will be, we render the fingerprinted filename in the map component file and [attach it to a data attribute](http://github.com/alphagov/frontend/blob/main/app/views/components/_map.html.erb#L35)
- we then [read the attribute in the component JS](https://github.com/alphagov/frontend/blob/main/app/assets/javascripts/components/map.js#L11) and pass it to the initialiser for the DEFRA map

## ES modules
The alternative to UMD is ES modules. We currently can't use [ES modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) on GOV.UK because of our asset pipeline, which compiles and fingerprints our JS files. Typical ES modules syntax such as the following will fail because the JS doesn't know the fingerprinted name of the file it is trying to import.

```JS
import { name as squareName, draw } from "./shapes/square.js"; // fails
// square.js is output from the asset pipeline as square-[unknown fingerprint].js
```

PR #5790, opened 2026-09-11
[DO NOT MERGE] Spike importmaps

andysellick:

## What
Spiking importmaps to try to allow the map component to work with ES modules.

References:

- https://github.com/rails/importmap-rails
- https://medium.com/jungletronics/mastering-import-maps-in-rails-8-5f6c32991aa8

Note: this requires the line `<%= javascript_importmap_tags %>` to be added to the components gem, specifically in the HEAD of the public layout component, and in the component guide layout template HEAD.

Here's why this is difficult.

- we can use importmaps to include the map component code as a module, this needs to be excluded from the normal sprockets compilation in `assets.rb`
- we then need to do the same for the `defra` and `maplibre` dependencies, which the map code needs to import
- they can't be called directly from `node_modules` so I created new files that `require` each of them, and put them into the new place for JS (`app/javascript`)
- however at that point `defra` or `maplibre` start using their own `import x from y` commands, which fail because those other things don't exist (`maplibre` relies on babel, for example)
- so then we need to add those dependencies into the mix as well, at which point we're manually maintaining an increasingly complex web of things, and dealing with new errors where variable names in these different files clash, for some reason

This isn't working yet, and I think it might be possible to get there, but the effort and maintenance burden is beginning to look a little high.

## Why
See https://github.com/alphagov/frontend/issues/5716

https://gov-uk.atlassian.net/jira/software/c/projects/PNP/boards/1356?selectedIssue=PNP-10149


andysellick:
Closing as a not viable solution.

PR #5792, opened 2026-09-11
Add analytics to the map component

andysellick:

**Note: this PR currently includes the addition of a map to the homepage of GOV.UK, for testing purposes (analytics doesn't work in the component guide). DO NOT MERGE until this has been removed.**

## What
Adds analytics events to the map component. These events are:

- when a user clicks on a map marker
- when a panel (popup) opens (note that this will happen immediately after a user clicks on a map marker - it is possible for a map marker to not have a panel, but that wouldn't be helpful in a real world usage of the map component)
- when a panel is closed
- when links are clicked in a panel
- when links are clicked in the list of panels (beneath the map)

## Why
https://gov-uk.atlassian.net/jira/software/c/projects/PNP/boards/1356?selectedIssue=PNP-10150

## Visual changes
None.


app/assets/javascripts/components/map.js
```diff
@@ -59,21 +60,52 @@
...
 
       /* istanbul ignore next */
       this.map.on('interact:selectionchange', (e) => {
         if (e.selectedMarkers.length > 0) {
           let marker = parseInt(e.selectedMarkers[0].replace('marker-', ''))
           marker = this.markers[marker]
+
+          const popupContent = this.createPopupContent(marker)
+          const el = document.createElement('div')
+          el.innerHTML = popupContent
```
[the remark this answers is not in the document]
andysellick: In this situation this isn't a concern. We're using the built in functionality of `textContent` to strip HTML tags out of the popupContent to put it into an analytics event. The element created for this purpose isn't appended to the DOM, and is immediately discarded once we've extracted the text-only content from it.

app/assets/javascripts/components/map.js
```diff
@@ -192,6 +223,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         })
       }
     }
+
+    analytics (data) {
```
AshGDS: Possibly a Nitpick, but I think we ideally want this to include a verb? E.g. `sendAnalytics`

app/views/homepage/_popular_links.html.erb
AshGDS, line 1: I won't approve until this is removed of course 😄

spec/javascripts/components/map.spec.js
```diff
@@ -439,4 +437,20 @@ describe('Map component', function () {
...
+    })
+
+    afterEach(function () {
+      delete window.dataLayer
+    })
+
+    it('detects that tracking should occur', function () {
+      expect(module.trackingEnabled).toEqual(true)
+    })
+  })
```
AshGDS, line 455: Non-blocking, but I wonder if we could extract the `this.analytics` calls to a different file, and then you'd be able to test the functions a bit better against a DOM you have control over, but as you said we're fairly certain it will work given our previous analytics code.

AshGDS (commented):
LGTM - is there a reason this link is unstyled? 

Image

app/assets/javascripts/components/map.js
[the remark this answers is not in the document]
AshGDS: Non-blocking and haven't fully investigated, but if the below code was converted to use `document.createElement` would that resolve it?
```
let popupContent = `<${heading} class="govuk-heading-s govuk-!-margin-bottom-2">${feature.properties.name}</${heading}>`
```

andysellick:
> LGTM - is there a reason this link is unstyled?

Yes this is just test data so there's a link to click. I'm not planning to put a map on the homepage 😁

app/assets/javascripts/components/map.js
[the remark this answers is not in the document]
andysellick: Thanks, but I think I might have an alternative way around this - I don't think it's necessary at all, the event should only include the name of the marker, not the whole text of the panel.

app/assets/javascripts/components/map.js
andysellick: Changed.

andysellick:
@AshGDS just to highlight I've added some new logic in around panels closing, seems the event was firing twice if the user clicks the 'x' on the panel to close it, hopefully this should prevent duplicate events from firing.

app/views/homepage/_popular_links.html.erb
andysellick: @AshGDS I've now removed this change.

AshGDS (dismissed):

andysellick:
@AshGDS sorry just thought of some extra tests to add - I've put them in a separate commit, but will merge them together ahead of your final review if you're happy with them 👍

AshGDS (dismissed):
LGTM 👍

AshGDS (approved):
LGTM 👍

PR #5801, opened 2026-09-15
Fix flaky JS code coverage

AshGDS:

## What/Why
Sometimes our CI would fail as there wasn't enough code coverage on certain JS files. For example: https://github.com/alphagov/frontend/actions/runs/34957215640/job/104342040663?pr=5798

I looked into this and it was complaining about our `url-params.js` component, and the map component.

I was able to get better line coverage by:

- Ignoring the `window.GOVUK = window.GOVUK || {}` declarations on both files
- Fixing some instances in the map code where the `if` statement always resulted to `true`. See commit messages for more detail.
- Coverage has now gone from 86.72% to over 90%.

## Testing
You can test the seed that would fail by modifying `package.json`'s `jasmine:ci` command with the seed that was failing: `"jasmine:ci": "yarn run jasmine:prepare && yarn run jasmine-browser-runner runSpecs --seed=32683 && yarn run nyc:text-report",`. It should pass OK now, but if you run that command on `main` it should fail.

## Still some uncovered lines

There's still some uncovered lines in `url-params.js` but I'm not sure how to resolve that. It's saying the whole `init` function is uncovered which is essentially the whole code. Might be one to look at collectively.

If you want to see the lines for the JS it's complaining about you'll need to navigate to your `public/assets/` folder to find the built JS file.

## Visual changes

Should be none.

--


app/assets/javascripts/modules/url-params.js
AshGDS, line 1: Sorry already marked this PR for review but I should probably rename this commit as I'm also ignoring `function UrlParams (module)`. Can do this once you've reviewed it.

andysellick (approved):
Looks good. The coverage problems seemed to be intermittent, suggesting a weirder problem, are you sure this will fix it?

AshGDS:
Thanks @andysellick - It's possible it will happen again, but if we know the seed we can see what lines aren't covered. It was a bit hard to diagnose why it happens too, the debug steps I found didn't provide any evidence.  It's possible a test in the codebase is not resetting correctly and might be causing false positives in most situations, and only failed on certain seeds when the false positive didn't happen?

PR #5802, opened 2026-09-15
[PNP-10161] Remove support for embedded about pages from topic

KludgeKML:

## What

Removes support for embedded about pages in topical events (added in https://github.com/alphagov/frontend/pull/5612). This is not a simple revert, because we are also removing flexible-page-style rendering from these about pages.

Before this PR, we had two methods of rendering about pages:
- standalone content items rendered as flexible pages
- embedded about pages rendered in the standard model->presenter->layout manner.

After this PR we will have one method:
- standalone content items rendered in the standard model->presenter->layout manner.

## Why

1) We're moving away from flexible pages as a concept internal to Frontend - it's better to have more standard and simpler methods of page-building, and flexibility is in the entire system of creation (there will be some support for re-orderable blocks, but they will be limited to body rendering and handled in a different way).
2) Embedded about pages were never actually implemented in/published from Whitehall, so this code is effectively orphaned.

Jira card: https://gov-uk.atlassian.net/browse/PNP-10161

## Visual changes

None

(Compare https://www.gov.uk/government/topical-events/western-balkans-summit-london-2018/about with https://govuk-frontend-app-pr-5802.herokuapp.com/government/topical-events/western-balkans-summit-london-2018/about)


sairamya93 (approved):
Looks good 👍

